PR #6, opened 2017-12-12
Fix color setting in summary plot

tempse:
Propagate the user color setting also in case of the non-violin
plot. (This prevents the call of the automatic matplotlib color
rotation routine and the creation of a potentially much too
colorful result.)


slundberg:
Thanks!

Issue #16, opened 2018-01-17
Distribution in summary plot


kodonnell:
Interested in me submitting a PR for this? We've used it internally to make it more interpretable. If the lowest kde subplot is for variable X, then dark blue means 'bottom 20% of X values' up to red which is 'top 20% of X values'. We find this more interpretable as e.g. in this case we can quickly see that low values of X tend to decrease SHAP values, while high values increase it, etc. Otherwise it was hard to explain to clients, especially for unbalanced kdes like the middle one (which might be 'the wrong way round').

![image](https://user-images.githubusercontent.com/8204904/35066104-5ff36e2e-fc34-11e7-8d88-e12785ea177e.png)

slundberg:
Hey, that is a great idea! I like the way it connects the feature values to their SHAP values while still only using a single dimension.

A PR would be much appreciated. If you have any questions just let me know.

kodonnell:
Awesome - I will submit one hopefully over the weekend.

PR #17, opened 2018-01-18
initial distributed summary plot

kodonnell:
As per #16, plus a few additions to support scatter, and a few tweaks. @slundberg - it's not finished yet, but can you give some initial feedback:

- function call: I put all the arguments at the end so as to be backward compatible, but it's not as clean
- do you have any idea about the sum of individual kdes not equalling the overall kde?
- any comments on the visuals and changes, including code style etc. 

Examples below. Note that the 2nd and 7th from bottom have only two unique values (i.e. one-hot encoding). These two don't quite seem to match the scatterplot, which makes me somewhat suspicious.

```python
hidden_names = [str(i) for i in range(len(X_train.columns))]
summary_plot(shap_values, hidden_names, max_display=20, features=X_train.as_matrix())
summary_plot(shap_values, hidden_names, color="#cccccc", max_display=20, features=X_train.as_matrix())
summary_plot(shap_values, hidden_names, max_display=10, violin=False, features=X_train.as_matrix(), alpha=0.01)
summary_plot(shap_values, hidden_names, max_display=10, violin=False, features=X_train.as_matrix(), width=0., alpha=0.01)
```
image
image
image
image


slundberg:
Thanks! Looking good so far:

- It would probably make sense to rename the second parameter of summary_plot to `features` then it can be either an array of feature names like currently, or it can be a panda's data frame in which case we can get both the feature values and names. We would also leave the `feature_names` parameter just in case we get a numpy array for features. I think that would make it a lot cleaner, what do you think?
- There seems to be a bug with the binary variables, not sure what it is at first glance.
- The fact the the kde's don't perfectly sum to the full kde does not surprise me since they are a density estimate that can change depending on what samples you give them (I assume it is not way off).
- For consistency I would like, if possible, to use the same color scheme by default as the rest of the package, the `red_blue` cmap.
- There is often a red "halo" around the blue regions, but not a blue one around the red regions. This makes me think it is just an anti-aliasing issue and not real signal in the data. Do you agree?

kodonnell:
> It would probably make sense to rename the second parameter 

That's exactly how I did it originally. Will add it in.

>  bug with the binary variables

Yes, that's what I was suspicious of with the 2nd and 7th. In my case it looked like the 'tail' of the distributions was being removed. I think this much just be down to kde parameters, but I'll look.

> `red_blue` cmap

Sure - I didn't know there was a default.

>There is often a red "halo" around the blue regions, but not a blue one around the red regions. This makes me think it is just an anti-aliasing issue and not real signal in the data. Do you agree?

This is just by nature that the blue bars are the low values so always drawn first (i.e. center of the violin), and the red ones high and drawn last (i.e. edge of the violin). This means the color order, going from center of violin out, must be monotonically increasing in the cmap - i.e. it's impossible to have a blue halo around the red. Make sense?

Will work on tidying it up in the next few days.

kodonnell:
Note to self: 

- [ ] edit documentation (readme.md etc.)
- [ ] update notebooks??
- [ ] tidy up code including docstring

slundberg:
> This is just by nature that the blue bars are the low values so always drawn first
Ah right, I should look for small blue inner lines instead for symmetry. (I think you mean blue is drawn last right?)

One other random thought is to compare with coloring the violin plot by a kde mean of the feature feature value of samples. It would not catch U shaped shap plots well, but it would be a visually simpler but similar idea. Anyway, don't let it slow down this PR

kodonnell:
> Ah right, I should look for small blue inner lines instead for symmetry. (I think you mean blue is drawn last right?)

Sorry, yes - I was thinking of the old approach where I drew from center of plot out, whereas now I draw outward in. But you get what I mean how blue can never be outside red?

> One other random thought is to compare with coloring the violin plot by a kde mean of the feature feature value of samples. It would not catch U shaped shap plots well, but it would be a visually simpler but similar idea. Anyway, don't let it slow down this PR

You've lost me there, sorry. With 2 bins you'll bet blue = bottom 50% and red = top 50% - is that what you mean?

slundberg:
Yes it makes sense why blue is not outside red.

The idea was going off your idea of coloring the summary plot by feature value. I applied it to the scatter plot style and got:

image

Each dot is colored by its feature value, and dots stack outwards instead of covering each other, so it shows density.

kodonnell:
> Each dot is colored by its feature value, and dots stack outwards instead of covering each other, so it shows density.

Ah - the coloring is already supported in this PR, but I just used a random y-value, as otherwise it gets too similar to a KDE. Would you like me to implement as with your example? Alternatively, I can just submit what I've currently got, and let you finish the PR? (I might not get to it for a few days, and it's mainly doco/API stuff, which you'll do a better job of anyway.) I got somewhat hung up trying to run the notebooks (the front-page one kept crashing my python etc.), so was probably going to bail out of updating documentation and all the notebooks, too.

slundberg:
@kodonnell sounds good. I'll merge what you have and also try and clean up the interface a bit. Thanks!

slundberg:
Just FYI...I ended up merging these ideas into the dot plot, but still need to check some things about the layer-style plot (which will happen after some deadlines in a couple weeks)

kodonnell:
Cool - might be a good idea to append your commits on this branch etc., or we can switch to another one?

slundberg:
I'll just merge this one once I get some time so your contributions are
recorded, then I'll change it around.

slundberg:
Since files were refactored merging wouldn't make sense now. So I just copied the code (and fixed what seemed to be broken about it after merging into plots.py). I still think I might have broken something based on what I see in the notebook example for violins https://github.com/slundberg/shap/blob/master/notebooks/Scatter%20Density%20vs.%20Violin%20Plot%20Comparison.ipynb

Are you using a working version of this currently?

Thanks!

kodonnell:
Does that mean my contributions won't be in recorded in git?

I was using a working version of this, but I think I'd developed it somewhat after the repo. If you'd like, I can submit a new PR with that code, which would hopefully fix the merge and attribution stuff too.

slundberg:
That would be great. Sorry for the trouble, the parts of the shap.py file
you updated were now in the plots.py file, so merging would just mean
merging and then deleting it from the file.

I noted you in the plots.py file where I added the code, so if you send a
PR with your latest then I'll merge it in there and it will also be in git.

Thanks.

kodonnell:
Great - will aim to do so this weekend.

slundberg:
Great!

kodonnell:
@slundberg - I've pushed the tweaks. Unfortunately I don't think my original work will be recognized (as you'll show as the original committer and the net change from this PR will be minor). In future, it's probably a good idea to avoid doing the copy-and-paste trick for this reason, especially when you [say](https://github.com/slundberg/shap/pull/17#issuecomment-362004925) that you were going to merge it. Anyway, a little discouraging, but let's leave it at that.

In these updates:

- fixed bug with not supporting categorical data correctly - see change in 'sex' and 's5_binned' in plots below (I just binned the `s5` variable to demonstrate this) - you can see it's "right" when comparing with dot plot
- fixed bug in coloring (needed a `-1` not a `-0`) - see change in 'sex' feature in plots below - you can see it's "right" when comparing with dot plot
- added support for single color or a map - only for `layered_violin`
- added layered_violin_max_num_bins as an option
- updated the `ipynb` to document the options a bit more, and explain their differences. I used the diabetes dataset as it gave nicer plots and explained more features.
- added the sidebar for the layered_violin
- removed the warning as I think it's OK (at least to the same level as the others)

Before:

image

After:

image

Dot plot for reference:

image


Questions:

- Can you explain the `plot_type='violin'` option ... I was confused by it, and I'm not sure it does anything better than the other options. For that reason, I'd propose calling the 'layered_violin' just the 'violin' plot, and call this one something else ('striped_violin'?).
- Can I add an example of the `layered_violin` plot to the readme? I think it looks great, so it's good to put in the readme.
- Happy to handle #59 in here too ...

As a general comment, the plotting code/API could probably do with a bit of tidying and documenting, and getting some easy wins (e.g. #62).

slundberg:
Thanks! Sorry for the delay, I must have accidentally archived the email notification of this update.

I'll delete the copy of your changes before merging this, that way you'll be listed as the author on github of the whole section. I don't want you to feel like any credit has been lost :)

I'll work to finish the merge this week. and respond to your questions (just heading out for today).

kodonnell:
Thanks @slundberg - that's a clever solution, and thanks for bringing it up.

slundberg:
@kodonnell can you fix the merge conflict that came when I cleared out the old version? If I fix the merge I am not sure how the attribution will pan out and it would be a pain to undo.

kodonnell:
> can you fix the merge conflict 

Done, I think.

slundberg:
Now merged! I'll look over the other suggestions next week.

PR #26, opened 2018-01-30
Minor fixes for summary_plot

perretv:
Hi Scott,

First of all, I would like to thank you for sharing your amazing work with the community.
I created this PR to solve 2 minor issues I noticed when using the summary plot function:

- Some points where missing due to NaN values in the `ys` array that positions the dots on the y-axis.
- Feature names could spill out of the figure if too long.


slundberg:
Thanks! Seems like I should have tested the dot plot on a small number of points.

PR #37, opened 2018-02-25
Added custom color maps to force plots

JuanCorp:
Added the ability to use custom colors for force plots, akin to those
available in matplotlib. In case a color map is not supplied, the
default color scheme is used. A notebook was added showing this feature.


slundberg:
Thanks! Changing the colors is a good feature, and looks like you did a good job. Two points:

1. Did you mean to change something in the front page example notebook? I couldn't see what was different.
2. The force_plot function should actually be calling the visualize function in iml, rather than just returning the object (my mistake). If we do that then you don't need to include the cmap in the explanation object. This would be good since the explanation object is just an explanation that shouldn't have display parameters in it.

Both are minor issues. If you clarify (1) then I'll just merge and update the force_plot function later.

JuanCorp:
For (1), I actually didn't change anything in the front page example notebook. The reason it says there are changes, is because I was using it to test the custom colors before creating the new notebook for the custom colors. After I tested it using that notebook, I reverted all changes, so it should work as it had before.

As for (2), I wasn't aware of that either, but it's good to know for future contributions!

Thanks!

PR #42, opened 2018-03-07
Python 2 urllib compatibility issue

brianhill11:
Hi there, 

Trying to run your package using Python 2.7.13, and having issues with import of urllib.request so I added a check that imports from urllib (Python2) if it can't find the Python3 version. 

Thanks,
Brian


slundberg:
Thanks! I don't run 2.7 much but I suppose I should probably add it to travis.

PR #43, opened 2018-03-09
Improving compliance to the PEP8 standard / Code cleaning

sbrugman:
For more information on PEP8:
https://www.python.org/dev/peps/pep-0008/


slundberg:
Thanks! The code was in need of some clean-up. I'll also see about adding pylint to travis.

PR #57, opened 2018-04-12
To allow shap to work with models that consume time series data and u…

xzzxxzzx:
…se pandas dataframe as input and output, I'm proposing this change. Because shap convert dataframe into ndarray and only work with ndarrays, I propose keep index information when flag keep_index is set to be True, and reconstruct the dataframe before shap calls model.f function. keep_index is set to False by default, so there should be no behavior change to existing use case. This change should go after the change in iml package. Please feel free to let me know if you have any questions/concerns. Thanks!


slundberg:
Thanks! I'll look this over soon.

shap/explainers/kernel.py
```diff
@@ -76,14 +78,21 @@ def shap_values(self, X, **kwargs):
...
         assert str(type(X)).endswith("'numpy.ndarray'>"), "Unknown instance type: " + str(type(X))
         assert len(X.shape) == 1 or len(X.shape) == 2, "Instance must have 1 or 2 dimensions!"
 
         # single instance
         if len(X.shape) == 1:
-            explanation = self.explain(X.reshape((1, X.shape[0])), **kwargs)
+            if self.keep_index:
+                data = X.reshape((1, X.shape[0]))
+                data = convert_to_instance_with_index(data, column_name, index_name, index_value)
+            explanation = self.explain(data, **kwargs)
```
slundberg: What if self.keep_index is False? there is no data value...

xzzxxzzx: Thanks for spotting this issue. It might be incurred by final cleanup. sorry for the careless mistake. I'll submitted a patch for this.

PR #73, opened 2018-05-04
update setup.py to bootstrap numpy installation

owlas:
Building the shap package in an empty python environment fails because running the `setup.py` file requires `numpy`.

The following fix was suggested on StackOverflow [here](https://stackoverflow.com/questions/19919905/how-to-bootstrap-numpy-installation-in-setup-py). The `build_ext` class is extended to import numpy and include the headers only after installation. The disadvantage of this approach is that numpy is added to the `setup_requires` list, which builds whenever setup is invocated.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,9 +1,20 @@
 from setuptools import setup, Extension
-import numpy
+from setuptools.command.build_ext import build_ext as _build_ext
 
 # to publish use:
 # > python setup.py sdist upload
 # which depends on ~/.pypirc
+
+# Extend the default build_ext class to bootstrap numpy installation
+# that are needed to build C extensions.
+# see https://stackoverflow.com/questions/19919905/how-to-bootstrap-numpy-installation-in-setup-py
+class build_ext(_build_ext):
+    def finialize_options(self):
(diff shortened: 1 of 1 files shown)


slundberg:
Getting that fixed would be great! seems like the build didn't work yet though on travis.

owlas:
Just going through the logs.

owlas:
Looks like this only works with `python setup.py build_ext --inplace` before running tests. `pip` isn't building the extension. My packaging knowledge is still a mess.

I'll work on a solution

slundberg:
thanks for looking into it!

slundberg:
Looks like it was because`include_dirs=[numpy.get_include()]` was still in the setup function so numpy was still being used before it was loaded. I fixed it and will merge it. Thanks for getting 90% of the way there!

owlas:
@slundberg thanks for the fix. Sorry for letting this hang.

Love the shap project, please let me know if there are any issues I can help with.

slundberg:
Glad it is useful to you! ...just proposing solutions to problems you run into is the most helpful :)

PR #74, opened 2018-05-06
Ability to change plot labels

sbrugman:
Added the ability to change plot labels. Used this to translate the labels, but it could also be used for changing terminology (i.e. `feature` to `independent variable`, depending on your context ([1])).

Approach: the text values are moved from the plot functions into global constants. This technique is called Monkey Patching ([2]). It is a simple pattern to enable to change the texts of the plots without changing the functions.

Example code:
```
import shap

shap.plots.SHAP_MAIN_EFFECT = "Translation for SHAP main effect value for\n%s"
shap.plots.SHAP_INTERACTION_VALUE = "Translation for SHAP interaction value"
shap.plots.SHAP_INTERACTION_EFFECT = "Translation for SHAP interaction value for\n%s and %s"
shap.plots.SHAP_VALUE = "Translation for SHAP value (impact on model output)"
shap.plots.SHAP_VALUE_FOR = "Translation for SHAP value for\n%s"
shap.plots.SHAP_PLOT_FOR = "Translation for SHAP plot for %s"
shap.plots.SHAP_FEATURE = "Translation for Feature %s"
shap.plots.SHAP_FEATURE_VALUE = "Translation for Feature value"
shap.plots.SHAP_FEATURE_VALUE_LOW = "Translation for Low"
shap.plots.SHAP_FEATURE_VALUE_HIGH = "Translation for High"
shap.plots.SHAP_JOINT_VALUE = "Translation for Joint SHAP value"

# plot using shap
```

[1]: https://en.wikipedia.org/wiki/Dependent_and_independent_variables#Statistics_synonyms
[2]: https://en.wikipedia.org/wiki/Monkey_patch

### DIFF
--- a/shap/plots.py
+++ b/shap/plots.py
@@ -132,7 +132,7 @@ def dependence_plot(ind, shap_values, features, feature_names=None, display_feat
         else:
             proj_shap_values = shap_values[:, ind2, :] * 2  # off-diag values are split in half
 
-        # TODO: remove sick ugly recursion
+        # TODO: remove recursion; generally the functions should be shorter for more maintainable code
         dependence_plot(
             ind1, proj_shap_values, features, feature_names=feature_names,
             interaction_index=ind2, display_features=display_features, show=False
--- a/shap/plots.py
+++ b/shap/plots.py
@@ -47,18 +47,19 @@ try:
 except ImportError:
     pass
 
-
-SHAP_MAIN_EFFECT = "SHAP main effect value for\n%s"
-SHAP_INTERACTION_VALUE = "SHAP interaction value"
-SHAP_INTERACTION_EFFECT = "SHAP interaction value for\n%s and %s"
-SHAP_VALUE = "SHAP value (impact on model output)"
-SHAP_VALUE_FOR = "SHAP value for\n%s"
-SHAP_PLOT_FOR = "SHAP plot for %s"
-SHAP_FEATURE = "Feature %s"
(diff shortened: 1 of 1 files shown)


slundberg:
Thanks! I can see why this would be helpful. I wonder if it would make more sense to put all those in a dictionary, then different locales would just be different dictionaries, and there would be less top-level namespace usage in the plots module. Something like `shap.plots.labels = { "main_effect": ... }` What do you think?

sbrugman:
I agree that is a more convenient way of implementing this. Eventually there are better practices for implementing this (via `gettext` I think). Anyway adding this feature to shap could help others until then.

In this implementation the labels are changed with the following code:

```
import shap

shap.plots.labels = {'MAIN_EFFECT': "SHAP value for independent variable\n%s",
    'INTERACTION_VALUE': "SHAP value for variable interaction",
    'INTERACTION_EFFECT': "SHAP value for variable interaction\n%s and %s",
    'VALUE': "SHAP value",
    'VALUE_FOR': "SHAP value for independent variable\n%s",
    'PLOT_FOR': "SHAP value for independent variable %s",
    'FEATURE': "Independent variable %s",
    'FEATURE_VALUE': "Value",
    'FEATURE_VALUE_LOW': "Low",
    'FEATURE_VALUE_HIGH': "High",
    'JOINT_VALUE': "N/A"
}

# shap plots
```

slundberg:
This should be enough for now I think. Thanks again!

PR #87, opened 2018-05-16
Fix pandas 0.23.0 deprecation of as_matrix

jeremiahpslewis:
```
FutureWarning: Method .as_matrix will be removed in a future version.
Use .values instead.
```
https://github.com/pandas-dev/pandas/pull/18458


slundberg:
Thanks!

PR #116, opened 2018-06-15
Add interaction calculations for LightGBM and sklearn trees

aokeson:


shap/explainers/tree.py
```diff
@@ -1,10 +1,11 @@
 import numpy as np
 import multiprocessing
 import sys
+from functools import partial
```
slundberg: This seems like it is not used?

shap/explainers/tree.py
```diff
@@ -1,10 +1,11 @@
...
 
 have_cext = False
 try:
-    from .. import _cext
+    from shap import _cext
```
slundberg: Sounds like it needs to switch back.

shap/explainers/tree.py
```diff
@@ -65,10 +66,12 @@ def __init__(self, model, **kwargs):
             self.trees = model.get_booster()
         elif str(type(model)).endswith("lightgbm.basic.Booster'>"):
             self.model_type = "lightgbm"
-            self.trees = model
+            self.model = model
+            self.trees = [Tree(e,num_trees=len(model.dump_model()["tree_info"])) for e in model.dump_model()["tree_info"]]
```
slundberg: Seems strange for a tree to need to know how many trees there are. Perhaps we could save this info elsewhere? Or just change it to a `rescale` parameter?

shap/explainers/tree.py
```diff
@@ -81,7 +84,7 @@ def __init__(self, model, **kwargs):
         else:
             raise Exception("Model type not yet supported by TreeExplainer: " + str(type(model)))
 
-    def shap_values(self, X, **kwargs):
+    def shap_values(self, X, tree_limit=-1, **kwargs):
```
slundberg, line 90: Great to have this argument. It should probably apply to all the model types though not just XGBoost and LightGBM

shap/explainers/tree.py
```diff
@@ -124,10 +129,9 @@ def shap_values(self, X, **kwargs):
             X = X.values
 
         assert str(type(X)).endswith("'numpy.ndarray'>"), "Unknown instance type: " + str(type(X))
-        assert len(X.shape) == 1 or len(X.shape) == 2, "Instance must have 1 or 2 dimensions!"
-
+        assert len(X.shape) == 1 or len(X.shape) == 2, "Instance must have 1 or 2 dimensions!" + str(len(X.shape))
```
slundberg: Should probably either be removed or be integrated into the text.

shap/explainers/tree.py
```diff
@@ -160,24 +164,69 @@ def shap_values(self, X, **kwargs):
...
+                tree_limit=0
+            phi = self.trees.predict(X, ntree_limit=tree_limit, pred_interactions=True)
             if len(phi.shape) == 4:
                 return [phi[:, i, :, :] for i in range(phi.shape[1])]
             else:
                 return phi
         else:
-            raise Exception("Interaction values not yet supported for model type: " + self.model_type)
+            if str(type(X)).endswith("pandas.core.series.Series'>"):
+                X = X.values
```
slundberg: Down below you assert X has 2 dimensions. So not sure what this does

shap/explainers/tree.py
```diff
@@ -160,24 +164,69 @@ def shap_values(self, X, **kwargs):
...
+            self.tree_shap(t, self._current_X[i,:], self._current_x_missing, phi, condition, condition_feature)
+        phi /= len(self.trees)
+        return phi
+
+    def _tree_shap_ind_interactions(self, i, condition=0, condition_feature=0):
+        phi = np.zeros((self._current_X.shape[1] + 1, self._current_X.shape[1] + 1, self.n_outputs))
+        phi_diag = np.zeros((self._current_X.shape[1] + 1, self.n_outputs))
+        for t in self.trees:
+            self.tree_shap(t, self._current_X[i,:], self._current_x_missing, phi_diag, condition, condition_feature)
+            unique_features = np.unique(t.features)
```
slundberg: seems like this should happen in the tree constructor not here where it happens again and again for every sample

shap/explainers/tree.py
```diff
@@ -199,16 +248,14 @@ def __init__(self, children_left, children_right, children_default, feature, thr
         self.features = feature.astype(np.int32)
         self.thresholds = threshold
         self.values = value
-        self.node_sample_weight = node_sample_weight
```
slundberg: why remove this if it is still and argument?

tests/test_tree.py
```diff
@@ -0,0 +1,19 @@
+import sklearn
+import numpy as np
+import shap
+
+def test_sklearn_interaction():
```
slundberg: There should probably be some kind of test here for the json loading in LightGBM. We can comment it if it makes travis unhappy.

slundberg (commented):
Great overall! I have left my notes.

slundberg:
I wanted to merge this so I didn't create lots of problems for you when I started changing things to support the deep explainer. It still would be good to add some kind of unit test to ensure your parsing code for the JSON is right. Thanks again!

shap/explainers/tree.py
aokeson: This is necessary to rescale `self.values` by the number of trees for lightGBM. Since currently only the current tree is passed into the tree init function, there isn't a great way of having a new parameter other than just passing it to the function.

PR #119, opened 2018-06-18
take external referencing variable as an argument

miaekim:
Issue: function kernel_shap is referencing variable "M" outside of  its function, this may cause NameError
solution: take external referencing variable as an argument


slundberg:
Thanks!

Issue #127, opened 2018-06-21
TreeSHAP for Apache Solr boosted trees json model 


alessandrobenedetti:
Hi all, 
I am a newbie in Shap, so
sorry to disturb!
I am exploring a way to parse a Xgboost model dump to memory to then be able to use it with TreeSHAP.
This is my first approach in explaining the Solr Json format for the boosted tree model (basically converting the Solr format to Xgboost dump and then loading the dump and using plain TreeSHAP).
it is to keep it simple and basically just use what is already there.

In case it is not going to work, I would happy to contribute the Solr Boosted Tree format to the list of models supported by TreeSHAP.
That would require some questions, explanations and discussion with the community to agree on the best approach.
For that case, how do you suggest me to engage? Directly here? 

This is an example of how Apache Solr expects a boosted trees model to be encoded in Json :

https://lucene.apache.org/solr/7_3_0//solr-ltr/org/apache/solr/ltr/model/MultipleAdditiveTreesModel.html

slundberg:
Hey! I personally think directly supporting the Solr Boosted Tree JSON format would be easier than trying to match the XGBoost dump format. A recent PR (#116) added support for loading the JSON format of LightGBM models, so this would be very similar.

The C extension that is part of shap takes a set of parallel arrays describing each tree. These follow the same format as the sklearn tree models. So the main task would be to convert the Solr json representation into the parallel array representation for each tree (which would happen in the Tree constructor in explainers/tree.py).

If you want to give it a shot, I am happy to answer questions along the way.

PR #129, opened 2018-06-22
solr json draft

alessandrobenedetti:
Hi Shap community,
First of all, apologies for the quality of the PR, it is just a draft though.
I am completely new to Python and Shap, so I would like to verify if this is the right direction.
Thank you in advance for your patience ( long description follows)

**Assumption** : Apache Solr is a search engine that implements support for Learning To Rank[1].
It is not a library to train a model, you are expected to train a model with the known libraries, convert the output model to a standard Solr Json format and push it to Solr to be used at re-ranking.

This assumption means that you may be able to run TreeShap **before** you convert the tree model in the Solr Json format ( which is much simpler than the output of machine learning libraries)

The reason for this patch is to enable a developer to run TreeShap from the Solr Json model as well ( let's assume the source model is not available anymore or it is in some complex legacy format and you want to explain it anyway)

I have few questions as well : 

1) 
Do we need this bit for the Solr Json Model as well ? In the tree constructor ?
assert have_cext, "C extension was not built during install!" + str(have_cext)
self.max_depth = _cext.compute_expectations(
    self.children_left, self.children_right, self.node_sample_weight,
    self.values
)

2)  at :  shap/explainers/tree.py:125
I do : phi = self.trees.predict(X, ntree_limit=tree_limit, pred_contribs=True)
This doesn't work, as no predict is available.
Do I need to implement it ( basically navigating the tree) or I can use some auxiliary function already available after you have parsed the model ?

3) In the tree constructor I see a paramter called scaling ( which is initialized for the other tree based model to : scaling=len(tree_info)) which is the number of trees.
In the Solr Json each tree has a weight.
I assumed the weight should go to the scaling param, please correct me if I have been incorrect.

Thank you very much again, I would happy to work more on this contribution and make treeShap available for an easy use for the Solr audience :)

### DIFF
--- a/shap/explainers/tree.py
+++ b/shap/explainers/tree.py
@@ -68,6 +68,11 @@ class TreeExplainer:
             self.model = model
             tree_info = self.model.dump_model()["tree_info"]
             self.trees = [Tree(e, scaling=len(tree_info)) for e in tree_info]
+        elif str(type(model)).endswith("<class 'dict'>"):
+            self.model_type = model["class"]
+            self.model = model
+            trees = self.model["params"]["trees"]
+            self.trees = [Tree(e, self.model["features"], e["weight"]) for e in trees]
         elif str(type(model)).endswith("lightgbm.sklearn.LGBMRegressor'>"):
             self.model_type = "lightgbm"
             self.model = model.booster_
@@ -114,6 +119,10 @@ class TreeExplainer:
             phi = self.trees.predict(X, ntree_limit=tree_limit, pred_contribs=True)
         elif self.model_type == "lightgbm":
             phi = self.model.predict(X, num_iteration=tree_limit, pred_contrib=True)
+            if phi.shape[1] != X.shape[1] + 1:
+                phi = phi.reshape(X.shape[0], phi.shape[1]//(X.shape[1]+1), X.shape[1]+1)
+        elif self.model_type == "org.apache.solr.ltr.model.MultipleAdditiveTreesModel":
+            phi = self.trees.predict(X, ntree_limit=tree_limit, pred_contribs=True)
             if phi.shape[1] != X.shape[1] + 1:
                 phi = phi.reshape(X.shape[0], phi.shape[1]//(X.shape[1]+1), X.shape[1]+1)
         elif self.model_type == "catboost": # thanks to the CatBoost team for implementing this...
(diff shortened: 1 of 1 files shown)


slundberg:
Hey! The general gist is that you need to get 7 parallel arrays just like the sklearn example below (you will just need another similar elif statement). Once you get the information into those 5 arrays then the code inside `_cext` will do the prediction. One thing to double check is that the Solr JSON fomat records the number of training samples that go down each branch of the tree. That is needed to fill in the `node_sample_weight` array. 

1) Yes, this precomputes expectations that are needed.
2) `self.trees.predict` is only called for XGBoost and LightGBM models where Tree SHAP is directly integrated into the packages. You don't need to change the `shap_values` method since it will automatically fall down to the rest of the method that uses the compiled `_cext` module.
3) That's right. The `_cext` module will average the outputs of all trees (like in a random forest), but if you want it to add the trees or weigh trees differently you can just pass a scaling parameter to each tree.

Hope that helps!

```
if str(type(tree)).endswith("'sklearn.tree._tree.Tree'>"):
    self.children_left = tree.children_left.astype(np.int32)
    self.children_right = tree.children_right.astype(np.int32)
    self.children_default = self.children_left # missing values not supported in sklearn
    self.features = tree.feature.astype(np.int32)
    self.thresholds = tree.threshold.astype(np.float64)
    if normalize:
        self.values = (tree.value[:,0,:].T / tree.value[:,0,:].sum(1)).T
    else:
        self.values = tree.value[:,0,:]


    self.node_sample_weight = tree.weighted_n_node_samples.astype(np.float64)

    # we compute the expectations to make sure they follow the SHAP logic
    self.max_depth = _cext.compute_expectations(
        self.children_left, self.children_right, self.node_sample_weight,
        self.values
    )
```

slundberg:
FYI I looked at the example of the Solr JSON you gave in #127 and it does not seem to encode how many training examples went down into each leaf. This is needed for computing SHAP values, since it describes the distribution of the data and so allows variables to be integrated out. Is this info recorded in the file you are trying to convert?

alessandrobenedetti:
Hi Scott,
first of all, thank you very much for your time, it's greatly appreciated.
I revised briefly the contribution and updated the Pull Request.

In regards to your last comment, you are right : 
The additive tree model that Solr takes in input, doesn't have any info about how many training example went down each leaf.
I have not studied yet the SHAP papers ( I will ) but from your comments I understand it is a value necessary to compute the shap values, but can't we use some approximation there, will it make any sense?

I am assuming that Solr users will use TreeShap on the original model ( XgBoost, LightGBM ...) after they training, if it is a format supported by SHAP.

In case the training library model format is not supported by TreeSHAP or they don't have the original model available anymore, they will resort to explore the Solr Json format directly.
Let me know if we can do anything here!

slundberg:
We can recompute the number of training examples if we have a training dataset. But without a training dataset approximating them means approximating the training dataset. Would it make sense in this use-case to assume the user has access to the Solr trees and some example data?

alessandrobenedetti:
That would make sense!
Is it needed anything from the implementation side?
Because checking the examples, when you explain your model you go with validation set already (which you can potentially substitute with the training set): 

_shap_values = shap.TreeExplainer(model).shap_values(Xv)_

_shap.force_plot(shap_values[0,:], Xv.iloc[0,:])_

_shap.summary_plot(shap_values, Xv)_

...
Is it needed anything implementation side to allow for the leaves weight approximation ?

slundberg:
If you are explaining an entire validation set then I agree you could use
that to compute the leaf weights. But often people just explain a single
example, which wouldn't be enough. I think the best approach would be to
build a solr JSON loader function that takes both the solr json and a
dataset to compute leaf weights, then outputs an array of shap Tree
objects. Once you have that, we can hook it into the current api.

alessandrobenedetti:
Hi Scott,
thank you again for your time.
I will implement it using vectorizations as much as possible.
I will update the Pull Request when I have something sensible.
I let you know here.

connortann:
Marking this PR as stale due to the lack of recent activity.

PR #130, opened 2018-06-22
Allow ExtraTreesClassifier

saaryalov:
Added elif statement to allow ExtraTreesClassifier. Passed basic test, I can provide more if needed.


slundberg:
Thanks! I'll merge this now with a note that we still need to make a unit test for that case.

PR #131, opened 2018-06-24
expose Tree object + fix breaking test of kernel shap for dataframe input

xzzxxzzx:
Change 1: expose Tree object so that I can run tree shap with my tree based model

Change 2: recent change in kernel shap broke my local test for the case of pandas dataframe input. Here I submit a proposed change along with my test case.

Please feel free to let me know if you have any concerns.
Thanks!

### DIFF
--- a/shap/__init__.py
+++ b/shap/__init__.py
@@ -6,7 +6,7 @@ from iml.links import Link, IdentityLink, LogitLink
 from iml.common import Instance, Model
 from .explainers.kernel import KernelExplainer, kmeans
 from .explainers.sampling import SamplingExplainer
-from .explainers.tree import TreeExplainer
+from .explainers.tree import TreeExplainer, Tree
 from .plots import visualize, plot, summary_plot, joint_plot, interaction_plot, dependence_plot, force_plot
 from iml.visualizers import initjs, SimpleListVisualizer, SimpleListVisualizer, AdditiveForceVisualizer, \
     AdditiveForceArrayVisualizer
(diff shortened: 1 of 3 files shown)


xzzxxzzx:
Hi Scott,

Looks like there's some import issue causing AppVeyor build to fail: pkg_resources.DistributionNotFound: The 'xgboost' distribution was not found and is required by the application. Any suggestion on what I should try here?

Thanks!

slundberg:
That's actually expected since XGBoost doesn't build from pip on windows. It looks like real problem is scipy failing to load some of it's modules. Windows builds have been flaky (because dependency packages like scipy randomly fail) so I would ignore it for now.

slundberg:
Thanks for the new test! I think I actually just fixed that in a different way but hadn't pushed it yet (sorry to have broken it):

```
return AdditiveExplanation(
    self.link.f(self.fnull if self.vector_out else self.fnull[0]),
    self.link.f(self.fx if self.vector_out else self.fx[0]),
    phi, phi_var, instance, self.link, self.model, self.data
)
```
(and removing the unwrap inside `if not self.vector_out` for both fx and fnull.

I am not sure about the desire to consider functions that output a vector with one element as "not vector", could you explain more your thoughts there? Thanks again.

xzzxxzzx:
Thanks for the quick response, Scott! The "not vector" change is also part of the fix to the case of pandas dataframe input. If your fix can pass my test, we can ignore this change. 

Now should I wait for your fix and then push again? or you want to include the __init__.py change to expose Tree + include the new test case and I can cancel this push request?

Thanks!

slundberg:
Just leave this PR and I'll edit it before merging and after I push my changes. You unit test seems to pass fine on my local copy.

PR #139, opened 2018-07-03
add args to keep index sorted -- required by some models

xzzxxzzx:
Because some of my models requires the DataFrame input to be ordered by index, the model reference in shap will break for these models. Here I added an argument (default to False) to allow sorting DataFrame by index after sampling. Please feel free to let me know if you have any questions/concerns.

Thanks!


slundberg:
Looks good. Thanks!

PR #161, opened 2018-07-16
Indexation bug on one sample cases fix

mizukasai:
Hello,

I noticed that when we want to explain one sample using `shap_values` or `shap_interactions_values` there is an indexation error "too many indexes error" 

This should fix it 

Cheers


slundberg:
Thanks! I'll also add a unit test to ensure this stays working :)

PR #177, opened 2018-07-24
fix the error found in kernel shap tests with dataframe input and fix test case

xzzxxzzx:
Sorry my original test case didn't turn on the keep_index flag so the test isn't done correctly. This change should be able to fix the test for kernel shap on dataframe input. 
Thanks!

### DIFF
--- a/shap/explainers/kernel.py
+++ b/shap/explainers/kernel.py
@@ -117,7 +117,7 @@ class KernelExplainer(object):
         else:
             model_null = self.model.f(self.data.data)
         if isinstance(model_null, (pd.DataFrame, pd.Series)):
-            model_null = model_null.values
+            model_null = np.squeeze(model_null.values)
         self.fnull = np.sum((model_null.T * self.data.weights).T, 0)
 
         # see if we have a vector output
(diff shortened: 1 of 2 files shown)


shap/explainers/kernel.py
```diff
@@ -117,7 +117,7 @@ def __init__(self, model, data, link=IdentityLink(), **kwargs):
         else:
             model_null = self.model.f(self.data.data)
         if isinstance(model_null, (pd.DataFrame, pd.Series)):
-            model_null = model_null.values
+            model_null = np.squeeze(model_null.values)
```
slundberg, line 120: Could you clarify what this change is designed to fix?

xzzxxzzx: Sorry for the confusing comments earlier, Scott. Looks like after earlier refactoring of kernel.py, for models with dataframe output, one extra dimension is somehow added, which eventually makes the resulting dimension of shap value different. This change fix the dimension issue and will only impact the specific case with dataframe output.

thanks.

slundberg: makes sense. Thanks!

PR #184, opened 2018-07-27
TreeExplainer.shap_values can take catboost.Pool as X argument.

PashaSmirnov:
Hello, I am a developer for catboost. This small fix allows TreeExplainer.shap_values to take catboost.Pool as X argument. Catboost works with categorical features, but current interface doesn't allow to provide categorical features description for dataset.


slundberg:
Great, thanks!

PR #196, opened 2018-08-01
enable summary_plot to return the figure if show=False

zkurtz:
I was trying to access the underlying pyplot figure so that I could export it to a .png. This solved my problem. Is there an even easier way?

### DIFF
--- a/shap/plots.py
+++ b/shap/plots.py
@@ -329,6 +329,11 @@ def summary_plot(shap_values, features=None, feature_names=None, max_display=Non
 
     plot_type : "dot" (default) or "violin"
         What type of summary plot to produce
+
+    show : `True` (default) or `False`
+        If `True`, display the plot in the notebook or interactive session. If `False`, return
+        the pyplot object without showing it, allowing the user to modify the plot and/or
+        write it to a png file, for example.
     """
 
     multi_class = False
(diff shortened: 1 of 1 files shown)


chrisadas:
I was just thinking about the same thing.

https://github.com/slundberg/shap/issues/197

slundberg:
Hmm...is just calling `pl.savefig(filename)` right after calling with `show=False` not working for you? You should (I think) be able to just call `pl.gcf()` right after the plot call if `show=False` was passed.

zkurtz:
At the moment I'm getting bizarre behavior regardless of which version I use - I have not been able to get your method to work, but may follow up later. Regardless, one argument in favor of passing back the figure explicitly is that the recipient need not have imported `pl` to access and use it.

slundberg:
I think I buy that. However I apologize that a recent refactor means that the update now needs to go in `plots/summary.py`. I am happy to copy it over there it if you prefer, but if you copy it over then github will correctly record you as the author.

zkurtz:
Thanks for checking in, but I'm going to pass this back to you as I don't think the solution I proposed is ultimately satisfactory. In the context of a notebook, I'm observing that the plot gets displayed regardless of whether show = True or False. Specifically, deleting every `pl.show()` command seems to have no effect in a notebook. (It works as expected, on the other hand, in a regular non-interactive script). I'm wondering if it's necessary to define a `fig` variable at the outset of creating the plot rather than picking it up at the last minute (as `pl.gcf()`) to really get full control.

slundberg:
Okay. If you end up with a solution let us know :)

Issue #138, opened 2018-06-30
Support for csr sparse matrices.


asutosh97:
Currently, KernelExplainer only supports pandas dataframe and numpy matrices. It should be extended to support `scipy.sparse.csr.csr_matrix`

slundberg:
This would be good. It will take some work though so I can't promise any time line. You could work around it by selecting just the non-zero entries for the background and the instance you are explaining and passing those as a dense vector. This would effectively assume the sparse entries will have zero SHAP values.

asutosh97:
Okay, cool. I'll give that a try. 
I also have one more doubt. Let's say I have a text data featured as BoW or TF-IDF. So, I have a separate list of mappings of feature number to the word that it corresponds to. Is there any way that I can pass this to get SHAP values of those words?

slundberg:
Not sure I fully understand the question. You have SHAP values for the features represented as a number and you want to get the SHAP value for the words? If there is a one-to-one mapping then SHAP values are the same.

psinger:
Having sparse functionality would be extremely valuable. At the moment I run in a lot of bottlenecks working with larger datasets in TreeExplainer, e.g., with BOW models.

slundberg:
@psinger yeah, that would be really nice. The core part would require updating the code `tree_shap.h` to handle a sparse feature vector and return a sparse attribution vector.

I should note that one way to mitigate this would be to find out what features the tree model actually depends on and make sure those are the only ones in your input dataset (i.e. prune unused words).

psinger:
@slundberg Wouldn't it be already enough to return a sparse matrix in `shap_values` and then update the plotting function accordingly?

zmabzug:
+1 for sparse support for TreeExplainer!

slundberg:
It depends on if we are talking about the XGBoost LightGBM or sklearn implementations. the XGBoost and LightGBM currently are embedded in those packages. So the whole phi matrix is returned before we have the chance to make it sparse. That means the C++ implementations should be updated. You are right that if we could just import the XGBoost models and use the tree_shap.h implementation we could do dense rows but make a sparse matrix. I don't know if building a sparse matrix in python/numpy will be fast or not. Could be worth a try :) ...I can't pursue it at this moment though, so if anyone else wants to take a shot at it that would be great!

psinger:
[...]

PR #200, opened 2018-08-02
Adding CSR support for KernelExplainer

imatiach-msft:
Adding support for sparse matrix in scipy CSR format to KernelExplainer in shap library
This is to fix issue #138


imatiach-msft:
Maybe instead of including the file, which is quite large, I should download it - I've created a zip here: 
https://publictestdatasets.blob.core.windows.net/data/datasets.zip
Also, there's some code to generate a much more sparse dataset, but currently it takes too long to evaluate it:

        remove = ('headers', 'footers', 'quotes')
        categories = [
            'alt.atheism',
            'talk.religion.misc',
            'comp.graphics',
            'sci.space',
        ]
        from sklearn.datasets import fetch_20newsgroups
        ngroups = fetch_20newsgroups(subset='train', categories=categories,
                               shuffle=True, random_state=42,
                               remove=remove)
        x_train, x_test, y_train, y_validation = train_test_split(ngroups.data, ngroups.target,
                                                                        test_size=0.001, random_state=42)
        from sklearn.feature_extraction.text import HashingVectorizer
        vectorizer = HashingVectorizer(stop_words='english', alternate_sign=False,
                                       n_features=2**16)
        x_train = vectorizer.transform(x_train)
        x_test = vectorizer.transform(x_test)

One way to boost performance might be to do count based feature selection and remove the columns that are both zero in the training set and the test example when going over each example (and modify the background dataset for each example) and assign a feature importance of zero for sparse features.  That may be one way to get around the high dimensionality of hashed features in that specific example.

imatiach-msft:
I noticed kmeans is returning a dense matrix instead of sparse when using shap kmeans method.  I've created a ticket and a PR here:
https://github.com/scikit-learn/scikit-learn/pull/11787
https://github.com/scikit-learn/scikit-learn/issues/11786

I'm not sure if using kmeans on sparse data is a good idea though, or what would be the best way to summarize the data.  Usually the clusters are not very good when the data is very sparse.

shap/explainers/kernel.py
```diff
@@ -370,7 +374,13 @@ def varying_groups(self, x):
         varying = np.zeros(len(self.data.groups))
         for i in range(0, len(self.data.groups)):
             inds = self.data.groups[i]
-            num_mismatches = np.sum(np.abs(x[0, inds] - self.data.data[:, inds]) > 1e-7)
+            x_group = x[0, inds]
+            if sp.sparse.issparse(x_group):
+                if i not in x.nonzero()[1]:
```
slundberg: `i` is a group index not an index into `x` (that is what `inds` is) so while they are the same when there is no grouping this won't work when some features are grouped.

slundberg: Also we need to make sure both the background data and the current x are 0 otherwise the feature does vary.

slundberg:
With sparse data it seems like just passing a background of all zeros would
make sense (i.e. just the median value of all features).

shap/explainers/kernel.py
imatiach-msft: I wonder if in the sparse case there should be some threshold for the background?  If the background is mostly zero but only has a single nonzero, and the feature in the evaluated sample is zero, then the feature importance will be nonzero but also very low, since most of the time we will be choosing zeros from the background.  To optimize execution time maybe we should just ignore those columns too?

imatiach-msft:
There could be weird cases where some columns could be dense but the majority are sparse.  That is actually the dataset one of our internal customers had.  Ideally I think you would have something like a table representation where some columns are dense numpy arrays, others sparse, etc.  But it seems most frameworks just have a single csr matrix and include the dense columns in the matrix.

shap/explainers/kernel.py
slundberg: I think if the user has decided to pass many background values then they have made the choice to spend lots of time comparing to many samples, so we shouldn't short-circuit that when the input matrix happens to be in the sparse format. That being said, I think for sparse data people should mostly just pass a background with one sample of all 0's since that is by far the most common value.

slundberg:
If only some columns are dense then just passing a background of the median sample value should still work. If you want more background samples for the dense columns you could always just zero-out the sparse feature columns in the background. Then all the sparse features will get pruned during the explanation process.

shap/explainers/kernel.py
imatiach-msft: great idea, will update the code, thanks!

imatiach-msft: done, changed to use inds instead of i when checking if group varies

imatiach-msft:
@slundberg I've updated the PR based on comments.  I've also added an additional test for very sparse case where we use feature hashing prior to shap (65k columns).  I modified the code that generates the background and the output scores from the background to use sparse matrix instead of dense and I optimized the code that adds/creates samples.  I did not implement grouping for sparse case, this is something I could add in a future PR.  I noticed that the biggest performance bottleneck in the code was addsample, and when varyingFeatureGroups was not jagged (eg all groups had same size or size of 1) I could get a large performance boost by using numpy arrays instead.

slundberg:
Thanks @imatiach-msft! That sounds great. It looks good from what I can tell, but I'll go over it more carefully and try and get it merged tomorrow morning after testing it out a bit more.

The speed up when we don't use groups is great idea!

slundberg:
Thanks again @imatiach-msft! Sorry it took a few days to review, but this is a big improvement that I'll tag in v0.23 sometime today :)

imatiach-msft:
@slundberg thank you for the review!  I noticed the url for the a1a sparse dataset needed to be fixed now that the PR is checked in, I created a PR here: https://github.com/slundberg/shap/pull/223
I can also change the code to put the test file in a zip and download if you prefer not to have large files checked into the repo.

suzinyou:
Hello, I have a question about the sparse case for the varying group function. 

When my query data into `shap_values` is sparse, in [shap/explainers/kernel.py L432](https://github.com/slundberg/shap/blob/47c6a7d76696a1f2fd2b333e594ac5ac0f30d358/shap/explainers/kernel.py#L432), "`if nonzero_rows:`" tries to get a truth value of an array (with more than one element). 

I suspect we should be checking for the length of the `nonzero_rows` to see if any entries are nonzero, and add the column to `remove_unvarying_indices` only if `num_mismatches == 0` rather than `> 0` to indicate that the column does not vary.

Is this correct? Please correct me if I'm not understanding it right. @imatiach-msft Thanks a lot in advance!

imatiach-msft:
@suzinyou thank you for finding this issue, I think you are right.  I will debug and verify that this is indeed an issue and send out a fix.

imatiach-msft:
@suzinyou I think I know why I never noticed this issue - we always pass the background as sparse 0's currently, which you would usually do for sparse data.  This is why I never noticed it.  This is background data: https://github.com/slundberg/shap/blob/47c6a7d76696a1f2fd2b333e594ac5ac0f30d358/shap/explainers/kernel.py#L430
so if it is all zeros, which is almost always the case (in fact in my code I always pass zeros currently to shap from [my package](https://pypi.org/project/azureml-explain-model/)), the bug will never appear.  I will send out a fix soon.  This is an edge case that I never hit in my testing on sparse case.

imatiach-msft:
@suzinyou PR has been checked in with fix & validation test case, thank you for finding the bug!

kjanko:
Any updates on this PR?

suzinyou:
> Any updates on this PR?

@kjanko 
If you're talking about the issue I raised, the relevant PR is here: https://github.com/slundberg/shap/pull/319 (it's been merged for a while)

kjanko:
@suzinyou ah okay, thanks for the link. I was trying to use shap with a dok_matrix - didn't notice the PR is for CSR support only. My apologies.

PR #207, opened 2018-08-08
fixed benchmark import error

mrandrewandrade:
Hey @slundberg ,

When you go to import shap you get an error:

```
  File python2.7/site-packages/shap/__init__.py", line 14, in <module>
    from . import benchmark
ImportError: cannot import name benchmark

```

This also causes the conda-forge CI to fail:

https://circleci.com/gh/conda-forge/staged-recipes/29466?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

I made the change in the PR and tested installing locally so hopefully this should be g2g!

### DIFF
--- a/shap/__init__.py
+++ b/shap/__init__.py
@@ -11,5 +11,5 @@ from .plots.dependence import dependence_plot
 from .plots.force import force_plot, initjs
 from .plots.image import image_plot
 from . import datasets
-from . import benchmark
+import benchmark
 from .explainers import other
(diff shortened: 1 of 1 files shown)


slundberg:
Thanks! Unfortunately the source of the problem is not the import format but the fact that the benchmark directory was not in the `packages` list in setup.py. It works locally because a git clone doesn't have the problem.

PR #216, opened 2018-08-13
add args to allow customized tree model

xzzxxzzx:
add args to allow customized tree model. The self.trees will be set by logic outside tree.py.

### DIFF
--- a/shap/explainers/tree.py
+++ b/shap/explainers/tree.py
@@ -51,8 +51,10 @@ class TreeExplainer(Explainer):
         self.less_than_or_equal = False # are threshold comparisons < or <= for this model
         self.base_offset = 0.0
         self.expected_value = None
-        self.trees = None
-        self.customized_tree = kwargs.get('customized_tree', False)
+        self.trees = kwargs.get('trees', None)
+
+        if self.trees is not None:
+            return
 
         # parse all the different possible supported model types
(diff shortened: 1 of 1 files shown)


slundberg:
I can see why you would want to have customized trees. Would it work to just add support for passing an array of Tree objects? That would match the interface of the other model types.

xzzxxzzx:
Yes, that would work too! Do you want to add it, or should I?

slundberg:
Go for it!

slundberg:
Thanks! Why not just pass the array of trees as the `model` parameter? It is the model, so that will keep it consistent with the other model types.

xzzxxzzx:
Hi Scott, sorry for the late reply. I was out on vacation. I understand your concern about the consistency. In many cases, I have no control of/access to the model implementation, so just like what you did with sklearn, I get the array of trees and pass it into TreeExplainer together with the model. Maybe there's a better way to handle this? please feel free to let me know if you have any suggestions :)

slundberg:
Right, it all looks good to me. The one change for the sake of consistency would be to not have a new keyword argument called `trees` but rather just check if the model object that was passed is an array of Tree objects. Then you can just pass that array of tree objects as though it was the the model (which makes sense because it is the model).

slundberg:
I just changed the code to accept the list of trees directly as the model object and will merge once the unit tests pass.

PR #236, opened 2018-08-23
improve performance for sparse matrix input in kernel explainer

imatiach-msft:
Improve performance for sparse matrix input in kernel explainer - by converting the background dataset to lil and the examples to lil format the performance was improved by 10X-20X.
Also, this PR changes the interface to allow any sparse matrix to be accepted.

### DIFF
--- a/shap/common.py
+++ b/shap/common.py
@@ -1,6 +1,7 @@
 import re
 import pandas as pd
 import numpy as np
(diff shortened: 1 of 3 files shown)


slundberg:
That's great, thanks Ilya! This will land in 0.23.2

PR #239, opened 2018-08-26
Added support for matplotlib additive plot

jverre:
Hi,

I have started porting the Additive visualizer code from JS to matplotlib (https://github.com/slundberg/shap/issues/27). It is in a state where the code currently works, at least on the examples I was able to test on. However some functionality is not supported: custom color maps for example.

As it is still a draft, the code can be improved by increasing modularity and test more edge cases.

To plot a using matplotlib instead of JS, use:
`shap.force_plot(explainer.expected_value, shap_values[i], X_test.iloc[i], link='logit', matplotlib=True, show=True)`
instead of:
`shap.force_plot(explainer.expected_value, shap_values[i], X_test.iloc[i], link='logit')`

### DIFF
--- a/shap/plots/force.py
+++ b/shap/plots/force.py
@@ -17,9 +17,10 @@ import warnings
 import re
 from . import labels
 from ..common import convert_to_link, Instance, Model, Data, DenseData, Link
+from ..plots.force_matplotlib import draw_additive_plot
 
 def force_plot(base_value, shap_values, features=None, feature_names=None, out_names=None, link="identity",
-               plot_cmap="RdBu"):
+               plot_cmap="RdBu", matplotlib=False, show=True, figsize=(20,3)):
     """ Visualize the given SHAP values with an additive force layout. """
 
     # auto unwrap the base_value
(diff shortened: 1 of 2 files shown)


slundberg:
Thanks!! This is a great feature to have. I'll review as soon as I can (which isn't today unfortunately).

slundberg:
Looking through this is doesn't change any current functionality, so I am going to merge it for now and I'll suggest any improvements once I have time to test it out fully.

slundberg:
Thanks again for adding this :)

I had a chance to try it out, and overall it looks great! A few things it would be good to address at some point (either you or I or someone else):
 - #275 seems like a simple fix
 - I added an error check if people try to use matplotlib = True when multiple samples are passed.
 - There a few differences from the JS version would be good to adjust for the sake of consistency:
   - Base value doesn't fall in the center of the plot.
   - labels are not centered under a bar segment.
   - font sizes seem a bit smaller
   - some minor spacing differences and line fade effects

If you have any thoughts on these ideas let me know!

Issue #191, opened 2018-07-30
Question: any plans on adding pyTorch compatibility for this tool?


shgidi:

slundberg:
I have thought about it. It would not be too hard I think for the GradientExplainer, but for DeepExplainer it would require building a custom back propagation graph just like we do for TensorFlow. If anyone is familiar with PyTorch models (I don't use them often) and willing to give either one a shot, that would be great!

connortann:
Closed by #262

PR #250, opened 2018-09-01
PyTorch Gradient Explainer

gabrieltseng:
This is a partial solution to issue #191 , which is a PyTorch version of the gradient explainer. I would appreciate any feedback.

Something to note is that because the network is not defined before runtime in PyTorch, its much harder to see the results of interim layers as happens here:
https://github.com/slundberg/shap/blob/master/notebooks/gradient_explainer/Explain%20an%20Intermediate%20Layer%20of%20VGG16%20on%20ImageNet.ipynb

Doing so would probably require a network to be split into two by the user.

I've also added a notebook demonstrating the changes; the results are less convincing than the intermediate layers, so @slundberg please let me know if they still make sense to you, or are indicative of a mistake.


shap/explainers/gradient.py
```diff
@@ -245,3 +267,119 @@ def run(self, out, model_inputs, X):
...
+        if not self.multi_input:
+            assert type(X) != list, "Expected a single tensor model input!"
+            X = [X]
+        else:
+            assert type(X) == list, "Expected a list of model inputs!"
+        # assert len(self.model_inputs) == len(X), "Number of model inputs does not match the number given!"
+
+        # rank and determine the model outputs that we will explain
+        with torch.no_grad():
+            model_output_values = self.model(*X)
```
slundberg: If we are not sorting then we could skip this forward pass. So perhaps just move it inside the if statement on the next line?

slundberg:
Thanks @GabrielTseng!! This is great!

The MNIST test seems to imply it is working, but you are right the VGG16 example in the notebook isn't reassuring. Data with strong input correlations like natural images can cause problems in the lower layers (something I hope we can solve in the future) but it would be good to make sure we get nice results for something. Could you also run the MNIST example from the front page and make sure you get sensible explanations there?

A minor point is that we should be able to auto-detect the framework based on the passed model object. But that's not something that needs to happen before we merge.

slundberg:
Also, I see why explaining a middle layer is harder. I am not a pytorch expert, but it seems like once you run a forward pass you should have a fully defined computation graph. If you trace that back to a set of intermediate inputs perhaps you could grab the gradients and forward values flowing through them (which is all we need). But that's something to worry about after merging.

gabrieltseng:
Hi @slundberg , thank you for your feedback!

I ran the MNIST example, and things seem a lot more reasonable; I replaced the Imagenet notebook with it.

I'll definitely take a look at getting intermediate outputs and autodetecting the framework.

slundberg:
Great! It seems like the intermediate outputs and framework auto-detection shouldn't change the API. So unless you would rather address those first, I can go ahead and merge what is here already (since it is all new functionality).

gabrieltseng:
That sounds good to me!

slundberg:
Great. I'll merge this and then once things have had a chance to settle out a bit we can add a note to the Readme.

Issue #275, opened 2018-10-01
force_matplotlib bug


omarelb:
https://github.com/slundberg/shap/blob/ed284b6278813c5292d83dc2a22976a0fdedd4ec/shap/plots/force_matplotlib.py#L221-L238

When creating an additive matplotlib plot for a single observation, the function crashes when either pos_features or neg_features is empty, which happens when all shapley values for a given observation are either negative or positive. It crashes because of 
```python

total_neg = np.max(neg_features[:, 0].astype(float)) - \
                    np.min(neg_features[:, 0].astype(float)
```
because neg_features and pos_features cannot be sliced if they are empty. I'm working around this with this code
```python
if len(neg_features) > 0:
        total_neg = np.max(neg_features[:, 0].astype(float)) - \
                    np.min(neg_features[:, 0].astype(
else:
        total_neg = 0
```
, similarly for pos_features, though I'm not too sure if that's the right way to handle it.

The same error is present in `update_axis_limits`, where i've replaced the code at 
https://github.com/slundberg/shap/blob/ed284b6278813c5292d83dc2a22976a0fdedd4ec/shap/plots/force_matplotlib.py#L299-L300 with
```python
    if len(pos_features) > 0:
        min_x = min(np.min(pos_features[:, 0].astype(float)), base_value) - padding
    else:
        min_x = 0
    if len(neg_features) > 0:
        max_x = max(np.max(neg_features[:, 0].astype(float)), base_value) + padding
    else:
        min_x = 0
    ax.set_xlim(min_x, max_x)
```

slundberg:
Thanks for pointing this out! @jverre kindly added the matplotlib version of force_plot very recently, and it looks like there are still a few issues to work out. Would you mind submitting this fix as a PR? Thanks!

PR #277, opened 2018-10-03
fixed bug in force_matplotlib

omarelb:
Fixed issue with force_matplotlib as described in #275.


slundberg:
Thanks! Since this is definite step forward I am going to merge it for now.

PR #290, opened 2018-10-18
Added support for Independent TreeSHAP

HughChen:
- Wrote a linear time implementation of TreeSHAP that computes Shapley values with **independence** between the conditioned and non-conditioned sets.  
    - This implementation works for a single reference, a single sample, and a single tree.  
    - The overall runtime is O(NLT), where N is the size of the reference set, L is the number of leaves in the tree, and T is the number of trees.
    - Wrote a c++ implementation which works approximately 100 times faster than the python implementation.
    - Added support for non-linear functions of the output, in order to support decomposing loss functions to particular features.
- Added an implementation for xgboost booster models.  In addition, wrote up an implementation to create tree.py Tree objects from the string dump of the booster models.
- Added tests to test out independent TreeSHAP
    - Verify that Independent TreeSHAP gives the same results as Kernel SHAP for simple models (which makes the same independence assumptions).
    - Verify that the output sums up correctly for models with 20 trees of max_depth 6.
    - Verify that the output sums up correctly for computing the attributions of mse, logistic, and logloss.

### DIFF
--- a/shap/explainers/tree.py
+++ b/shap/explainers/tree.py
@@ -46,8 +46,7 @@ class TreeExplainer(Explainer):
     implementations either inside the externel model package or in the local compiled C extention.
     """
 
-    def __init__(self, model, feature_dependence = "tree_path_dependence", 
-                 model_output = "margin", data = None, **kwargs):
+    def __init__(self, model, data = None, model_output = "margin", feature_dependence = "tree_path_dependence", **kwargs):
         """ 
         Parameters
         ----------
--- a/shap/explainers/tree.py
+++ b/shap/explainers/tree.py
@@ -46,14 +46,14 @@ class TreeExplainer(Explainer):
     implementations either inside the externel model package or in the local compiled C extention.
     """
 
-    def __init__(self, model, direct = False, output_type = "margin", 
-                 ref_X = None, **kwargs):
+    def __init__(self, model, feature_dependence = "per_tree_dependent", 
+                 model_output = "margin", ref_X = None, **kwargs):
         """ 
         Parameters
         ----------
         model : Several sklearn, xgboost, lightbgm, and catboost models are supported.
-        direct : Denotes to directly generate Trees from the xgboost booster model.
-        output_type : Can be "margin", "mse", "logistic", or "logloss".
+        feature_dependence : Can be "per_tree_dependent" or "independent".
+        model_output : Can be "margin", "mse", "logistic", or "logloss".
         """        
         self.model_type = "internal"
         self.less_than_or_equal = False # are threshold comparisons < or <= for this model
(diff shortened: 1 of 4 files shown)


shap/explainers/tree.py
```diff
@@ -46,12 +46,22 @@ class TreeExplainer(Explainer):
...
 
-    def __init__(self, model, **kwargs):
+    def __init__(self, model, direct = False, output_type = "margin", 
+                 ref_X = None, **kwargs):
+        """ 
+        Parameters
+        ----------
+        model : Several sklearn, xgboost, lightbgm, and catboost models are supported.
+        direct : Denotes to directly generate Trees from the xgboost booster model.
+        output_type : Can be "margin", "mse", "logistic", or "logloss".
```
slundberg: I think `model_output` would be a clearer name that highlights how this is connected to the model. Also we need to add checks that flag a problem if someone calls something other than "margin" when doing non-independent tree shap (since it's not supported then).

shap/explainers/tree.py
```diff
@@ -46,12 +46,22 @@ class TreeExplainer(Explainer):
     implementations either inside the externel model package or in the local compiled C extention.
     """
 
-    def __init__(self, model, **kwargs):
+    def __init__(self, model, direct = False, output_type = "margin",
```
slundberg: What's the point of direct here? It seems like a user shouldn't need to care how we parse the model.

slundberg: We should add an option here about what kind of dependence assumptions we are making: `feature_dependence=independent`, `feature_dependence=global_tree_dependent`, or `feature_dependence=per_tree_dependent`

shap/explainers/tree.py
```diff
@@ -97,8 +107,17 @@ def __init__(self, model, **kwargs):
             self.trees = [Tree(e.tree_, scaling=scale) for e in model.estimators_[:,0]]
             self.less_than_or_equal = True
         elif str(type(model)).endswith("xgboost.core.Booster'>"):
-            self.model_type = "xgboost"
-            self.trees = model
+            if (not direct):
```
slundberg: This should now be done based on the `feature_dependence` option.

shap/explainers/tree.py
```diff
@@ -366,6 +385,150 @@ def recurse(node):
 
         recurse(0)
 
+    def gen_trees(self, model):
```
slundberg, line 396: All this stuff should be in the Tree Object and based on the JSON parsing but I can change that after merging since I already have that implemented.

shap/explainers/tree.py
```diff
@@ -366,6 +385,150 @@ def recurse(node):
...
+                features[key] = feature
+                thresholds[key] = threshold
+                node_sample_weights[key] = node_sample_weight
+        tree_dict = {"children_left":children_left, "children_right":children_right,
+                     "children_default":children_default, "feature":features,
+                     "threshold":thresholds, "value":values[:,np.newaxis],
+                     "node_sample_weight":node_sample_weights}
+        return(Tree(tree_dict))
+
+    def independent_treeshap(self,x,y=None):
```
slundberg, line 481: we won't expect users to call this directly since with the `feature_dependence` option in the constructor the `shap_values` method will know to call this when we are assuming independence.

slundberg:
We will also need to figure out why travis is failing.

PR #302, opened 2018-10-27
Added support for missing values and optimizations to independent treeshap

HughChen:
- Added support for missing values (encoded as np.nans) with a check in independent treeshap.
- Added optimizations to c++ code underlying independent treeshap.
  - Pulled memory allocations out for various arrays (pos_lst, neg_lst, node_stack, feat_hist).
  - Stopped zeroing out pos_lst and neg_lst.
  - Added memoized calculation of weights.
  - Made a mytree array that contains all the state we need at a particular node (Nodes have shorts and floats in order to keep the size of the elements small).
  - Removed parent calculations and moved out some checks.


slundberg:
Great! I'll push my updates to the code after this merge.

PR #319, opened 2018-11-09
fix sparse nonzero background case in shap and add validation test

imatiach-msft:
This fixes the issue found by @suzinyou here:
https://github.com/slundberg/shap/pull/200
When adding a test, I found several additional problems that made the sparse result differ from dense when there were non-zero values in the background dataset.  I never noticed this issue previously because I've only tested the case with sparse all-zero background values.
I also manually validated locally by printing the varying indices returned by the [varying groups method](https://github.com/slundberg/shap/blob/169fe9d889308fb1266ae1a42f36eea6d5c1d3c6/shap/explainers/kernel.py#L408) in both the sparse and dense case and manually comparing the two to ensure that they have the exact same indices.  This was partly how I was able to find and resolve the other issues.


slundberg:
Thanks!

Issue #301, opened 2018-10-27
Possible Compatibility Issues in PyTorchDeepExplainer (tested in PyTorch 1.0 Preview)


ceshine:
Hi, 

I wasn't able to run PyTorchDeepExplainer in my environment, so I followed the error messages and made some changes to it. [Here are the changes](https://github.com/ceshine/shap/commit/d7ad868c9e64ef4ddfb4d746482b4bbc287f2625). (The model I trained has a single regression output.)

It's very likely that I did not fix the problem correctly, hence no PR, just an issue here as FYI. 

[Example (Kaggle Kernel)](https://www.kaggle.com/ceshine/feature-importance-from-a-pytorch-model/notebook)

The error messages were (paraphrasing):

- PyTorchDeepExplainer does not have the property `num_outputs`.
- Cannot multiply a PyTorch tensor by a numpy array.

austinmw:
Were you able to get `PyTorch Deep Explainer MNIST example.ipynb` working with PyTorch 1.0, and if so can you please share?

ceshine:
Yes, I was able to get it working with both PyTorch 1.0 and 0.4.1. 

* 0.4.1 (Kaggle kernel): https://www.kaggle.com/ceshine/pytorch-deep-explainer-mnist-example
* 1.0.0 (Github gist): https://gist.github.com/ceshine/37dcc304f22b0ad40b01e47c70f555ec

The version of SHAP package (0.24.0) that comes with Kaggle Kernel does not work, though: https://www.kaggle.com/ceshine/pytorch-deep-explainer-mnist-example-unmodified

slundberg:
That's because PyTorch support is new in 0.25 which I am working to get pushed soon.

ceshine:
Yes. The output "Using TensorFlow backend." from the 0.24.0 kernel kinda gave it away. I was trying to demonstrate a possible scenario of failure, as @austinmw did not mention whether he has successfully got the notebook runing with PyTorch 0.4.1 or not. Sorry for not being clear enough. [Here's a notebook](https://www.kaggle.com/ceshine/pytorch-deep-explainer-mnist-example-latest) confirming the notebook works with a install from the official master branch.

[...]

PR #321, opened 2018-11-10
Ensure PyTorch Deep Explainer can handle single output models

gabrieltseng:
A fix to #301

### DIFF
--- a/shap/explainers/deep/deep_pytorch.py
+++ b/shap/explainers/deep/deep_pytorch.py
@@ -26,6 +26,7 @@ class PyTorchDeepExplainer(Explainer):
         self.input_handle = None
         self.interim = False
         self.interim_inputs_shape = None
+        self.expected_value = None  # to keep the DeepExplainer base happy
         if type(model) == tuple:
             self.interim = True
             model, layer = model
@@ -48,6 +49,7 @@ class PyTorchDeepExplainer(Explainer):
         self.model = model.eval()
 
         self.multi_output = False
+        self.num_outputs = 1
         with torch.no_grad():
             outputs = model(*data)
             if outputs.shape[1] > 1:
--- a/tests/explainers/test_deep.py
+++ b/tests/explainers/test_deep.py
@@ -236,3 +236,69 @@ def test_pytorch_mnist_cnn():
 
     # clean up
     shutil.rmtree(root_dir)
+
+
+def test_pytorch_regression():
+    """Testing regressions (i.e. single outputs)
+    """
+    try:
+        import torch
+        from torch import nn
+        from torch.nn import functional as F
+        from torch.utils.data import TensorDataset, ConcatDataset, DataLoader
+        from sklearn.datasets import load_boston
+    except Exception as e:
+        print("Skipping test_pytorch_regression!")
+        return
+    import shap
+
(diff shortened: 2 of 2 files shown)


slundberg:
Great!

AvantiShri:
Hi @gabrieltseng @slundberg,

How stable is the pytorch implementation of DeepExplainer? We're thinking of using it to analyze pytorch models in the Kipoi repository (http://kipoi.org) for a hackathon happened at BDS 2018 right now. @Avsecz

gabrieltseng:
@AvantiShri, it should be fairly stable, but its functionality is limited, mainly because:

1. The multipliers are only calculated on a module level, instead of functions. See here for more information: https://github.com/slundberg/shap/pull/262

2. Only a few operations have been implemented so far. They are all listed at https://github.com/slundberg/shap/blob/master/shap/explainers/deep/deep_pytorch.py#L316

AvantiShri:
Thank you @gabrieltseng - could I trouble you to add `op_handler['Conv1d'] = linear_1d` to the list at https://github.com/slundberg/shap/blob/1e770087279455784329fb98bbae931b0685784e/shap/explainers/deep/deep_pytorch.py#L320-L325 - you already have MaxPool1d and AvgPool1d, so leaving out Conv1d seems unintentional.

slundberg:
@AvantiShri I don't have a 1D model in front of me to verify it works, but if you find adding Conv1d works for you I am happy to merge a one-line PR as soon as the unit tests pass.

AvantiShri:
Thanks Scott, will do if we end up going in that direction.

One feature that I am planning to implement (let me know if I shouldn't) is functionality to generate a set of reference on-the-fly for each input example (rather than having a fixed `self.data` reference for all the examples). This is relevant because we've found that a reference consisting of dinucleotide-shuffled versions of the input sequence works better than a fixed all-zeros or a GC-content-frequency reference. It looks like the current DeepExplainer implementation fixes `data` as an attribute of the class:

https://github.com/slundberg/shap/blob/1e770087279455784329fb98bbae931b0685784e/shap/explainers/deep/deep_tf.py#L105

However, unlike marcoancona/DeepExplain (where the reference is used to precompute static values in the computation graph), in your implementation `self.data` is evidently fed as an input to the computation graph in the form of `joint_input`, so it seems as though it would be fairly straightforward to modify the code to generate a different reference in `joint_input` dynamically for each input:

https://github.com/slundberg/shap/blob/1e770087279455784329fb98bbae931b0685784e/shap/explainers/deep/deep_tf.py#L247-L250

Please let me know if I'm mistaken!

slundberg:
@AvantiShri that would be great! The only issue I can foresee is that `self.expected_value` will no longer be a single number for all the inputs (since each sample will have it's own background distribution). For now I would just leave `self.expected_value` as `None` when there is not a fixed background set, people can easily compute it anyway as the difference between the model prediction and the sum of the attributions.

I am not sure the best way to expose that functionality, but perhaps letting the `data` argument in the constructor be a function that gets the data sample and produces the background set?

AvantiShri:
Yup, that's what I was thinking too (letting the data argument in the constructor be a function that gets the data sample and produces the background set). Ok, great that you don't foresee any dealbreaker issues.

PR #322, opened 2018-11-11
Adding dynamic reference functionality to tensorflow DeepExplainer

AvantiShri:
Added functionality to determine the reference on-the-fly according to the input data. This is implemented by allowing the user to supply a function as the `data` parameter. I benchmarked  the change using the genomics model+simulated data from the DeepLIFT repo, and the results look good. The reference generated for each sequence is a collection of dinucleotide-shuffled versions of the sequence. A Colab notebook comparing the output using the dynamic reference API and the standard API (where the standard API requires instantiating a new DeepExplainer object for each new reference) is here: https://gist.github.com/AvantiShri/8a3a0a03f4c4a578ee7909e3989467cc.

I added an example notebook to the `notebooks`  folder, and bumped the version to reflect the change - feel free to roll back these modifications at your discretion.


slundberg:
Thanks @AvantiShri! All looks good to me, my only change was to delete the .gitignore file.

PR #325, opened 2018-11-13
 Add optional x_jitter to dependence_plot

ihopethiswillfi:
Optionally add random noise to feature values if feature has less than 20 unique values.  May increase plot readability when feature is non-continuous.

Attached example picture with noise set to 0.33.
Without the noise, most of the dots overlap, especially where X == 2.0
plot_interaction_businesstravel_vs_age

### DIFF
--- a/shap/plots/dependence.py
+++ b/shap/plots/dependence.py
@@ -149,7 +149,7 @@ def dependence_plot(ind, shap_values, features, feature_names=None, display_feat
         xvals = np.unique(xv)
         smallest_diff = np.min(np.diff(np.sort(xvals)))
         jitter_amount = x_jitter * smallest_diff
-        for i in range(len(xv)):
+        xv += (np.random.ranf(size = len(xv))*jitter_amount) - (jitter_amount/2)
 
     # the actual scatter plot, TODO: adapt the dot_size to the number of data points?
     if interaction_index is not None:
--- a/shap/plots/dependence.py
+++ b/shap/plots/dependence.py
@@ -150,7 +150,6 @@ def dependence_plot(ind, shap_values, features, feature_names=None, display_feat
         smallest_diff = np.min(np.diff(np.sort(xvals)))
         jitter_amount = x_jitter * smallest_diff
         for i in range(len(xv)):
-            xv[i] += (np.random.ranf(size=1)[0] * jitter_amount) - (jitter_amount/2)
 
     # the actual scatter plot, TODO: adapt the dot_size to the number of data points?
     if interaction_index is not None:
--- a/shap/plots/dependence.py
+++ b/shap/plots/dependence.py
@@ -146,7 +146,7 @@ def dependence_plot(ind, shap_values, features, feature_names=None, display_feat
     # optionally add jitter to feature values
     if x_jitter > 0:
         if x_jitter > 1: x_jitter = 1
-        xvals = list(set(xv))
+        xvals = np.unique(xv)
         smallest_diff = np.min(np.diff(np.sort(xvals)))
         jitter_amount = x_jitter * smallest_diff
         for i in range(len(xv)):
--- a/shap/plots/dependence.py
+++ b/shap/plots/dependence.py
@@ -144,7 +144,7 @@ def dependence_plot(ind, shap_values, features, feature_names=None, display_feat
         color_norm = matplotlib.colors.BoundaryNorm(bounds, colors.red_blue.N)
         
     # optionally add jitter to feature values
-    if (x_jitter > 0) and (len(set(xv)) < 20):
+    if x_jitter > 0:
         if x_jitter > 1: x_jitter = 1
         xvals = list(set(xv))
         smallest_diff = np.min(np.diff(np.sort(xvals)))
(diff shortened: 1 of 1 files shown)


shap/plots/dependence.py
```diff
@@ -138,6 +142,15 @@ def convert_name(ind):
     if categorical_interaction and clow != chigh:
         bounds = np.linspace(clow, chigh, chigh - clow + 2)
         color_norm = matplotlib.colors.BoundaryNorm(bounds, colors.red_blue.N)
+        
+    # optionally add noise if feature has less than 20 unique values
+    if (noise > 0) and (len(set(xv)) < 20):
+        if noise > 1: noise = 1
+        xvals = list(set(xv))
+        smallest_diff = np.min([xvals[i] - xvals[i-1] for i in range(1, len(xvals))])
```
JacobMcc:
```suggestion
        smallest_diff = np.min(np.diff(np.sort(xvals)))
```
You could use `np.min(np.diff(xval))` here to avoid the comprehension. Also `set(xv)` may not preserve ordering, so you may want to call `sort` on the `xvals`. If `xvals` isn't ordered then you might end up with a negative value for `smallest_diff`.

shap/plots/dependence.py
```diff
@@ -138,6 +142,15 @@ def convert_name(ind):
...
+        noise_amount = noise * smallest_diff
+        for i in range(len(xv)):
```
JacobMcc:
```suggestion
        xv += (np.random.ranf(size = len(xv))*jitter_amount) - (jitter_amount/2)
```
You could get ride of this for loop and take advantage of vectorized operations here. 
Overall I like your contribution to the dependence plot btw, great idea!

slundberg:
Thanks @ihopethiswillfi ! This looks good, we should name it `x_jitter` though since jitter is the name of this kind of noise and we are only adding it to the x-axis. The vectorization suggestions from @JacobMcc are great, you can also consider using `np.unique`.

One other thing, I am not sure about the limit on 20 unique values. It seems that we shouldn't assume a user doesn't want jitter for 30 values (especially if they make a wider plot). It is also a bit strange to turn on jitter only to have it turn off automatically. If at some point in the future we auto-detect categorical features then we could auto turn on jitter for those.

Once these changes are in I am happy to merge :)

shap/plots/dependence.py
ihopethiswillfi: Thanks!  Good point on the vectorization.  But I think you should also remove the line just under the FOR statement (cause you replaced it).  Can you do that or let me know how to.  Sorry I'm kinda a git rookie.

ihopethiswillfi:
@slundberg Agreed with everything there.  
- will rename to x_jitter
- will remove arbitrary cap of 20.  I also wasn't sure about that either when committing it.  I originally thought I needed this to get what I wanted but I now found a way around it.  I can just detect the number of unique values of a feature before I call dependence_plot and base my jitter decision on that.

shap/plots/dependence.py
```diff
@@ -138,6 +142,15 @@ def convert_name(ind):
...
         color_norm = matplotlib.colors.BoundaryNorm(bounds, colors.red_blue.N)
+        
+    # optionally add jitter to feature values
+    if x_jitter > 0:
+        if x_jitter > 1: x_jitter = 1
+        xvals = list(set(xv))
+        smallest_diff = np.min(np.diff(np.sort(xvals)))
+        jitter_amount = x_jitter * smallest_diff
+        for i in range(len(xv)):
+            xv[i] += (np.random.ranf(size=1)[0] * jitter_amount) - (jitter_amount/2)
```
JacobMcc:
```suggestion

```
Here is the deletion of the line previously in the for loop. For future reference you can just change the code on your machine and then commit/push to GH with the changes and the Pull Request will automatically update.

shap/plots/dependence.py
```diff
@@ -138,6 +142,15 @@ def convert_name(ind):
     if categorical_interaction and clow != chigh:
         bounds = np.linspace(clow, chigh, chigh - clow + 2)
         color_norm = matplotlib.colors.BoundaryNorm(bounds, colors.red_blue.N)
+        
+    # optionally add jitter to feature values
+    if x_jitter > 0:
+        if x_jitter > 1: x_jitter = 1
+        xvals = list(set(xv))
```
JacobMcc:
```suggestion
        xvals = np.unique(xv)
```
Per @slundberg 's recommendation

shap/plots/dependence.py
JacobMcc: No problem. See comment below.

slundberg:
Thanks! The windows test fail has to do with an unrelated C++ VLA stack issue.

I have been thinking more about this, and since it happens for categorical data perhaps we should just add a flag denoting categorical data. With `categorical="auto"` by default. Then we could jitter automatically for small category counts. This would be similar to your 20 threshold, but it would also then enable us to do interesting things with the x jitter, such as doing non-uniform jitter to show the number of samples in each category (width of the jitter can be proportional to sample counts). This would simultaneously convey histogram information without adding mush visual complexity, since the width of the jitter would be the same as the height of a category histogram. This would be a much more precise and robust way of communicating this information than trying to pick the right opacity so we can tell where the most points are.

Anyway, let me think about this a bit more before merging, because if we go with a `categorical` flag then that would change this api.

slundberg:
I am going to merge this. If we add support for categorical features later we can update things then. Thanks again!

PR #334, opened 2018-11-21
ADD: speed comparison test for gradient boosting libraries

rushkv:


rushkv:
Hi @slundberg !

Could you please check my PR? I added speed tests for different gradient boosting libraries

slundberg:
Thanks, it looks interesting! Have you checked though to make sure the trees built by each library have similar numbers of leaves? I wouldn't want people to get the wrong idea if some of the differences are differences in model size. Also do you think the fact it is on a GPU matters? Is there a cuda implementation of Tree SHAP in CatBoost? I don't think there is in LightGBM or XGBoost.

rushkv:
1. I fix ``max_bin`` parameter equal to 128 and other parameters are default for every library
2.  XGBoost does pruning while training and i think there is no way to cancel it 
3. CatBoost makes oblivious trees and CatBoost calculates shap values on CPU

slundberg:
Sounds good. If you could just clarify in the notebook that all the tests were run on the CPU (or just don't mention the GPU), that would be great. Also for the sake of consistency could you name it "Performance comparison"? Then I'll go a head and merge this :)

rushkv:
@slundberg  no problem, done it

PR #345, opened 2018-12-03
Enable sum of errors for categorical classes instead of absolute error sum.

axsaucedo:
This PR contains a proposed modification in the code that performs the comparison of the classes with an equals match as opposed to an absolute error sum.

This change ensures that categorical variables are only compare, and errors are weighted the same for any class mismatch (as opposed to taking the numerical difference between the categorical values which may not make sense).

### DIFF
--- a/shap/explainers/kernel.py
+++ b/shap/explainers/kernel.py
@@ -416,7 +416,10 @@ class KernelExplainer(Explainer):
                         varying[i] = False
                         continue
                     x_group = x_group.todense()
-                num_mismatches = np.sum(np.abs(x_group - self.data.data[:, inds]) > 1e-7)
+                if type(x_group[0]) is str: 
+                    num_mismatches = np.sum(x_group[0] == self.data.data[:, inds])
+                else:
+                    num_mismatches = np.sum(np.abs(x_group - self.data.data[:, inds]) > 1e-7)
                 varying[i] = num_mismatches > 0
             return np.nonzero(varying)[0]
         else:


slundberg:
Thanks! Would you mind posting an example (as a unit test in `test_kernel.py`) that shows how this fixes an issue? It also looks like `x_group[0]` means you assume the categorical features are not grouped, but we would need to extend that before merging.

axsaucedo:
Good point. I am going through this right now - given that this change is in the generic KernelExplainer, in order to test whether the columns to compare are categorical, sometimes these are passed as INT, so just checking for string may not be enough. Perhaps a solution could be to provide another `**kwargs` parameter that allows the user to point out which are the columns that are `categorical`, which would then be ones that will be compared as exact matches. marcotcr's [LIME](https://github.com/marcotcr/lime)'s function allows for a `categorical_columns` parameter. What are your thoughts on this? 

It could be something along the lines of `shap_explainer.shap_values(shap_x, categorical_cols=["col1", "col2", ...)`. 

Thoughts?

slundberg:
Sorry for the slow reply. I think if people pass an INT that's fine, because we don't need to do anything special for categorical features as far as the method is concerned. If it is a "categorial feature" with 3 distinct INT values that's fine, or if it is a "continuous" feature with 1 million different INT values makes no difference.

I see this PR as supporting strings as an alternative to INTs (which are already supported). That seems like a nice thing to me, just want to make sure we get the groups right :)

PR #346, opened 2018-12-05
Add in individual reference attributions (for model stacking) for Independent Tree SHAP

HughChen:


slundberg:
Looks good. Remind me again why it is a constructor argument and not a `shap_values` argument?

HughChen:
I think the only reason was because of computing expected values.  However, I think it's fine if it is computed within the `shap_values` function (which I have changed it to be), since the expected value will be computed differently depending on the `model_stack` flag.  

For instance, if we are doing model stacking, it will compute the expected value for each individual reference.

I've added this to be reflected in the `test_model_stack_indep` test too.

HughChen:
I also added basic support for a stack of a neural network embedding into a tree model into `deep_tf` by allowing the `DeepExplainer` to accept starting gradients for the references (`model_stack_ref_grad`).

More explicit support for model stacking would probably be important down the line too.

Diyago:
hope this PR gets merged

PR #347, opened 2018-12-06
x_jitter is now aware of NaNs

ihopethiswillfi:
If x-values contained one or more NaNs, x_jitter would choke and cause a blank graph as output.  This is now fixed.  

We're also checking that we have at least 2 unique X-values after removing the NaNs, cause the code just below fails if that's not the case.

I'm not sure if
`if isinstance(xvals[0], float)`
is the best way to check if the elements inside the ndarray are floats.  I tried also `xvals.dtype` but this always returned 'object' even if the contents are floats.


slundberg:
Looks good to me!

PR #353, opened 2018-12-10
Add support for sklearn gradient boosting quantile estimator

dmilad:
- added support for quantile estimator for sklearn's GBR
- added unit tests for sklearn's GBR

I wasn't sure if you wanted separate tests added for mean and quantile estimators, but I didn't see a similar distinction for other models so I kept it simple.


slundberg:
Thanks!

PR #359, opened 2018-12-18
Use tight_layout to avoid clipping of feature names

HHest:
Without this, matplotlib could clip the feature names.

image

vs.

image


slundberg:
Sounds good, though it looks like it is causing some test failures at the moment.

HHest:
Yes, unsure why at moment... Test runs on my computer; I am using the same version 3.0.2 of matplotlib. Will try to reproduce.

HHest:
I give up. Cannot figure out why the tests fail. Change isn't much, so not worth the effort.

PR #361, opened 2018-12-19
updated pytorch class for pytorch 1.0.0

vladmarinescu:
I tried to calculate the shap values for my pytorch 1.0.0 model, but got the following errors:
1) TypeError: mul(): argument 'other' (position 1) must be Tensor, not numpy.ndarray 
Which I fixed by transforming sample_phis from a numpy array to a tensor on the calculation.

2) AttributeError: '*some torch.nn.Module*' object has no attribute 'y' 
And I wrapped the tensors in torch.nn.Parameter which puts them in '_parameters': OrderedDict of the module.

### DIFF
--- a/shap/explainers/deep/deep_pytorch.py
+++ b/shap/explainers/deep/deep_pytorch.py
@@ -132,13 +132,13 @@ class PyTorchDeepExplainer(Explainer):
                 if func_name in ['maxpool', 'nonlinear_1d']:
                     # only save tensors if necessary
                     if type(input) is tuple:
-                        setattr(module, 'x', input[0].detach())
+                        setattr(module, 'x', torch.nn.Parameter(input[0].detach()))
                     else:
-                        setattr(module, 'x', input.detach())
+                        setattr(module, 'x', torch.nn.Parameter(input.detach()))
                     if type(output) is tuple:
-                        setattr(module, 'y', output[0].detach())
+                        setattr(module, 'y', torch.nn.Parameter(output[0].detach()))
                     else:
-                        setattr(module, 'y', output.detach())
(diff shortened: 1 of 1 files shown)


slundberg:
Seems good to me, but @gabrieltseng do you have any comments since you wrote this part?

gabrieltseng:
This looks good to me too! Thanks for the update, @vlaqtz

gabrieltseng:
Actually, when I run the [example notebook for MNIST](https://github.com/slundberg/shap/blob/master/notebooks/deep_explainer/PyTorch%20Deep%20Explainer%20MNIST%20example.ipynb), for PyTorch 0.4 with an interim layer (`e = shap.DeepExplainer((model, model.conv_layers[0]), background)` instead of `e = shap.DeepExplainer(model, background)`), I also get a type error when calling `mul()`:

(on line 231)
```
TypeError: mul() received an invalid combination of arguments - got (numpy.ndarray), but expected one of:
 * (Tensor other)
      didn't match because some of the arguments have invalid types: (numpy.ndarray)
 * (Number other)
      didn't match because some of the arguments have invalid types: (numpy.ndarray)
```
I think because the `output` argument, which makes the `x` and `data` lists, is also a numpy array. @vlaqtz, could you try running the explainer with an interim layer?

vladmarinescu:
@gabrieltseng I did not test the notebook with an interim layer, turns out all the variables on line 231 were numpy arrays, so I removed the `torch.from_numpy()` and it works fine now.

slundberg:
Thanks! The appveyor issue seems to be related to a change in their system so it's unrelated to this.

Issue #429, opened 2019-02-05
memory corruption error from TreeExplainer in latest release


imatiach-msft:
When upgrading to latest release of shap some of our tests which use TreeExplainer started to fail in native layer randomly, causing python processes to fail.  I was able to reproduce snippets of the test cases in our library in shap and narrow down the issue to the native code in _cext.cc where the output tuple returned was using an object that had its reference counter decremented to 0 (so the object could be randomly garbage collected at any point in time, and it just so happened that some of our tests would hit this edge case).  It seems this error is more likely to be reproduced when using a large dataset.

imatiach-msft:
Issue fixed with merge of PR #430

ElSaico:
This consistently affects some of the models I try to create, which forces me to run the development version.
Would it be possible to make a new release?

slundberg:
Done!

PR #430, opened 2019-02-05
fixed memory corruption error from TreeExplainer

imatiach-msft:
resolves issue #429 and contains test case to reproduce error


slundberg:
Thanks!! That must have been a rare corner case since I think that's been around a while. Glad you were able to track it down!

imatiach-msft:
@slundberg thanks for merging - you are probably right, it may have been there for a long time; perhaps the recent refactor just made it more likely for the garbage collection to occur at that time

slundberg:
Sorry you had to find it! seems like it could have been a pain to debug.

PR #448, opened 2019-02-14
Make kernel and sampling explainers have consistent nsamples defaults

jklaise:
In writing a wrapper for the shap library so people can specify default solver (e.g. kernel or sampling) I ran into an error which turned out to be an inconsistency in the "nsamples" default value between the two methods. This PR makes nsamples="auto" be the common default.


slundberg:
Great, thanks!

PR #449, opened 2019-02-15
Adding support for matplotlib forceplot with text rotation

vatsan:
When using matplotlib for rendering the force plot (ex: in an environment where javascript is not trusted or if there is no interest access), it is painful if the input dataset has really long feature names. In this case the plot is cluttered with feature names. It will be helpful to rotate the feature names so that they are legible.

I have added support for this, updated the README.md with an example and also added a test suite to verify if this works (in addition to separately verifying in a jupyter notebook). The full test suite passes.


slundberg:
Thanks, this is great! I am bit hesitant though about using a lot of space for it at the top of the readme. I'm going to comment that part out for the moment, then think about where to put more detailed examples.

PR #456, opened 2019-02-22
Allow limited customizations to image_plot

verdimrc:
When input image is small, the output figure can be large and pixelated, and takes quite an amount of space when displayed on a Jupyter notebook. Example is to plot shap values of the MNIST with `ranked_outputs=2`.


slundberg:
Thanks!

PR #468, opened 2019-03-01
Added test verifying that linear shap gives the correct result.

jorgecarleitao:
Documentation of `LinearExplainer` state that

`Assuming features are independent leads to interventional SHAP values which for a linear model are coef[i] * (x[i] - X.mean(0)[i]) for the ith feature`, but this is never tested.

This PR contains a unit-test confirming this.


slundberg:
Thanks! I appreciate the PR test

PR #470, opened 2019-03-01
Added test checking that DeepExplainer of a linear model (in tf keras) is correct.

jorgecarleitao:
This test consists in:

1. generate data with 2 independent features and a target `y = 1 + 2 x1 + x2 + small noise`
2. define a linear model with Keras
3. fit the model to the data
4. confirm that the model fits the data correctly
5. explain the fitted model
6. confirm that the shap values correspond to the expected values of a linear model, namely, what is described in the docs as `"Assuming features are independent leads to interventional SHAP values which for a linear model are coef[i] * (x[i] - x.mean(0)[i]) for the ith feature"`.

This test currently fails (check with tensorflow installed) because the shap values do not correspond to the equation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
   - '3.6'
 install:
   - pip install .
+  - pip install tensorflow
 script: python setup.py nosetests
 before_script:
   - export DISPLAY=:99.0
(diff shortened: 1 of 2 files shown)


jorgecarleitao:
Interestingly, it passes on the CI system (I wrote the above assuming the behavior from my computer). So, all seems good!

slundberg:
Thanks for taking the time to verify this! It looks like the issue was that you passed `(model.layers[0].input, model.layers[-1].input)` as the model to explain to DeepExplainer. Since there is only one layer `model.layers[-1].input` is actually the input data, not the output of the model (which is `model.layers[-1].output`)

The tolerances we are also too low given the amount of noise in the problem, so I changed the test to use the fitted coefficients (not the true ones) since we are explaining what the model actually fit.

PR #473, opened 2019-03-01
Fixed CI by installing tensorflow -- #472

jorgecarleitao:
Also improved test skipping with nose.

Addressing #472

PR #474, opened 2019-03-01
Made pytorch tests not be skipped in CI.

jorgecarleitao:
Same as #473, but for pytorch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ python:
   - '3.6'
 install:
   - pip install .
-  - pip install tensorflow keras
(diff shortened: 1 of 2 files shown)


jorgecarleitao:
From the logs, pytorch does not provide a pypi installation. This means that we likely need another way to test it, e.g. through miniconda.

slundberg:
Yeah I tried several things, but so far to no avail.

thatlittleboy:
Irrelevant now, torch is being installed in CI and pytorch tests are running.

PR #494, opened 2019-03-14
Add BN to deep torch and fix a small Warning statement

JiechengZhao:
Add BatchNorm layer to the deep torch. When the eval is open it is simply a linear layer.

Fix a bug in a warning statement.


slundberg:
Thanks! I think it's safe to assume that people are not explaining a batch norm that is not fixed. Otherwise we would need to explicitly check for that.

PR #504, opened 2019-03-22
fixed broken keras multi input model in gradient explainer

moritzaugustin:
`DeepExplainer` worked nicely with my multi input keras model, but in order to make the `GradientExplainer` work as well I had to copy the respective lines from `deep/deep_tf.py` to `gradient.py`.

Without the fix I would be receiving the following errors when calling the `shap_values` method of the `GradientExplainer` object:
```
File "shap/explainers/gradient.py", line 196, in shap_values
    assert type(X) != list, "Expected a single tensor model input!"

### DIFF
--- a/shap/explainers/gradient.py
+++ b/shap/explainers/gradient.py
@@ -128,13 +128,13 @@ class _TFGradientExplainer(Explainer):
 
         # determine the model inputs and outputs
         if str(type(model)).endswith("keras.engine.sequential.Sequential'>"):
-            self.model_inputs = model.layers[0].input
+            self.model_inputs = model.inputs
             self.model_output = model.layers[-1].output
         elif str(type(model)).endswith("keras.models.Sequential'>"):
-            self.model_inputs = model.layers[0].input
+            self.model_inputs = model.inputs
             self.model_output = model.layers[-1].output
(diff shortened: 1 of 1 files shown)


moritzaugustin-tiplu:
( having changed via force-pushing only my mail address to a noreply address ;-) )

slundberg:
Thanks! Sorry for the slow reply, I'll work on reviewing this soon.

moritzaugustin:
The last two commits can be ignored if they are not wanted... ( I should have created a separate branch )

slundberg:
Thanks again. Both those last two commits look good to me. So let's keep them. I'll merge this for now, and at some point we need to add a unit test for the multi-input test case.

niloofardadras:
@moritzaugustin  I am trying to use gradient ex plainer but i get the error still

assert type(X) != list, "Expected a single tensor model input!"

is this change supposed to fix the error?

moritzaugustin:
in my case it did but i have not tested the current version with tf2 support yet. this code part seems to have been slightly refactored

FarzanT:
Hi,
I'm using Pytorch 1.7 and I still get this error when I pass a list of tensors at the `e.shap_values` call, and interestingly not at the `shap.DeepExplainer` call.  Would you please have a look?

PR #507, opened 2019-03-23
Add leaky ReLU and other conv layers to pytorch deep explainer

gabrieltseng:
Adds the other convolutional layers (1d and 3d), as well as an additional nonlinearity (Leaky ReLU).


slundberg:
Thanks!!

Issue #466, opened 2019-03-01
required dependencies -> optional dependencies?


jorgecarleitao:
> pytz, pandas, PyWavelets, pillow, decorator, networkx, cloudpickle, toolz, dask, scikit-image, tqdm, pickleshare, ptyprocess, pexpect, appnope, parso, jedi, pygments, backcall, wcwidth, prompt-toolkit, ipython-genutils, traitlets, ipython, shap

I am not sure why should I need to install `pillow` or `scikit-image` `networkx` if I am not working with image data nor networks.

I understand that this rep contains everything, all methods and plotting. But it seems unreasonble to ask for all these dependencies if the user just wants shap for a specific problem.

For example, for me, the use-case is to compute the shap values for feature importance and instance interpretability on a restful API supporting a frontend containing vizualization (that does not depend on shap visualizations). The backend does not need any of the vizualisation, nor pandas, only numpy, tensorflow and spark. I do not want to expose data to third-party libraries that I do not trust (e.g. `appnope`) to a production system.

One way to approach this is to split the `install_requires` in multiple `extra_requires`, and make only the mandatory packages part of the `install_requires`, as per the [official recommendation to use `install_requires`](https://packaging.python.org/guides/distributing-packages-using-setuptools/#install-requires):

> “install_requires” should be used to specify what dependencies a project minimally needs to run.

slundberg:
I bet many of those are driven by the use of 'scikit-image'. That package is only used for color space conversions, but those conversions are needed for all the plotting methods. We could pull out that dep by getting Lch color space methods another way

afiodorov:
This currently make deployment of the library in production somewhat inconvenient. Would be nice to have the lean version that doesn't pull in plotting available.

slundberg:
I can see why that would be nice. If anyone has ideas for an elegant way to do that without splitting up the repo and making maintenance harder please speak up.

jorgecarleitao:
One way is to use [extras](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies), that are supported by pip: `pip install shap[all]`, `pip install shap`, `pip install shap[images]`.

jorgecarleitao:
I will take a look at this one.

slundberg:
[...]

PR #508, opened 2019-03-24
WIP: Made most dependencies be declared as extras in setup.py.

jorgecarleitao:
The goal of this PR is to reduce the number of dependencies that are installed when installing shap, see #466 

Essentially, there are two types of dependencies

1. core dependencies: `['numpy', 'scipy', 'scikit-learn', 'pandas', 'tqdm']`
2. non-core dependencies

Non-core dependencies are dependencies of a given `feature` of the package. In Python, this can be achieved through `extras_require` keyword on `setup.py`. This PR does exactly that. It also generalizes the testing to gracefully handle uninstallable dependencies (this was limited to `xgboost` and `lightgbm`).

This PR proposes that the package contains a `extra_requires` that enumerates the features of this package:

```
extras_require = {
    'plots': ['matplotlib', 'scikit-image', 'ipython'],
    'xgboost': ['XGBoost>=v0.81'],  # A bug in XGBoost fixed in v0.81 makes XGBClassifier fail to give margin outputs
    'lightgbm': ['lightgbm'],
    'tensorflow': ['keras>=2.1.0', 'tensorflow>=1.4.0'],
    'torch': ['pytorch>=0.4'],
    'torchvision': ['pytorch>=0.4', 'torchvision']
}
```

**This change changes the installation script:** the user now has to run `pip install shap[plots]` to obtain the same behavior as before. On the other hand, `pip install shap[plots,tensorflow]` is now possible :)

There are still a couple of changes for this PR to be completed:

[ ] Document new way of installing package
[ ] Add extra dependency `catboost`, that this package also depends on.
[ ] Check that tests run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,6 @@ os:
 python:
   - '2.7'
   - '3.6'
-install:
-  - pip install .
-  - pip install tensorflow keras
 script: python setup.py nosetests
 before_script:
   - export DISPLAY=:99.0
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,7 +15,6 @@ environment:
 install:
   # We need wheel installed to build wheels
   - "%PYTHON%\\python.exe -m pip install wheel"
-  - "%PYTHON%\\python.exe -m pip install tensorflow keras"
 
 build: off
 
--- a/setup.py
+++ b/setup.py
@@ -25,11 +25,23 @@ if sys.platform == 'darwin':
         if python_target < '10.9' and current_system >= '10.9':
             os.environ['MACOSX_DEPLOYMENT_TARGET'] = '10.9'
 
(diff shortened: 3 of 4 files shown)


slundberg:
Thanks! I have a few concerns though that we should think about:
1) This could break any current dependency setups that assume plotting works by default.
2) It is uncommon for the default use case to require installing an 'extra'.
3) This won't match the conda-forge version since I don't think that supports extras.

https://github.com/pypa/setuptools/issues/1139 is what we really want. But unfortunately it seems to not work.

Thoughts?

maddiehawkinson:
@jorgecarleitao This is exactly the error I am having as I try to use this package, thank you for identifying and building this solution! Do you know when this will be merged and any instructions on how to use it? 

@slundberg thank you so much for this package, I think this will revolutionize a lot of my research.

Issue #477, opened 2019-03-05
problem with pytorch model


eranbio:
I am trying to use shap to interpret genomic deep learning model in PyTorch. The input tensors to shap are three dimensional and have shapes like torch.Size([100, 4, 2000]) for the background and 
torch.Size([5, 4, 2000]) for the input samples for shap_values().

I get the following error message:

  File "/home/users/bioinf/erane/work/unity/tools/gedai/dna_pytorch_shap.py", line 668, in <module>
    main()
  File "/home/users/bioinf/erane/work/unity/tools/gedai/dna_pytorch_shap.py", line 215, in main
    shap_values = e.shap_values(xdata)
  File "/evogene/software/anaconda3/envs/py3.6-env-erane/lib/python3.6/site-packages/shap/explainers/deep/__init__.py", line 119, in shap_values
    return self.explainer.shap_values(X, ranked_outputs, output_rank_order)
  File "/evogene/software/anaconda3/envs/py3.6-env-erane/lib/python3.6/site-packages/shap/explainers/deep/deep_pytorch.py", line 221, in shap_values
    sample_phis = self.gradient(feature_ind, joint_x)
  File "/evogene/software/anaconda3/envs/py3.6-env-erane/lib/python3.6/site-packages/shap/explainers/deep/deep_pytorch.py", line 166, in gradient
    grads = [torch.autograd.grad(selected, x)[0].cpu().numpy() for x in X]
  File "/evogene/software/anaconda3/envs/py3.6-env-erane/lib/python3.6/site-packages/shap/explainers/deep/deep_pytorch.py", line 166, in <listcomp>
    grads = [torch.autograd.grad(selected, x)[0].cpu().numpy() for x in X]
  File "/evogene/software/anaconda3/envs/py3.6-env-erane/lib/python3.6/site-packages/torch/autograd/__init__.py", line 144, in grad
    inputs, allow_unused)
**RuntimeError: hook 'deeplift_grad' has changed the size of value**

On the same machine I can successfully use shap with examples of pytorch MNIST image classification code (like https://www.kaggle.com/ceshine/pytorch-deep-explainer-mnist-example) .  Any hint will be well appreciated. 

Eran

slundberg:
Do you get the same error for GradientExplainer as well?

eranbio:
[...]

PR #516, opened 2019-03-26
Fix maxpool1d in pytorch deep explainer

gabrieltseng:
This is a partial fix to the issue raised in https://github.com/slundberg/shap/issues/477 .

The problem is that the MaxPool1d module doesn't have the same behavior as other modules when a backward hook is added. Specifically, a backward hook receives (as arguments) a `grad_input`, which is the gradient of the inputs, and `grad_output`, which is the gradient of the outputs. `grad_input` is the same shape as the input, and `grad_output` is the same shape as the output, as can be seen for this example with MaxPool2d:

```python
>>> import torch
>>> from torch import nn
>>> test_2d_input = torch.ones(40, 10, 10, 13, requires_grad=True)
>>> test_2d_output = torch.ones(40, 10, 5, 6, requires_grad=True)
>>> maxpool2d = nn.MaxPool2d(kernel_size=2)
>>> def print_grads(module, grad_input, grad_output):
...     print(module)
...     print([g.shape for g in grad_input])
...     print([g.shape for g in grad_output])
...
>>> maxpool2d.register_backward_hook(print_grads)
<torch.utils.hooks.RemovableHandle object at 0x109059860>
>>> o_2d = maxpool2d(test_2d_input)
>>> o_2d.backward(test_2d_output)
MaxPool2d(kernel_size=2, stride=2, padding=0, dilation=1, ceil_mode=False)
[torch.Size([40, 10, 10, 13])]
[torch.Size([40, 10, 5, 6])]
```

For maxpool1d, `grad_input` actually only contains the gradients for the "selected" inputs (i.e. the maximum of a kernel), which explains the following behaviour, where `grad_input` and `grad_output` have the same number of elements.
```python
>>> test_1d_input = torch.ones(40, 1, 13, requires_grad=True)
>>> test_1d_output = torch.ones(40, 1, 6, requires_grad=True)
>>> maxpool1d = nn.MaxPool1d(kernel_size=2)
>>> def print_grads(module, grad_input, grad_output):
...     print(module)
...     print([g.shape for g in grad_input])
...     print([g.shape for g in grad_output])
...
>>> maxpool1d.register_backward_hook(print_grads)
<torch.utils.hooks.RemovableHandle at 0x11c505ef0>
>>> o_1d = maxpool1d(test_1d_input)
>>> o_1d.backward(test_1d_output)
MaxPool1d(kernel_size=2, stride=2, padding=0, dilation=1, ceil_mode=False)
[torch.Size([40, 1, 1, 6])]
[torch.Size([40, 1, 6])]
```
I have fixed the bug so that the output is the right shape, but there is now an issue where if the deep explainer assigned a non-zero gradient to an element which was not selected in the maxpooling layer, there is no way for that gradient to be propagated back, meaning the shap values will be wrong.

@slundberg , for now I have added a warning that is printed when a maxpool1d layer is encountered. We could throw an error instead; let me know which you prefer.

Thanks!

### DIFF
--- a/shap/explainers/deep/deep_pytorch.py
+++ b/shap/explainers/deep/deep_pytorch.py
@@ -283,6 +283,7 @@ def maxpool(module, grad_input, grad_output):
     grad_input[0] = torch.where(torch.abs(delta_in) < 1e-7, torch.zeros_like(delta_in),
                            (xmax_pos + rmax_pos) / delta_in).repeat(dup0)
     if module.__class__.__name__ == 'MaxPool1d':
+        print("Warning: Results of DeepExplainer are unstable with MaxPool1d")
         grad_input[0] = torch.gather(grad_input[0], -1, indices).unsqueeze(1)
     # delete the attributes
     del module.x
--- a/shap/explainers/deep/deep_pytorch.py
+++ b/shap/explainers/deep/deep_pytorch.py
@@ -57,7 +57,7 @@ class PyTorchDeepExplainer(Explainer):
                 self.num_outputs = outputs.shape[1]
 
     def add_target_handle(self, layer):
-        input_handle = layer.register_forward_hook(self.get_target_input)
+        input_handle = layer.register_forward_hook(get_target_input)
         self.target_handle = input_handle
 
     def add_handles(self, model, forward_handle, backward_handle):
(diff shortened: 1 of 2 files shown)


gabrieltseng:
By registering a backward hook on the tensor instead of the module, the issue described above has been fixed.

slundberg:
Thanks!! Sorry for the slow review, I am getting back from stuff tonight and will look this over tomorrow

slundberg:
Thanks again!

gabrieltseng:
Thank you! I think this means https://github.com/slundberg/shap/issues/477 can be closed now

florianst:
Dear @gabrieltseng, I'm still running into the issue reported in #477 (hook deeplift_grad has changed) with MaxPool1d layers within a Sequential layer. No error message / warning comes up. Any chance this could be fixed?

gabrieltseng:
Hi @florianst, which versions of shap and pytorch are you using?

The [pytorch single input test](https://github.com/slundberg/shap/blob/master/tests/explainers/test_deep.py#L350), which uses a MaxPool1d layer, passes if I run it from master.

I also tried wrapping it in a Sequential layer, so that [line 376](https://github.com/slundberg/shap/blob/master/tests/explainers/test_deep.py#L376) is
```python
self.maxpool2 = nn.Sequential(nn.MaxPool1d(kernel_size=2))
```
and this passes for me too.

florianst:
I've updated to pytorch 1.1 and the current shap master branch version (0.29.3.dev0), and the hook 'deeplift_grad' error still persists. My layers look like this:
```
self.ConvLayer = nn.Sequential(
            nn.Conv1d(256, 512, kernel_size=2, stride=1, padding=0),
            nn.MaxPool1d(kernel_size=3, padding=1, stride=1),
            nn.ReLU()
            )
```
A difference from my code to the test is that my model requires _multiple inputs_. As there is no test for that implemented, is that supported for convolutional models?
For completeness' sake, here is the full error message I get:
```

  File "train.py", line 420, in <module>
    shap_values = explainer.shap_values(to_explain)
  File "/users/florianst/.local/lib/python3.6/site-packages/shap-0.29.3.dev0-py3.6-linux-x86_64.egg/shap/explainers/deep/__init__.py", line 119, in shap_values
    return self.explainer.shap_values(X, ranked_outputs, output_rank_order)
  File "/users/florianst/.local/lib/python3.6/site-packages/shap-0.29.3.dev0-py3.6-linux-x86_64.egg/shap/explainers/deep/deep_pytorch.py", line 172, in shap_values
    sample_phis = self.gradient(feature_ind, joint_x)
  File "/users/florianst/.local/lib/python3.6/site-packages/shap-0.29.3.dev0-py3.6-linux-x86_64.egg/shap/explainers/deep/deep_pytorch.py", line 115, in gradient
    for idx, x in enumerate(X)]
  File "/users/florianst/.local/lib/python3.6/site-packages/shap-0.29.3.dev0-py3.6-linux-x86_64.egg/shap/explainers/deep/deep_pytorch.py", line 115, in <listcomp>
    for idx, x in enumerate(X)]
  File "/users/florianst/.local/lib/python3.6/site-packages/torch/autograd/__init__.py", line 149, in grad
    inputs, allow_unused)
RuntimeError: hook 'deeplift_grad' has changed the size of value
```
Any ideas / any chance of reproducing that with the pytorch test?

gabrieltseng:
Hi @florianst ,

I've been able to reproduce the problem! I'll work on finding a fix; thanks for pointing this out.

PR #517, opened 2019-03-27
Set expected values in pytorch

gokceneraslan:
Is this just missing or is there a special reason not to set self.expected_value in the constructor?

### DIFF
--- a/shap/explainers/deep/deep_pytorch.py
+++ b/shap/explainers/deep/deep_pytorch.py
@@ -55,6 +55,7 @@ class PyTorchDeepExplainer(Explainer):
             if outputs.shape[1] > 1:
                 self.multi_output = True
                 self.num_outputs = outputs.shape[1]
+                self.expected_value = outputs.mean(0).numpy()
 
     def add_target_handle(self, layer):
         input_handle = layer.register_forward_hook(self.get_target_input)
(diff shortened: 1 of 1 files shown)


gabrieltseng:
Hi!

The pytorch deep explainer doesn't actually use an expected value, so it is set to `None`:
https://github.com/slundberg/shap/blob/master/shap/explainers/deep/deep_pytorch.py#L29

gokceneraslan:
Yes, but neither does the TF class. However, expected_value is still set in the TF explainer class:

https://github.com/slundberg/shap/blob/master/shap/explainers/deep/deep_tf.py#L139

Isn't the idea to keep the information about the expected value of the background coupled with the explainer? 

I think, it's also needed for consistency. `shap.force_plot` codes in TF notebooks do not work with PyTorch explainer for example, since expected value is not set in PyTorch class:

```python
shap.force_plot(explainer.expected_value, shap_values[0,:], X.iloc[0,:])
```

gabrieltseng:
Ah that makes sense. Thanks!

DavidRuhe:
Hi,

Sorry, I'm quite new to SHAP, but how would one use force plot with pytorch now? Compute the expected values manually? If so, how?

slundberg:
Thanks! This looks good so I am going to merge it.

slundberg:
@DavidRuhe you can now just clone master to get this (or do it manually by checking out the one-line change in this PR).

PR #520, opened 2019-03-28
Fix fatal crash on xgb gpu_hist model load

### LINKED ISSUE #452: TreeExplainer with xgboost model trained on GPU dies.
Hi, I've train quite a huge model using GPU, and save/load it before using with `TreeExplainer()`. The problem is the jupyter kernel dies when I call `TreeExplainer(model)`.

I supposed that it's because the model is too big to fit GPU memory, so I tried to change the model's parameter to 'cpu_predictor' using `set_params` method, so that the SHAP internally use CPU & RAM for the calculation. 

But it doesn't work as I expected. Even if I changed the predictor to use CPU, the jupyter kernel still dies. There's no error log so I couldn't attach here. The program just dies. What can I do with this?

Here's my code
---
```python
def load_model(fname):
    model = xgb.XGBClassifier()
    model.load_model(fname)
    with open(fname.replace('.xgboost', '.json'), encoding='utf-8') as fin:
        params = json.load(fin)
    model.set_params(**params)
    return model

model = load_model('./model/model_2-gram_2019-02-20T15-10-38.xgboost')

params = {
     'tree_method': 'hist',
     'nthread': 4,
     'predictor': 'cpu_predictor', 
     'n_gpus': 1
}
model.set_params(**params)

# compute the SHAP values for every prediciton in the validation dataset
# DIES HERE!
```

gnodabb:
I'm trying to resolve #452. When using an xgboost model trained with parameter
'tree_method':'gpu_hist' the shap.TreeExplainer crashes.
This crash is on the call to the C++ lib on line 873
function _cext.compute_expectations. This appears to be
due to the zeros that occur in the children_left and/or
children_right variables. The fix replaces the zeros with
-1's after the data is loaded from the xgboost gpu_hist model.

### DIFF
--- a/shap/explainers/tree.py
+++ b/shap/explainers/tree.py
@@ -716,6 +716,8 @@ class Tree:
             self.thresholds = tree["threshold"]
             self.values = tree["value"] * scaling
             self.node_sample_weight = tree["node_sample_weight"]
+            self.children_left[self.children_left==0]=-1
+            self.children_right[self.children_right==0]=-1
 
         elif type(tree) == dict and 'tree_structure' in tree:
             start = tree['tree_structure']


slundberg:
Thanks!! So the GPU trained model uses 0 and not -1 to denote leaves? It seems odd that would change from the CPU models. Do you know if that is the case and if so why? (Just trying to make sure I understand the bug this is fixing)

gnodabb:
So there is a difference between ‘gpu_exact’ and ‘gpu_hist’. ‘gpu_exact’ behaves as expected (consistent with cpu models). For some reason ‘gpu_hist’ was setting leaves to zero only on the ‘children_right’. ‘children_left’ still had ‘-1’. 

There may be a better fix. Particularly since the crash was in the cpu lib. But the python parser was easier for me to follow. Also, I should probably file an issue with xgboost just to make sure this is expected behavior.

slundberg:
Interesting. Yeah I think we should wait until you hear back from XGBoost on if the behavior is expected. If not then I fear this could be masking some other issue with the GPU format.

gnodabb:
I totally understand. I will update with xgboost's response. Thanks!

gnodabb:
I have opened an issue with the xgboost team to get clarification  on expected behavior.
[https://github.com/dmlc/xgboost/issues/4314](url)

kyoungrok0517:
Oh so what happened next? Is the problem with `gpu_hist` solved?

gnodabb:
It's been a long time. However, I'm currently waiting for a response from the XGBoost team. XGBoost issue 4314.

trivialfis:
I must forgot about this, sorry.  GPU exact is removed since hist is much more efficient.  I'm currently refactoring the model IO.  Will look into this later.  Feel free to ping me here

slundberg:
@trivialfis Thanks! After the IO gets refactored could be a good time to sort this out. I wrote a custom binary loader for the XGB models based on reverse engineering the XGB c++ code. I think this issue is because there is something different when gpu_models are saved that my loader does not handle. If you could let us know when that refactor is done, then we can dig into this issue (since doing right before you change things seems like a waste).

kyoungrok0517:
Great to hear that there’s some progress recently :) hope the problem is solved soon.

trivialfis:
I looked into this briefly today.  Will rewrite the parser here after XGBoost-1.0 is released.

kyoungrok0517:
Great. Thanks for checking!

2019년 12월 3일 (화) 오후 12:47, Jiaming Yuan <notifications@github.com>님이 작성:

trivialfis:
@kyoungrok0517 Should be fixed once https://github.com/dmlc/xgboost/pull/5084 is merged.  For the long term, shap may consider using the format introduced in https://github.com/dmlc/xgboost/pull/4732 .  I will create a schema later with an example parser.

trivialfis:
Fixed in https://github.com/dmlc/xgboost/pull/5084 .

slundberg:
Since this got fixed in the XGBoost code base I going to close this.

slundberg:
@trivialfis happy to use the JSON API once we can, though we will need to consider backward compatibility.

YannDubs:
Hi @slundberg did you ever reconsider using JSON. Some important features of xgboost don't currently work with SHAP, eg, using categorical features: see #2662

PR #536, opened 2019-04-05
Add warning when model.objective == None

mm5631:
For objectives not supported by SHAP, add user warning for clarity

### DIFF
--- a/shap/explainers/tree.py
+++ b/shap/explainers/tree.py
@@ -7,6 +7,7 @@ import struct
 from distutils.version import LooseVersion
 from .explainer import Explainer
 from ..common import assert_import, record_import_error
+import warnings
 
 try:
     from .. import _cext
@@ -531,6 +532,7 @@ class TreeEnsemble:
             self.objective = objective_name_map.get(model.objective, None)
             self.tree_output = tree_output_name_map.get(model.objective, None)
             if model.objective is None:
+                warnings.warn('{model.objective} is not yet supported by TreeExplainer, defaulting to identity link function')
                 self.objective = "squared_error"
                 self.tree_output = "raw_value"
         elif str(type(model)).endswith("lightgbm.sklearn.LGBMClassifier'>"):
--- a/shap/explainers/tree.py
+++ b/shap/explainers/tree.py
@@ -532,7 +532,7 @@ class TreeEnsemble:
             self.objective = objective_name_map.get(model.objective, None)
             self.tree_output = tree_output_name_map.get(model.objective, None)
             if model.objective is None:
-                warnings.warn('{model.objective} is not yet supported by TreeExplainer, defaulting to identity link function')
+                warnings.warn(f'{model.objective} is not yet supported by TreeExplainer, defaulting to identity link function')
                 self.objective = "squared_error"
                 self.tree_output = "raw_value"
         elif str(type(model)).endswith("lightgbm.sklearn.LGBMClassifier'>"):


slundberg:
Thanks! However, this code is meant to capture that the default of LGBMRegressor is linear regression. So it doesn't mean we don't know the objective.

mm5631:
My bad, should have dug a little more. My issue is with the fact that SHAP defaults to certain link functions for the shap_values that won't necessarily be the same as the one used with poisson/tweedie GBM objectives. How do you suggest tackling this?

mm5631:
LGBM is simply using a log-link as the function so the following does the trick:
```
import shap
from lightgbm import LGBMRegressor
from sklearn.datasets import load_boston

X, y = load_boston(return_X_y=True)
lgbm = LGBMRegressor(objective='tweedie')
lgbm.fit(X, y)

explainer = shap.TreeExplainer(lgbm)
shap_contrib = np.sum(shap_values[0]) # first pred

print('contrib sum:', shap_contrib + explainer.expected_value)
# shap contrib: 3.1898915721235372

print('predict:', lgbm.predict(X[0:1])[0])
# predict: 24.28579404325386

print('shap correct:', np.exp(shap_contrib + explainer.expected_value))
# shap correct: 24.285794043253837


```
The issue is caused by the fact that when it gets mapping from objective_map it defaults to linear as per your comment above. However this linear parameter is ultimately used to select the link function in common.py

Would be great to add support for log-link or at least a warning in the meantime.
```
class Link:
    def __init__(self):
        pass


class IdentityLink(Link):
    def __str__(self):
        return "identity"

    @staticmethod
    def f(x):
        return x

    @staticmethod
    def finv(x):
        return x


class LogitLink(Link):
    def __str__(self):
        return "logit"

    @staticmethod
    def f(x):
        return np.log(x/(1-x))

    @staticmethod
    def finv(x):
        return 1/(1+np.exp(-x))
    
# class LogLink(Link):
#     def __str__(self):
#         return "log"
#    blah


def convert_to_link(val):
    if isinstance(val, Link):
        return val
    elif val == "identity":
        return IdentityLink()
    elif val == "logit":
        return LogitLink()
#     elif val == 'log':
#         return LogLink()
    else:
        assert False, "Passed link object must be a subclass of iml.Link"
```

slundberg:
I agree we should support a log link (and other link functions). The behavior you see right now is because by default TreeExplainer explains the margin of the model (before the link function). There is an option to explain the transformed version for probabilities (logit link), but other links should also be supported. I think that will require me doing some updates to the way tree.py handles transforms to make it more general than just for probabilities.

mm5631:
Sounds good. That and some comments around the fact that the values would be more like SHMP (SHapley Multiplicative exPlanations) 😆.
Thanks for the hard work Scott, love the package. Feel free to close this.

mm5631:
@slundberg Could you point me to where the code is for the following:

> There is an option to explain the transformed version for probabilities (logit link), but other links should also be supported.

Thanks a lot

slundberg:
@maximemerabet that's referring to TreeExplainer's `output` option. By default it is `output="margin"` but it can also be `output="probability"` which works for models like XGBoost with a logistic loss.

A163460:
@mm5631  You code below is very helpful
```

import shap
from lightgbm import LGBMRegressor
from sklearn.datasets import load_boston

X, y = load_boston(return_X_y=True)
lgbm = LGBMRegressor(objective='tweedie')
lgbm.fit(X, y)

explainer = shap.TreeExplainer(lgbm)
shap_contrib = np.sum(shap_values[0]) # first pred

print('contrib sum:', shap_contrib + explainer.expected_value)
# shap contrib: 3.1898915721235372

print('predict:', lgbm.predict(X[0:1])[0])
# predict: 24.28579404325386

print('shap correct:', np.exp(shap_contrib + explainer.expected_value))
# shap correct: 24.285794043253837
```

Do you know how i can transform the value when there is a background data provided to the TreeExplainer? 

`explainer = shap.TreeExplainer(lgbm, data=referencedata) `

I found the above transformation no longer work.

PR #549, opened 2019-04-11
Update README.md

ricoms:
Suggestion to change the first example to a working example with lightgbm (version `2.2.3`) and shap (version `0.28.5`)

The previous example did not work even with xgboost package at `0.7.post4` version.

Thanks for all the great work, shap is an excellent package.

Sorry, correcting `X` and `y` variable names from previous `a`, `b`

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ conda install -c conda-forge shap
 While SHAP values can explain the output of any machine learning model, we have developed a high-speed exact algorithm for tree ensemble methods ([Tree SHAP arXiv paper](https://arxiv.org/abs/1802.03888)). Fast C++ implementations are supported for *XGBoost*, *LightGBM*, *CatBoost*, and *scikit-learn* tree models:
 
 ```python
-import xgboost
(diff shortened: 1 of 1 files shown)


imatiach-msft:
@ricoms sorry, two things:
1.) shouldn't we fix shap to work with xgboost if it is not working right now?  I'm also not aware of shap not working with xgboost.
2.) even if we decide to use lightgbm for the default example, we should at least update the surrounding documentation.  The example still says:
```# train XGBoost model```
and also:
```# (same syntax works for LightGBM, CatBoost, and scikit-learn models)```

ricoms:
hello @imatiach-msft .

Don't worry, I'm sorry too, it seems that my xgboost version was not up to date enough. This PR is related to this closed issue: https://github.com/slundberg/shap/issues/28

I had installed the `0.7` and there, @slundberg  mentioned that the example code should work for version `0.7`, but it does not.

Although I now tried with xgboost version `0.82` and the original code works. Please don't consider this PR anymore.

Thanks for the great work!

PR #552, opened 2019-04-12
make shap play nicer with sklearn<0.18

stillmatic:
an alternative is to pin sklearn >= 0.18, otherwise there are some import errors with older versions


slundberg:
Thanks!

PR #554, opened 2019-04-18
dependence_plot uses color arg + optional cmap

rcarneva:
`dependence_plot()` had an orphaned `color` arg that I slotted into the fallback scatterplots, and I also added an optional `cmap` arg that allows for overriding the default `colors.red_blue`


slundberg:
Thanks! Those args needed cleaned up :)

PR #557, opened 2019-04-18
fixed multirow sparse background case for KernelExplainer and added test

imatiach-msft:


slundberg:
Thanks! LGTM, will wait for the tests to run though.

PR #578, opened 2019-05-07
add SHAP values for binary classification to CatBoost using TreeExplainer with model_output='probibility' and tree_dependence='Independent'

dvpolyakov:
Hello, @slundberg)

I have added calculation of SHAP values for binary classification in CatBoost Classifier using TreeExplainer with model_output='probibility' and tree_dependence='Independent' to address this https://github.com/catboost/catboost/issues/668 and this https://github.com/slundberg/shap/issues/576 issues. 

The main work done in the CatBoostTreeModelLoader (which is analogous to XGBTreeModelLoader). The wrapper saves the CatBoost model and loads it to a unified format as it is done in XGBTreeModelLoader.

The tutorial to added feature is here: https://github.com/dvpolyakov/CatBoost_issues/blob/master/Illustration_examples.ipynb.

Currently this mode works only on data without categorical features.

I am open to discussion and comments.

### DIFF
--- a/shap/explainers/tree.py
+++ b/shap/explainers/tree.py
@@ -579,6 +579,11 @@ class TreeEnsemble:
             assert_import("catboost")
             self.model_type = "catboost"
             self.original_model = model
+            self.dtype = np.float32
+            cb_loader = CatBoostTreeModelLoader(model)
+            self.trees = cb_loader.get_trees(data=data, data_missing=data_missing)
+            self.tree_output = "log_odds"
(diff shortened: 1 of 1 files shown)


slundberg:
Thanks for this! Sorry I have been on travel etc. for a few weeks, but should get to this tomorrow.

slundberg:
Thanks again for this! It looks good. (long story, but it took a couple extra days to be able to review this). We don't have CatBoost unit tests at the moment, but I'll leave that as future work.

PR #596, opened 2019-05-17
Adds the ability to export summary graph

xEcEz:
Adds the ability to export a summary graph with a specific width to the given export_path.

### DIFF
--- a/shap/plots/summary.py
+++ b/shap/plots/summary.py
@@ -6,7 +6,6 @@ from __future__ import division
 import warnings
 import numpy as np
 from scipy.stats import gaussian_kde
-
 try:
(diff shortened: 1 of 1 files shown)


slundberg:
Thanks! The ability to set the width seems great. I am less sure about the export_path. How does that improve over just calling `pl.savefig` after calling summary plot with the `show=False` option?

xEcEz:
Well, I am not sure why, but even when I set `show=False` the plot gets printed, hence is not available to be saved afterward. At least that is what I've experienced working in a notebook.

slundberg:
Perhaps that is because you have an interactive matplotlib mode set for the notebook? (`%matplotlib notebook `)

PR #599, opened 2019-05-17
Support passing in an Axes object to dependence_plot.

mqk:
I often find it very convenient to define a matplotlib figure and axes object externally to the plotting routine, and to simply pass the axes object in as an argument. That way it is quite easy to construct a grid of many independent plots, for example. It also makes it easier to modify the resulting plot (axes labels, plot limits, add annotations, etc.) before displaying it.

In this PR I add this capability to the `dependence_plot` method. One could imagine doing something similar for the other plot types (summary, etc.), but that seems more involved, and I haven't had the need for it so far.


slundberg:
Thanks! Looks good to me.

PR #600, opened 2019-05-18
relaxing tolerances to fix failing test

imatiach-msft:
[WIP] scikit learn was recently upgraded which may be causing some test failures due to precision of test assertions.  Relaxing the tolerance to see if it will fix the test.


slundberg:
Interestingly, I can't seem to reproduce the problem even with the same 0.21.1 sklearn version on my own machine.

imatiach-msft:
@slundberg I was able to reproduce the issue on an ubuntu machine.  Relaxing the tolerance seems to have fixed it.  Strangely, downgrading scikit-learn did not fix the issue on the machine, so I'm not sure why the results were so different.  Increasing the l1 regularization parameter for kernel explainer also fixed the issue.

slundberg:
Thanks! I expect to be back here covering issues etc by next week.

Issue #593, opened 2019-05-16
DeepExplainer fails for PyTorch Pretrained Alexnet Model


kracekumar:
```python
from torchvision import transforms
import torchvision.models as models

from torch.utils.data.dataset import Dataset
from torch.utils.data import DataLoader

alexnet = models.alexnet(pretrained=True).eval()
transforms = transforms.Compose([
            transforms.ToTensor(),
            #transforms.Normalize([0.5, 0.5, 0.5], [0.5, 0.5, 0.5])
        ])
dataloader = DataLoader(ImageDataset('images', transforms=transforms), batch_size=16)
batch = next(iter(dataloader))
images, _ = batch
e = shap.DeepExplainer(alexnet, images[0:1])
shap_values = e.shap_values(images[1:])
````

The output of the `e.shap_values` call is below
```
Warning: unrecognized nn.Module: <class 'type'>
Warning: unrecognized nn.Module: <class 'type'>
Warning: unrecognized nn.Module: <class 'type'>
...
```

The architecture of alexnet 

```python
print(alexnet)
AlexNet(
  (features): Sequential(
    (0): Conv2d(3, 64, kernel_size=(11, 11), stride=(4, 4), padding=(2, 2))
    (1): ReLU(inplace)
    (2): MaxPool2d(kernel_size=3, stride=2, padding=0, dilation=1, ceil_mode=False)
    (3): Conv2d(64, 192, kernel_size=(5, 5), stride=(1, 1), padding=(2, 2))
    (4): ReLU(inplace)
    (5): MaxPool2d(kernel_size=3, stride=2, padding=0, dilation=1, ceil_mode=False)
    (6): Conv2d(192, 384, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1))
    (7): ReLU(inplace)
    (8): Conv2d(384, 256, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1))
    (9): ReLU(inplace)
    (10): Conv2d(256, 256, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1))
    (11): ReLU(inplace)
    (12): MaxPool2d(kernel_size=3, stride=2, padding=0, dilation=1, ceil_mode=False)
  )
  (avgpool): AdaptiveAvgPool2d(output_size=(6, 6))
  (classifier): Sequential(
    (0): Dropout(p=0.5)
    (1): Linear(in_features=9216, out_features=4096, bias=True)
    (2): ReLU(inplace)
[...]

PR #609, opened 2019-05-24
Add Adaptive Average Pooling and Transposed Convolutions to the Pytorch Deep Explainer

gabrieltseng:
Adds adaptive average pooling (prevents the warning in #593 ) and transposed convolutional layers to the PyTorch Deep Explainer


slundberg:
Sorry I have been out of the loop recently @gabrieltseng ...I am going to merge this under the assumption that travis will be happy now that @imatiach-msft fixed the XGB issue. If it causes a travis failure then we can deal with that since these are not massive changes code-wise

Thanks!

Issue #615, opened 2019-05-29
Broken Tree Explainer in latest version


prempiyush:
Currently, when we use TreeExplainer for a GradientBoostingClassifier.. we are always getting an error:

```
File "/usr/lib/python3.6/site-packages/shap/explainers/tree.py", line 96, in __init__
    self.model = TreeEnsemble(model, self.data, self.data_missing)
  File "/usr/lib/python3.6/site-packages/shap/explainers/tree.py", line 484, in __init__
    assert False, "Unsupported init model type: " + str(type(model.init_))
AssertionError: Unsupported init model type: <class 'sklearn.ensemble.gradient_boosting.LogOddsEstimator'>
```

This is because of a recent change to `tree.py`:

```
# currently we only support the logs odds estimator
if str(type(model.init_)).endswith("ensemble.gradient_boosting.LogOddsEstimator'>"):
    self.base_offset = model.init_.prior
    self.tree_output = "log_odds"
if str(type(model.init_)).endswith("sklearn.dummy.DummyClassifier'>"):
    self.base_offset = scipy.special.logit(model.init_.class_prior_[1]) # with two classes the trees only model the second class
    self.tree_output = "log_odds"
else:
    assert False, "Unsupported init model type: " + str(type(model.init_))
```

The `else` block is now associated with second `if` only.. The second `if` should be an `elif`

imatiach-msft:
@prempiyush closing this issue as your fix has been merged to master, feel free to reopen if there are still any issues

PR #616, opened 2019-05-29
Fix for #615

prempiyush:
Changed the second `if` to an `elif`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,15 @@ os:
   - linux
 python:
   - '3.6'
+before_install:
+  - sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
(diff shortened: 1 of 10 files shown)


imatiach-msft:
@prempiyush this is a great fix, could you please rebase to latest master which fixes the build?  Also, it would be great if you could add a test corresponding to this change, if not that's ok too, I can merge the change once the build is green.

prempiyush:
@imatiach-msft Hi. I have rebased the latest master. Needed to merge it as well to be able to push it.

slundberg:
Thanks!

prempiyush:
@slundberg Thanks for the merge. When do you think this will be available on pypi?

slundberg:
It is on there now :)

PR #618, opened 2019-05-29
Fix shap import errors on python environment with a faulty matplotlib

hchandola:
The try, except blocks in colors.py take care of matplotlib failures but the function dependence_plot outside try, except, accesses the red_blue property in colors - which results in failed shap import on a system with misbehaving matplotlib (due to a missing backend).

This PR contains a fix for this. I have verified the fix manually in an environment that contains a missing matplotlib backend.


imatiach-msft:
@hchandola can you please update the PR with latest master?  The build has been fixed.

imatiach-msft (approved):
discussed offline - LGTM

PR #619, opened 2019-05-30
fix travis build due to xgboost install error

imatiach-msft:
using suggestion from here:
https://github.com/dmlc/xgboost/issues/1786
specifically:
https://github.com/dmlc/xgboost/issues/1786#issuecomment-299387401

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,15 @@ os:
   - linux
 python:
   - '3.6'
+before_install:
+  - sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
+  - sudo apt-get update
+  - sudo apt-get install -y gcc-4.9 g++-4.9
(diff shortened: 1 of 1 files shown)


slundberg:
Thanks for working on this!! I have been finishing up job stuff and am defending next week so I have been behind on issues.

imatiach-msft:
I found the same error here:
https://forums.synfig.org/t/solved-gomp-4-0-not-found-ubuntu-14-04-derivated-distros-related/9137/3
Which suggested that the issue was that with ubuntu 14.04 gomp 4.0 does not work with the default older version of gcc/g++, at least version 4.9 is required.
This led to the suggestion from here:
https://askubuntu.com/questions/466651/how-do-i-use-the-latest-gcc-on-ubuntu/581497#581497
Which fixed the build.

imatiach-msft:
@slundberg I think this PR is good to merge to master, will merge this tomorrow morning if there are no new comments to unblock other builds for PRs

slundberg:
Thanks again! This was a much needed fix!

PR #622, opened 2019-05-30
fixed using console tqdm in jupyter notebooks by using tqdm.autonotebook selecting the right variant automatically

KOLANICH:


slundberg:
Thanks!

PR #643, opened 2019-06-12
Fix sklearn GradientBoostingRegressor bug

nasir-bhanpuri:
if to elif
After making this change locally, the code executes as expected.


imatiach-msft (approved):
nice fix!

slundberg:
Thanks!

PR #656, opened 2019-06-22
PyTorch Deep Explainer: Make sure everything is on the same device

### LINKED ISSUE #629: DeepPytorch for cpu and cuda tensors in same time.
I've created `shap.DeepExplainer` object with input data torch.Tensor on cuda.
```python
tensor = torch.Tensor(data).cuda()
loader = DataLoader(TensorDataset(tensor, tensor), batch_size=128 shuffle=False)

batches = []
for xdata, _ in loader:
     batches.append(xdata)
shap_input = torch.cat(batches, 0).cuda()

explainer = shap.DeepExplainer(model, shap_input)
```
But if I call `explainer.shap_values(X_test)`, where X_test - cpu tensor, I will receive next exceprion:
```
joint_x = [torch.cat((tiled_X[l], self.data[l]), dim=0) for l in range(len(X))]
RuntimeError: Expected object of backend CPU but got backend CUDA for sequence element 1 in sequence argument at position #1 'tensors'
```

gabrieltseng:
This resolves #629 , by making sure the test tensors are on the same device as the model and the background tensors

### DIFF
--- a/shap/explainers/deep/deep_pytorch.py
+++ b/shap/explainers/deep/deep_pytorch.py
@@ -52,6 +52,10 @@ class PyTorchDeepExplainer(Explainer):
         self.num_outputs = 1
         with torch.no_grad():
             outputs = model(*data)
+
+            # also get the device everything is running on
+            self.device = outputs.device
+
             if outputs.shape[1] > 1:
                 self.multi_output = True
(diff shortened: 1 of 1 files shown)


slundberg:
Thanks!

PR #668, opened 2019-06-26
Fix force plot UnicodeEncodeError, add reduced HTML option

### LINKED ISSUE #665: save_html() - UnicodeEncodeError with Force Plot (Partial Solution Found)
Hi,

I've been trying to use the save_html() function to save a force plot returned from DeepExplainer. 
I have no problem saving the plot as such:
 `plot =shap.force_plot(
                    explainer.expected_value[0],
                    shap_values[0][0], features = original_feature_values,
                    feature_names= feature_names)`
It produces an ipython HTML object as expected.

However when I call `shap.save_html('explainer.html', plot)` on it I get the following error `UnicodeEncodeError: 'charmap' codec can't encode character '\u0100' in position 34699: character maps to <undefined>`

I looked into the error and it's usually the result of not explicitly encoding characters as 'utf-8' .
I went through the code in force.py and into the `save_html()` function on line 217 and found by changing the following:
`if type(out_file) == str:
        out_file = open(out_file, "w")
        internal_open = True
    out_file.write("<html><head><script>\n")`
on line 222 to:
`      if type(out_file) == str:
        out_file = open(out_file, "w",encoding='utf-8')
        internal_open = True
    out_file.write("<html><head><script>\n")`
I can get it to work again. 

This is has solved the error for me and the `explainer.html` file is now being generated, however I think this is only a partial solution as the force plot is rendering somewhat incorrectly (see the attached image) 
Capture_new

Any thoughts on what might be going on here ? Thanks.

Rypo:
Fixes #665 by including utf-8 encoding on file write and adding
meta tag with charset information into head tag string. Adds optional
parameter to save_html to allow reduced HTML output. This is useful
for those who wish to embed plots into an already existing webpage.
Parameter defaults to full HTML to avoid changing current behavior.

### DIFF
--- a/shap/plots/force.py
+++ b/shap/plots/force.py
@@ -214,26 +214,45 @@ def initjs():
         "<script>{bundle_data}</script>".format(bundle_data=bundle_data)
     ))
 
-def save_html(out_file, plot_html):
+def save_html(out_file, plot_html, full_html=True):
     """ Save html plots to an output file.
+    
+    Parameters
+    ----------
+    out_file : str or file
+        Location or file to be written to
+    plot_html : HTML Object
+        HTML object returned by shap.force_plot()
+    full_html : boolean (default: True)
+        If True, writes a complete HTML document starting 
+        with an <html> tag. If False, only script and div
+        tags are included.
     """
 
     internal_open = False
     if type(out_file) == str:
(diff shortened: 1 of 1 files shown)


slundberg:
Thanks!

PR #671, opened 2019-06-29
give an expected_value for each output of multi-target estimators

Kylecrif:
changes made in https://github.com/slundberg/shap/commit/f616f76b14a19e61c9a2087d039fc93dfa1e3012 (version `0.29.3`) caused `explainer.expected_value` to only output the first expected_value for multi-target estimators, which was not the case in version `0.29.2`

* only unwraps a 1-element array rather than truncating a sequence
* adds testing for multi-target regression


slundberg:
Sorry for the slow review. This is great, thanks!

PR #674, opened 2019-07-01
fix edge case with lightgbm 2.2.1 where trees may exist but have no nodes causing error in TreeExplainer

imatiach-msft:
In some edge cases, eg where a regressor is trained on constant labels, lightgbm can produce empty trees which fails with TreeExplainer, giving the error:

>       return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
E       ValueError: zero-size array to reduction operation maximum which has no identity

with stack trace:

..\..\shap\explainers\tree.py:96: in __init__
    self.model = TreeEnsemble(model, self.data, self.data_missing)
..\..\shap\explainers\tree.py:614: in __init__
    max_nodes = np.max([len(t.values) for t in self.trees])
c:\users\ilmat\appdata\local\continuum\miniconda3\envs\sh\lib\site-packages\numpy\core\fromnumeric.py:2505: in amax
    initial=initial)

Note this only seems to occur with lightgbm older than 2.2.1, as latest lightgbm 2.2.3 doesn't seem to have this issue.  However, the check seems simple enough to be a good edge case scenario to add in for robustness.

### DIFF
--- a/shap/explainers/tree.py
+++ b/shap/explainers/tree.py
@@ -610,7 +610,7 @@ class TreeEnsemble:
             raise Exception("Model type not yet supported by TreeExplainer: " + str(type(model)))
         
         # build a dense numpy version of all the tree objects
-        if self.trees is not None:
+        if self.trees is not None and self.trees:
             max_nodes = np.max([len(t.values) for t in self.trees])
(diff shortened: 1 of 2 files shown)


imatiach-msft:
It looks like this PR would also fix this issue for xgboost, which was closed but still seems like a good edge case to fix:
https://github.com/slundberg/shap/issues/400

slundberg:
Thanks! LGTM

PR #676, opened 2019-07-03
remove keys

aquiire:
for security reasons

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,6 @@ before_script:
 #   provider: pypi
 #   user: "shap_poster"
 #   password:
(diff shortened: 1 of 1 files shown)


slundberg:
We could do this, the trouble is that anyone can submit a PR that will print `$Some_Variable` so it is not really more secure. Thoughts?

jamesmyatt:
Surely this is fine since the key is encrypted here. Only Travis can decrypt it.

Although, maybe switching CI provider is better in the long term: https://www.youtube.com/watch?v=9G2s1TN9QQY

thatlittleboy:
Closing since we no longer use travis ci.

PR #684, opened 2019-07-08
Pytorch DeepExplainer: retain graph for multiple inputs

gabrieltseng:
Currently, if the DeepExplainer is used with multiple inputs, the following error is raised:
```
RuntimeError: Trying to backward through the graph a second time, but the buffers have already been freed. Specify retain_graph=True when calling backward the first time.
```
This PR ensures the graph is retained if necessary, and adds a test for a multi-input model.


slundberg:
Excellent, thanks!

PR #686, opened 2019-07-09
Do not open xserver in Travis CI

trams:
How to reproduce:

1. Clone the repo
2. Set up Travis CI with a fork

it fails due to the fact that xvfb is not present

What I want to get: green tests

I am quite new to Travis CI so I do not exactly understand why it works for you but not for me.
My guess there is either some configuration not commited to repo or just cached build images which for some reason have this `xvfb`

Also I did not quite get why do we use it. Here I may be doing a stupid mistake by removing `xvfb`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,10 +13,6 @@ install:
   - pip install .
   - pip install tensorflow keras xgboost
 script: python setup.py nosetests
-before_script:
-  - export DISPLAY=:99.0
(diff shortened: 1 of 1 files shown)


slundberg:
Thanks! Not sure why it caused you issues, but it is not used right now, and so we can cut it out. (It was put in to support plotting tests in Travis that didn't work out for other reasons.)

PR #690, opened 2019-07-09
fix output of shap TreeExplainer with lightgbm to be list of ndarray in binary case to conform to other explainers and scikit learn output

imatiach-msft:
Fix output of shap TreeExplainer with lightgbm to be list of ndarray in binary case to conform to other explainers and scikit learn output.

Currently SHAP TreeExplainer outputs a single ndarray of shap values for LightGBMClassifier in binary classification case.  However, for scikit-learn models in both binary and multiclass classification the result is always a list of ndarray, one for each class (in binary case there are two ndarrays).  This PR fixes this scenario to conform to the scikit learn API and other shap explainers.  In general, we should expect SHAP to output list of ndarray for classification case and a single ndarray in regression case.


slundberg (changes requested):
This all looks good to me. But since it is a significant API change would you mind putting in a warning when this is triggered? We can leave that warning in for only one version, but I think flagging the change is important.

Also have you looked at any of the example notebooks that might be effected by this? Just want to make sure we don't break them :)

imatiach-msft:
@slundberg yes, it looks like this notebook broke: "Census income classification with LightGBM"
will update

imatiach-msft:
@slundberg I've updated the notebook, thanks

slundberg (approved):

slundberg:
Looks great, thanks!

PR #691, opened 2019-07-09
[minor] fix retrieving default session for tensorflow in TF DeepExplainer for 1.11.0

imatiach-msft:
I didn't add a test for this - to repro just run the existing tests with tensorflow version 1.11.0.  It seems that with latest 1.14.0 the code ```tf.compat.v1.keras``` works though.


slundberg:
Thanks! Looks great.

slundberg (approved):

Issue #384, opened 2019-01-09
Shap 0.27.0 incompatible with lightgbm 2.2.2


sei-dleung:
when trying to call shap.TreeExplainer() with a Booster object from lightgbm, I get the following error:

Error in py_call_impl(callable, dots$args, dots$keywords) : 
  AttributeError: 'Booster' object has no attribute '_Booster__name_inner_eval'

Detailed traceback: 
  File "/Users/dleung/miniconda3/lib/python3.6/site-packages/shap/explainers/tree.py", line 91, in __init__
    self.model = TreeEnsemble(model)
  File "/Users/dleung/miniconda3/lib/python3.6/site-packages/shap/explainers/tree.py", line 491, in __init__
    self.objective = objective_name_map.get(model._Booster__name_inner_eval[0], None)
Calls: <Anonymous> -> py_call_impl
Execution halted

This issue doesn't exist when using shap 0.21.0 with lightgbm 2.2.2

slundberg:
Thanks! That part is only need for explaining the loss function, so I wrapped it to fail silently in the new lightgbm. Later we can look for how the naming changed

slundberg:
This should now be fixed for good on v0.28

BlaBlaPer:
Hi @slundberg, I encountered a similar problem 'AttributeError: 'Booster' object has no attribute 'params'' when I use lightGBM version == 2.1.2 and shap version == 0.28.4. When I use shap version == 0.21.0, this problem solved~

slundberg:
@BlaBlaPer I think that might be because of a change in LightGBM 2.2.x could you see if that fixes the issue?

DSXiangLi:
@slundberg I am using shap 0.28.5, and lightgbm 2.2.3. I still see this error message: AttributeError: 'Booster' object has no attribute 'params'. Hasn't been fixed in 0.28.x version ?

bbennett36:
@DSXiangLi have you figured out a solution to this?

Issue #278, opened 2018-10-03
Remove matplotlib from install requirements in setup.py


jack1728:
In setup.py
install_requires=['numpy', 'scipy', 'scikit-learn', 'matplotlib', 'pandas', 'tqdm', 'ipython'],

matplotlib is not strictly required, since you have all these try/catch statements around importing.

It might be a good idea to remove matplotlib from the install_requires list, so that people who are trying to use this library without the plotting and visualization functions don't need to install matplotlib

slundberg:
Just a note, for others interested, follow the discussion in #270

slundberg:
Solved now, see #466

PR #708, opened 2019-07-20
Made pandas, matplotlib, ipython and scikit-image optional (on the source code alone)

jorgecarleitao:
This PR changes the source code for it to only depend on `['numpy', 'scipy', 'scikit-learn', 'tqdm']`. This is suitable for a production system. It partially addresses #466 and #278, and fully addresses #384.

Note that this does not change `setup.py`, only the source code in `shap/shap`. Consequently, it does not completely solve the problem (it is currently unsolvable), it only helps solving it.

## Rational

Currently, installing shap adds a considerable overhead and a small security concern as it depends on many packages (#466). One way to approach this is to split the `install_requires` in multiple `extra_requires`.

However, as @slundberg pointed out in #508, conda-forge does not support extras, and many users expect `pip install shap` to be installed with plotting functionality, and pip does not support a default extra, see [here](https://github.com/pypa/setuptools/issues/1139)). Thus, changing `install_requires` `setup.py` is not feasible at this point as it breaks backward compatibility and drops support for conda-forge.

Another solution is to create a new repository, `shap-core`, that only contains the explainers, and leave this package for all things combined. However, this adds some maintenance cost that @slundberg would have to be willing to maintain.

The current compromise is for developers that want to use a minimal version of this package to fork this repository, change it, and deploy it from their own fork. 

## Main change

This PR introduces a set of changes to the source code that make this a bit easier: it makes the source code work without `['pandas', 'matplotlib', 'scikit-learn', 'ipython', 'scikit-image']` installed.
I.e. with these changes, the only thing that the fork has to do is to change the `setup.py` for those dependencies not be part of the `install_requires`.
    
This change makes shap depend only on: `['numpy', 'scipy', 'scikit-learn', 'tqdm']`.

### DIFF
--- a/shap/benchmark/plots.py
+++ b/shap/benchmark/plots.py
@@ -416,22 +416,23 @@ def make_grid(scores, dataset, model):
     
     return row_keys, col_keys, data
 
-from matplotlib.colors import LinearSegmentedColormap
-red_blue_solid = LinearSegmentedColormap('red_blue_solid', {
-    'red': ((0.0, 198./255, 198./255),
-            (1.0, 5./255, 5./255)),
 
-    'green': ((0.0, 34./255, 34./255),
-              (1.0, 198./255, 198./255)),
-
-    'blue': ((0.0, 5./255, 5./255),
-             (1.0, 24./255, 24./255)),
-
-    'alpha': ((0.0, 1, 1),
-              (1.0, 1, 1))
-})
-from IPython.core.display import HTML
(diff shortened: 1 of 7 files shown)


iwanbolzern:
Hi all, thank you for your great work! Is there any progress on this?

slundberg:
See #466, but I finally got around to dealing with this, it is now on master and will land in 0.31 so I'll close this PR. Thanks @jorgecarleitao for pushing on this!

PR #709, opened 2019-07-21
Added test that KernelExplainer reduces to linear case in linear model.

jorgecarleitao:
This is similar to #470, and helps at having trust in the explainer.


slundberg:
Thanks @jorgecarleitao !

PR #711, opened 2019-07-22
Small interaction plot improvements

tyler-romero:
Allow `x_jitter` and `alpha` to be used in interaction plots. When plotting main effect, don't color by `interaction_index` (there isn't one).


slundberg:
Thanks! I just made a small formatting change before merging.

PR #713, opened 2019-07-24
Update force.py for fixing an index bug of force_plot()

yunchuankong:
Hi Thanks for the awesome tool for ML explanation. I encountered a bug when plotting force plots and did a quick fix:

The index "i" is doubly used in the for loop of line 144, causing confusion in subsequent lines. 

The bug was detected when I tried to force plot a shap matrix with 100 rows and 346 features. An IndexError popped up saying "345 is out of the range of 100". This was probably due to that "i" had been used to loop over all features, and been re-used to indexing instances. 

I guess it would be a better to change the index variable of the for loop to avoid the issue.


slundberg:
Great catch! Thanks!

Issue #576, opened 2019-05-06
Binary CatBoost classifier using TreeExplainer with model_output='probability' and tree_dependence='Independent'


jayantsahewal:
I know there are similar issues already which talk about multi-class and/or not using feature_dependence='Independent'. I put a comment in one of the issues as well but that has been tagged with label "To Do". I wasn't sure whether someone will be looking at that issue or not. So, I thought of creating an issue here since it's slightly different.

In my case, I am getting the same issue even with a binary CatBoost classifier. Based on the code base, it looks like if model output is not margin and feature_dependence is set to "Independent" even then it will always throw an error because objective attribute for CatBoost classifiers hasn't been defined and it will always remain None. Is it possible to make it work for at least binary CatBoost classifiers?

```
if model_output != "margin":
    if self.model.objective is None and self.model.tree_output is None:
        raise Exception("Model does have a known objective or output type! When model_output is " \
"not \"margin\" then we need to know the model's objective or link function.")


elif str(type(model)).endswith("catboost.core.CatBoostClassifier'>"):
    assert_import("catboost")
    self.model_type = "catboost"
    self.original_model = model
```

dvpolyakov:
@jayantsahewal , luckily I have implemented this feature) I have opened a pull request https://github.com/slundberg/shap/pull/578 and I hope that it will be merged by Scott soon, and the updates will be available in the closest release

jayantsahewal:
@dvpolyakov Thanks for implementing it!

I looked at your code in the pull request and it looks fine to me. Just one comment, I would suggest using np.float64 instead of np.float32 for the dtype. I have experienced many issues with np.float32. Though, those are edge case. But, it's still better to avoid it if possible.

pmeier-tiplu:
minor detail: it should be "probability" (with an `a`)

PR #726, opened 2019-08-05
Removed unnecessary tree building for CatBoost

lupusomniator:
There is no need to build trees, since all shap values are [available from the CatBoost](https://github.com/slundberg/shap/blob/cec69079422afaf9cba35eccefa70092b26744bf/shap/explainers/tree.py#L200). I suggest just get everything back as it was in 0.29.1. The need arose in connection with issue https://github.com/slundberg/shap/issues/662

### DIFF
--- a/shap/explainers/tree.py
+++ b/shap/explainers/tree.py
@@ -598,11 +598,6 @@ class TreeEnsemble:
             assert_import("catboost")
             self.model_type = "catboost"
             self.original_model = model
-            self.dtype = np.float32
-            cb_loader = CatBoostTreeModelLoader(model)
-            self.trees = cb_loader.get_trees(data=data, data_missing=data_missing)
-            self.tree_output = "log_odds"
-            self.objective = "binary_crossentropy"
         elif str(type(model)).endswith("catboost.core.CatBoost'>"):
             assert_import("catboost")
             self.model_type = "catboost"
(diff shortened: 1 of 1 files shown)


slundberg:
I should probably check and make sure that all the same cases are supported, but if the tree building is failing we should at least make it a soft-fail and use CatBoost whenever we can. Perhaps just wrap this in a try rather than a full delete for the moment?

dvpolyakov:
Hello, @lupusomniator . I have implemented the code that you suggest to delete. This tree building is necessary for calculation of SHAP values for binary classification. While the code you are referring to calculates SHAP values only for regression task. Just look at these issues: https://github.com/catboost/catboost/issues/668
https://github.com/slundberg/shap/issues/576

@lupusomniator, hold your horses. To break - not to build.

lupusomniator:
> Hello, @lupusomniator . I have implemented the code that you suggest to delete. This tree building is necessary for calculation of SHAP values for binary classification. While the code you are referring to calculates SHAP values only for regression task. Just look at these issues: [catboost/catboost#668](https://github.com/catboost/catboost/issues/668)
> #576
> 
> @lupusomniator, hold your horses. To break - not to build.

Yeap, I was wrong. I'll close this review and will fix the existing problem by the next commit.

PR #732, opened 2019-08-06
decision plot updates

floidgilbert:
- Updated `decision_plot` notebook to support changes in TreeExplainer output for LightGBM models.
- Fixed documentation error.
- Removed incompatibility with Python 2.
- When `matplotlib=False`, raise more informative error indicating that an interactive version of the plot is not yet supported and what to do about it.

**Suggest:** Create https://slundberg.github.io/shap/notebooks/plots/decision_plot.html and insert associated link in README.md under "Documentation notebooks" next to `shap.dependence_plot`.


slundberg:
Looks great! I am also all for putting the doc notebook as an html link from the ReadMe. You want to add that as another PR? For now the html should go in https://github.com/slundberg/shap/tree/master/docs/notebooks/plots to match the other notebooks.

PR #736, opened 2019-08-08
Fix bug with CatBoost dense tree builder

lupusomniator:
https://github.com/slundberg/shap/issues/662
Previously, it was not taken into account that the features can be categorical. Added support for categorical features when building a dense tree view.


slundberg:
Thanks!

PR #737, opened 2019-08-08
removing import of pyplot as it was not used and created an dependenc…

satylogin:
there was an unnecessary import for matplotlib.pyplot
this created an dependency for tkinter.
removing this as this was not needed

### DIFF
--- a/shap/plots/colors.py
+++ b/shap/plots/colors.py
@@ -7,7 +7,6 @@ import numpy as np
 import skimage.color
 
 try:
-    import matplotlib.pyplot as pl
     import matplotlib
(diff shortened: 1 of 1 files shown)


slundberg:
Thanks. Happy to remove that unneeded line, but since pyplot is used elsewhere I doubt it will change the dependencies of shap.

satylogin:
thanks for the merge :) ... actually it was creating an dependency during import itself, since colors.py was imported during import of shap. It was leading to import error. I think you have it at some other place also where you have put it in a try catch block where it gives a warning.

slundberg:
Ah! That makes sense now. Thanks!

PR #744, opened 2019-08-13
decision_plot - remove matplotlib param. add html documentation.

floidgilbert:
1. HTML documentation for `decision_plot`
a. Added `/docs/notebooks/plots/decision_plot.html`
b. Added link to `/README.md` referencing slunderberg.github.io
2. Removed `matplotlib` parameter. This is not a backwards-compatible change. but no one is using the plot yet. So, best to clean up the API now.


slundberg:
Looks good. I agree that making matplotlib default is good and cleaner.

Issue #745, opened 2019-08-14
IndexError: list index out of range


ibuda:
I am running the following code:
```
from catboost.datasets import *
train_df, _ = catboost.datasets.amazon()
ix = 100
X_train = train_df.drop('ACTION', axis=1)[:ix]
y_train = train_df.ACTION[:ix]
X_val = train_df.drop('ACTION', axis=1)[ix:ix+20]
y_val = train_df.ACTION[ix:ix+20]
model = CatBoostClassifier(iterations=100, learning_rate=0.5, random_seed=12)
model.fit(X_train, y_train, eval_set=(X_val, y_val), verbose=False, plot=False)
shap.TreeExplainer(model)
```
I get the following error:
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-2-6d52aef09dc8> in <module>
      8 model = CatBoostClassifier(iterations=100, learning_rate=0.5, random_seed=12)
      9 model.fit(X_train, y_train, eval_set=(X_val, y_val), verbose=False, plot=False)
---> 10 shap.TreeExplainer(model)

~/prb/anaconda3/lib/python3.6/site-packages/shap/explainers/tree.py in __init__(self, model, data, model_output, feature_dependence)
     94         self.feature_dependence = feature_dependence
     95         self.expected_value = None
---> 96         self.model = TreeEnsemble(model, self.data, self.data_missing)
     97 
     98         assert feature_dependence in feature_dependence_codes, "Invalid feature_dependence option!"

~/prb/anaconda3/lib/python3.6/site-packages/shap/explainers/tree.py in __init__(self, model, data, data_missing)
    594             self.dtype = np.float32
    595             cb_loader = CatBoostTreeModelLoader(model)
--> 596             self.trees = cb_loader.get_trees(data=data, data_missing=data_missing)
    597             self.tree_output = "log_odds"
    598             self.objective = "binary_crossentropy"

~/prb/anaconda3/lib/python3.6/site-packages/shap/explainers/tree.py in get_trees(self, data, data_missing)
   1120 
   1121             # load the per-tree params
-> 1122             depth = len(self.loaded_cb_model['oblivious_trees'][tree_index]['splits'])
   1123 
   1124             # load the nodes

IndexError: list index out of range
````
This error was spotted with Catboost version 0.15.2, I upgraded to the latest version (0.16.4 as of today), but the error persists.
I have Shap version: '0.29.3'

ibuda:
[...]

PR #749, opened 2019-08-15
Fix for  CatBoostClassifier TreeExplainer with evaluation and early stopping

ibuda:
Fix for get_trees() method of CatBoostTreeModelLoader tree explainer (causes issue #745).
The number of trees is not always equal to the number of iterations. In cases when there is early stopping, and/or evaluation set is used, then the number of trees is equal to model.tree_count_ or just the size of the model['oblivious_trees'] list in the model's json file dict.

### DIFF
--- a/shap/explainers/tree.py
+++ b/shap/explainers/tree.py
@@ -1192,7 +1192,7 @@ class CatBoostTreeModelLoader:
         self.loaded_cb_model = json.load(open("cb_model.json", "r"))
 
         # load the CatBoost oblivious trees specific parameters
-        self.num_trees = self.num_trees = len(self.loaded_cb_model['oblivious_trees'])
+        self.num_trees = len(self.loaded_cb_model['oblivious_trees'])
         self.max_depth = self.loaded_cb_model['model_info']['params']['tree_learner_options']['depth']
 
     def get_trees(self, data=None, data_missing=None):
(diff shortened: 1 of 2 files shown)


shap/explainers/tree.py
```diff
@@ -1192,7 +1192,7 @@ def __init__(self, cb_model):
         self.loaded_cb_model = json.load(open("cb_model.json", "r"))
 
         # load the CatBoost oblivious trees specific parameters
-        self.num_trees = self.loaded_cb_model['model_info']['params']['boosting_options']['iterations']
+        self.num_trees = self.num_trees = len(self.loaded_cb_model['oblivious_trees'])
```
AbhinavanT: Might want to remove the double `self.num_trees = self.num_trees =`

ibuda: Good point. Done.

slundberg:
Thanks @ibuda! To make sure this says fixed could you also add a short unit test to test_tree.py that would fail before you added this?

ibuda:
@slundberg Please check if the test looks good to you. I had no access to CatBoostTreeModelLoader.num_trees to assert equals it against the model's tree_count_, so I just called the class initialization of TreeExplainer on the model that triggers the error.

slundberg:
Sorry for losing track of this! I really appreciate you sending this in. Staying on top of Github issues has been a challenge, but I didn't realize I had let any PRs get stuck.

Thanks for doing this @ibuda!

PR #754, opened 2019-08-16
fix edge case where background data is dense and evaluation data is sparse

imatiach-msft:
follow-up bug fix for https://github.com/slundberg/shap/issues/569 and https://github.com/slundberg/shap/issues/138
There is an edge case where the background data is dense and the test data is sparse.  In this case, we convert the test data to dense prior to setting it on the background when generating samples.


slundberg:
Looks great!

PR #755, opened 2019-08-16
fixed linear shap explainer shape for classifiers to be list of ndarray

imatiach-msft:
follow-up to discussion on https://github.com/slundberg/shap/pull/645 , specifically:
https://github.com/slundberg/shap/pull/645#discussion_r293589975 
was the workaround needed for anything in particular?  I validated the tests pass with the multi-class workaround removed.

### DIFF
--- a/shap/explainers/linear.py
+++ b/shap/explainers/linear.py
@@ -57,13 +57,8 @@ class LinearExplainer(Explainer):
 
         # sklearn style model
         elif hasattr(model, "coef_") and hasattr(model, "intercept_"):
-            # work around for multi-class with a single class
-            if len(model.coef_.shape) > 1 and model.coef_.shape[0] == 1:
-                self.coef = model.coef_[0]
-                self.intercept = model.intercept_[0]
-            else:
-                self.coef = model.coef_
-                self.intercept = model.intercept_
+            self.coef = model.coef_
+            self.intercept = model.intercept_
         else:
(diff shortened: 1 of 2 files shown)


slundberg:
I'll need to look more at this. I think it was for some sklearn models that one would think are single class but are stored with a single dimension that makes it look like a multi-class output (with one class).

imatiach-msft:
@slundberg ah, yes, in some cases we have binary classifiers with only one output, and in very rare cases we can have a multiclass classifier with one class (which scikit learn just fails for in their classifiers).  Either could create this 2D coef, but I was mostly concerned about the output of the shap values being an ndarray instead of a list of ndarray, since the difference between the classifier and regressor in our contract is that classifiers output a list.  There is another issue where binary classifiers should really have a 2D coef array with two classes instead of one.
That's exactly what happens with logistic regression in scikit learn:
https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html
It stores the coefficients as a 2D list of one class in binary case, eg as an example:
[[0.2, 0.4, -0.1, 0.7]]

slundberg:
Makes sense. If this is critical to merge soon let me know. Otherwise I'll snooze this for a few weeks until I have some solid time to consider the implications for current usage. Thanks!

imatiach-msft:
not critical at all, thanks!

PR #763, opened 2019-08-21
Handle 1d output tensor

jotsif:
Don't crash on 
```
  File "/usr/local/lib/python3.7/site-packages/shap/explainers/gradient.py", line 73, in __init__
    self.explainer = _PyTorchGradientExplainer(model, data, batch_size, local_smoothing)
  File "/usr/local/lib/python3.7/site-packages/shap/explainers/gradient.py", line 361, in __init__
    if outputs.shape[1] > 1:
IndexError: tuple index out of range
```
when the output tensor from the pytorch model is one-dimensional.

### DIFF
--- a/shap/explainers/gradient.py
+++ b/shap/explainers/gradient.py
@@ -358,7 +358,7 @@ class _PyTorchGradientExplainer(Explainer):
 
         multi_output = False
         outputs = self.model(*self.model_inputs)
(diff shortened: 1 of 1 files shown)


slundberg:
Thanks for the fix! (and sorry for the slow merge, was out of town for a bit)

PR #768, opened 2019-08-24
Fixed kernel explainer expected value with single value output

parkerzf:
The expected_value of kernel explainer is dependent to fnull but it is initialized before the change of fnull w.r.t. single value output predict function. 

Example to show the unexpected case:

```python
import pandas as pd
import shap

def predict(X):
    return X.sum(axis=1)

X = pd.DataFrame({(1,), (2,)})

explainer = shap.KernelExplainer(predict, X)
```
explainer.expected_value is `array(1.5)` while the correct value should be `array([1.5])`. So I moved the initialization of expected_value to the end of the init function to fix it.

### DIFF
--- a/shap/explainers/kernel.py
+++ b/shap/explainers/kernel.py
@@ -118,7 +118,6 @@ class KernelExplainer(Explainer):
         if isinstance(model_null, (pd.DataFrame, pd.Series)):
             model_null = np.squeeze(model_null.values)
         self.fnull = np.sum((model_null.T * self.data.weights).T, 0)
-        self.expected_value = self.linkfv(self.fnull)
 
         # see if we have a vector output
         self.vector_out = True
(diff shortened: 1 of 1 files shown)


slundberg:
Thanks! But why should it be `array([1.5])`?

parkerzf:
Hi @slundberg ,
because index access to retrieve value from array(1.5)
```python
shap.force_plot(explainer.expected_value, shap_value[0], X.iloc[0])
```

shows exceptions: `IndexError: too many indices for array`.

slundberg:
Ah. I think it should actually be `1.5` then and not `array([1.5])`. The issue is that the result is left as a zero dimensional numpy array. See if the change I added addresses your concern. Thanks again!

parkerzf:
Cool, works perfectly!

PR #778, opened 2019-08-27
Fix broken link in README.md

floidgilbert:
Line-break in README.md broke the link to the `decision_plot` documentation and displayed markdown code. Normally, I wouldn't submit such a small change, but it affects the project's main page.


slundberg:
Ah! Good catch and thanks!

PR #784, opened 2019-08-30
Add sklearn isolation forest shap values

JiechengZhao:
Add isolation forest shap values

the value in sklearn isolation forest tree is not right. recalculate them based on https://github.com/scikit-learn/scikit-learn/blob/1495f6924/sklearn/ensemble/iforest.py .

### DIFF
--- a/shap/explainers/tree.py
+++ b/shap/explainers/tree.py
@@ -1011,12 +1011,12 @@ class IsoTree(Tree):
     In sklearn the tree of the Isolation Forest does not calculated in a good way.
     """
     def __init__(self, tree, normalize=False, scaling=1.0, data=None, data_missing=None):
-        super().__init__(self, tree, normalize, scaling, data, data_missing)
+        super(IsoTree, self).__init__(tree, normalize, scaling, data, data_missing)
         if str(type(tree)).endswith("'sklearn.tree._tree.Tree'>"):
             from sklearn.ensemble.iforest import _average_path_length
 
             def _recalculate_value(tree, i , level):
-                value = level + _average_path_length(tree.n_node_samples[i])
+                value = level + _average_path_length(np.array([tree.n_node_samples[i]]))
                 self.values[i, 0] =  value
                 if tree.children_left[i] > 0 :
                     _recalculate_value(tree, tree.children_left[i] , level + 1)
--- a/shap/explainers/tree.py
+++ b/shap/explainers/tree.py
@@ -1011,7 +1011,7 @@ class IsoTree(Tree):
     In sklearn the tree of the Isolation Forest does not calculated in a good way.
     """
     def __init__(self, tree, normalize=False, scaling=1.0, data=None, data_missing=None):
-        super().__init__(self, tree, normalize, scaling, data)
+        super().__init__(self, tree, normalize, scaling, data, data_missing)
         if str(type(tree)).endswith("'sklearn.tree._tree.Tree'>"):
             from sklearn.ensemble.iforest import _average_path_length
 
@@ -1027,10 +1027,6 @@ class IsoTree(Tree):
             if normalize:
                 self.values = (self.values.T / self.values.sum(1)).T
             self.values = self.values * scaling
-
-
-
-
 
  
 def get_xgboost_json(model):
--- a/shap/explainers/tree.py
+++ b/shap/explainers/tree.py
@@ -424,6 +424,12 @@ class TreeEnsemble:
             self.trees = [Tree(e.tree_, scaling=scaling, data=data, data_missing=data_missing) for e in model.estimators_]
             self.objective = objective_name_map.get(model.criterion, None)
             self.tree_output = "raw_value"
+        elif str(type(model)).endswith("sklearn.ensemble.iforest.IsolationForest'>"):
+            self.dtype = np.float32
+            scaling = 1.0 / len(model.estimators_) # output is average of trees
+            self.trees = [IsoTree(e.tree_, scaling=scaling, data=data, data_missing=data_missing) for e in model.estimators_]
+            self.objective = objective_name_map.get('mse', None)
+            self.tree_output = "raw_value"
         elif str(type(model)).endswith("skopt.learning.forest.RandomForestRegressor'>"):
             self.dtype = np.float32
             scaling = 1.0 / len(model.estimators_) # output is average of trees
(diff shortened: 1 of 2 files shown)


slundberg:
Thanks! However, I am having trouble figuring out what value is being explained. It is not the output of `iso.score_samples(X)`. Is there a transform that will make it match that?

JiechengZhao:
It should be 2**(value / average path length). but when I try add assert sentence, the result doesn't match. I will try to figure that out later.

slundberg:
Thanks. Let me know if you run into questions.

JiechengZhao:
In previous version the average_path_length of a parent node is estimated by its samples count. This leads to `node_value * node_sample_count  != left_child_value * left_child_sample_count  +  right_child_value * right_child_sample_count`。 fix this by set node_value of a parent vie its childrens value and count.

JiechengZhao:
besides not quite understand the usage of `self.objective = objective_name_map.get('mae', None)`, but looks required by the program.

slundberg:
Thanks! This looks good now :) ...as for the objective parameter...it is only used when explaining the loss of a model, and needs C++ support. So I just removed it since we don't support explaining the loss of an iso forest right now (not setting it leaves it as `None`, which denotes a lack of support for that feature).

JiechengZhao:
Is there anything else need to add before merge?

legchikov:
@slundberg Could you please comment when this pull request will be merged?

ssankarasubramanian:
@JiechengZhao Thank you for working on this. I was looking for this solution.

@slundberg Any update on when this might be merged to code base?.

slundberg:
Sorry I lost track of this! Merging now :) ...and thanks again @JiechengZhao !

adrianperezbote:
@JiechengZhao Thank you very much. I have used this code on a ML project and has been really helpful

JiechengZhao:
@slundberg Thank you for maintaining the great tool. Now I can use the code through public repo. 😄

srikiran-cpn:
@JiechengZhao : can I ask why you choose to explain 2**(value / average path length)? I would expect something like explaining the output of IsolationForest.predict() i.e., simply using the 1/-1.

pjgao:
> @JiechengZhao : can I ask why you choose to explain 2**(value / average path length)? I would expect something like explaining the output of IsolationForest.predict() i.e., simply using the 1/-1.

In fact, `IsolationForest`'s `predict` is calculated by `score_sample`. `score_sample` gives an anomaly score calculated by the average path of the data sample.

```python
    scores = 2 ** (
        -depths
        / (len(self.estimators_)
            * _average_path_length([self.max_samples_]))
    )
```
Currently, `shap` for `IsolationForest` explains `depths/(len(self.estimators_)` (mean split path length).

We can see that:
```python
import numpy as np
import shap
from sklearn.ensemble import IsolationForest
from sklearn.ensemble.iforest import _average_path_length

X, y = shap.datasets.boston()
model = IsolationForest(max_features=1.0,random_state=1)
model.fit(X)


test = X.iloc[[0],:]
score = model.score_samples(test)

shap_explained = explainer.shap_values(test).sum()+explainer.expected_value
#score = 2 ** (iso_calculated/ _average_path_length([self.max_samples_]))
iso_calculated = -np.log2(-score) * \
    _average_path_length(np.array([model.max_samples_]))
print(shap_explained, iso_calculated)
# [11.52547704] [11.52547704]
```
from `shap` explained to `sklearn` output, we can:
```python
print(2**(-shap_explained/_average_path_length([model.max_samples_])))
print(score)
# [0.45849862]
# [-0.45849862]
```
for expected value, it is calculated by:

```python
def recalculate_node_value(tree):
    values = tree.value.reshape((tree.value.shape[0]), 1)

    def _recalculate_value(tree, i, level):

        if tree.children_left[i] == -1 and tree.children_right[i] == -1:
            value = level + \
                _average_path_length(np.array([tree.n_node_samples[i]]))[0]
            #print(tree.n_node_samples[i])
            values[i, 0] = value
            return value * tree.n_node_samples[i]
        else:
            value_left = _recalculate_value(
                tree, tree.children_left[i], level + 1)
            value_right = _recalculate_value(
                tree, tree.children_right[i], level + 1)
            values[i, 0] = (value_left + value_right) / tree.n_node_samples[i]
            return value_left + value_right
    _recalculate_value(tree, 0, 0)
    return values

expected_value = sum((recalculate_node_value(est.tree_) /
                      model.n_estimators)[0][0] for est in model.estimators_)
print(expected_value)
print(explainer.expected_value)
# 11.908404682235796
# [11.90840468]
```

JiechengZhao:
> @JiechengZhao : can I ask why you choose to explain 2**(value / average path length)? I would expect something like explaining the output of IsolationForest.predict() i.e., simply using the 1/-1.

The true might be simpler than expected. When I implemented the function, I just tried to follow [the suggestions](https://github.com/slundberg/shap/issues/237#issuecomment-415786788) here. To do this, I should reuse lots of logic in `IsolationForest` in Sklearn and SHAP in explaining other tree models. The result was the calculated value is a raw value, not the value given by the predict. But it is O.K. Actually, it is common that SHAP explain a raw value of the final result. For example [catboost](https://github.com/slundberg/shap/blob/9d7de4dee742ce10684400171257c66a86e4c7a6/tests/explainers/test_tree.py#L417) and [lightgbm](https://github.com/slundberg/shap/blob/9d7de4dee742ce10684400171257c66a86e4c7a6/tests/explainers/test_tree.py#L364). 

It might be a good idea to embed this part `_average_path_length(np.array([iso.max_samples_]))[0]` to the shap value. I will try it later.

WaterKnight1998:
> > @JiechengZhao : can I ask why you choose to explain 2**(value / average path length)? I would expect something like explaining the output of IsolationForest.predict() i.e., simply using the 1/-1.
> 
> The true might be simpler than expected. When I implemented the function, I just tried to follow [the suggestions](https://github.com/slundberg/shap/issues/237#issuecomment-415786788) here. To do this, I should reuse lots of logic in `IsolationForest` in Sklearn and SHAP in explaining other tree models. The result was the calculated value is a raw value, not the value given by the predict. But it is O.K. Actually, it is common that SHAP explain a raw value of the final result. For example [catboost](https://github.com/slundberg/shap/blob/9d7de4dee742ce10684400171257c66a86e4c7a6/tests/explainers/test_tree.py#L417) and [lightgbm](https://github.com/slundberg/shap/blob/9d7de4dee742ce10684400171257c66a86e4c7a6/tests/explainers/test_tree.py#L364).
> 
> It might be a good idea to embed this part `_average_path_length(np.array([iso.max_samples_]))[0]` to the shap value. I will try it later.

Did you continue this at the end?

JiechengZhao:
> > > @JiechengZhao : can I ask why you choose to explain 2**(value / average path length)? I would expect something like explaining the output of IsolationForest.predict() i.e., simply using the 1/-1.
> > 
> > 
> > The true might be simpler than expected. When I implemented the function, I just tried to follow [the suggestions](https://github.com/slundberg/shap/issues/237#issuecomment-415786788) here. To do this, I should reuse lots of logic in `IsolationForest` in Sklearn and SHAP in explaining other tree models. The result was the calculated value is a raw value, not the value given by the predict. But it is O.K. Actually, it is common that SHAP explain a raw value of the final result. For example [catboost](https://github.com/slundberg/shap/blob/9d7de4dee742ce10684400171257c66a86e4c7a6/tests/explainers/test_tree.py#L417) and [lightgbm](https://github.com/slundberg/shap/blob/9d7de4dee742ce10684400171257c66a86e4c7a6/tests/explainers/test_tree.py#L364).
> > It might be a good idea to embed this part `_average_path_length(np.array([iso.max_samples_]))[0]` to the shap value. I will try it later.
> 
> Did you continue this at the end?

Not yet. Focusing on other works. This would be a simple change, just divide `_average_path_length([model.max_samples_])` at https://github.com/slundberg/shap/blob/46b3800b31df04745416da27c71b216f91d61775/shap/explainers/_tree.py#L1379 
, but need to be careful not to break the current feature. Some people are using current result and the change may break their work. It might be a better idea to add some new options somewhere. You can go a head if you are interested in.

PR #790, opened 2019-09-03
AdditiveForceVisualizer: Use locally imported lodash filter

jaller94:
I found a usage of lodash's `filter` which was using the global `_.filter` instead of the imported function `filter`.


slundberg:
Sorry I missed merging this. Thanks!

PR #808, opened 2019-09-15
Handle the multiclass case with no intercept

### LINKED ISSUE #807: LinearExplainer cannot handle models with no intercept
When I train a model with no intercept I get an error stemming from the following code:
```
shap/explainers/linear.py in __init__(self, model, data, nsamples, feature_dependence)
     50             if len(model.coef_.shape) and model.coef_.shape[0] == 1:
     51                 self.coef = model.coef_[0]
---> 52                 self.intercept = model.intercept_[0]
```
When a model is learned without an intercept then the value of model.intercept_ becomes 0.0 so model.intercept_[0] throws a loat' object is not subscriptable error.

gabrieltseng:
Closes #807


slundberg:
Thanks for fixing this!

PR #819, opened 2019-09-22
Add dill in requirements

rohitgr7:
[dill](https://github.com/uqfoundation/dill)  should be in requirements. It's not available by default.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,7 +88,7 @@ def run_setup(with_binary=True, test_xgboost=True, test_lightgbm=True, test_catb
         package_data={'shap': ['plots/resources/*', 'tree_shap.h']},
         cmdclass={'build_ext': build_ext},
         setup_requires=['numpy'],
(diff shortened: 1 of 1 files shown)


jorgecarleitao:
I would disagree with this change, since dill is only used in benchmarks. My proposal is that shap does not import benchmarks on its `__init__`, so that dill becomes a optional requirement. Alternatively, make `dill` to be imported inside the function it is used, instead of in the top of the module.

rohitgr7:
But benchmark is imported in `__init__.py`.https://github.com/slundberg/shap/blob/a6cb866b4f912c1cc0717a6e4356d350bdacbf11/shap/__init__.py#L20

slundberg:
Thanks! I agree that making the benchmark dependencies not imported in __init__ seems best, since most people will not be running the benchmark code. I am switching machines this week as part of starting at MSR, but once I get things all happy on my end I can either help edit this PR or push a separate change.

slundberg:
I am closing this since 0.30.2 fixes this. Thanks again!

PR #820, opened 2019-09-22
Added accuracy and consistency test to kernel explainer.

jorgecarleitao:
This commit adds testing of KernelExplainer against the properties on the paper (local accuracy and consistency/monotonicity).

Note that this currently does not pass, as the KernelExplainer is not guaranteeing consistency as outlined in the paper. Let's figure out why and act accordingly.

I have validated that the consistency test is correct by writing the brute force version of shapley values (without local approximations), and validating the test against that explainer (with multiple training data and non-linear functions).

### DIFF
--- a/tests/explainers/common.py
+++ b/tests/explainers/common.py
@@ -1,0 +1,63 @@
+import typing
+import itertools
+
+import numpy as np
+
+
+def drop(x: np.ndarray, to_drop: typing.Iterable[int]):
+    """
+    Returns x with ``to_drop`` features set to their average value.
+    """
+    x = np.copy(x)
+    for c in to_drop:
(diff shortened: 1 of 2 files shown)


slundberg:
The kernel explainer (given enough samples) is consistent, there is just a typo in here at the moment. Once that's fixed I would love to merge the new test :)

jorgecarleitao:
Great to know, because I have been assuming that the KernelExplainer is consistent and that I made a mistake on the test. :P

slundberg:
I sent an updated file your way, let me know if you need to update this PR

slundberg:
It looks like this is still failing. I thought I remember we had this sorted out though over email. What's the status here? (and thanks again!)

PR #821, opened 2019-09-22
Simplified test_linear of KernelExplainer.

jorgecarleitao:
Just a small cleanup to make the test clearer, since we do not need to fit a model to explain it.


slundberg:
Thanks! Sorry for slowness, I was traveling and then starting a new researcher role. But am back now :)

PR #853, opened 2019-10-16
setup: Add extras-require

AbdealiLoKo:
extras-require lets the user decide on any extra packages they may
want to install. This way users can just request for [all] and all
packages that shap uses internally will be installed in one go.

@slundberg this is somewhat of a follow up for https://github.com/slundberg/shap/issues/823

I understand that most packages are "optional", but in local/analytical environments some folks don't like tracking all dependencies and just want everything to work. So, added a `extras_require` as per setuptools convention to maintain a list of such dependencies.
These can be installed with:

`pip install shap[all]`

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,7 +15,7 @@ environment:
 install:
   # We need wheel installed to build wheels
   - "%PYTHON%\\python.exe -m pip install wheel"
-  - "%PYTHON%\\python.exe -m pip install tensorflow keras matplotlib ipython"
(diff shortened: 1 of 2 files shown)


slundberg:
Thanks! I think the CI issue is with TensorFlow 2 (so unrelated to this PR). I like the idea of doing this for 'plots' and perhaps 'other', but I am not sure it makes sense for the modeling packages. For modeling packages users can't even try to use SHAP until they have those installed (since they won't have a model object).

Issue #863, opened 2019-10-23
Support for subclasses of supported classifiers


zoj613:
Shap's tree explainer does not seem to work at all for "custom" classifiers based to the supported ones.

Say I have:
```python
# modelling.py
from  sklearn.ensemble.gradient_boosting import GradientBoostingClassifier
class MyCustom(GradientBoostingClassifier, SomeMixin):
    def custom_func(self, ...):
        pass
   # other methods
```
If i used the shap `TreeExplainer` I get the error:
`AssertionError: <class 'src.modelling.MyCustom'> is not currently a supported model type!`

Is there a way to support custom classifiers that are subclasses of the supported ones if very minimal changes to the parent class are made??

Looking at the source code it seems like the string output of `str(type(model))` is the check that determines if an exception should be thrown:
https://github.com/slundberg/shap/blob/e475795cbd88fb5c0fc9b110540b872ce8456e4c/shap/explainers/tree.py#L407-L664

Maybe there could be something done to see if the custom classifier has inherited from the supported parent instead?

zoj613:
ping @slundberg

peterdhansen:
Is there any reason not to apply the fix in #863 to the plotting functions as well?
So if I have a pandas dataframe subclass (e.g. geopandas), it will still be recognized?

PR #871, opened 2019-10-28
#863 subclass support

parsatorb:
Added subclass support for issue #863 by making a helper function which checks if an object is an instance of a class without having to explicitly import the parent module. This ensures a user doesn't need to have all supported frameworks installed

### DIFF
--- a/shap/common.py
+++ b/shap/common.py
@@ -359,11 +359,10 @@ def safe_isinstance(obj, class_path_str):
         spec = importlib.util.find_spec(module_name)
     except:
         spec = None
-
     if spec is None:
         return False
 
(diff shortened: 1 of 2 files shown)


parsatorb:
Hold off on this for now, ill make necessary changes soon

slundberg:
This is great. Thanks!!

zoj613:
Has this PR been merged into the latest release? because it still does not work with custom classifiers. For example using subclasses of `tensorflow.keras.wrappers.scikit_learn.KerasClassifier` with deep explainer does not work. Even models wrapped with sklearn's `CalibratedClassifier` do not work even if they linear or tree-based with their explainers. The only explainer that works is KernelExplainer.

PR #877, opened 2019-10-30
Fix Eager Mode execution for Gradient Explainer

bottydim:
tf.executing_eagerly() checks whether eager mode is active and enables users with Tensorflow version <2 to also use the GradientExplainer.


slundberg:
Thanks!

PR #878, opened 2019-10-31
Update README.md

troied-solo:


LakshmanKishore:
Hello @connortann,

We suggest closing this pull request as we have recently updated the readme file, and the specific line mentioned in the proposed change is not present in the latest version. Therefore, the update is not deemed necessary.

Thank you.

Issue #890, opened 2019-11-08
TreeExplainer with feature_dependence='independent' should raise an error if no background dataset passed


ssaamm:
Hi there,

The [docs](https://github.com/slundberg/shap/blob/18267b69b22089853e4d0d419e494fee10406899/shap/explainers/tree.py#L62-L63) and [code](https://github.com/slundberg/shap/blob/18267b69b22089853e4d0d419e494fee10406899/shap/explainers/tree.py#L92-L95) make it seem like a `TreeExplainer` object initialized with `feature_dependence='independent'` requires the `data` parameter to also be specified.

This code seems to run just fine, though, despite not passing `data`:

```python
import xgboost
import shap

X,y = shap.datasets.boston()
model = xgboost.train({"learning_rate": 0.01}, xgboost.DMatrix(X, label=y), 100)

explainer = shap.TreeExplainer(model, feature_dependence='independent')
shap_values = explainer.shap_values(X)
print(len(shap_values))
```

```$ python shap_test.py
506
```

Here are relevant versions, if they help:

```
$ python -V
Python 3.6.8
$ pip freeze
joblib==0.14.0
numpy==1.17.3
pandas==0.25.3
pkg-resources==0.0.0
python-dateutil==2.8.1
pytz==2019.3
scikit-learn==0.21.3
scipy==1.3.1
shap==0.32.1
six==1.13.0
tqdm==4.37.0
xgboost==0.90
```

Thanks,
Samuel

ssaamm:
Would y'all be open to a PR swapping out these asserts for raising some exception?

There are [times when asserts are skipped](https://docs.python.org/3/reference/simple_stmts.html#assert), so an exception might be more appropriate (see Caveat #1 [here](https://dbader.org/blog/python-assert-tutorial)).

slundberg:
Thanks @ssaamm for pointing this out! This is due to a change a few days ago. We auto flip over to using feature_dependence = "tree_path_dependent" when no data is given. I am planning to add a warning when this happens but am waiting until I do an upcoming API change when I rename  `feature_dependence = "independent"` to `feature_perterbation = "interventional"`. This renaming is part of highlighting what independence is not really an assumption about the input features, but is about interventionally perturbing the data. Anyway, I'll write more about this. But for now I'l leave a todo tag for adding the warning.

As for a PR converting asserts to Exceptions I am all for that, thanks!

ssaamm:
[...]

PR #894, opened 2019-11-12
Use exceptions rather than assertions in TreeExplainer.__init__

ssaamm:
Some prior [discussion](https://github.com/slundberg/shap/issues/890#issuecomment-551919580) in #890

Let me know if there's anything I can do (e.g. add tests?) to make this a more helpful PR.


slundberg:
Thanks @ssaamm !

PR #897, opened 2019-11-13
Update image.py

Jimbotsai:
seems like if the image is not a square, the image_plot function has the opposite width and height position for the image


slundberg:
Thanks @Jimbotsai !

Issue #884, opened 2019-11-07
Error with Pyspark GBTClassifier


allard-jeff:
@QuentinAmbard 

I just installed Shap from PyPi (0.32.0) and running a version of your test still produces the same error - shown below. Is there something that I am missing in the use of Shap with a pyspark model?

```
import pyspark
print(pyspark.__version__)
import shap
print(shap.__version__)
import sklearn.datasets
from pyspark.sql import SparkSession
from pyspark import SparkContext, SparkConf
from pyspark.ml.feature import VectorAssembler, StringIndexer
from pyspark.ml.classification import RandomForestClassifier, DecisionTreeClassifier, GBTClassifier
import pandas as pd

iris_sk = sklearn.datasets.load_iris()
iris = pd.DataFrame(data= np.c_[iris_sk['data'], iris_sk['target']], columns= iris_sk['feature_names'] + ['target'])[:100]
spark = SparkSession.builder.config(conf=SparkConf().set("spark.master", "local[*]")).getOrCreate()

col = ["sepal_length","sepal_width","petal_length","petal_width","type"]
iris = spark.createDataFrame(iris, col)
iris = VectorAssembler(inputCols=col[:-1],outputCol="features").transform(iris)
iris = StringIndexer(inputCol="type", outputCol="label").fit(iris).transform(iris)

classifier = GBTClassifier(labelCol="label", featuresCol="features")
model = classifier.fit(iris)
explainer = shap.TreeExplainer(model)
X = pd.DataFrame(data=iris_sk.data, columns=iris_sk.feature_names)[:100] # pylint: disable=E1101
shap_values = explainer.shap_values(X)
```

```

---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
<ipython-input-31-f47b3a56c25f> in <module>
     23 explainer = shap.TreeExplainer(model)
     24 X = pd.DataFrame(data=iris_sk.data, columns=iris_sk.feature_names)[:100] # pylint: disable=E1101
---> 25 shap_values = explainer.shap_values(X)

/mnt1/anaconda3/lib/python3.7/site-packages/shap/explainers/tree.py in shap_values(self, X, y, tree_limit, approximate, check_additivity)
    283 
    284         if check_additivity and self.model_output == "margin":
--> 285             self.assert_additivity(out, self.model.predict(X))
    286 
    287         return out

[...]

PR #905, opened 2019-11-19
#884 add spark in setup.py tests and fix spark issue with additivity check

QuentinAmbard:
Because pyspark wasn't in the test dependencies, the spark unit test was ignored, leading to errors like #884 
This commit add spark as test dependency and fix an issue ignoring check_additivity as it requires to run the prediction, which isn't implemented yet.
(Long term fix is to support the prediction directly with spark, starting a local spark cluster)


slundberg:
Thanks for working on this! Looks there are still some spark errors, so I'll hold off on merging right now.

QuentinAmbard:
> ERROR Instrumentation: java.lang.IllegalArgumentException: Unsupported class file major version 55

I suspect this is an issue with the java version used in the travis build, spark is still on java8 and this error suggest that the jdk is running on something else, like java11 

@slundberg I'm not super familiar with travis, should I add the following to the .travis.yaml file and try to re-run the build?

```
jdk:
  - openjdk8
```

slundberg:
Thanks! I was trying to figure out why this fails on python 36 and 37 on windows. Do you have thoughts there?

QuentinAmbard:
Making slow progress here...
- It was failing on Ubuntu because we need java8 and not 11, I fixed that
- On windows the winutils lib was missing (required by spark), so I added that too
- Still have an issue on windows with python. For some reason it's looking for ctype under C:\Python27 and is failing, probably because C:\Python27 is 32b and not 64b.

```
File "c:\python36-x64\lib\site-packages\numpy\_distributor_init.py", line 26, in <module>
    WinDLL(os.path.abspath(filename))
  File "C:\Python27\lib\ctypes\__init__.py", line 366, in __init__
    self._handle = _dlopen(self._name, mode)
```

I thought adding `$env:PYTHONPATH = "%PYTHON%\Lib;%PYTHON%\DLLs"` in appveyor would solve this issue but there is something else. I'll try again tomorrow, but any idea is welcome!

QuentinAmbard:
@slundberg all good now!
However I had to remove `- PYTHON: "C:\\Python38-x64"`: tensorflow doesn't work with python 3.8, how did you managed to make it works in the other build?

slundberg:
Thanks @QuentinAmbard this is great!! I have been meaning to remove py38 on windows for a while, it was not working on TF1.14 and I need to re-add it once we bump up the version of TF we are testing.

PR #907, opened 2019-11-20
Draw colorbar on individual axes when doing dependence_plot

lassevalentini:
When using dependence plot with a subplot, the colorbar is currently drawn in the right part of the figure rather than the individual axes.

Examples of before/after using the below code:

```python
fig, axs = plt.subplots(1, 2, figsize=(20,8))
shap.dependence_plot("DateDaysAgo-1", shap_values, X, ax=axs[0], show=False)
shap.dependence_plot("CouponRate", shap_values, X, ax=axs[1], show=False)
```

Before PF:
Before PR
After PR:
After PR


slundberg:
Excellent fix, thanks!

PR #873, opened 2019-10-29
add `return_variances` to `_PyTorchGradientExplainer.shap_values` 

s6juncheng:
`_PyTorchGradientExplainer.shap_values` is called by `GradientExplainer.shap_values` with the argument `return_variances`, which is not implemented in _PyTorchGradientExplainer. This causes the following error. 

```
shap/shap/explainers/gradient.py in shap_values(self, X, nsamples, ranked_outputs, output_rank_order, rseed, return_variances)
    103         were chosen as "top".
    104         """
--> 105         return self.explainer.shap_values(X, nsamples, ranked_outputs, output_rank_order, rseed, return_variances)
    106 
    107 

TypeError: shap_values() takes from 2 to 6 positional arguments but 7 were given
```

I added  `return_variances` to the `_PyTorchGradientExplainer`, consistent with the `_TFGradientExplainer` class.

PR #918, opened 2019-11-29
Updates to the PyTorch explainers

gabrieltseng:
Two updates:

1. Previously, if the input tensor passed to the explainer has `requires_grad == True`, then the DeepExplainer throws the following RunTime error:
```
Can't call numpy() on Variable that requires grad. Use var.detach().numpy() instead.
```
This is fixed by ensuring all tensors are detached when they are passed to the explainer.

2. ~The PyTorch gradient explainer is updated to handle `return_variances`.~ This is fixed in #873


slundberg:
Thanks @gabrieltseng !

Issue #406, opened 2019-01-24
shap.summary_plot displaying gray plot


terryyylim:
![screen shot 2019-01-23 at 8 56 37 pm](https://user-images.githubusercontent.com/25025366/51649259-a4d8e280-1f51-11e9-9a78-3124b47d7fb6.png)

I'm facing this issue where the features are not getting the expected blue and red colors. Does anyone have any idea why this might be so? Thank you!

dynamik1703:
I'm getting a mix of blue, red and grey lately.

Version: 0.18.2

slundberg:
Grey is for NaN feature values (there are never NaN SHAP values), but I would be surprised if your whole dataset is NaN.

dynamik1703:
I checked my feature values. There are no nans included.  Nevertheless, I get grey dots. This is only since the last update with me.

shap==0.28.2
numpy==1.16.0
Win

xthomas8888:
I got similar displays to @dynamik1703 when plotting summary plot using SHAP. Specifically, I checked a feature which has no NaNs in values. But still got a mixture of blue, red and gray dots in the summary plot.

shap==0.28.2
numpy==1.16.0

Thanks,
Thomas

dynamik1703:
I'm not 100% sure but it's probably been happening since I updated numpy to 1.16.0

slundberg:
Thanks for pointing that out! ...it looks like the recent color range update to better plot NaN values in dependence_plots caused the dots outside of the color bar range to be gray in the summary plot. I am working on a fix now...

slundberg:
0.28.3 now has this fixed.

xthomas8888:
Thanks for the fix, @slundberg. Although this is closed, want to follow up with a question: if they are NaN values in the data, will they be marked in gray in the summary_plot? If not, will they be marked in blue (low values) or red(high values)? In my case, I did have NaN in my data, but I didn't see gray color indicated for NaN in summary plot. Thanks for your attention on this!

slundberg:
@xthomas8888 they should be gray. If you have an example where they are not please share it and I can take a look.

alvarosalgado:
Hi,

I know this is a closed issue, but this problem is happening to me and I don't know what might be causing it.
My dataset is a pd.DataFrame containing genomic data (n genomes vs. p nucleotides). The attributes contain categorial values, i.e., "A, C, T or G". Please see screen prints bellow.

[...]

Issue #59, opened 2018-04-17
default to violin in summary_plot


kodonnell:
@slundberg - may I suggest defaulting to the violin plot for the distributed `summary_plot`? Generally they look better, and are more easily interpretable (i.e. the height means something, and the visualization isn't dependent on the order you draw the dots - which is always a problem for overlapping objects). That said, there's a chance the smoothing from the violin plot misleads people (though it should be easy to reduce that smoothing).

slundberg:
hmmm...I originally switched from the violin plot because I felt like the smoothing was hiding features in the data, and that choosing the right kernel size was not foolproof. I'll compare the plots again and think about it. Your point about the order of drawing makes sense. As for the height being meaningful I think the same holds in both plots, just one is smoothed.

kodonnell:
Cool, we understand each other. 

One thing I'll add is that the purpose of these graphs is visually summarizing a general trend, and smoothed violins are good for this, especially when communicating to clients. Dot plots add a lot more noise, and if you're trying to use these to identify specific features that are small enough to be hidden in the violin plot, then you should be using something more accurate than graphs anyway.

Also, this doesn't matter for people who know about both options. However, I'd posit that the average user is more interested in violin type plots, and will be less harmed by not knowing about dot plots. The technical user who might want dot plots is also more likely to know about them. Hence it might be safer to default to violin. (Of course, the opposite could easily be argued.)

Anyway - will leave you to close this as you see fit - I'm not personally worried either way. It just came up when I saw a blog by someone using our new graphs, and I just wished they'd used the violin plots as they look much nicer (in my opinion).

slundberg:
I added a new notebook to compare the plot types: https://github.com/slundberg/shap/blob/master/notebooks/Scatter%20Density%20vs.%20Violin%20Plot%20Comparison.ipynb

Which type of violin plot were thinking of? (plain, colored, or layered)

kodonnell:
I updated the notebook as part of #17 - see [here](https://github.com/awaregroup/shap/blob/98630d08d2026a11de1be9cf5c1dbd15e12fb07f/notebooks/Scatter%20Density%20vs.%20Violin%20Plot%20Comparison.ipynb). I'd personally be defaulting to the layered one (sorry, I hadn't realised there were now two), in red/blue - i.e. like this:

![image](https://user-images.githubusercontent.com/8204904/39046343-e9c9b114-44e9-11e8-9602-6a23a0ead67a.png)


[...]

PR #919, opened 2019-11-30
Clip vmin and vmax for scatter markers in violinplots

joelostblom:
This addresses the latest issues reported in #406 where some individual scatter dots show up in grey in summary violin plots, but not in summary dot plots. I have just copied over the logic from the dotplot for how the colors are clipped to vmin/vmax.

(I am not sure if the nan part is needed since `gaussian_kde` doesn't handle nans anyways, but I might misunderstand how nans in the `feature` variable relate to nans in the `shap` variable, so I left the nan logic in there for now)

### DIFF
--- a/shap/plots/summary.py
+++ b/shap/plots/summary.py
@@ -341,7 +341,7 @@ def summary_plot(shap_values, features=None, feature_names=None, max_display=Non
                 for i in range(len(xs) - 1):
                     if ds[i] > 0.05 or ds[i + 1] > 0.05:
                         pl.fill_between([xs[i], xs[i + 1]], [pos + ds[i], pos + ds[i + 1]],
-                                        [pos - ds[i], pos - ds[i + 1]], color=colors.red_blue(smooth_values[i]),
+                                        [pos - ds[i], pos - ds[i + 1]], color=colors.red_blue_no_bounds(smooth_values[i]),
                                         zorder=2)
 
         else:
--- a/shap/plots/summary.py
+++ b/shap/plots/summary.py
@@ -318,8 +318,21 @@ def summary_plot(shap_values, features=None, feature_names=None, max_display=Non
                     if vmin == vmax:
                         vmin = np.min(values)
                         vmax = np.max(values)
-                pl.scatter(shaps, np.ones(shap_values.shape[0]) * pos, s=9, cmap=colors.red_blue, vmin=vmin, vmax=vmax,
-                           c=values, alpha=alpha, linewidth=0, zorder=1)
+
+                # plot the nan values in the interaction feature as grey
+                nan_mask = np.isnan(values)
+                pl.scatter(shaps[nan_mask], np.ones(shap_values[nan_mask].shape[0]) * pos,
+                           color="#777777", vmin=vmin, vmax=vmax, s=9,
+                           alpha=alpha, linewidth=0, zorder=1)
+                # plot the non-nan values colored by the trimmed feature value
+                cvals = values[np.invert(nan_mask)].astype(np.float64)
+                cvals_imp = cvals.copy()
+                cvals_imp[np.isnan(cvals)] = (vmin + vmax) / 2.0
+                cvals[cvals_imp > vmax] = vmax
+                cvals[cvals_imp < vmin] = vmin
+                pl.scatter(shaps[np.invert(nan_mask)], np.ones(shap_values[np.invert(nan_mask)].shape[0]) * pos,
+                           cmap=colors.red_blue, vmin=vmin, vmax=vmax, s=9,
+                           c=cvals, alpha=alpha, linewidth=0, zorder=1)
                 # smooth_values -= nxp.nanpercentile(smooth_values, 5)
                 # smooth_values /= np.nanpercentile(smooth_values, 95)
                 smooth_values -= vmin


joelostblom:
Current behavior (note that the grey dots are non NaNs):
image

Suggested behavior:
image

joelostblom:
The failing test seems unrelated:

```
%PYTHON%\python.exe -m pip install tensorflow==1.14 keras matplotlib ipython
ERROR: Could not find a version that satisfies the requirement tensorflow==1.14 (from versions: none)
ERROR: No matching distribution found for tensorflow==1.14
Command exited with code 1
```

joelostblom:
I have also noticed that part of the violins are also greyed out. Again, this is not because of NaNs, but seems to be because the value in `smooth_values` goes under zero and can be fixed by adjusting this up to zero, but I am afraid this might introduce some artifact.

image

The color of the violins also does not correspond well to the color of the dots in all case. For example, compare the three bottom-most violins in the plot above with the corresponding dotplot below. There are red areas in the dot plot that are completely blue in the violinplot.

image

The layered violins correspond better to the dot plot, although it apparently enhances the amount of red in the three bottom-most violins;

image

Briefly, I can think of a couple of reasons why the standard violins differ:

1. There are just a few red dots that happen to be plotted on top in the dot plot, but when they are averaged/smoothed to create the violin, the overwhelming amount of blue dots in the same area causes the violin to be blue.
2. The values are not depicted in color equally in the two different plots. This might be related to the clipping of low and high values when creating the dot plot and/or the lack of clipping when creating the violinplot.

I should also say that I just started using shap so it is possible that I have misunderstood something on how the above is supposed to work.

slundberg:
Thanks for this! I added one update to fix the gray regions in the violin plots (by using a color range that does not have grey for out-of-bounds). As for the color differences I have not carefully tested the consistency there since I primarily use the scatter option. Perhaps some simulated data would help illuminate things? We randomized the layering of the dots in the scatter plot so it would be unlikely that all the red dot's would show up on top.

joelostblom:
:+1:  I can try to create some simulation data and compare some time next week.

Btw, is there any reason that the layered violinplot uses a different colormap by default compared to the dotplot and normal violin (`'coolwarm'`, instead of `red_blue`)? The transition through white/grey instead of purple emphasizes the extreme feature values and could hide middle values against the white background (which might be an advantage, I am just not sure why it is different). It seems like the `'coolwarm'` colormap was implemented in https://github.com/slundberg/shap/pull/69, (I modified my plot above to use `red_blue` instead just to ease the comparison)

slundberg:
Great. I will merge this fix for now. I think coolwarm was just a default left by @kodonnell . I don't recall there being a specific reason.

kodonnell:
> I think coolwarm was just a default left by @kodonnell . I don't recall there being a specific reason.

Yup, I've totally forgotten. My suspicion is that `red_blue` implies (though I'm not sure if it's the case) "low numbers are red, high numbers are hot" which is, I believe, the opposite of `coolwarm` (and what's expected).

> Briefly, I can think of a couple of reasons why the standard violins differ:
> - There are just a few red dots that happen to be plotted on top in the dot plot, but when they are averaged/smoothed to create the violin, the overwhelming amount of blue dots in the same area causes the violin to be blue.

Yup, this is why I prefer violins and not dots, and am suggesting them as the default - #59 . I personally don't think it's a good idea when these graphs (which can drive some important decisions) are (as far as I know) arbitrarily dependent on the order plots are drawn (i.e. the ordering of your data). You can check our more comments in the notebook referred there. I'd be interested in your thoughts.

NB: I think I once tried randomising the ordering of the dots, but that introduces other issues. Better to just use the layered violin, which I also think looks better = )

joelostblom:
Thanks for merging @slundberg !

Thanks for the reply @kodonnell , I had a brief look in the notebook, but will have to go through it more thoroughly with simulation data. I agree that the layered violins look very cool! I haven't encountered them previously so I need to test and think about if they represent data according to my expectations when using them.

In addition to the concerns I mentioned above with randomly ordered points, and what you two already discussed around smoothing parameters and the sometimes strange colorscale of the standard striped violins, a few (maybe minor) things come to mind:

- In the layered violins, the order of the layering means that blue (low values) will always pair up giving rise to a bigger continuous block of blue. Red (high values) layers are often separated, could this mean that the same amount of red and blue is perceived differently? I'm mostly thinking that there might appear to be more blue since they are joined together. A filled density curve/half violin, would not have this potential issue, but there will always be some difference between the layers since high and low values are plotted at different locations.
- Wouldn't a traditional sequential colormap (such as viridis, BuPu or single color) be more appropriate for visualizing a sequence of feature values from low to high? Diverging maps are most often used for values with a natural center point (such as the shap values). Or is it intentionally selected to highlight extremes?
- Regarding violins not being suitable for a smaller number of observations, there could be a threshold of points which changes the default between dot and violin plot.
- In general, I like swarm plots where the dots are stacked without overlapping, but I appreciate that there is no room for that in these plots. 

> Yup, I've totally forgotten. My suspicion is that red_blue implies (though I'm not sure if it's the case) "low numbers are red, high numbers are hot" which is, I believe, the opposite of coolwarm (and what's expected).

Both `red_blue` and `"coolwarm"` colors low numbers blue and high numbers red.

kodonnell:
Thanks @joelostblom - I might move this conversation over to #59 so it's all captured there.

slundberg:
All great points for discussion! One note on:

> I think I once tried randomising the ordering of the dots

The layering of the dots is randomized already so the behavior of the plots does not depend on the ordering of your input dataset.

PR #936, opened 2019-12-10
Add docs badge

willprice:
I couldn't see whether there was a ref docs page when skimming the README. A badge like this would be handy.


slundberg:
This is great...except I need to get the docs in much better shape before pointing people there. I am going to leave this for a bit until then.

lrjball:
Hey Will, it looks like this already went in when the docs were updated and the badge was added, so I'll close this for now. Cheers for the PR though :+1:

PR #937, opened 2019-12-11
fix shap import error on matplotlib with missing backend

imatiach-msft:
fix shap import error on matplotlib with missing backend

to test locally on a non-macos machine (setting to missing macos backend will then force matplotlib to throw error on import):

(sh) PS C:\shap> python
>>> import matplotlib
>>> matplotlib.use('MacOSX')
>>> import shap

### DIFF
--- a/shap/plots/bar.py
+++ b/shap/plots/bar.py
@@ -1,4 +1,9 @@
-import matplotlib.pyplot as pl
+import warnings
+try:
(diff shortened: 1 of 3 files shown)


rihorn2 (approved):
Lgtm

slundberg:
Good catch thanks!

PR #962, opened 2019-12-18
Allows retaining original class order in multi-class summary plot

sjstreicher:
I needed the colors used in the multi-class stacked bar summary plot to be consistent across many different scenarios and predictable (I'm using a custom cmap where the colours make sense wrt the classes). This trivial change allows for the order of the stacks per variable to be the same as the original class order (or can be entirely custom by providing an index list in `class_inds`).


slundberg:
Looks great, thanks!

Issue #759, opened 2019-08-18
CRIC data missing


ibuda:
I noticed that the CRIC data is missing from the repo when calling:
`X, y = shap.datasets.cric()`
and getting:
`
HTTPError: HTTP Error 404: Not Found
`
Lurking around, I found that you've "gitignored" CRIC data files in this commit 01771891384d4c856240499b78ee0ff0ed56297f

ibuda:
Closing, pull request #1011 was accepted and merged.

PR #1011, opened 2020-01-23
Removed references to CRIC dataset

ibuda:
This is related to #759. I removed the method loading CRIC data.
The CRIC data is no longer available, as it was "git ignored" in this 0177189 commit.


slundberg:
Thanks, it is a good idea to remove this.

PR #1012, opened 2020-01-23
Make package compatible with IBM zOS

DorianCzichotzki:
Not sure if the stdlib include is necessary but alloca.h is not available on zOS.

An additional compile option is needed for the xlC compiler to enable long long type.


slundberg:
Great thanks!

PR #1041, opened 2020-02-12
Compute shap value with tweedie objective function in xgboost

jfrery:
Following: /issues/454

Tested to see if I could get back the prediction with the shap values computed and it works.

only works with feature_perturbation = "interventional"

Would be nice to have it working with "tree_path_dependent".


slundberg:
Thanks @jfrery ! This looks good. The only chance is could you add a unit test into the test_tree.py file just to make sure this continues to work in the future?

narendramukherjee:
@slundberg @jfrery Thanks for your work with the tweedie objective. Just wondering what the status of this is, and what contributions are needed to move this forward?

statgirl30:
Hello, Just wanted to follow up on this. can we transform SHAP marginal values to original values when Tweedie is used?

narendramukherjee:
@statgirl30 I haven't gotten a response from the contributors on this thread, but what I figured out is that Tweedie basically uses a log-link. So you can convert the shap values you get with `link="identity"` to Tweedie specific values by just exponentiating them. It will be great to merge this fix (as it is almost complete), but it would be great to get some feedback from @slundberg on what else remains to be done.

statgirl30:
@narendramukherjee Thanks for your respond. I am using [this ](https://cheevahagadog.github.io/posts/model_explanations/) link to calculate the 5 highest contributions and simply exponentiating each value would not be correct.

jfrery:
@statgirl30 @narendramukherjee You can't use the exponential on the shap values obtained with the identity link to find the original shap values as this would assume the following:

X = s1 + s2 + ... sN 
exp(X) = exp(s1) + exp(s2) + ... + exp(sN)
which is not true. 

This fix allows you to have the correct shap values when xgboost uses the tweedie objective function but I did not dig enough to make the tree path dependent work (fastest method).

jfrery:
This doesn't work anymore. Seems like the sum of the shap_values + expected_value is off by a factor of 3.... Not sure what happens there :/. @slundberg  any idea why this would happen ? The link between the model predictions with output_margin and without output_margin is still the log. I wonder what could cause the shap values to be off by a constent factor.

mosscoder:
At this point can anyone offer insight as to how we transform SHAP values for Tweedie objectives for a given feature back to the native units of the response? I'm struggling with this in [lightgbm's implementation](https://stackoverflow.com/questions/64793802/transformation-of-lightgbm-shap-values-necessary-with-tweedie-objective). A generalized transformation formula would benefit not only this package, but the greater community as well.

jfrery:
@mosscoder we can't really do this as I am not aware of any mathematical trick to get the shap values back on sync with the predictions. You could get a rough estimation considering that 1- the original shap values should follow the same order and 2- the sum of the original shap values should be equal to the prediction.

mosscoder:
I last night found [this somewhat old thread discussing ](https://github.com/slundberg/shap/issues/29)the analog challenge of transforming SHAP values to the probability space for binary classification, which appears to have been integrated into this package at this point? Anyway I encountered this function (slightly-modified):

```python
from scipy.special import expit #Importing the logit function for the base value transformation

def shap_transform_scale(shap_values, expected_value, model_prediction):
    
    #Compute the transformed base value, which consists in applying the logit function to the base value    
    #expected_value_transformed = expit(expected_value) #transform for binary classification
    expected_value_transformed = numpy.log(expected_value) #appropriate transformation for Tweedie and Gamma?

    
    #Computing the original_explanation_distance to construct the distance_coefficient later on
    original_explanation_distance = sum(shap_values)

    #Computing the distance between the model_prediction and the transformed base_value
    distance_to_explain = model_prediction - expected_value_transformed

    #The distance_coefficient is the ratio between both distances which will be used later on
    distance_coefficient = original_explanation_distance / distance_to_explain

    #Transforming the original shapley values to the new scale
    shap_values_transformed = shap_values / distance_coefficient

    return shap_values_transformed, expected_value_transformed

```

Two years ago it was referred to as a 'hack', and I'm wondering how perspectives on this approach may have evolved?

jfrery:
@mosscoder this is the kind of trick I was referring to transform the shap values you would get from a model trained with a tweedie objective function. However, the problem is that this shap_transform_scale function assume a linear transformation of the shap values is possible which is wrong. The resulting shap values do sum up to the correct prediction however, their scale is completely wrong. 

If we want to be able to work properly with tweedie models in shap, we need the shap values to be computed over the actual prediction and not the log ones as it's currently done. The branch I created does this well when feature_perturbation = "interventional".

mosscoder:
So, for the raw scores of the SHAP values (those that aren't back-transformed to the native units of the response), are we safe to assume that negative values, by some unknown magnitude, subtract from the final prediction in native units, and positive values the converse? Can we also safely compare magnitudes across raw SHAP values?

jfrery:
> So, for the raw scores of the SHAP values (those that aren't back-transformed to the native units of the response), are we safe to assume that negative values, by some unknown magnitude, subtract from the final prediction in native units, and positive values the converse?

You can find a solution where the raw shap values + the expected value multiplied by a factor give you the model prediction. However, these shap values will be wrong and I don't think you can use them to explain the model prediction as they are trying to explain the log(model_prediction) rather than the model_prediction. I don't even think that you can assume that the order of the raw shap values would be the same than the actual shap values.

>  Can we also safely compare magnitudes across raw SHAP values?

You can safely compare the magnitudes across the shap values but extrapolating these magnitude to your original problem of explaining the model predictions would be wrong. 

To do this you will have to solve the following  equation:
y_pred = s_1 + s_2 + ... s_N + expected_value
 such that:
exp(y_pred) = exp(s_raw_1 + s_raw_2 + ... + s_raw_N + raw_expected_value)

where N is the number of features. And y_pred is the actual model prediction. 

But again, I don't think there is any mathematical tool to solve this as there isn't a unique solution.

samstiv:
Hello, after much digging I have found this [stackoverflow answer](https://stackoverflow.com/questions/64793802/transformation-of-lightgbm-shap-values-necessary-with-tweedie-objectiver) (see under "Demo" in the answer) and it seems to be working for my xgboost with reg:tweedie. Any thoughts?

jfrery:
@SquaredS44 It works to get the original prediction. However you can't infer the value of the original shap values. The trick they use is simply the one I mentioned before:
original_prediction = exp(s_raw_1 + s_raw_2 + ... + s_raw_N + raw_expected_value)

They use exp(s_raw_1) * exp(s_raw_2)*...*exp(s_raw_N)*exp(raw_expected_value) in their code but this doesn't mean that exp(s_raw_1) = s1 (your original shap value for the first feature).

samstiv:
@jfrery I see, thank you for the clarification. But despite not being able to infer original shap value for the n-th feature, is it safe to assume that the relative importance of the s_raw_n value w.r.t. the other s_raw is the same of the n-th original shap values w.r.t to the rest of the original shap values? (I'm not sure if that's what you were referring to with "However, the problem is that this shap_transform_scale function assume a linear transformation of the shap values is possible which is wrong." and "the original shap values should follow the same order")

mosscoder:
@jfrery thank you for working on this. I recently downloaded your shap branch and have two questions:

1) Does this...
 ```python
 explainer = shap.TreeExplainer(bst, dat, feature_perturbation="interventional", model_output="log_link")
 ```
...only work for XGBoost, or would it also work for lightgbm to get the SHAP values in native units?

2) Would the above work with categorical features, presuming a lightgbm model, or continuous only? I ask because I get the following:

Screen Shot 2020-11-13 at 9 52 34 AM

jfrery:
@slundberg The demand for having the tweedie handled in shap seems to increase as I get many requests in that sense. It seems that this fix works for xgboost==0.90 but not for more recent versions (the sum of the shap values + expected value is higher than the model prediction by a factor of 3.3). Have you ever noticed a similar behavior before?

jukiewiczm:
Are there any plans on getting this solved? Being able to use tweedie distribution while having shap values in native units would be really useful.

A163460:
Follow up on this issue. Is there a solution available?

PR #1046, opened 2020-02-13
fix the shap values shape from xgboost to be the same as scikit-learn and lightgbm

imatiach-msft:
fix the shap values shape from xgboost to be the same as scikit-learn and lightgbm for binary classification case

### DIFF
--- a/shap/explainers/tree.py
+++ b/shap/explainers/tree.py
@@ -173,6 +173,13 @@ class TreeExplainer(Explainer):
 
         return self.model.predict(self.data, np.ones(self.data.shape[0]) * y).mean(0)
 
+    def _binary_reshape_phi(self, phi, classifier_name):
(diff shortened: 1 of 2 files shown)


imatiach-msft:
FYI I looked at the two test failures, they seem to be unrelated to this change and are appearing in other PRs

slundberg:
Thanks @imatiach-msft ! I want to think this over a bit more after talking to sam and harsha more about potential API stuff. Just want to make sure we don't make a breaking change on the output here more than once.

harshaneekhra:
Why the fix is implemented for LightGBM and not for XGBoost? 
It would be good to have the same fix for XGBoost

PrimozGodec:
@slundberg Can this fix be considered and merged? Is there anything that needs to be done before merging?

Note: the same error persists for [sklearn.ensemble](https://scikit-learn.org/stable/modules/classes.html#module-sklearn.ensemble)

PR #1056, opened 2020-02-20
Fix for Tensorflow 2.1

Zahlii:
See also https://github.com/slundberg/shap/issues/1055

Using this version, I was able to successfully calculate SHAP values for a complicated NN based on TF 2.1.0


shap/explainers/deep/deep_tf.py
```diff
@@ -146,8 +151,7 @@ def __init__(self, model, data, session=None, learning_phase_flags=None):
             else:
                 self.expected_value = tf.reduce_mean(self.model(self.data), 0)
 
-        if not tf.executing_eagerly():
-            self._init_between_tensors(self.model_output.op, self.model_inputs)
+        self._init_between_tensors(self.model_output.op, self.model_inputs)
```
slundberg: Did you try this in eager mode?

Zahlii:
Hi, while running our internal test suite I still noticed some problems related to shap, will try to fix them and update this MR accordingly.

shap/explainers/deep/deep_tf.py
FRUEHNI1: Yes - thats why I included this change. If I do not run this code, I get the following error downstream: 
image

Zahlii:
Closing since its now obsolete, will follow-up with a newer PR.

PR #1062, opened 2020-02-24
QoL change. Takes into account default 'objective' param in LightGBM,

ggmblr:
Hi, I just added a preliminary check to see if the params dictionary for a model in LightGBM has its 'objective' key populated. Since when not defining an objective its default value is 'regression' ( as can be seen here: [https://lightgbm.readthedocs.io/en/latest/Parameters.html#objective](url) ). But this does not create an 'objective' entry in the params dictionary unless explicitly specified. Thus when trying to run TreeExplainer(...).shap_values(...) it throws a KeyError, as seen below:

shap_keyError

This is just a QoL change to avoid that anyone else has to dig into why this happens, and allow a more fluid workflow.

P.S. This is my first PR and first time interacting in the open source environment. If there is something else i should have done, considered, please let me know.

### DIFF
--- a/shap/explainers/tree.py
+++ b/shap/explainers/tree.py
@@ -245,9 +245,10 @@ class TreeExplainer(Explainer):
                 assert not approximate, "approximate=True is not supported for LightGBM models!"
                 phi = self.model.original_model.predict(X, num_iteration=tree_limit, pred_contrib=True)
                 # Note: the data must be joined on the last axis
-                if self.model.original_model.params['objective'] == 'binary':
-                    warnings.warn('LightGBM binary classifier with TreeExplainer shap values output has changed to a list of ndarray')
-                    phi = np.concatenate((0-phi, phi), axis=-1)
+                if self.model.original_model.params.get('objective') is not None:
+                    if self.model.original_model.params['objective'] == 'binary':
+                        warnings.warn('LightGBM binary classifier with TreeExplainer shap values output has changed to a list of ndarray')
+                        phi = np.concatenate((0-phi, phi), axis=-1)
                 if phi.shape[1] != X.shape[1] + 1:
                     try:
                         phi = phi.reshape(X.shape[0], phi.shape[1]//(X.shape[1]+1), X.shape[1]+1)


shap/explainers/tree.py
```diff
@@ -245,9 +245,10 @@ def shap_values(self, X, y=None, tree_limit=None, approximate=False, check_addit
                 assert not approximate, "approximate=True is not supported for LightGBM models!"
                 phi = self.model.original_model.predict(X, num_iteration=tree_limit, pred_contrib=True)
                 # Note: the data must be joined on the last axis
-                if self.model.original_model.params['objective'] == 'binary':
-                    warnings.warn('LightGBM binary classifier with TreeExplainer shap values output has changed to a list of ndarray')
-                    phi = np.concatenate((0-phi, phi), axis=-1)
+                if self.model.original_model.params.get('objective') is not None:
```
imatiach-msft, line 248: can you add a test for this, I can't repro this issue.  I also sent a PR here, but the test test_lightgbm_mse_regressor doesn't actually fail when I undo the fix, so I am a bit confused as to why it is failing on some environments:
https://github.com/slundberg/shap/pull/1063

imatiach-msft: also I'm confused by this description:

```
But this does not create an 'objective' entry in the params dictionary unless explicitly specified
```

If that was the case, then the test "test_lightgbm" under test_tree.py would always fail...

ggmblr: Let me look into that when i'm back from work. As for reproducing the error, maybe the issue is I wasn't using LGBMRegressor, but instead called directly the .train() function.
Something like:

```import lightgbm as lgb

lgb_train = lgb.Dataset(x_train,y_train)

lgb_test = lgb.Dataset(x_test,y_test, reference=lgb_train)

params = {
    'num_threads': 4,
    'n_estimators': 8000,
    'early_stopping_round':50, 
    'metric': ['rmse'],
    'random_state': C_SEED,
    'verbose': 1,
}

evals_result = {} 

gbm = lgb.train(params,
                lgb_train,
                valid_sets=[lgb_train, lgb_test],
                evals_result=evals_result,
                verbose_eval=30
               )
```

I'm new to this library, are both interchangeable if objective is regression?


Edit: Looked at the tests and indeed everything is done with LGBMRegressor. Thus I assume LightGBM treats both of these diferently when creating the default parameters. I'm not sure if shap should accomodate or if a change in LightGBM is called for.

Edit 2:
From LightGBMs sklearn.py:
```
class LGBMModel(_LGBMModelBase):

(...)

if self._objective is None:
            if isinstance(self, LGBMRegressor):
                self._objective = "regression"
            elif isinstance(self, LGBMClassifier):
                self._objective = "binary"
            elif isinstance(self, LGBMRanker):
                self._objective = "lambdarank"
            else:
                raise ValueError("Unknown LGBMModel type.")
```

Whereas for LightGBMs engine.py (where .train() is defined), there is no such assignment when objective is None.

So there lies the confusion as why you couldn't reproduce it. I'm open for whichever approach seems better to you guys. 

(Now I'm really of to work, i just couldn't stop thinking about it haha)

imatiach-msft: @ggmblr thank you, I added your suggested code and was able to 1.) reproduce the issue 2.) validate the issue is fixed with the code changes

imatiach-msft: https://github.com/slundberg/shap/pull/1063

j-adamczyk:
@ggmblr @imatiach-msft any update on this? I am running into this error for LGBMClassifier (scikit-learn API). Adding workaround from this PR works.

ggmblr:
@j-adamczyk No news from my side. I think i modified this on my local environment to avoid the bug, but shortly after i changed positions/projects so i completely lost track of this. As far as i can see here and the linked PR they solved it but didn't merge to master.

PR #1063, opened 2020-02-24
fix for missing objective error for LGBMRegressor

imatiach-msft:
https://github.com/slundberg/shap/issues/1042

For some reason I can't reproduce this issue locally.  The objective is populated on the model.


slundberg:
Thanks for digging into this @imatiach-msft ! Feel free to merge this or the other PR whenever you can verify it fixes it :)

imatiach-msft:
@slundberg I was able to reproduce the issue when using the non-sklearn API from lightgbm.  There might be other cases where this happens as well (maybe save/reload model?).  I added more test coverage for the non-sklearn API.  The current fix does seem to resolve the issue for non-sklearn API so I think it can be merged.

DawidDabkowski:
Hello! Is there a chance of reviving this Pull Request and merging with Shap? It'd be great to solve this bug, I just ran into this issue on my current project.

j-adamczyk:
Any update on this @slundberg @imatiach-msft? This is a pretty SHAP-breaking bug for me.

CloseChoice:
I cannot reproduce the issue on the latest master but I don't see any harm in here and it adds more tests, which I think, is a plus. Therefore I'll merge this ;)

PR #1088, opened 2020-03-06
get summary plot

Benjamin15:
Hi,

When i use the function summary_plot, sometimes i want to get the summary plot.

Currently, i think this is not possible, so i suggest this PR.

In the function summary_plot, I had a parameter, get_png. When the parameter get_png is True, we return a BytesIO object which contains the summary plot.

### DIFF
--- a/shap/plots/dependence.py
+++ b/shap/plots/dependence.py
@@ -15,7 +15,8 @@ from ..common import convert_name, approximate_interactions
 def dependence_plot(ind, shap_values, features, feature_names=None, display_features=None,
                     interaction_index="auto",
                     color="#1E88E5", axis_color="#333333", cmap=None,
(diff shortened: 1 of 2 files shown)


slundberg:
Sorry for the delay I have been out sick. Is there a reason you can't do this after drawing by using the `show=False` option?

Benjamin15:
Hey, 
No problem,
Yes, I want to generate a HTML website with the plot, not just a jupyter.
I have two ways to do this : 
- Use the matplotlib object
- Use a BytesIO which contains a png file

I think the second choice is the better way.

slundberg:
Right but can't you run the plot with `show=False` and then just follow that with?: 
```
file = BytesIO()
pl.savefig(file, format='png', bbox_inches="tight")
```

...since savefig will just work with the current active figure.

Benjamin15:
That work ! I didn't know that matplotlib is a global variable or a Singleton. Thanks for the learning !

PR #1103, opened 2020-03-16
Returning fig and axis for matplotlib force plots to allow for customizations and image exports externally

thampiman:


slundberg:
thanks! Does `plt.gca()` not currently work for what you need to get the axis?

PR #1114, opened 2020-03-20
changed uses of equality to 'is' for None, True, and False

jameslamb:
Thanks for this great project!

In this PR, I'd like to propose a small change for safety. When comparing a value to `None`, `False`, or `True`, using `is` is preferred to `==`if you want to strictly check that something is one of those builtins.

For example:

```
>>> 1 == True
True
>>> 1 is True
False
```

Thanks for your time and consideration.


slundberg:
Sounds great, thanks!

PR #1135, opened 2020-04-02
[FIX] kmeans: support sparse

PrimozGodec:
### Issue
`kmeans` function does not support sparse data. Since KernelExplainer, which is used in the combination with the `kmeans`, supports sparse data, it would be nice that `kmeans` support them too. 

`kmeans` do not support data with missing values.

### Changes
- With this PR I change one line of the function and now it perfectly works with sparse data.
- `kmeans` is extended to impute values when missing before performing clustering.


slundberg:
Thank you! Looks good.

PR #1141, opened 2020-04-07
Prevent unexpected `Warning: unrecognized nn.Module`

rightx2:
In these days, there are many codes or programs where coders define their own custom torch module classes inheriting `nn.Module` to build or customize various of models easily. While doing this, a bunch of nested custom Module structure could appear, such as the one I've written in test codes, but `PyTorchDeepExplainer` could not handle it and showed warning like this: `Warning: unrecognized nn.Module` and return `input_gradient`

So, I tried to fix it by recursion and wrote the test code as well.

Please review the code.

### DIFF
--- a/shap/explainers/deep/deep_pytorch.py
+++ b/shap/explainers/deep/deep_pytorch.py
@@ -70,12 +70,13 @@ class PyTorchDeepExplainer(Explainer):
         Recursively for non-container layers
         """
         handles_list = []
-        for child in model.children():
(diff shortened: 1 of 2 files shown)


rightx2:
I've also checked that `d / np.abs(diff).sum()` values (in test_deep.py) are remain same even after changing the codes

slundberg:
Thanks! @gabrieltseng this look good to you?

gabrieltseng:
Yes - this looks good! Thank you @rightx2

PR #1148, opened 2020-04-10
Fixes #1145: Force_plot function with link=logit & matplotlib=True, displays incorrect results for feature importance

### LINKED ISSUE #1145: Force_plot function with link=logit displays incorrect results for feature importance
**Description**
Force_plot does not display correct feature importance in case of link=logit
- identical logit shap values result in different importance
- base value + feature importance effects <> output value

**Steps/Code to Reproduce**
```
import shap
import numpy as np
shap.force_plot(base_value=-3,
                shap_values=np.array([1,1,-1]),
                feature_names=['PosFeat1', 'PosFeat2', 'NegFeat1'],
                matplotlib=True,
                link="logit")
```
afbeelding

**Expected Results**
PosFeat1, PosFeat2 and NegFeat1 all have exactly the same shap value (either positive or negative).
Each feature is represented by same size effect in plot.
Difference in length of negative and positive effects should equal difference base value vs output value

**Actual Results**
PosFeat2 effect is smaller than PosFeat1 effect.
Sum PosFeat effect - NegFeat1 effect < Output value - Base value

**Analysis**
Logit function is not linear, therefore the calculation of the effect is dependent of the order (and base value) of the calculation.

**Proposed solution**
1. Approximate the logistic function by a linear function
2. Calculate the slope: (proba output value - proba base value) / sum shap values
(Special case sum=0; use derivative of logistic: exp(shap) / (exp(shap) + 1)^2
3. Multiply all shap values by the slope

This way identical shap values result in identical probability effects.
Base value and feature effects sum up to the output value.

ehuijzer:
Fixes #1145
Force_plot function with link=logit displays incorrect results for feature importance

### DIFF
--- a/shap/plots/force_matplotlib.py
+++ b/shap/plots/force_matplotlib.py
@@ -227,36 +227,42 @@ def format_data(data):
         convert_func = lambda x: 1 / (1 + np.exp(-x))
     else:
         assert False, "ERROR: Unrecognized link function: " + str(data['link'])
-    
+
+    # Convert output value and base value and calculate slope of linear approximation of link function
+    sum_shap = data['outValue'] - data['baseValue']
+    data['outValue'] = convert_func(data['outValue'])
+    data['baseValue'] = convert_func(data['baseValue'])
+    if sum_shap != 0:
+        slope = (data['outValue'] - data['baseValue']) / sum_shap
+    else:
+        if data['link'] == 'identity':
+            slope = 1
+        else: # data['link'] == 'logit':
+            slope = -(data['outValue']-1)*data['outValue']
+
     # Convert negative feature values to plot values
     neg_val = data['outValue']
     for i in neg_features:
         val = float(i[0])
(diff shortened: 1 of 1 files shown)


slundberg:
Fixing this is great! But would you be able to try and tackle updating the axis tick marks and leaving the pixel-level measurements in logit space? This way it will exactly match what the JS version does. The linear approximation has some subtle issues and only is reasonable if you do what we do in DeepExplainer and linearize individually for each background data point (not something that should be happening during plotting).

ehuijzer:
I will (try to) update the code the have matplotlib match the JS version (which clearly isn't the case at this time)
afbeelding

ehuijzer:
Changing the axis appeared not to be too difficult.
JS version and matplotlib version do differ in tick marks displayed.
Please let me know if this is an issue.

afbeelding

slundberg:
Excellent! Thank you so much, that looks great.

PR #1154, opened 2020-04-15
Refactor force plots to decouple from IPython using _repr_html_.

timothyrenner:
Decouples the force plots from IPython displays by implementing a `_repr_html_` method. They still display in Jupyter notebooks as expected, but are also now able to be saved as raw HTML / JS for use in Jinja templates or other HTML based templating environments. My main use case is to produce SHAP reports, including force plots, as part of an end-to-end ML training pipeline, so non-interactivity is important. I'll file an issue describing in more detail in case there's a better way to implement this.

The changes:

* There's a new function, `shap.getjs()` that gets the javascript string. This is necessary for rendered HTML templates to work. `shap.initjs()` still works as expected in Notebooks.
* The `Visualizer` classes now inherit from `BaseVisualizer`, used for type checks.
* Visualizer classes now have `_repr_html_`, which displays them in Jupyter environments as expected without explicitly registering a function or returning IPython.display classes.
* `force_plot` now returns the Visualizer object directly. The Visualizer object renders as HTML in a notebook as expected, but now the `.html()` method returns a raw HTML string, allowing it to be injected into HTML templates.
* `save_html` takes the Visualizer object instead of an IPython HTML object, but otherwise works exactly the same way.

### DIFF
--- a/shap/__init__.py
+++ b/shap/__init__.py
@@ -31,7 +31,7 @@ if have_matplotlib:
     from .plots.summary import summary_plot
     from .plots.decision import decision_plot, multioutput_decision_plot
     from .plots.dependence import dependence_plot
-    from .plots.force import force_plot, initjs, save_html
+    from .plots.force import force_plot, initjs, getjs, save_html
     from .plots.image import image_plot
     from .plots.monitoring import monitoring_plot
     from .plots.embedding import embedding_plot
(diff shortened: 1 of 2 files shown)


timothyrenner:
Related issue: https://github.com/slundberg/shap/issues/1155

Also, I had some tests fail on my local box from additivity checks but these appear to be unrelated to the changes I made.

slundberg:
Just got a chance to go through this. Looks great, thanks!

PR #1180, opened 2020-04-26
Added ability to select which features to display.

azanovivan02:
Added ability to select which features to display. Possible to select features using either names, indices or their combination.

### DIFF
--- a/shap/__init__.py
+++ b/shap/__init__.py
@@ -28,7 +28,7 @@ try:
 except ImportError:
     have_matplotlib = False
 if have_matplotlib:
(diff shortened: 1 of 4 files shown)


slundberg:
Thanks for this and sorry for the slow review, I was out sick. This is related to an API upgrade I have been working on, so I'll get back to this shortly I can compare how they should mesh. Basically we are looking at having a smarter sliceable explanation object (which should be cool :) ) and that object might have some of this functionality, so need to compare them once it is finalized.

Thanks again!

PR #1211, opened 2020-05-14
Enable KernelShap support for string type input

YotamElor:
SKLearn pipelines use encoders to support string input e.g. OneHotEncoder and OrdinalEncoder. In some cases it is desired to calculate the shap values in the original input space i.e. before encoding it into numeric values. While there is no theoretical limitation, currently KernelExplainer does not support string type input due to the limitations of a single utility function. The utility function varying_groups() relies on numpy functions that assume numeric input. After fixing it to support string type input, KernelExplainer fully supports string values.

### DIFF
--- a/shap/explainers/kernel.py
+++ b/shap/explainers/kernel.py
@@ -423,6 +423,11 @@ class KernelExplainer(Explainer):
 
         return phi
 
+    def not_equal(self, i, j):
+        if isinstance(i, str):
+            return 0 if i is None or j is None or i == j else 1
+        return 0 if np.isclose(i, j, equal_nan=True) else 1
+
     def varying_groups(self, x):
         if not sp.sparse.issparse(x):
             varying = np.zeros(self.data.groups_size)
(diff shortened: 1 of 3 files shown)


niloofardadras:
@YotamElor  hi, has this been implemented yet?
I am looking to use string features for kernel explainer and I am not able to do anything except label encoding but shap results do not make sense for them. please let me know if there is any solution for it.

YotamElor:
Yes, it is implemented, feel free to use this code

niloofardadras:
@YotamElor i have tried make the change in package but it gave following error. i use tensorflow and my pipeline 1.uses dataframe with both string and floats 2. passes dataframe to create input tensor of string and floats 3. passes the input for prediction. i get following error when i make the changes you have in kernel.py.

"tensorflow.python.framework.errors_impl.InternalError: Unable to get element as bytes"

slundberg:
Thanks @YotamElor ! I added a quick sum test to the unit test. Assuming that checks out I'll go ahead and merge this :)

niloofardadras:
@slundberg   Hi scott,
is there any plan to have support for string type input with deep explainer maybe? as i have tried seems tonot be supported with deep explainer also.

Best regards

rchatwin1:
@YotamElor: Your not_equal function does not handle missing data. If you call not_equal(i, j) with i = nan, then isinstance(i, str) will resolve to False. If at the same time j is a string, then the call to np.isclose will fail. This is a common use case when using pandas dataframes where all null data is set to nan regardless of whether the column data type is 'object' or some numeric type. 

Also, the behavior with missing data is different when comparing strings than when comparing numerics: with numerics, not_equal returns 1 when one entry is missing but the other is not, whereas with strings not_equal returns 0 if either entry is None.

YotamElor:
@rchatwin1 , thank you for noting that. I fixed that behavior and added tests to verify consistent behavior for string and numeric values

Issue #335, opened 2018-11-22
AdaBoost Support


DNCoelho:
Hello,

First of all, thank you very much for developing this amazing tool.

I just wanted to ask if there was any plans to add support to AdaBoost as the title implies.
Or if support is not planned and this is not a too great pain in the ass a quick explanation of how one could go about adding it.

Many thanks in advance.

slundberg:
Thanks! I don't have plans to add support, but it shouldn't be too hard. TreeExplainer creates a TreeEnsemble object from whatever model type we are trying to explain, and then works with that downstream. So all you would need to do is and add another if statement in the TreeEnsemble constructor similar to the one for gradient boosting: https://github.com/slundberg/shap/blob/4bd41ea9592475370905eadd47f6b4879eeebf49/shap/explainers/tree.py#L433-L443

If you have specific questions let me know and I'll be happy to help.

DNCoelho:
Tank you for the quick response!

maxencebrochard:
Hi @DNCoelho,

Did you manage to implement Adaboost? If so, would be very interested :) 

Thanks

hdnh2006:
I left an answer in this [link](https://stackoverflow.com/questions/60433389/how-to-calculate-shap-values-for-adaboost-model/61108156#61108156) that surely could help.

slundberg:
@hdnh2006 if you added support for AdaBoost to your version please consider a PR :)

M-Schreurs:
Since @hdnh2006 solution in the comments [here](https://stackoverflow.com/questions/60433389/how-to-calculate-shap-values-for-adaboost-model/61108156#61108156), the shap module has updated again.
Does someone know how to modify it to get it running again?

tudstudent:
This has been fully implemented in #4271, which adds `AdaBoostClassifier` (SAMME) support to `TreeExplainer`.

The implementation correctly decomposes AdaBoost's weighted hard votes into per-leaf SHAP-compatible values and adds a new C++ softmax vector transform for the multiclass probability output. Both binary and multiclass are supported with `model_output="raw"` (decision function) and `model_output="predict_proba"` (probabilities), with full SHAP additivity verified in tests.

PR #1219, opened 2020-05-20
add AdaBoostClassifier

tk27182:
I had issues using AdaBoostClassifier with your package, and came across #335 and this [Stack Overflow question and answer](https://stackoverflow.com/questions/60433389/how-to-calculate-shap-values-for-adaboost-model/61108156#61108156). It seemed like the OP never added a pull request, so I figured I'd add it with the updated code that uses `safe_isinstance` instead. 

Edits are in lines 598-610 in the `shap/explainers/tree.py` file.

### DIFF
--- a/shap/explainers/tree.py
+++ b/shap/explainers/tree.py
@@ -595,6 +595,19 @@ class TreeEnsemble:
             self.trees = [Tree(e.tree_, normalize=True, scaling=scaling, data=data, data_missing=data_missing) for e in model.estimators_]
             self.objective = objective_name_map.get(model.criterion, None)
             self.tree_output = "probability"
+            
+        ### Added AdaBoostClassifier based on the outdated StackOverflow response and Github issue here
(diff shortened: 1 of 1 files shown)


slundberg:
Thanks!! Could you also add a quick sum check to make sure there are no bugs now or in the future? See this example for the random forest: https://github.com/slundberg/shap/blob/fc30c661339e89e0132f5f89e5385e3681090e1f/tests/explainers/test_tree.py#L605-L619

tudstudent:
This has been fully implemented in #4271, which adds `AdaBoostClassifier` (SAMME) support to `TreeExplainer`.

The implementation correctly decomposes AdaBoost's weighted hard votes into per-leaf SHAP-compatible values and adds a new C++ softmax vector transform for the multiclass probability output. Both binary and multiclass are supported with `model_output="raw"` (decision function) and `model_output="predict_proba"` (probabilities), with full SHAP additivity verified in tests.

PR #1248, opened 2020-06-03
use safe_instance for plots fixes #1247

### LINKED ISSUE #1247: use safe_instance to check for DataFrames in plot functions
The current method for checking if dataframe was passed will fail to detect subclasses 
PR #871  added a method `safe_instance` which can solve this issue (without requiring pandas)

peterdhansen:
uses safe_instance to detect pandas in plotting functions


peterdhansen:
this no longer applies since plots take Explanation object

PR #1267, opened 2020-06-13
Bypass warning for feature_perturbation being "tree_path_dependent"

InterferencePattern:
when feature perturbation is manually set to "tree_path_dependent", issue 1236


Thomas9292:
LGTM!

lrjball:
Looking on master, it looks like the warning has since been removed (or at least commented out) so this fix may not be required

PR #1306, opened 2020-07-09
Fix doc after param name change

candalfigomoro:
Change `feature_dependence="independent"` to `feature_perturbation="interventional"` which is the new param/value.


lrjball:
@candalfigomoro Would it be possible for you to update this PR since `tree.py` has since been renamed to `_tree. py`? Renaming should fix the merge conflict and then this can hopefully go in

candalfigomoro:
`logloss` also changed to `log_loss`

thatlittleboy:
Closing in favour of #3006 since this PR is now stale.

PR #1307, opened 2020-07-09
Update tree.py

guidiandrea:
commented out self.original_model = model in case of pyspark model

After a talk on ISSUE #884, I created this pull request. Please @QuentinAmbard review it and accept if you think it's reasonable.

:)


QuentinAmbard:
Thanks! shouldn't we just delete the line instead ?
We should also add a unit test to make sure it's serializable to prevent this to happen again (when we implement the predictions for example).

guidiandrea:
Right! 
Also, I noticed that a test caused a build to fail; also the previous PR got the same problem

PR #1314, opened 2020-07-14
Update summary plot to support custom colormaps

nasir-bhanpuri:
Addresses https://github.com/slundberg/shap/issues/62 for summary plots.

Creates a new argument cmap that replaces hard-coded colors.red_blue in several places.
After making these change locally, I am able to customize the colormap on summary plots.


slundberg:
Looks great, thanks!

PR #1327, opened 2020-07-23
Enable KernelShap support for string type input

YotamElor:
SKLearn pipelines use encoders to support string input e.g. OneHotEncoder and OrdinalEncoder. In some cases it is desired to calculate the shap values in the original input space i.e. before encoding it into numeric values. While there is no theoretical limitation, currently KernelExplainer does not support string type input due to the limitations of a single utility function. The utility function varying_groups() relies on numpy functions that assume numeric input. After fixing it to support string type input, KernelExplainer fully supports string values.

### DIFF
--- a/shap/explainers/_kernel.py
+++ b/shap/explainers/_kernel.py
@@ -381,6 +381,12 @@ class Kernel(Explainer):
 
         return phi
 
+    @staticmethod
+    def not_equal(i, j):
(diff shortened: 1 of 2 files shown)


slundberg:
This looks great, thanks for kindly redoing it for a clean merge!

YotamElor:
Thank you so much for merging!

Yotam

PR #1350, opened 2020-08-06
Correct samples per feature calculation in SamplingExplainer

### LINKED ISSUE #1349: SamplingExplainer draws incorrect number of samples.
The `SamplingExplainer` doesn't draw the number of samples specified by the `nsamples` keyword argument in the `.shap_values` method. 

The reason is in [these two lines](https://github.com/slundberg/shap/blob/0.35.0/shap/explainers/sampling.py#L115-L116). During the second round of sampling, remaining samples are allocated to features according to variance in estimates from the first round of sampling, represented by `phi_var` which has shape `(n_features, n_model_outputs)`. `phi_var` is first normalised by `phi_var.sum()`, then the number of samples for each feature is weighted by `phi_var.mean(1)`, i.e. the mean (taken over model outputs) variance for that feature. However since the entire array has been normalised to sum to 1, the columns sum to `1 / n_model_outputs` rather than `1`, resulting in the number of samples for round two being dramatically reduced.

Here's a minimal example using a simple MNIST classifier. In the second round, the number of remaining samples is reduced by 90% (as the model has 10-dimensional output) resulting in a fraction of the requested samples being drawn. I wrote a little wrapper class to keep track of both the number of times the model is evaluated and the number of data points it is evaluated on.

```python
import numpy as np
import shap
import tensorflow as tf

# load MNIST data from Keras
(x_train, y_train), (x_test, y_test) = tf.keras.datasets.mnist.load_data()

x_train = x_train.reshape(-1, 784).astype(np.float32) / 255
y_train = y_train.astype(np.int32)
x_test = x_test.reshape(-1, 784).astype(np.float32) / 255
y_test = y_test.astype(np.int32)

# train simple feed forward network
model = tf.keras.Sequential([
    tf.keras.layers.Dense(100, activation="relu"),
    tf.keras.layers.Dense(10, activation="softmax"),
])

model.compile(
    tf.optimizers.Adam(),
    "sparse_categorical_crossentropy",
    metrics=["accuracy"],
)

model.fit(
    x_train,
    y_train, 
    batch_size=128,
    validation_split=0.25,
    epochs=200,
    callbacks=[tf.keras.callbacks.EarlyStopping(patience=2, restore_best_weights=True)],
)

# wrap model.predict to keep track of how many model evaluations and how many
# spliced data points are passed to the model for evaluation
class PredictCounter:
    def __init__(self, predict_fn):
        self.predict_fn = predict_fn
        self.reset()
    
    def __call__(self, x, *args, **kwargs):
        self.eval_count += 1
        self.rows += x.shape[0]
        return self.predict_fn(x, *args, **kwargs)
    
    def reset(self):
        self.eval_count = 0
        self.rows = 0


model.predict = PredictCounter(model.predict)

# create the explainer
explainer = shap.SamplingExplainer(model.predict, x_test)

# sample the Shapley values for a particular data point
res = explainer.shap_values(x_test[0], nsamples=500_000)

# report model evaluations and number of data points evaluated
print("Evaluation count:", model.predict.eval_count)
print("Rows evaluated:", model.predict.rows)
```

The output I got was 

```
Evaluation count: 1338
Rows evaluated: 121433
```

Showing that the model was evaluated on 120k spliced data points rather than the expected and requested 500k.

I have submitted #1350 to address this issue.

As a side note, I notice that the number of samples that this method will produce is actually `nsamples // 2`, as model evaluations with the feature of interest turned on and turned off are [counted separately](https://github.com/slundberg/shap/blob/0.35.0/shap/explainers/sampling.py#L173-L174). I would have expected a sample to correspond to a single coalition of features and the difference in resulting model evaluations. Would be happy to include a fix to that also in my PR if you think it's worth correcting.

tcbegley:
As reported in #1349 (see there for details), the number of samples drawn by SamplingExplainer does not match that specified by `nsamples` due to an error in the variance normalisation. This small PR corrects that bug.

As I pointed out in the issue, I think that `nsamples` is additionally out by a factor two, as it refers to model evaluations (of which there are two per coalition of features sampled + spliced data point) rather than coalition samples as I would have expected. Happy to add a fix for that to this PR if you think it's worth fixing.

Closes: #1349

### DIFF
--- a/shap/explainers/_sampling.py
+++ b/shap/explainers/_sampling.py
@@ -131,7 +131,7 @@ class Sampling(Kernel):
             # optimally allocate samples according to the variance
             if phi_var.sum() == 0:
                 phi_var += 1 # spread samples uniformally if we found no variability
-            phi_var /= phi_var.sum()
+            phi_var /= phi_var.sum(0)[np.newaxis, :]
             nsamples_each2 = (phi_var[self.varyingInds,:].mean(1) * round2_samples).astype(np.int)
             for i in range(len(nsamples_each2)):
                 if nsamples_each2[i] % 2 == 1: nsamples_each2[i] += 1


slundberg:
Thanks! This is probably because it was tested just on single output models. I would be happy to merge a nsamples /2 PR. Since you are using SamplingExplainer, you might also check out the new `shap.explainers.Permutation` explainer. It uses the new API that is going through final stabilization still, but it should run much faster than SamplingExplainer and converge faster for models dominated by additive effects.

tcbegley:
Thanks! I'll prepare a quick second PR and definitely check out the permutation explainer.

PR #1353, opened 2020-08-07
fixed minor doc issue in notebook

imatiach-msft:
should be 5000 samples, not 500, if I understood the notebook correctly


imatiach-msft:
hmm seeing some build failures, not quite sure what the root cause is

slundberg:
Thanks!

PR #1368, opened 2020-08-17
BLD: Cleaned imports and some refactoring

JoElfner:
While trying to get a local fork of shap running, I realized that many imports depend on an installed version of shap.
F.i. in ` shap/plots/_partial_dependence.py`:
```
import shap
from ..utils import convert_name, sample, safe_isinstance
```
This results in requiring checks as `safe_isinstance` to compare against a possibly not yet installed shap module, while being in the shap module env. Furthermore, changes in the module env are not reflected in the current build, as long as the module is not installed using `setup.py`.

Furthermore I started refactoring some parts of `plots._summary.py`.

### DIFF
--- a/shap/explainers/_deep/__init__.py
+++ b/shap/explainers/_deep/__init__.py
@@ -1,6 +1,6 @@
 from .deep_pytorch import PyTorchDeep
 from .deep_tf import TFDeep
-from shap.explainers._explainer import Explainer
+from .._explainer import Explainer
 
 
 class Deep(Explainer):
--- a/shap/explainers/_deep/deep_pytorch.py
+++ b/shap/explainers/_deep/deep_pytorch.py
@@ -1,6 +1,6 @@
 import numpy as np
 import warnings
-from shap.explainers._explainer import Explainer
+from .._explainer import Explainer
 from distutils.version import LooseVersion
 torch = None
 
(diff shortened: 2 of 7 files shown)


slundberg:
This looks good in general. The problem is that _summary is now split into in _bar and _violin and _beeswarm (since there really were 3 different plot types in _summary). Could you adjust this PR to resolve the conflicts? I could do some of this manually but then I can't merge this PR (and so appropriately credit you for the changes)

JoElfner:
> 
> 
> This looks good in general. The problem is that _summary is now split into in _bar and _violin and _beeswarm (since there really were 3 different plot types in _summary). Could you adjust this PR to resolve the conflicts? I could do some of this manually but then I can't merge this PR (and so appropriately credit you for the changes)

Thanks! I pulled your changes made a clean new local branch, since otherwise implemented the big amount of new code would lead to many conflicts... Thus it'll take me some time to re-implement it.

Edit: Done. Also corrected some references to the now non-existing summary function.
But I am still not sure what the new API is supposed to look like. Is the summary plot, now called violin, going to change to only plot violins? And If yes, is the beeswarm meant to replace the dot plot? Meaning I'd have to implement this in beeswarm?

slundberg:
The current/old summary_plot function really has two plot types inside it, a beeswarm "dot" plot, and a violin plot (with some variations). The goal of the new API was to be more transparent about the type of plot, by naming functions by plot type and not by purpose. So the summary_plot is now in two files, and plots.beeswarm is the one the matches the current default behavior of summary_plot, while plots.violin will match the current summary_plot set to violin mode. The separation still has some cleanup left at the code level.

Issue #58, opened 2018-04-17
Support summary_plot for categorical features


annaveronika:
We have implemented shap_values in catboost (http://catboost.yandex or https://github.com/catboost/catboost), it works really cool, thank you very much!
It'll be on pypi in tomorrow's release.

One problem with visualization is that you cannot visualize summary_plot if some features are categorical.
I suggest in this case to draw all the dots in gray color for example. It will still make a lot of sense, because you see the effect of categorical feature value on the objects in the train dataset.

Here is a code example that fails:
```
X_adult,y_adult = shap.datasets.adult(display=True)
cat_features = [1, #Work
                    3,#Marital status
                    4,#Occupation
                    5,#Relationship
                    6,#Race
                    7,#Sex
                    11 #Country
]
from catboost import *
model = CatBoostClassifier(
    iterations=300, learning_rate=0.1, random_seed=123,
    max_ctr_complexity=1, # Currently we don't calculate shap values for models with categorical feature combinations.
                          # This will be changed in one of the next releases.
)
model.fit(X_adult, y_adult, cat_features=cat_features, verbose=False)
shap_values = model.get_feature_importance(data=Pool(X_adult, y_adult, cat_features=cat_features), fstr_type='ShapValues')
shap.summary_plot(shap_values, X_adult)
```

We have also added a tutorial to our github, it uses a summary_plot for cat features, and it's informative (are you ok with us taking the first part from Shap values?): https://github.com/catboost/catboost/blob/master/catboost/tutorials/advanced_tutorials/shap_values_tutorial.ipynb

alexandregerbeaux:
Hi Annaveronika,
Thank you for the post! Wanted to have a look on your notebook, is it on the master branch yet?

annaveronika:
Thank you for a quick answer! Yes, it'll sync there from our internal repo in about half an hour. Here is the file: 
[shap_values_tutorial.txt](https://github.com/slundberg/shap/files/1920978/shap_values_tutorial.txt)
You need to change it from txt to ipynb first, github doesn't allow ipynb.

annaveronika:
Actually it's already on github, you don't need the file

annaveronika:
Plots are not shown for ipynb, so I'm converting the file to html. The link will be
https://github.com/catboost/catboost/blob/master/catboost/tutorials/advanced_tutorials/shap_values_tutorial.html

slundberg:
[...]

PR #1370, opened 2020-08-17
ENH `plots._summary.summary_plot`: Added functionality to plot categoric features with separate markers

JoElfner:
I added the funcitonality to plot categoric features (single, multiple, arbitrary number of value combinations) to `summary_plot`. Scanning #58 shortly, this could be a fix for the issue (didn't go into detail while looking at the requested functionality).

To enable marking values of a categoric feature with separate markers, set `categoric=feat_name`, where `feat_name` can be:
- str or list/tuple of str for `isinstance(features, pd.DataFrame)`
- int or list/tuple of int for all types
- None (default) to disable categoric plotting

This should work with `pd.DataFrame` and `np.ndarray` inputs. The `dtype` can be anything, also `str` values in columns are supported.

`categoric_names` allows specifying the names of the categoric features when no `pd.DataFrame` is given.
`cat_legend` controls whether to display a legend with the categoric combinations.
`markers` specifies the markers to use for the categorics. Since `linewidth=0` is the default plotting style, only "thick" markers will be displayed. I didn't want to change this default, but of course this should be an easy task.

This works only for `plot_type in ('dot', 'compact_dot)`.

All changes specific to this PR are in `plots._summary.py`. Other changed files are actually part of #1368 .

Any remarks, objections etc?

### DIFF
--- a/shap/plots/_beeswarm.py
+++ b/shap/plots/_beeswarm.py
@@ -43,7 +43,6 @@ def beeswarm(shap_values, max_display=10, order=Explanation.abs.mean(0),
 
     # support passing an explanation object
     if str(type(shap_values)).endswith("Explanation'>"):
-        breakpoint()
         shap_exp = shap_values
         base_values = shap_exp.base_values
         values = shap_exp.values
--- a/shap/plots/_beeswarm.py
+++ b/shap/plots/_beeswarm.py
@@ -35,7 +35,7 @@ def beeswarm(shap_values, max_display=10, order=Explanation.abs.mean(0),
 
     plot_size : "auto" (default), float, (float, float), or None
         What size to make the plot. By default the size is auto-scaled based on the number of
-        features that are being displayed. Passing a single float will cause each row to be that 
+        features that are being displayed. Passing a single float will cause each row to be that
         many inches high. Passing a pair of floats will scale the plot by that
         number of inches. If None is passed then the size of the current figure will be left
         unchanged.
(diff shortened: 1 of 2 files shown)


JoElfner:
Added fixes for cases where `features=None` and features is a np.ndarray and no categories are used.

JoElfner:
Ehm... No idea why the build is failing now. Travis CI was running smoothly before removing `itertools` import. Now it fails during install of `numba`/`llvmlite`. Seems unconnected to this PR.
The failing of AppVeyor at `tests.explainers.test_tree.test_lightgbm_ranking` is imho also not connected to this PR.

JoElfner:
@slundberg So what do you think?

JoElfner:
Since there have been 24 merged commits since I opened this PR 21 days ago and there was not even a single reaction on this PR, I take this as an indirect way of saying: "No, I don't want this feature."
Is this assumption correct?

slundberg:
@JoElfner So sorry about this! I wasn't sure how to move forward when I first read this, and then just failed to come back in midst of pushing out a bunch of things. I'll look this over tomorrow more carefully and get back to you.

JoElfner:
> 
> 
> @JoElfner So sorry about this! I wasn't sure how to move forward when I first read this, and then just failed to come back in midst of pushing out a bunch of things. I'll look this over tomorrow more carefully and get back to you.

No problem! There is no rush, as long as I know that you'll take a look at it sometime in the next weeks. :)

JoElfner:
I think I now understand what the new API is supposed to look like. `beeswarm` is supposed to be the replacement of the dot-type summary plot, now called `summary_legacy`, while `violin` is going to be the violin plot sub-type replacement.
So the categoric plots for dot-type summary should be placed in beeswam, right?

slundberg:
That's right! Sorry I was looking for this in the other PR and missed your answer here :)

JoElfner:
Sorry for my late answer... Thanks for the clarification! I guess then I'll wait until the final implementation of `beeswam` is merged before I start working on this PR again. Could you please ping me as soon as the beeswarm implementation is finalized?

Issue #1347, opened 2020-08-05
Can't run genomics example


f6v:
Running `DeepExplainer`  example with genomics data results in following:

> TypeError                                 Traceback (most recent call last)
> <ipython-input-9-1ac975dbed95> in <module>()
>      14 seqs_to_explain = onehot_data[[0,3,9]] #these three are positive for task 0
>      15 dinuc_shuff_explainer = shap.DeepExplainer((keras_model.input, keras_model.output[:,0]), shuffle_several_times)
> ---> 16 raw_shap_explanations = dinuc_shuff_explainer.shap_values(seqs_to_explain)
>      17 
>      18 #project the importance at each position onto the base that's actually present
> 
> 1 frames
> /usr/local/lib/python3.6/dist-packages/shap/explainers/deep/deep_tf.py in shap_values(self, X, ranked_outputs, output_rank_order, check_additivity)
>     316             else:
>     317                 model_output = self.model(X)
> --> 318             for l in range(len(self.expected_value)):
>     319                 if not self.multi_input:
>     320                     diffs = model_output[:, l] - self.expected_value[l] - output_phis[l].sum(axis=tuple(range(1, output_phis[l].ndim)))
> 
> TypeError: object of type 'NoneType' has no len()

I'm having this issue on Colab with tf 1.15.2 and on Jupyter with tf 1.13.1 and Python 3.6.6.
With tf 2.3.0 on Colab there's different issue:

> NameError                                 Traceback (most recent call last)
> <ipython-input-9-1ac975dbed95> in <module>()
>      13 
>      14 seqs_to_explain = onehot_data[[0,3,9]] #these three are positive for task 0
> ---> 15 dinuc_shuff_explainer = shap.DeepExplainer((keras_model.input, keras_model.output[:,0]), shuffle_several_times)
>      16 raw_shap_explanations = dinuc_shuff_explainer.shap_values(seqs_to_explain)
>      17 
> 
> 1 frames
> /usr/local/lib/python3.6/dist-packages/shap/explainers/deep/deep_tf.py in __init__(self, model, data, session, learning_phase_flags)
>     105             if type(model) is tuple or type(model) is list:
>     106                 assert len(model) == 2, "When a tuple is passed it must be of the form (inputs, outputs)"
> --> 107                 self.model = Model(model[0], model[1])
>     108             else:
>     109                 self.model = model
> 
> NameError: name 'Model' is not defined

Would appreciate any tips!

suragnair:
[...]

PR #1407, opened 2020-09-03
Missing import

suragnair:
Should solve #1347


slundberg:
Thanks @suragnair ! Not sure how that bug went uncaught, but glad you fixed it :)

PR #1412, opened 2020-09-06
fix isolation forest when max_features < 1

### LINKED ISSUE #1389: Isolation forest values look incorrect when features are sampled
When a scikit learn IsolationForest is created with max_features  < 1.0 the shap values are incorrect
I am using this formula to compare score_samples and shape_values

```python
-2 **(-(explainer.shap_values(fit_data).sum(axis=1) + explainer.expected_value) / _average_path_length([model.n_estimators]))
```

zhanjiezhu:
fix #1389 

Issue is caused by each tree gets a different set of features when max_features < 1, the solution is to re-number the feature indexes with the attribute "estimators_features_".

Also updated the test to include this case.


slundberg:
Excellent, thanks @zhanjiezhu !

ssoftcheck:
Thanks @zhanjiezhu !

Issue #1369, opened 2020-08-17
Documentation


BrunoBarre:
Hello, is this normal that since a few weeks, the documentation, [Doc page,](https://shap.readthedocs.io/en/latest/) is almost empty. Now it only contains the different function names, but no more.

Is there a better page (couldn't find it yet), or is it possible to come back to a previous version of this page, which had better documentation?

Thanks a lot.

slundberg:
Thanks for bringing this up! I didn't think about how this would happen during the refactor...

The issue is that we just pushed a significant new/cleaner API for SHAP and everything has been moved around. So that must have messed up Sphinx. I will take a look at this as soon as I can, but I expect it will take a week or two since I am about to switch over the docs to the new API (note the old API should all still work fine), and that will change all these pages.

jonnor:
The documentation page is up again now, and have been for a year or so.
This issue can be closed, CC @slundberg

PR #1428, opened 2020-09-12
Docs update

lrjball:
There is currently an issue with the docs not being displayed properly (#1369). This PR should fix this issue, and a working version has been built using my fork: https://shap-lrjball.readthedocs.io/en/docs_update/

The main issue was that autodoc doesn't work properly when aliases are used, like `from .explainers._tree import Tree as TreeExplainer`, and instead of showing the docs it will just show 'alias of Tree'. This is fixed by adding a setup to `conf.py` which reassigns the `__name__` attr for all of the aliased classes. 

I've also made a couple of additional tweaks to the layout and content to aid usability, but can happily make updates if needed.


slundberg:
This is excellent! Thank you!! I looked it over and it all looks great. One forward looking issue is that 0.36.0 has a new API, that will inform how we arrange things here, but that will just be a natural future update on top of all these fixes. I am going to merge now, and leave future iterations for the future :)

lrjball:
No worries. Yeah I saw there were a few changes coming up, but it should be easy to update now that it is in

lrjball:
Also, would it be possible to add the docs page in the website field of the about section on the repo? Could make it easier for people to find the docs

PR #1442, opened 2020-09-18
Fixed summation of token attribution for multiclass along the class axis

ryserrao:
Presently when we run sum on the explanation object for multiclass, it flattens features and creates new_values which are a set of arrays(with len=no. of classes) corresponding to every unique token but doesn't sum across individual dimension and instead produces one overall score as feature contribution for a token. Hence modified that to sum contribution separately for each class so we now have contribution of every token towards each class


slundberg:
Thanks!

PR #1443, opened 2020-09-18
Updates for images multi-class output

anusham1990:
Changes made to files:
1. _image.py: Added an if to check if labels is None, only then assign the shap output names
2. _partition.py: Added output names as an argument to super class constructor, commented out assigning output names again in constructor and added output_indices as a key to dictionary returned in explain_row
3. _explainer.py: Added output_names as an argument when creating Partition explainer, got output_indices from row_result, collapsed the output_indices list, obtained sliced_labels from output_names using output_indices, passed sliced_labels to output_names of Explanation object 
4. _explanation.py: Added a similar check as feature names for output_names for Alias


slundberg:
Thanks!

PR #1447, opened 2020-09-19
Change SamplingExplainer's use of nsamples

tcbegley:
Previously `nsamples` in `SamplingExplainer` corresponded to the number of model evaluations, of which there were two per coalition of features. With this change `nsamples` instead counts the number of sampled coalitions (meaning there will be `2 * nsamples` model evaluations).

This change was previously discussed in #1349 and #1350.


slundberg:
Thanks!

PR #1455, opened 2020-09-21
Multiclass text explanation notebook

ryserrao:
1. Added notebook for multiclass partition text explainer on emotion dataset 
2. Passing feature_names and output_names as list to Alias as it expects a list when slicing the explanation object
3. Update to feature ordering in _bar.py to not order features based on abs() value. This helps to visualize raw values of features negatively contributing to the output class


slundberg:
Thanks! Could you remove the `_bar.py` edits though? Rather than hard coding that, we should really allow any sorting through the `order` parameter like:
`shap.plots.bar(shap_values[0], order=shap.Explanation.identity)` which is now supported :)

ryserrao:
Thanks Scott! I have merged the changes from the master so does not have abs() any more in ordering. And updated some of the plots in the notebook to now use the order=shap.Explanation.identity parameter instead

PR #1470, opened 2020-09-25
Non-Fast Tokenizer Support - Text Masker

vivekchettiar:
This change allows for users to pass non-fast tokenizers to the explainers. Since these tokenizers do not return output mapping indices, we instead clean and concatenate tokens to produce masked inputs.


slundberg:
Looks great!

PR #1481, opened 2020-09-28
Use pseudo-inverse instead of inverse for non-singluar matrices in linear regression

### LINKED ISSUE #240: SingularMatrixException with Shap KernelExplainer on SVM
Hello, 

Sorry if this is a duplicate post, but I searched for it and didn't find the answer. 

I'm getting a SingularMatrixException with the KernelExplainer on SVM. My setup is that I do a k-fold cross-validation on SVM and after the training, I run the test samples on the SVM to get the predicted probabilities and the feature importances through Shap's KernelExplainer. I use a single sample of mean features as my background sample. 

                       background = np.zeros((1, len(features)))
                        try:
                            estimator.predict_proba(background)
                            explainer = shap.KernelExplainer(estimator.predict_proba, background, link="logit")
                            shap_values = explainer.shap_values(x_test, nsamples=100)
                        except Exception as e:
                            print("Got exception: " + str(e) + " on " + str(data_set))
                            shap_values = []

From time to time, I get the following exception: 

  File "linear_stats_plots.py", line 482, in <module>
    shap_values = explainer.shap_values(x_test, nsamples=100)
  File "/home/ubuntu/miniconda3/lib/python3.6/site-packages/shap/explainers/kernel.py", line 203, in shap_values
    explanations.append(self.explain(data, **kwargs))
  File "/home/ubuntu/miniconda3/lib/python3.6/site-packages/shap/explainers/kernel.py", line 366, in explain
    vphi, vphi_var = self.solve(self.nsamples / self.max_samples, d)
  File "/home/ubuntu/miniconda3/lib/python3.6/site-packages/shap/explainers/kernel.py", line 483, in solve
    tmp2 = np.linalg.inv(np.dot(np.transpose(tmp), etmp))
  File "/home/ubuntu/miniconda3/lib/python3.6/site-packages/numpy/linalg/linalg.py", line 532, in inv
    ainv = _umath_linalg.inv(a, signature=signature, extobj=extobj)
  File "/home/ubuntu/miniconda3/lib/python3.6/site-packages/numpy/linalg/linalg.py", line 89, in _raise_linalgerror_singular
    raise LinAlgError("Singular matrix")
numpy.linalg.linalg.LinAlgError: Singular matrix

One odd thing is that I don't always get this exception: it depends on the iteration. I would appreciate any insights on this. Thank you very much!

PrimozGodec:
Fixes https://github.com/slundberg/shap/issues/240

With this PR I changed the use of inverse to pseudoinverse in cases when a non-singular matrix appears in the linear regression equation. This is kind of approximation for the inverse in cases when it is not possible to compute the inverse. This way we avoid the problem when code fails because the result of the computation is not invertible since it is not nice that error appears just because the result of some computation is not appropriate.

I know that pseudoinverse is not the ideal solution but it is the best I can imagine in these cases and users get warned that it happened so they can do one of the proposed solutions to avoid this. **As far as I know also other Python-based implementations of the linear regression use pseudoinverse instead of the inverse.**

### DIFF
--- a/shap/explainers/_kernel.py
+++ b/shap/explainers/_kernel.py
@@ -568,7 +568,18 @@ class Kernel(Explainer):
 
         # solve a weighted least squares equation to estimate phi
         tmp = np.transpose(np.transpose(etmp) * np.transpose(self.kernelWeights))
-        tmp2 = np.linalg.inv(np.dot(np.transpose(tmp), etmp))
+        etmp_dot = np.dot(np.transpose(tmp), etmp)
+        try:
+            tmp2 = np.linalg.inv(etmp_dot)
+        except np.linalg.LinAlgError:
+            tmp2 = np.linalg.pinv(etmp_dot)
+            warnings.warn(
+                "Linear regression equation is singular, Moore-Penrose pseudoinverse is used instead of the regular inverse.\n"
+                "To use regular inverse do one of the following:\n"
+                "1) turn up the number of samples,\n"
+                "2) turn up the L1 regularization with num_features(N) where N is less than the number of samples,\n"
+                "3) group features together to reduce the number of inputs that need to be explained."
+            )
         w = np.dot(tmp2, np.dot(np.transpose(tmp), eyAdj2))
         log.debug("np.sum(w) = {0}".format(np.sum(w)))
         log.debug("self.link(self.fx) - self.link(self.fnull) = {0}".format(


slundberg:
Thanks! I think this is a good step forward. I believe the pseudo inverse produces the solution with the lowest squared coefficient values among all possible solutions, so it is like a minimal amount of ridge regression needed to avoid singularity in the inversion. This doesn't mean you are getting a good estimate of the Shapley values, but I think returning something with a warning is a good choice.

PR #1482, opened 2020-09-29
fix conversion warning

wangjoshuah:
Following from https://github.com/scikit-image/scikit-image/pull/4731
Fix warning 
```
Converting `np.inexact` or `np.floating` to a dtype is deprecated. The current result is `float64` which is not strictly correct.
```


slundberg:
Thanks!

PR #1483, opened 2020-09-29
Tensorflow V2.1 and later fixes

Zahlii:
This is still a work in progress. Especially the LSTM-based test is still failing since all values are zero.


Zahlii:
Possible tickets that cover the same area:
https://github.com/slundberg/shap/issues/548
https://github.com/slundberg/shap/issues/1079
https://github.com/slundberg/shap/issues/930

shap/explainers/tf_utils.py
```diff
@@ -43,7 +43,7 @@ def _get_graph(explainer):
     if not tf.executing_eagerly():
         return explainer.session.graph
     else:
-        return explainer.model_output.graph
+        return getattr(explainer.model_output, "graph", None)
```
Zahlii, line 46: Starting with I think tf2.3, this will not contain a reference to the graph, or any of its ops anymore

shap/explainers/_deep/deep_tf.py
```diff
@@ -130,33 +130,35 @@ def __init__(self, model, data, session=None, learning_phase_flags=None):
...
+                for op in self.graph.get_operations():
+                    if 'learning_phase' in op.name and op.type == "Const" and len(op.outputs[0].shape) == 0:
+                        if op.outputs[0].dtype == tf.bool:
+                            self.learning_phase_ops.append(op)
+                self.learning_phase_flags = [op.outputs[0] for op in self.learning_phase_ops]
+            else:
+                self.learning_phase_ops = [t.op for t in learning_phase_flags]
         else:
-            self.learning_phase_ops = [t.op for t in learning_phase_flags]
+            self.learning_phase_ops = self.learning_phase_flags = []
```
Zahlii, line 171: This is a "dirty" workaround for the fact that later TF versions do not expose ops or a graph anymore

shap/explainers/_deep/deep_tf.py
```diff
@@ -130,33 +130,35 @@ def __init__(self, model, data, session=None, learning_phase_flags=None):
...
-                if type(self.model)is tuple:
-                    sel.fModel(cnn.inputs, cnn.get_layer(theNameYouWant).outputs)
                 self.expected_value = tf.reduce_mean(self.model(self.data), 0)
 
         if not tf.executing_eagerly():
-            self._init_between_tensors(self.model_output.op, self.model_inputs)
-
+            self._init_between_tensors(getattr(self.model_output.op, self.model_output), self.model_inputs)
+        else:
+            self.between_tensors = []
```
Zahlii, line 195: Again, dirty workaround

shap/explainers/_deep/deep_tf.py
```diff
@@ -130,33 +130,35 @@ def __init__(self, model, data, session=None, learning_phase_flags=None):
...
+        if hasattr(self.data, '__call__'):
             self.expected_value = None
         else:
             if self.data[0].shape[0] > 5000:
                 warnings.warn("You have provided over 5k background samples! For better performance consider using smaller random sample.")
             if not tf.executing_eagerly():
                 self.expected_value = self.run(self.model_output, self.model_inputs, self.data).mean(0)
             else:
-                if type(self.model)is tuple:
-                    sel.fModel(cnn.inputs, cnn.get_layer(theNameYouWant).outputs)
```
Zahlii, line 154: I do think this line is completely unnecessary, even if it was to work?

shap/explainers/_deep/deep_tf.py
```diff
@@ -704,6 +706,9 @@ def break_dependence(explainer, op, *grads):
 op_handlers["TensorArrayScatterV3"] = passthrough
 op_handlers["TensorArrayReadV3"] = passthrough
 op_handlers["TensorArrayWriteV3"] = passthrough
+op_handlers["TensorListStack"] = passthrough
+op_handlers["While"] = passthrough
+op_handlers["TensorListFromTensor"] = passthrough
```
Zahlii, line 855: Quite possible many of the reported issues fall under this category. I ama ssuming that maybe we need to overwrite the "While" gradient (?)

shap/explainers/_deep/deep_tf.py
```diff
@@ -276,37 +328,54 @@ def shap_values(self, X, ranked_outputs=None, output_rank_order="max", check_add
...
                     bg_data = self.data
 
                 # tile the inputs to line up with the background data samples
-                tiled_X = [np.tile(X[l][j:j+1], (bg_data[l].shape[0],) + tuple([1 for k in range(len(X[l].shape)-1)])) for l in range(len(X))]
+                tiled_x = [
+                    np.broadcast_to(
+                        X[input_idx][sample_index],
+                        shape=bg_data[input_idx].shape
+                    )
+                    for input_idx in range(len(X))
```
Zahlii, line 359: I changed this for me to better understand what was going on and hope this code it semantically the same but easier to read.

shap/explainers/_deep/deep_tf.py
```diff
@@ -209,27 +246,35 @@ def _init_between_tensors(self, out_op, model_inputs):
...
-        """
+        """Return which inputs of this operation are variable (i.e. depend on the model inputs)."""
         if op not in self._vinputs:
             out = np.zeros(len(op.inputs), dtype=np.bool)
-            for i,t in enumerate(op.inputs):
+            for i, t in enumerate(op.inputs):
                 out[i] = t.name in self.between_tensors
+                if not out[i]:
+                    if "lstm/while/lstm_cell" in t.name:
+                        out[i] = True
```
Zahlii, line 256: This could be shortened into one line, but since the additivity doesn't fully hold I am keeping it here so its easier to debug via printing

huberl:
Any update on this?

slundberg:
Sorry to have left this hanging! Now that TF 2.4 is out I ran back across this, and would like to figure out how to get it merged. Looks like there is an additivity check problem right now.

shap/explainers/_deep/deep_tf.py
slundberg: The workaround here is just to not look for learning phase flags....this could cause problems if we need to find those flags during the eval process.

shap/explainers/_deep/deep_tf.py
slundberg: Odd, yes it seems like a left over typo. Thanks!

shap/explainers/_deep/deep_tf.py
slundberg: Can you explain the logic behind this change? It looks a bit like a hack to get around the fact that we no longer have self.between_tensors since we don't have the graph. Is that right?

FRUEHNI1: I actually posted an example explanation and deeper insights here: https://github.com/slundberg/shap/issues/930#issuecomment-700781695 . As stated there, without this SHAP assumes that all LSTM calculations do not impact the model output, thus the SHAP values are all zero. However, even with this "fix"
the results violate the additivity constraints, but since my matrix calculus is a bit rusty I was not able to figure out which gradients would need to be patched to fix this for good. As also explained, we do have the between tensors available 8see also comment), but the lstm_cells are not in them (for whatever reason).

Zahlii: I actually posted an example explanation and deeper insights here: https://github.com/slundberg/shap/issues/930#issuecomment-700781695. 
As stated there, without this SHAP assumes that all LSTM calculations do not impact the model output, thus the SHAP values are all zero. However, even with this "fix" the results violate the additivity constraints, but since my matrix calculus is a bit rusty I was not able to figure out which gradients would need to be patched to fix this for good. As also explained, we do have the between tensors available (see also comment), but the lstm_cells are not in them (for whatever reason).

Zahlii:
@slundberg The additivity check failing in my new tests is expected - I need help with fixing this, see my comment above.

Zahlii:
I just recognized that as of 2.4.0, more errors appeared in my local SHAP copy:

[2020-12-17T15:43:25.306Z]     def _get_graph(explainer):
[2020-12-17T15:43:25.306Z]         """ Common utility to get the graph for the tensorflow-based explainer.
[2020-12-17T15:43:25.306Z]     
[2020-12-17T15:43:25.306Z]         Parameters
[2020-12-17T15:43:25.306Z]         ----------
[2020-12-17T15:43:25.306Z]         explainer : Explainer
[2020-12-17T15:43:25.306Z]     
[2020-12-17T15:43:25.306Z]             One of the tensorflow-based explainers.
[2020-12-17T15:43:25.306Z]         """
[2020-12-17T15:43:25.306Z]         _import_tf()
[2020-12-17T15:43:25.306Z]         if not tf.executing_eagerly():
[2020-12-17T15:43:25.306Z]             return explainer.session.graph
[2020-12-17T15:43:25.306Z]         else:
[2020-12-17T15:43:25.307Z] >           return explainer.model_output.graph
[2020-12-17T15:43:25.307Z] E           AttributeError: 'KerasTensor' object has no attribute 'graph'

slundberg:
Yeah I went through this and it looks like you essentially turned off the detection of intermediate nodes. I expect this would break some of the back propagation steps because you then don't know which op inputs come from the source (and hence vary) and which don't. The real fix here need to be something that gets access to the graph again, or some equivalent representation so we can determine what inputs for each op depend on the original input. I spent some time yesterday looking at this, but I couldn't find a way to get at the graph. It seems TF just keeps hiding the graph more and more into their C++ internals...

matthiasschuurmans:
Hi @Zahlii and @slundberg , thanks so much for this fantastic package, I've been a huge fan for some years now and consider it an essential part of any data scientist's toolkit. 

I'm using TF 2.4 and ran into the issue reported in https://github.com/slundberg/shap/issues/1694 , found this PR and figured that Scott's comment about not having access to the `graph` object is exactly the problem. So I tried to find the `graph` object in the tensorflow library, and found a reference to it in https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/engine/base_layer.py#L1128 . I figured I'd give it a simple try for my own situation, by modifying `_get_graph()` in https://github.com/slundberg/shap/blob/master/shap/explainers/tf_utils.py :

```
def _get_graph(explainer):
    _import_tf()
    if not tf.executing_eagerly():
        return explainer.session.graph
    else:
        from tensorflow.python.keras import backend
        graph = backend.get_graph()
        return graph
        #return explainer.model_output.graph
```

Ran my code and got some beautiful shap values! I don't know anything about how this works under the hood, so apologies in advance if this would break all sorts of things, but I figured it might just be useful in some way or another.

janesser:
@matthiasschuurmans great fix. i placed here https://github.com/janesser/shap/blob/master/shap/explainers/tf_utils.py onto 0.37.0 tag. Can confirm that calculating shap_values is restored by that.
Still struggle to plot a graph from this, do you have sample of how you do that?

@slundberg any chance to have a 0.37.1 release with this shipped?

matthiasschuurmans:
@janesser Thanks! You mean like a summary or force plot? I just tried a force plot, which didn't immediately crash with
`shap.force_plot(explainer.expected_value.numpy()[0], shap_values, features.to_numpy())` but has been running now for almost an hour already so not really working either.

What is the error you're getting? I noticed `expected_value` of the DeepExplainer comes as an `EagerTensor` object, so I call `numpy()` and get the first (and only) element. Or are you maybe passing a dataframe for the feature matrix? If so, try passing it as arrays?

I made my own plots based on the shap values that work just fine, you might consider the same.

janesser:
@matthiasschuurmans thanks for the prompt response. i need to mention that i am on kaggle and related or not the force-render wouldn't show up, no explicit error or anything in browser console, whilst "sham.initjs()" is present.

i have like a multi dimension story on the sham values where force plot requires selection of a single row.
neither will i become happy forcing mathplotlib.pyplot cause its integration capabilities are limited.

though on the bottom line probably need to extend my fork by a few more enhancements. diving deeper as xmas schedule allows ;-) suggestions are of course welcome.

merry xmas to you folks.

janesser:
bad news first: i am afraid tf2.4 has evolved so that the workaround is no more working: KerasTensor is not holding graph anymore.
good news then: tf2.3.1 works nicely with the workaround provided by @matthiasschuurmans 

see below code where i am stepwise extracting the graph generation to see which part is troubling.
in this case it is a classification model with 11 possible outcomes.

shape of shap_values is (11, (3, 19563))

and it crashes somehow like that, which idk how to surpass:
```
<ipython-input-10-4d4d0d775a64> in <module>
     34 display((e.effects, e.data, e.data.group_names[:5]))
     35 
---> 36 for i in filter(lambda j: e.effects[j] != 0, range(len(e.data.group_names))):
     37     print(i)
     38 

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

post-training code
```
!pip uninstall -y shap
!pip install --upgrade --no-deps git+https://github.com/janesser/shap.git@master

import shap
from shap.utils._legacy import convert_to_link, Instance, Link, Model, DenseData, Data
from shap.plots._force import AdditiveExplanation
        
model = models[0]
feature_names = range(maxClipLength)
features = X_train
out_names = 'f(x)'

exp = shap.DeepExplainer(
    model, 
    X_train[np.random.choice(
        X_train.shape[0], 
        5, 
        replace=False
    )]
)
base_value = exp.expected_value
shap_values = exp.shap_values(X_test)
print(shap_values)
print((len(shap_values), shap_values[0].shape))

instance = Instance(np.zeros((1, len(feature_names))), features)
e = AdditiveExplanation(
    base_value=base_value,
    out_value=np.sum(shap_values[0]) + base_value,
    effects=shap_values[0],
    effects_var=None,
    instance=instance,
    link=convert_to_link('identity'),
    model=Model(None, out_names),
    data=DenseData(np.zeros((1, len(feature_names))), list(feature_names))
)
display((e.effects, e.data, e.data.group_names[:5]))

for i in filter(lambda j: e.effects[j] != 0, range(len(e.data.group_names))):
    print(i)
    
shap.plots._force.visualize(e, matplotlib=True)
```

matthiasschuurmans:
Hi @janesser , in response to your first comment, the force-render not showing up is probably the same problem I have, that it just runs for a very long time. I don't know how to fix that, sorry.

I don't know about your use cases, but the shap values are on individual rows and features, you basically have a value for every single feature and datapoint, so you can make whatever plot you want. Force plots are a nice visualisation of it, but "selecting a single row" is also possible from the shap values themselves. In my case, I'm working on time-based regressions, so I put the shap values back into a dataframe `effects` with the date in the index and each column being a feature with its shap values on that date, and my code to generate a stacked bar plot is 

```
import plotly.graph_objects as go
fig = go.Figure()
for col in effects.columns:
    trace = go.Bar(x=effects.index.get_level_values('date'), y=effects[col], name=col)
    fig.add_trace(trace)
```

In response to your second comment, I don't think I understand. `KerasTensor` didn't hold `graph` in the first place (at least in TF 2.4), that's why it crashed and that's what my workaround is supposed to fix. The `get_graph()` function in `backend` is still there, see https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/backend.py#L762 . Is that giving an error now?

The crash related to `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()` I think may be because `e.effects[j]` is an array, so it cannot check `e.effects[j] != 0` as it would be checking multiple elements. I've never worked with `AdditiveExplanation` directly though, so I'm not sure what `e.effects` holds.

darrahts:
It doesn't look like there has been recent activity on this PR, but the following error directs me here, this is still a work in progress?

```
keras is no longer supported, please use tf.keras instead.
Your TensorFlow version is newer than 2.4.0 and so graph support has been removed in eager mode and some static graphs may not be supported. See PR #1483 for discussion.
```

Here is a MWE that produces this error:

```
import os
import sys
import tensorflow as tf
from tensorflow import keras
from tensorflow.keras import layers
from tensorflow.keras.datasets import mnist
from tensorflow.keras import backend as K
#sys.path.append('..')
#from package import utils

import shap
import numpy as np


#utils.check_gpu()


(x_train, y_train), (x_test, y_test) = mnist.load_data()

BATCH_SIZE = 32
EPOCHS = 2

img_rows, img_cols = 28, 28

x_train = x_train.reshape(x_train.shape[0], img_rows, img_cols, 1).astype("float32") / 255.0
x_test = x_test.reshape(x_test.shape[0], img_rows, img_cols, 1).astype("float32") / 255.0
input_shape = (img_rows, img_cols, 1)

print(x_train.shape)
print(input_shape)


model = keras.Sequential(
    [
        layers.Input(shape=(28, 28, 1)),
        layers.Conv2D(64, 3, padding='same'),
        layers.ReLU(),
        layers.Conv2D(128, 3, padding='same'),
        layers.ReLU(),
        layers.Flatten(),
        layers.Dense(10),
    ],
    name='model',
)

model.compile(
    optimizer=keras.optimizers.Adam(),
    loss = keras.losses.SparseCategoricalCrossentropy(from_logits=True),
    metrics=["accuracy"]
)

model.fit(x_train, y_train, batch_size=BATCH_SIZE, epochs=EPOCHS)
model.evaluate(x_test, y_test, batch_size=BATCH_SIZE)


########## example code from shap readme ###########

# select a set of background examples to take an expectation over
background = x_train[np.random.choice(x_train.shape[0], 100, replace=False)]

# explain predictions of the model on four images
e = shap.DeepExplainer(model, background)
# ...or pass tensors directly
# e = shap.DeepExplainer((model.layers[0].input, model.layers[-1].output), background)
shap_values = e.shap_values(x_test[1:5])

# plot the feature attributions
shap.image_plot(shap_values, -x_test[1:5])
```

Thanks in advance.

Relevant package versions:

```
keras                     2.7.0                    pypi_0    pypi
keras-preprocessing       1.1.2                    pypi_0    pypi
keras-tuner               1.1.0                    pypi_0    pypi
numba                     0.54.1                   pypi_0    pypi
numpy                     1.20.3                   pypi_0    pypi
numpy-base                1.21.2           py38h79a1101_0  
python                    3.8.12               h12debd9_0  
shap                      0.40.0                   pypi_0    pypi
shapely                   1.7.1            py38h1728cc4_0  
tensorflow-gpu            2.7.0                    pypi_0    pypi
```

letalvoj:
Is everyone really using PyTorch? Or how is it that this PR does not have any traction? 😿

Aneeta-u60879:
> ```
> keras is no longer supported, please use tf.keras instead.
> Your TensorFlow version is newer than 2.4.0 and so graph support has been removed in eager mode and some static graphs may not be supported. See PR #1483 for discussion.


@darrahts I am facing this issue when I try to get shap values for my CNN model. Please share a solution if you cracked it. :) Thanks

kchen315:
For anyone looking for a workaround, this solution has worked for me, at least for tabular data: https://stackoverflow.com/questions/66814523/shap-deepexplainer-with-tensorflow-2-4-error

- use tf.keras instead of keras
- add tf.compat.v1.disable_v2_behavior()
- get shap values from numpy array vs dataframe

```
import tensorflow as tf
import shap
tf.compat.v1.disable_v2_behavior()

#load data

input_shape = X.shape[1:]
def build_model(n_hidden=2, n_neurons=50, dropout=0.2, learning_rate=3e-2):
    model = tf.keras.models.Sequential()
    model.add(tf.keras.layers.Flatten(input_shape=input_shape))
    model.add(tf.keras.layers.BatchNormalization())
    for layer in range(n_hidden):
        model.add(tf.keras.layers.Dense(n_neurons))
        model.add(tf.keras.layers.BatchNormalization())
        model.add(tf.keras.layers.Dropout(dropout))
        model.add(tf.keras.layers.Activation("elu"))
    model.add(tf.keras.layers.Dense(1, activation="sigmoid"))

    optimizer = tf.keras.optimizers.Adam(learning_rate=learning_rate)
    
    model.compile(loss=tf.keras.losses.BinaryCrossentropy(), metrics=['AUC'], optimizer=optimizer)
    return model
    
nn = build_model()

history = nn.fit(X_train, y_train,
                 batch_size=512, epochs=1000,
                 validation_data=(X_valid, y_valid),
                 callbacks=[keras.callbacks.EarlyStopping(monitor='val_loss', restore_best_weights=True)]
                 )

explainer = shap.DeepExplainer(nn, X_test.sample(n=1000))
shap_values = explainer.shap_values(X_test.sample(n=1000).values)
shap.summary_plot(shap_values, X_valid.sample(n=1000))
```

PR #1489, opened 2020-10-01
Benchmark Utility Framework (with unit tests) and Sample Notebooks

maggiewu19:
This PR includes the benchmark framework using sequential perturbation. The major changes are in the following files: 

- perturbation.py
- framework.py 
- test_benchmark.py 

The perturbation and framework files are created to replace the old benchmark files (metrics.py, measures.py, methods.py, models.py) such that users are able to benchmark their own explainers and maskers. These are no longer pre-written to allow greater variations in testing. 

The test_benchmark file includes unit tests used to check that the framework file is working as expected. 

Sample notebooks are created under notebooks folder for various user-defined maskers (mean, impute, and resample). These maskers are previously defined in the old benchmark files, which is why we chose them as examples to demonstrate the benchmark utility. Other metrics and maskers can also be applied following the same format as defined in these notebooks.


notebooks/benchmark/Default Mask Benchmark.ipynb
```diff
@@ -0,0 +1,283 @@
...
+   "outputs": [],
+   "source": [
+    "benchmark_masker = X"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "metadata": {},
+   "outputs": [
```
imatiach-msft, line 85: are we keeping outputs in the notebooks?  the thing I don't understand is that some cells have outputs while others do not.  We should be consistent about this.

slundberg:
Thanks! In the Resample Mask Benchmark.ipynb it looks like the permutation explainer is getting a nan value for one of the benchmarks. Do you know why that is? Thanks!

notebooks/benchmark/Default Mask Benchmark.ipynb
```diff
@@ -0,0 +1,283 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": []
+  }
+ ],
+ "metadata": {
+  "kernelspec": {
+   "display_name": "Python 3.7.9 64-bit",
```
imatiach-msft: we should make sure metadata is consistent across all shap notebooks (to keep things clean)

notebooks/benchmark/Default Mask Benchmark.ipynb
```diff
@@ -0,0 +1,283 @@
+{
```
imatiach-msft, line 1: it would be nice to have both documentation in the notebooks and in the folder in general

notebooks/benchmark/Default Mask Benchmark.ipynb
imatiach-msft: nevermind, it looks like the ones without outputs just didn't output anything, but were still run, so this is fine

notebooks/benchmark/Default Mask Benchmark.ipynb
```diff
@@ -0,0 +1,283 @@
...
+     },
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "compare_plot(benchmarks)"
+   ]
+  },
+  {
+   "cell_type": "code",
```
imatiach-msft: remove empty cell at end of notebooks

notebooks/plots/text.ipynb
```diff
@@ -35,6 +35,110 @@
...
+      "Dataset imdb downloaded and prepared to C:\\Users\\maggiewu\\.cache\\huggingface\\datasets\\imdb\\plain_text\\1.0.0\\76cdbd7249ea3548c928bbf304258dab44d09cd3638d9da8d42480d1d1be3743. Subsequent calls will reuse this data.\n"
+     ]
+    },
+    {
+     "ename": "KeyError",
+     "evalue": "0",
+     "output_type": "error",
+     "traceback": [
+      "\u001b[1;31m---------------------------------------------------------------------------\u001b[0m",
+      "\u001b[1;31mKeyError\u001b[0m                                  Traceback (most recent call last)",
```
imatiach-msft: it looks like there is some key error here?

notebooks/plots/text.ipynb
```diff
@@ -84,24 +188,11 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": null,
```
imatiach-msft: some cells are run and some are not, this is confusing

shap/__init__.py
```diff
@@ -68,7 +68,7 @@ def unsupported(*args, **kwargs):
 from . import utils
 from . import links
 
-#from . import benchmark
+from . import benchmark
 
-from .utils._legacy import kmeans
-from .utils import sample, approximate_interactions
+# from .utils._legacy import kmeans
```
imatiach-msft: remove commented out code

shap/benchmark/__init__.py
```diff
@@ -4,13 +4,13 @@
 from . import models
 from .. import datasets
 
-from .metrics import consistency_guarantees
+# from .metrics import consistency_guarantees
```
imatiach-msft: remove commented out code

shap/benchmark/__init__.py
```diff
@@ -4,13 +4,13 @@
...
-from .metrics import remove_negative_retrain
-from .metrics import keep_positive_retrain
-from .metrics import keep_negative_retrain
+# from .metrics import remove_positive_retrain
+# from .metrics import remove_negative_retrain
+# from .metrics import keep_positive_retrain
+# from .metrics import keep_negative_retrain
 from .metrics import remove_positive_mask
 from .metrics import remove_negative_mask
 from .metrics import keep_positive_mask
```
imatiach-msft, line 11: I think these imports from .metrics could be put on the same line?

imatiach-msft: is there some pep style that suggests it should be done this way?

imatiach-msft (commented):
some initial comments

notebooks/benchmark/Default Mask Benchmark.ipynb
maggiewu19: Yep. Only the last two should be plotting so for now I'll keep it this way. Thank you!

notebooks/benchmark/Default Mask Benchmark.ipynb
maggiewu19: I'll add some more comments in the notebooks and the folder to help clarify the steps.

shap/benchmark/__init__.py
maggiewu19: These are for the old API and the new framework would only depend on framework.py and perturbation.py to benchmark the explainers.

notebooks/benchmark/Default Mask Benchmark.ipynb
maggiewu19: Will do. Thank you

notebooks/benchmark/Default Mask Benchmark.ipynb
```diff
@@ -0,0 +1,283 @@
...
+    "metrics = {'sort_order': ['positive', 'negative'], 'perturbation': ['keep']}"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "model_generator = lambda: xgboost.XGBRegressor(n_estimators=100, subsample=0.3)\n",
```
imatiach-msft: instead of lambda just create the model:
untrained_model = xgboost.XGBRegressor(n_estimators=100, subsample=0.3)

notebooks/benchmark/Default Mask Benchmark.ipynb
```diff
@@ -0,0 +1,283 @@
...
+      ]
+     },
+     "metadata": {
+      "needs_background": "light"
+     },
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "trend_plot(benchmarks)"
```
imatiach-msft, line 239: for some reason I don't even see the orange line in the first trend plot, any way to fix it/make it obvious which line it is (since there are 3 lines)

imatiach-msft: I guess since they are so similar maybe it doesn't even matter.

Ah, I think I was worried about this notebook instead:
Impute Mask Benchmark.ipynb
here two lines diverge but the third one doesn't appear, so I don't know which one it is more similar to.  However in general they are pretty similar so it doesn't really matter I guess.

notebooks/benchmark/Default Mask Benchmark.ipynb
```diff
@@ -0,0 +1,283 @@
...
+   ]
```
imatiach-msft, line 240: what are the x and y axes on the plots? hard to tell.
also is there any way to see the execution time, since I would imagine they would differ most on that?

imatiach-msft (commented):
more comments

shap/benchmark/__init__.py
imatiach-msft: will the old API be removed eventually?  if we keep it, maybe we should just clean up some of the existing code?

shap/benchmark/framework.py
```diff
@@ -0,0 +1,110 @@
...
+def update(model, X, y, explainer, masker, sort_order, score_function, perturbation, scores):
+    metric = perturbation + ' ' + sort_order
+    sp = SequentialPerturbation(model, masker, sort_order, score_function, perturbation)
+    x, y, auc = sp.score(explainer, X, y=y)
+    scores['metrics'].append(metric)
+    scores['values'][metric] = [x, y, auc] 
+
+def get_benchmark(model, X, y, explainer, masker, metrics, exp_num=1, *args):
+    # convert dataframes
+    if safe_isinstance(X, "pandas.core.series.Series") or safe_isinstance(X, "pandas.core.frame.DataFrame"):
```
imatiach-msft, line 20: I think these kinds on strings should be moved to one constants file:
"pandas.core.series.Series"
this will make sure:
1.) if we write it incorrectly it will be easier to catch since it will be same everywhere
2.) if we need to update it we only need to update in one place

shap/benchmark/framework.py
```diff
@@ -0,0 +1,110 @@
...
+    # in case the explainer doesn't have a name
+    try: 
+        name = explainer.name 
+    except: 
+        name = 'explainer' + str(exp_num) 
+        exp_num += 1 
+
+    # record scores per metric 
+    scores = {'name': name, 'metrics': list(), 'values': dict()}
+    for sort_order in metrics['sort_order']:
```
imatiach-msft: I think you can use itertools.product here:
https://docs.python.org/3/library/itertools.html
it's
1.) more efficient
2.) reduces nesting

shap/benchmark/framework.py
```diff
@@ -0,0 +1,110 @@
...
+            update(model, X, y, explainer, masker, sort_order, score_function, perturbation, scores)
+
+    return scores 
+
+def get_metrics(benchmarks, selection):
+    # select metrics to plot using selection function
+    explainer_metrics = set()
+    for explainer in benchmarks: 
+        scores = benchmarks[explainer]
+        if len(explainer_metrics) == 0: explainer_metrics = set(scores['metrics'])
```
imatiach-msft: put if/else body on separate line.
Hmm I think this is something flake8 would catch, are we running flake8 in the builds?

shap/benchmark/framework.py
```diff
@@ -0,0 +1,110 @@
...
+    # plot all curves if metric exists 
+    for metric in explainer_metrics:
+        plt.clf()
+
+        for explainer in benchmarks: 
+            scores = benchmarks[explainer]
+            try: 
+                x, y, auc = scores['values'][metric]
+                plt.plot(x, y, label='{} - {}'.format(round(auc, 3), explainer))
+            except:
```
imatiach-msft: any reason for the try/except with a pass?

shap/benchmark/framework.py
```diff
@@ -0,0 +1,110 @@
...
+def compare_plot(benchmarks):
+    explainer_metrics = get_metrics(benchmarks, lambda x, y: x.intersection(y))
+    explainers = list(benchmarks.keys())
+
+    # dummy start to evenly distribute explainers on the left 
+    # can later be replaced by boolean metrics 
+    aucs = dict()
+    for i in range(len(explainers)): 
+        explainer = explainers[i]
+        aucs[explainer] = [i/(len(explainers)-1)]
```
imatiach-msft: nit: move len(explainers) to a variable and reuse on 78 and 76

notebooks/benchmark/Default Mask Benchmark.ipynb
maggiewu19: Will fix that. The old API requires a model generator, which is no longer present with the new framework. I didn't catch that while I was writing the code. Thanks for pointing out!

shap/benchmark/framework.py
```diff
@@ -0,0 +1,110 @@
...
+        for explainer in explainers: 
+            scores = benchmarks[explainer] 
+            _, _, auc = scores['values'][metric]
+            min_auc = min(auc, min_auc)
+            max_auc = max(auc, max_auc)
+        
+        for explainer in explainers: 
+            scores = benchmarks[explainer] 
+            _, _, auc = scores['values'][metric]
+            aucs[explainer].append((auc-min_auc)/(max_auc-min_auc))
```
imatiach-msft, line 103: suggestion:
maybe just store
```
             aucs[explainer].append(auc) 
             min_auc = min(auc, min_auc)
             max_auc = max(auc, max_auc)
```
then in the loop below:
```
    auc_diff = max_auc-min_auc
    for explainer in explainers: 
        plt.plot(np.linspace(0, 1, len(explainer_metrics)+1), (aucs[explainer]-min_auc)/auc_diff, '--o')
```
and remove one of the loops above:
```
       for explainer in explainers: 
            scores = benchmarks[explainer] 
            _, _, auc = scores['values'][metric]
            min_auc = min(auc, min_auc)
            max_auc = max(auc, max_auc)
```

shap/benchmark/methods.py
```diff
@@ -88,6 +90,16 @@ def tree_gain(model, data):
     """
     return other.TreeGainExplainer(model).attributions
 
+def permutation_shap_1(model, data):
+    """ Permutation Explainer
+    """
+    return lambda X: PermutationExplainer(model.predict, data).shap_values(X, npermutations=1)
```
imatiach-msft: hmm why does this need to return a lambda?
Also, why is the name _1 here and _10 below, is that corresponding to the npermutations?  Maybe add a comment in the method.

notebooks/benchmark/Default Mask Benchmark.ipynb
maggiewu19: Will change the kernel to be Python 3 (same as all the rest of the notebooks within shap)

notebooks/plots/text.ipynb
maggiewu19: Not meant to update this notebook so will clear output.

notebooks/plots/text.ipynb
maggiewu19: Not meant to update this notebook so will clear output.

shap/__init__.py
maggiewu19: Will do. Thank you

shap/benchmark/__init__.py
maggiewu19: Will do. Thank you

shap/benchmark/methods.py
```diff
@@ -88,6 +90,16 @@ def tree_gain(model, data):
...
+
+def permutation_shap_10(model, data):
+    """ Permutation Explainer
```
imatiach-msft: I'm wondering if we need to add docs here, eg model and data should ideally be documented.
It does seem like shap explainers are documented very well, for example:
https://github.com/slundberg/shap/blob/master/shap/explainers/_tree.py
I would think these utilities should have similar nice documentation, especially for public methods

shap/benchmark/metrics.py
```diff
@@ -42,22 +42,23 @@ def runtime(X, y, model_generator, method_name):
         model.fit(X_train, y_train)
 
         # evaluate each method
-        start = time.time()
-        explainer = getattr(methods, method_name)(model, X_train)
-        build_time = time.time() - start
+        # start = time.time()
```
imatiach-msft: remove commented code

shap/benchmark/metrics.py
```diff
@@ -73,298 +74,298 @@ def score_function(X_train, X_test, y_train, y_test, attr_function, trained_mode
...
-        "maple": 0.8,
-        "tree_maple": 0.8,
-        "deep_shap": 0.6,
-        "expected_gradients": 0.6
-    }
-
-    return None, guarantees[method_name]
+    return None, __score_method(X, y, None, model_generator, score_function, explainer)
+
+# def consistency_guarantees(X, y, model_generator, method_name):
```
imatiach-msft: remove commented code?  but I thought consistency guarantees was useful, any reason we are removing this now?

shap/benchmark/metrics.py
```diff
@@ -73,298 +74,298 @@ def score_function(X_train, X_test, y_train, y_test, attr_function, trained_mode
...
-    ylabel = "Mean model output"
-    transform = "identity"
-    sort_order = 12
-    """
-    return __run_measure(measures.remove_retrain, X, y, model_generator, method_name, -1, num_fcounts, __mean_pred)
-
-def __run_measure(measure, X, y, model_generator, method_name, attribution_sign, num_fcounts, summary_function):
+    return __run_measure(measures.remove_mask, X, y, model_generator, explainer, 0, num_fcounts, sklearn.metrics.roc_auc_score)
+
+# def keep_positive_retrain(X, y, model_generator, method_name, num_fcounts=11):
```
imatiach-msft: remove commented code?

shap/benchmark/perturbation.py
```diff
@@ -38,16 +39,22 @@ def __init__(self, f, masker, sort_order, score_function, perturbation):
         self.score_values = []
         self.score_aucs = []
         self.labels = []
+
+    def maskedmodel(self):
+        self.masked_model = MaskedModel(self.f, self.masker, shap.links.identity) 
     
-    def score(self, attributions, X, y=None, label=None, silent=False):
+    def score(self, explainer, X, y=None, label=None, silent=False):
+        # if explainer is already the attributions
```
imatiach-msft, line 47: "         # if explainer is already the attributions "
this seems like a naming issue, the explainer should be different from an explanation/attributions.  In which cases is this actually the attributions now, whereas before this was an explainer?

shap/benchmark/perturbation.py
```diff
@@ -93,15 +101,12 @@ def score(self, attributions, X, y=None, label=None, silent=False):
...
-            curve_sign = -1
-        else:
-            curve_sign = 1
-        
-        self.score_aucs.append(np.array([
-            sklearn.metrics.auc(np.linspace(0, 1, len(svals[i])), curve_sign*(svals[i] - svals[i][0]))
-            for i in range(len(svals))
-        ]))
+        if self.sort_order == "negative": curve_sign = -1
+        else: curve_sign = 1
```
imatiach-msft: hmm this seems like it would be a flake8/pep issue, I'm surprised this isn't caught

shap/benchmark/plotting.py
```diff
@@ -0,0 +1,57 @@
+import matplotlib.pyplot as plt 
+import pandas as pd 
+import numpy as np 
+
+root = 'C:/Users/maggiewu/Desktop/'
```
imatiach-msft: hmm seeing these kind of constants is concerning, any way to use relative path or make the code more generic?

shap/explainers/_explainer.py
```diff
@@ -57,7 +57,7 @@ def __init__(self, model, masker, link=links.identity, algorithm="auto", output_
             the Explanation objects produced by this explainer will not have any output_names, which could effect
             downstream plots.
         """
-
+        self.name = ''
```
imatiach-msft: might be good to still give it some name, perhaps "tabular explainer" or "selector explainer" or something along those lines?

shap/explainers/_permutation.py
```diff
@@ -48,7 +49,7 @@ def explain_row(self, *row_args, max_evals, main_effects, error_bounds, batch_si
 
         # by default we run 10 permutations forward and backward
         if max_evals == "auto":
-            max_evals = 10 * 2 * len(fm)
+            max_evals = 10 * 2 * len(fm) + 1
```
imatiach-msft, line 52: what was the reasoning behind the +1 here?

shap/maskers/_tabular.py
```diff
@@ -221,7 +221,7 @@ def __init__(self, data, max_samples=100):
 class Partition(Tabular):
     """ This masks out tabular features by integrating over the given background dataset.
 
-    Unlike Independent, Partition respects a hierarchial structure o
+    Unlike Independent, Partition respects a hierarchial structure
```
imatiach-msft: hierarchial -> hierarchical

tests/benchmark/test_benchmark.py
```diff
@@ -0,0 +1,45 @@
+import numpy as np 
+import shap
+
+model = lambda x: np.array([np.linalg.norm(x)])
+X = np.array([[3, 4], [5, 12], [7, 24]])
+y = np.array([5, 13, 25])
+explainer = np.array([[-1, 2], [-4, 2], [1, 2]])
+masker = X
+
+def test_update():
```
imatiach-msft, line 10: great to see some tests!

imatiach-msft: are these configured to run in the build already?  these aren't actually running the benchmarking from what I'm reading, it's just a few small tests of the benchmark class (update/get_benchmark/get_metrics), so execution time isn't a concern.

imatiach-msft (commented):
added even more comments

shap/benchmark/__init__.py
maggiewu19: It will be eventually. I will for now get rid of the commented code and keep the rest.

shap/benchmark/framework.py
maggiewu19: I will convert this to an if statement instead.

shap/benchmark/plotting.py
maggiewu19: File will be removed. Meant to only be a temporary plotting file. Thank you for catching.

shap/explainers/_explainer.py
maggiewu19: Sounds good.

shap/maskers/_tabular.py
maggiewu19: Thank you.

shap/benchmark/framework.py
maggiewu19: I'm not sure if flake8 is being run but I've changed this.

shap/explainers/_permutation.py
maggiewu19: I believe this is to ensure that even if len(fm) is 0, we would at least evaluate once. However, I'm not absolutely sure about it.

notebooks/benchmark/Default Mask Benchmark.ipynb
maggiewu19: I will add the labels for x and y axes. 
We've left out the runtime metric for the time being because the focus is mostly on the algorithmic metrics. We are considering adding additional performance metrics such as 'runtime' and 'memory' later into an evaluation class, which will also include the sequential perturbation object.

shap/benchmark/framework.py
maggiewu19: Agreed! Thank you for the suggestion!

shap/benchmark/framework.py
maggiewu19: Will do. Thank you.

shap/benchmark/framework.py
maggiewu19: The issue is that auc_diff differs for each metric. Therefore, if I were to plot each explainer after all metrics have been processed, I would have to store the min_auc and auc_diff for each metric. I'm not sure this would make the code more readable so I will keep it as it is for now

shap/benchmark/methods.py
maggiewu19: Will remove this because it is for testing and no longer valid.

shap/benchmark/methods.py
maggiewu19: Will remove this because it is for testing and no longer valid.

shap/benchmark/metrics.py
maggiewu19: These were originally commented out to test the old API, but the old API can remain for now so will be uncommented

shap/benchmark/metrics.py
maggiewu19: This was originally commented out as we are temporarily leaving it out for the new API, but old files will generally be removed later so it will for now be uncommented

shap/benchmark/perturbation.py
maggiewu19: This is for the case where users input explainer.values instead of explainer into the input. It is also valid for testing. However, we anticipate users to typically input the original explainer so it is unlikely to happen.

shap/benchmark/metrics.py
maggiewu19: This was originally commented out because we would not be implementing them for the new API. However, all old APIs will be cleaned up together later once the new API can completely replace the old one.

shap/benchmark/perturbation.py
maggiewu19: I'm not entirely sure if flake8/pep is called, but I will change this

maggiewu19:
> Thanks! In the Resample Mask Benchmark.ipynb it looks like the permutation explainer is getting a nan value for one of the benchmarks. Do you know why that is? Thanks!

This comes from setting the curr_val to be None. When attribution is all negative (for positive sort order) or all positive (for negative sort order), curr_val is never updated. This causes the score function to append None and leads to NaN during the computation. It is updated by setting it to the default fully masked output.

shap/benchmark/framework.py
maggiewu19: Will work on this in another PR.

tests/benchmark/test_benchmark.py
maggiewu19: Tests are now included in the build pipeline

shap/benchmark/perturbation.py
```diff
@@ -93,15 +103,14 @@ def score(self, attributions, X, y=None, label=None, silent=False):
             
         self.score_values.append(np.array(svals))
         
-        if self.sort_order == "negative":
+        if self.sort_order == "negative":
```
imatiach-msft, line 106: I think strings like "negative" should be moved to a constant

shap/explainers/_additive.py
```diff
@@ -31,6 +31,7 @@ def __init__(self, model, masker):
             note that this structure information has no effect on the explanations of additive models.
         """
         super(Additive, self).__init__(model, masker)
+        self.name = 'additive explainer'
```
imatiach-msft, line 34: can we add name to the base explainer and pass this through the init() method?

imatiach-msft: same thing for other explainers below

imatiach-msft (approved):
added some minor comments

shap/benchmark/perturbation.py
maggiewu19: Yep, moving things into string file is another ongoing process and will be taken care of in the next PR :)

shap/explainers/_additive.py
maggiewu19: Sure. Will be taken care of in the next PR as well.

imatiach-msft (approved):
It would be good to fix the ordering of the graphs in the notebooks and add some readme explaining the benchmarking notebooks more for anyone who looks at them (or info in the notebooks).  Also the comparison plot is very confusing, but maybe that's fine for now.

shap/benchmark/framework.py
lrjball: Still seems like this can be made simpler, as the lines 
```python
        for explainer in explainers: 
            scores = benchmarks[explainer] 
            _, _, auc = scores['values'][metric]
```
are exactly repeated. Something like:
```python
    for metric in explainer_metrics:
        auc_list = [benchmarks[explainer]['values'][metric][2] for explainer in explainers]
        min_auc = min(auc_list)
        max_auc = max(auc_list)
        auc_diff = max_auc - min_auc
        for auc, explainer in zip(auc_list, explainers): 
            aucs[explainer].append((auc-min_auc)/auc_diff)
```
should be faster and reduces the duplication

shap/benchmark/framework.py
```diff
@@ -0,0 +1,123 @@
...
+    
+    # plot common curves
+    ax = plt.gca()
+    for explainer in explainers: 
+        plt.plot(np.linspace(0, 1, len(explainer_metrics)+1), aucs[explainer], '--o')
```
lrjball, line 108: Maybe this could be
```python
x = np.linspace(0, 1, num_metrics + 1)
for explainer in explainers:
    plt.plot(x, aucs[explainer], '--o')
```
to save rebuilding `x` every time. And the `x` can also be reused later on for the xticks

maggiewu19: Sounds good. Thanks for the comment!

slundberg:
Happy to merge this soon. But why is notebooks/plots/text.ipynb changed? It seems unrelated to this PR. Thanks!

maggiewu19:
The changes to text.ipynb are indeed unrelated. I believe it came from running the cells but nothing major has been changed in the file.

slundberg:
Sounds good, if you could just revert the text.ipynb changes then we can merge :)

maggiewu19:
Reverted the text.ipynb so should be ready to merge once checks are completed!

PR #1493, opened 2020-10-03
Testing upgrades

lrjball:
Thought it would be useful to use pytest as the test suite, as it comes with a load of great features which I have also added:
- image testing via pytest-mpl: This is really useful for the plotting functionality in shap. Essentially a baseline image is generated the first time which can be checked for accuracy, then every subsequent run of the test is checked against that baseline image.
- importorskip: allows tests to be easily skipped when import is missing, and the test will show as such in the report
- tmpdir: temp directories are auto created for storing temporary data for tests and neatly cleaned up at the end

This PR also adds test coverage reports, integrated with codecov which can be viewed for example on my fork here: https://codecov.io/gh/lrjball/shap/tree/b05fae9372d67bffda146e04a305c9b55d75b56b/shap

Also, I have fixed a couple of things in the tests such as filenames which weren't being discovered, and random seeds not being set causing certain tests to occasionally fail randomly (specifically the pytorch ones).

This PR also adds some extra tests of various shap plots using the pytest mpl plugin and improves overall test coverage.


slundberg:
This is great. Thanks! Sorry it took a while to review.

PR #1536, opened 2020-10-19
Local text-to-text Visualization

vivekchettiar:
Changes:
1) Added Saliency Plot for text-to-text explaination
2) Added Heatmap for text-to-text explaination
3) text-to-text explaination can be passed to shap.plots.text for viz


slundberg:
Thanks! I am sure we will iterate more on this, but it is looking great!

PR #1554, opened 2020-10-27
Adding fixed_context as a parameter

anusham1990:


shap/explainers/_explainer.py
```diff
@@ -6,7 +6,7 @@
 
 
 class Explainer():
-    def __init__(self, model, masker=None, link=links.identity, algorithm="auto", output_names=None):
+    def __init__(self, model, masker=None, link=links.identity, algorithm="auto", output_names=None, fixed_context = 1):
```
imatiach-msft: it's weird that fixed_context isn't an enum, but not related to this PR.  Any reason it can't be an enum @slundberg ?

shap/explainers/_partition.py
```diff
@@ -142,8 +142,8 @@ def explain_row(self, *row_args, max_evals, main_effects, error_bounds, batch_si
         self.values = np.zeros(out_shape)
         self.dvalues = np.zeros(out_shape)
 
-        fixed_context = 1
-        self.owen(fm, self._curr_base_value, f11, max_evals // 2 - 2, outputs, fixed_context, batch_size, silent)
+        # fixed_context = 1
```
imatiach-msft: remove commented out code

imatiach-msft (approved):

shap/explainers/_explainer.py
slundberg: We don't need fixed_context until explain_row right? So it would probably better be a parameter of the `__call__` method just for the Partition class.

anusham1990: I have updated the files and sent a new PR. Request you to please take a look.

PR #1563, opened 2020-11-03
Text to text viz

michaelamoako:


slundberg (approved):
Looks good. We will want to merge text_to_text into just text() once we can get all the values we want into the Explanation object.

slundberg:
Can you take a look at the merge conflict?

vivekchettiar:
Looks like this PR has some unrelated commits from the text-to-viz work which isn't meant to be in this PR (Raised a new one for that).

PR #1564, opened 2020-11-03
Pylint tree tests

RAMitchell:
- Remove tests path from pytest.ini so we can run individual tests instead of running all tests every time 

- Reduce the duration of a few longer running tree tests

- Add .pylintrc and lint tests/explainers/test_tree.py

- Tests that fail optional imports now get marked as skipped instead of succeeded

- If pyspark cluster cannot be instantiated, mark as skipped instead of succeeded

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,9 @@ install:
 - python setup.py develop
 - pip install .[test]
 - pip install tensorflow xgboost matplotlib ipython codecov torch torchvision
-script: python -m pytest .
+script:
+  - python -m pytest tests
+  - pylint tests/explainers/test_tree.py
 after_success:
   - codecov
 deploy:
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_install:
 install:
 - python setup.py develop
 - pip install .[test]
-- pip install tensorflow xgboost matplotlib ipython codecov torch torchvision
+- pip install tensorflow xgboost matplotlib ipython codecov torch torchvision pylint
 script:
   - python -m pytest tests
   - pylint tests/explainers/test_tree.py
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,2 +1,2 @@
 [pytest]
-addopts = --mpl --cov=shap tests --cov-report=term-missing --durations=20
+addopts = --mpl --cov=shap --cov-report=term-missing --durations=20
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,2 +1,2 @@
 [pytest]
-addopts = --mpl --cov=shap tests --cov-report=term-missing
+addopts = --mpl --cov=shap tests --cov-report=term-missing --durations=20
--- a/tests/explainers/test_tree.py
+++ b/tests/explainers/test_tree.py
@@ -1,15 +1,19 @@
-import matplotlib
-matplotlib.use('Agg')
+# pylint: disable=missing-function-docstring,too-many-lines,fixme
+"""Test tree functions."""
+import itertools
+import math
+import pickle
+import numpy as np
+import pandas as pd
+import pytest
+import sklearn
+import sklearn.pipeline
+from sklearn.experimental import enable_hist_gradient_boosting  # pylint: disable=unused-import
+import shap
 
 
 def test_front_page_xgboost():
-    try:
-        import xgboost
-    except:
-        print("Skipping test_front_page_xgboost!")
-        return
-    import shap
-    import numpy as np
+    xgboost = pytest.importorskip('xgboost')
 
     # load JS visualization code to notebook
     shap.initjs()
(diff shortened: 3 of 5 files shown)


tests/explainers/test_tree.py
```diff
@@ -452,39 +402,29 @@ def test_catboost():
...
-    y_val = train_df.ACTION[ix:ix+20]
-    model = catboost.CatBoostClassifier(iterations=100, learning_rate=0.5, random_seed=12)
+        "SHAP values don't sum to modThisel output!"
+
+    X, y = sklearn.datasets.load_breast_cancer(return_X_y=True)
+    model = catboost.CatBoostClassifier(iterations=10, learning_rate=0.5, random_seed=12)
     model.fit(
-        X_train,
-        y_train,
-        eval_set=(X_val, y_val),
```
lrjball, line 467: Any reason for eval_set being removed here?

tests/explainers/test_tree.py
```diff
@@ -610,37 +530,29 @@ def test_lightgbm_binary():
...
+    X_train, _, Y_train, _ = sklearn.model_selection.train_test_split(*shap.datasets.iris(),
+                                                                      test_size=0.2, random_state=0)
+    rforest = sklearn.ensemble.RandomForestClassifier(n_estimators=100, max_depth=None,
+                                                      min_samples_split=2,
+                                                      random_state=0)
     model = rforest.fit(X_train, Y_train)
 
     # verify symmetry of the interaction values (this typically breaks if anything is wrong)
     interaction_vals = shap.TreeExplainer(model).shap_interaction_values(X)
-    for i in range(len(interaction_vals)):
```
lrjball, line 627: Is this change necessary? Feels like range(len) is already fine here. Same for below

pytest.ini
```diff
@@ -1,2 +1,2 @@
 [pytest]
-addopts = --mpl --cov=shap tests --cov-report=term-missing
\ No newline at end of file
+addopts = --mpl --cov=shap --cov-report=term-missing --durations=20
```
lrjball, line 2: Is this causing timeout issues in the CI? Don't think we want to skip tests by default in the pytest.ini Maybe that setting could be applied by a user if the tests are taking too long, or in the offending CI config file if the CI is failing

.travis.yml
```diff
@@ -18,7 +18,9 @@ install:
 - python setup.py develop
 - pip install .[test]
 - pip install tensorflow xgboost matplotlib ipython codecov torch torchvision
-script: python -m pytest .
+script:
+  - python -m pytest tests
+  - pylint tests/explainers/test_tree.py
```
lrjball, line 23: Has this been left in accidentally? I guess we would either lint the whole project as part of the CI or none of it, rather than just this one file

lrjball (commented):
Couple of small points raised, but other than that LGTM

tests/explainers/test_tree.py
RAMitchell: It has no purpose and takes time to evaluate.

tests/explainers/test_tree.py
RAMitchell: This one is a pylint failure. I agree that there is not much difference, but IMO it's better to conform.

pytest.ini
RAMitchell: This only prints the durations of the 20 longest tests.

.travis.yml
RAMitchell: I spent about 3 hours conforming this one file, because I will be working on it. It's just not possible to do the entire code base in one go, It's too much work and the diff is too large.

RAMitchell:
@slundberg

.travis.yml
lrjball: I wasn't suggesting you reformat the entire code base, don't worry! One thing I would say though is that there may be better ways to contribute than to reformat all of the test files. Ultimately formatting is a choice and can often lead to a lot of debate, but doesn't add too much value to the code base one way or another. 

I think adding test skips is definitely worthwhile, and moving imports too, but other than that it is just another potential source of a bug creeping into code which already works.

pytest.ini
lrjball: Ah my bad, good shout!

tests/explainers/test_tree.py
lrjball: pylint is only really a list of guidelines, and doesn't necessarily have to be followed religiously. Remember that every change takes time both to make, and to have someone else review, so if there is no difference either way my opinion would always be to leave it how it is.

tests/explainers/test_tree.py
lrjball: I guess including it does some basic testing of eval_set to make sure that nothing errors when it is set, but this could definitely be improved by adding some further testing to make sure eval_set behaves as intended.

slundberg (approved):
Thanks @RAMitchell ! This looks like a good start to separating out the code to support GPU testing. I had orgininally left the imports inside each function to make it easy to copy and paste them into a notebook for debugging, but I think moving them to the top is a good choice.

I expect there may be some linting changes for the future, but happy have this file conform as a start.

Thanks for taking a look @lrjball !

PR #1571, opened 2020-11-09
TreeShap using GPUs

RAMitchell:
This is a rough work in progress PR, compiling a c extension with GPUTreeShap. I have subclassed TreeExplainer as GPUTreeExplainer, overriding `shap_values` to provide a GPU accelerated version.


slundberg:
Looking good! I wonder if it would make sense to just add a `compute="gpu"` option to the Tree class? Then we could write a wrapper around the cext that we use directly and that could have a setting for the compute type the user wants. This would eliminate repeated code in the GPUTree class. Thoughts?

RAMitchell:
It's a possibility, I'm just afraid some things will diverge e.g. there will be some special algorithms implemented for GPU only, or some parameters needed only by the GPU version.

slundberg:
Okay, I am fine leaving it as GPUTree. In that case we should probably call the _cext through a property on the Tree object, then the GPUTree object only needs to set that property to _gpu_cext and everything should be set (no need to override the shap_values function)...

RAMitchell:
@slundberg one problem with that approach is that the default tree explainer often calls down into the underlying model (beyond our control), so we get a situation where the user calls GPUTreeExplainer.shap_values() and it computes on the CPU.

Instead I am refactoring the boilerplate code into helper functions of TreeExplainer, so there is virtually no code in GPUTreeExplainer except calling into _cext_gpu.

RAMitchell:
This PR supports the following, the rest can wait. 
feature_perturbation="tree_path_dependent"
Xgboost - regression, binary classification
LightGBM - regression, binary classification
LightGBM - regression, binary classification
sklearn rf - regression

I think it's more or less ready for review.

setup.py
```diff
@@ -149,7 +184,14 @@ def try_run_setup(**kwargs):
             try_run_setup(**kwargs)
         elif kwargs["with_binary"]:
             kwargs["with_binary"] = False
-            print("WARNING: The C extension could not be compiled, sklearn tree models not supported.")
+            print(
```
slundberg, line 209: this is fine. But just FYI I increased the pylint line limit just to avoid wraps like this.

shap/__init__.py
```diff
@@ -16,6 +16,7 @@
 from .explainers._kernel import Kernel as KernelExplainer
 from .explainers._sampling import Sampling as SamplingExplainer
 from .explainers._tree import Tree as TreeExplainer
+from .explainers._gpu_tree import GPUTree as GPUTreeExplainer
```
slundberg, line 19: I think this is the right choice, but my goal is to point users at explainers.GPUTree eventually.

shap/cext/_cext_gpu.cc
```diff
@@ -0,0 +1,188 @@
+#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
```
slundberg, line 1: Did you run valgrind memory leak checks (or something similar) on the GPU version?

shap/cext/gpu_treeshap.h
```diff
@@ -0,0 +1,1231 @@
+/*
```
slundberg, line 1: I didn't read through this whole thing, so I will trust your CUDA code based on the unit tests.

shap/cext/tree_shap.h
```diff
@@ -301,7 +305,7 @@ inline void tree_saabas(tfloat *out, const TreeEnsemble &tree, const Explanation
 /**
  * This runs Tree SHAP with a per tree path conditional dependence assumption.
  */
-void dense_tree_saabas(tfloat *out_contribs, const TreeEnsemble& trees, const ExplanationDataset &data) {
+inline void dense_tree_saabas(tfloat *out_contribs, const TreeEnsemble& trees, const ExplanationDataset &data) {
```
slundberg, line 308: did you notice a perf change with the inline? just curious.

shap/cext/_cext_gpu.cc
RAMitchell: The GPUTreeShap repo is heavily tested, I haven't run anything specific to its integration here. I can look into it.

shap/cext/tree_shap.h
RAMitchell: It's not for performance reasons, rather when you link two objects that both include this file the function definitions clash. Using inline or templates means this header function gets a unique name in each object.

RAMitchell:
Still to do on this:

- Get it working on linux (works on windows)

- Try to make setup.py code calling nvcc more portable and less fragile

dantegd (commented):
I would suggest using the `CUDA_PATH` environment variable for looking for nvcc and specifying some common architectures (Pascal+? not sure ig gputreeshap requires any particular compute capability) during compilation.

RAMitchell:
Review comments have been addressed, I think this is ready to go.

setup.py
```diff
@@ -1,163 +1,223 @@
...
+
+
+def compile_cuda_module(host_args):
+    libname = '_cext_gpu.lib' if sys.platform == 'win32' else 'lib_cext_gpu.a'
+    lib_out = 'build/' + libname
+    nvcc = os.path.join(os.path.abspath(os.environ['CUDA_PATH']), 'bin/nvcc')
+    arch_flags = "-arch=sm_60 " + \
+                 "-gencode=arch=compute_70,code=sm_70 " + \
+                 "-gencode=arch=compute_75,code=sm_75 " + \
+                 "-gencode=arch=compute_75,code=compute_75"
```
dantegd: I would add 80 if possible (Ampere), that would cover the bulk of GPU architectures probably

dantegd (commented):
Had a super minor comment/question

setup.py
RAMitchell: I was trying to avoid discovering the cuda toolkit version for now, as I would have to do this manually and 80 presumably won't work on cuda 10.1.

dantegd: ah totally true, I blanked for a sec about that requirement, then I think this is the best choice.

RAMitchell:
Fixed a couple of bugs relating to stacked models. Xgboost and LightGBM were having their leaf values put an incorrect place in the C++ representation when models were stacked. Also the LightGBM model loader was not detecting stacked multiclass models.

Now GPU is tested as working for XGBoost/LightGBM/RF, regression/binary/multiclass, shap values and interaction values.

The major missing feature is feature_perturbation='interventional', which I still need to implement in cuda, but I should save that for another PR.

slundberg:
Thanks again @RAMitchell ! This looks good to me, and I agree the interventional option will be great, but best left for another PR.

PR #1572, opened 2020-11-09
Add TreeExplainer support for PyOD IForest

munhouiani:
PyOD is a Python toolkit for outlier detection. Internally, a PyOD IForest uses Scikit-learn IsolationForest implementation. 

Shap's TreeExplainer already supports Scikit-learn IsolationForest, I slightly changed the existing code to support PyOD IForest.


slundberg (approved):
Looks great. Thanks!

PR #1591, opened 2020-11-16
Image-To-Text Visualization Initial Implementation

vivekchettiar:
This is initial implementation of the Image-To-Text Visualization, some details:
1) Can be called by: shap.plots.image_to_text(shap_values) ( This is separated from the shap.plots.image() API for now, to differentiate multioutput with text output - this can be changed later).
2) Allows for Zooming and Setting Opacity of the SHAP Overlay.


tests/plots/test_image.py
```diff
@@ -14,3 +14,26 @@ def test_random_multi_image():
     """
 
     shap.image_plot([np.random.randn(3, 20,20) for i in range(3)], np.random.randn(3, 20,20), show=False)
+
+def test_image_to_text_single():
+    """ Just make sure the image_to_text function doesn't crash.
+    """
+    
+    class MockImageExplaination:
```
imatiach-msft: nit: MockImageExplaination -> MockImageExplanation

shap/plots/_image.py
```diff
@@ -127,3 +132,381 @@ def image(shap_values, pixel_values=None, labels=None, width=20, aspect=0.2, hsp
     cb.outline.set_visible(False)
     if show:
         pl.show()
+
+
+def image_to_text(shap_values):
+
+    from IPython.core.display import display, HTML
```
imatiach-msft: we should put all imports at top, I think that is more standard practice - if this is not required we can have a try/catch around it

shap/plots/_image.py
```diff
@@ -127,3 +132,381 @@ def image(shap_values, pixel_values=None, labels=None, width=20, aspect=0.2, hsp
...
+
+    if len(shap_values.values.shape) == 5:
+        for i in range(shap_values.values.shape[0]):
+            display(HTML("<br/><b>"+ordinal_str(i)+" instance:</b><br/>"))
```
imatiach-msft: nit: I think format would like nicer here, eg:
https://www.w3schools.com/python/ref_string_format.asp

shap/plots/_image.py
```diff
@@ -127,3 +132,381 @@ def image(shap_values, pixel_values=None, labels=None, width=20, aspect=0.2, hsp
...
+    
+    # creating input html tokens
+    
+
+    model_output = shap_values.output_names
+
+    output_text_html = ''
+
+    for i in range(model_output.shape[0]):
+        output_text_html += "<div style='display:inline; text-align:center;'>" \
```
imatiach-msft, line 157: nit: hmm I don't think you need the \ at end of line if you are using + operator (?)

shap/plots/_image.py
```diff
@@ -127,3 +132,381 @@ def image(shap_values, pixel_values=None, labels=None, width=20, aspect=0.2, hsp
...
+    
+    
+    image_data = shap_values.data
+    image_height = image_data.shape[0]
+    image_width = image_data.shape[1]
+    
+    
+    # computing gray scale image
+    
+    image_data_gray_scale = np.ones((image_height, image_width, 4)) * 255 * 0.5
```
imatiach-msft, line 181: I see "4" used here when allocating np.ones, but only 0, 1, 2 are specified below (not 3rd axis), is 3rd axis supposed to be all 1's?

also, these two lines are the same, is that meant to be the case:
    image_data_gray_scale[:,:,1] = image_data_gray_scale[:,:,0]
    image_data_gray_scale[:,:,2] = image_data_gray_scale[:,:,0]

shap/plots/_image.py
```diff
@@ -127,3 +132,381 @@ def image(shap_values, pixel_values=None, labels=None, width=20, aspect=0.2, hsp
...
+        shap_values_color_dict[f'{uuid}_output_flat_token_{index}'] = (colors.red_transparent_blue(0.5 + 0.5 * shap_values_color_maps[:,:,index]/ max_val) * 255).astype(int).tolist()
+
+    
+    # converting to json to be read in javascript
+    
+    image_data_json = json.dumps(shap_values.data.astype(int).tolist())
+    shap_values_color_dict_json = json.dumps(shap_values_color_dict)
+    image_data_gray_scale_json = json.dumps(image_data_gray_scale.astype(int).tolist())
+    
+    image_viz_html = f"""
```
imatiach-msft, line 203: nit: I'm wondering if the html code should be moved to a separate file?  it's quite a bit of code.

shap/plots/_image.py
```diff
@@ -127,3 +132,381 @@ def image(shap_values, pixel_values=None, labels=None, width=20, aspect=0.2, hsp
...
+
+        <div id="{uuid}_image_viz" class="{uuid}_image_viz_content">
```
imatiach-msft, line 205: why do you use {uuid} on all of the divs?  wondering if that could be refactored out somehow, perhaps moved to a function that appends the {uuid} string to the given parameter string

shap/plots/_image.py
```diff
@@ -127,3 +132,381 @@ def image(shap_values, pixel_values=None, labels=None, width=20, aspect=0.2, hsp
...
+            </div>
+          </div>
+        </div>
+
+    """
+    
+    
+    
+    image_viz_script = f"""
+        <script>
```
imatiach-msft, line 251: I think there might be a better way to do this than embed the html/script code as strings, not sure how much effort we want to put into it though.  main problem is this is harder to read/understand as a string and there is no syntax highlighting.

imatiach-msft (approved):

shap/plots/_image.py
slundberg: The uuid is so there can be multiple plots in the same Jupiter notebook. Not sure how to make a cleaner prefix here, this might be best.

shap/plots/_image.py
slundberg: not opposed to this eventually, though for now I think it is fine.

vivekchettiar: I agree too, having it in another file is clean. I followed the convention in the existing viz. We'd need to pass in the raw shap_values data into JS and preprocess there to create the correct divs (that we currently, generate in python)

shap/plots/_image.py
vivekchettiar: Yes the 3rd axis represent alpha, its multiplied by 0.5 to make it look translucent, to allow for the shap overlay to be visible. 

Ans yes the two lines are meant to be the same, its just calculating average value of RGB and set it to all 2 channels (to make it grayscale).

shap/plots/_image.py
vivekchettiar: I think its required, otherwise the next line gets an unexpected indentation error

shap/plots/_image.py
vivekchettiar: done

tests/plots/test_image.py
vivekchettiar: Corrected the typo

PR #1618, opened 2020-12-02
Text to text explanation based on new api design

ryserrao:
This PR contains following work:
3 explanation notebooks (Text prediction, Machine Translation and Abstractive Summarization)
FixedCompositeMasker
Auto wrapping of TeacherForcingLogits and FixedCompositeMasker class
Text Infilling using collapse mask token
Support for sentencepiece tokenizer
Utility functions for transformers
Support for dynamic output names
Remove unwanted special tokens in text plots
Check for wrapped inputs and unpacking multiple args into model call function
Fix for self.background check in partition explainer while computing new base values for each row
Models (TeacherForcingLogits and TextGeneration)
Tests


slundberg:
This looks good to me. There are small things to tweak, but given the scope of this PR already, I think it is best to merge this now and then iterate on future updates. Thanks again, this is great!! :)

ryserrao:
sounds good, thanks!

PR #1620, opened 2020-12-03
Several bug fixes to call SHAP from EconML

heimengqi:
- the `explain_row()` method under linear explainer should return the values of shape (dx,dy) in multiple outcome, it will be in line with other explainers

- add extra argument to allow feature names being passed through

- fix variable name typo in beeswarm plot


slundberg:
Thanks!! This looks good to me.

PR #1626, opened 2020-12-07
When plotting categorical interactions, ensure we don't request more bounds than available color buckets

jeffreyftang:
Due to automatic detection of categorical interactions, certain datasets (specifically, those with extreme outlier values) can cause us to generate a bounds array with more elements than the number of color buckets available to us. This leads to an error from matplotlib saying "ncolors must equal or exceed the number of bins".

This PR caps the number of bounds generated to be at most the number of bins available in the colormap.


slundberg:
looks good. Thanks!

PR #1629, opened 2020-12-07
Specify numpy version requirements to keep python3.6 compatibility

mathPi:
This pr solves the issue: https://github.com/slundberg/shap/issues/1628.
Because of the future 1.20 numpy version (https://numpy.org/devdocs/release/1.20.0-notes.html) shap won't allow to be installed on python3.6.

In this PR, I propose to set a requirements to numpy < 1.20.0 in order to let time to move from python 3.6 to higher versions.


slundberg:
I am hesitant to do this because then people can't use SHAP with numpy 1.20

PR #1634, opened 2020-12-08
Interventional GPU algorithm

### LINKED ISSUE #1617: Inconsistency of threshold condition in TreeShap
TreeShap uses <= to evaluate thresholds, this is correct for models imported from sklearn.RandomForest but not xgboost. The prediction result is then incorrect when the feature value is the same as the split condition.

Reproducing script:

```python
import numpy as np
import xgboost as xgb
import shap

n = 2
m = 1
rng = np.random.RandomState(6)
X = rng.random_sample((n, m))
y = rng.random_sample(n)

dmat = xgb.DMatrix(X, y)
bst = xgb.train({"tree_method":"hist","lambda":0}, dmat, 1)

for t in bst.get_dump():
    print(t)

print(X)
ex = shap.TreeExplainer(bst, X)
xgb_pred = bst.predict(dmat)
print(xgb_pred)
shap_pred = ex.model.predict(X)
print(shap_pred)
```
Output:
```
0:[f0<0.892860174] yes=1,no=2,missing=2
	1:leaf=-0.137491018
	2:leaf=0.0963687375

[[0.89286015]
 [0.33197981]]
[0.59636873 0.36250898]
[0.36250898 0.36250898]
```

RAMitchell:
Complete the feature set of the GPUTreeExplainer class by adding interventional algorithms.

Also makes a slight change to the way xgboost models are imported to account for the difference in split condition (< vs <=), fixes #1617.

Adds GPUTreeExplainer to docs.


shap/explainers/_tree.py
```diff
@@ -1517,7 +1517,11 @@ def get_trees(self, data=None, data_missing=None):
                     self.children_default[i,j] = self.node_cright[i][j]
                 self.features[i,j] = self.node_sindex[i][j] & ((np.uint32(1) << np.uint32(31)) - np.uint32(1))
                 if self.node_cleft[i][j] >= 0:
-                    self.thresholds[i,j] = self.node_info[i][j]
+                    # Xgboost uses < for thresholds where shap uses <=
+                    # Move the threshold down by the smallest possible increment
+                    self.thresholds[i, j] = np.nextafter(self.node_info[i][j],
```
slundberg: Perhaps the real problem is that at https://github.com/slundberg/shap/blob/f88d4f28bc7a7a9e863d26bbf5362b1b83255c60/shap/explainers/_tree.py#L1023-L1024

`np.inf` should be `-np.inf` ?

slundberg:
Thanks! This looks good! The one issue is with the <= vs < and if it has been fixed in the right spot. I have tried to address that issue multiple times, but apparently it is still there or the last fix never got merged or something.

shap/explainers/_tree.py
RAMitchell: Yes, that should do it, will update that now.

RAMitchell: Adjusting the threshold at line 1024 creates a bug because this adjustment happens after node statistics get calculated (in Tree constructor), and it needs to happen before. I suggest removing the `less_than_or_equal` parameter (which  is currently only used for xgb) and adjusting thresholds as necessary in each constructor.

Also, for xgb the nudge has to happen in float32, nudging the threshold in float64 seems insufficient.

slundberg: Ah! I bet that is something I missed before. Good catch. I agree that sounds like a good approach.

slundberg:
The only fails here are for plot diffs, so I am going to merge and then address those plot fails if they pop up again on master.

RAMitchell:
Thanks! I will take a look at the plot issue today.

Some things I'd like to do next:
- Set up some form of gpu CI
- Add some info to the front page readme
- See if this can be bundled in the next pip release

PR #1638, opened 2020-12-09
Multi-sample text input with global bar visualization 

maggiewu19:
- Added "shap.models" to packages within setup.py so the folder is part of the build process 
- Modified _explainer.py and _explanation.py to allow multi-sample text input


maggiewu19:
Created a separate PR for this change.

PR #1639, opened 2020-12-09
Multi-sample text input with global bar visualization tested

maggiewu19:
Added "shap.models" to packages within setup.py so the folder is part of the build process
Modified _explainer.py and _explanation.py to allow multi-sample text input


slundberg:
This looks good! I am going to merge it and then play around with it a bunch and let you know if I find any corner cases

maggiewu19:
> This looks good! I am going to merge it and then play around with it a bunch and let you know if I find any corner cases

Thank you!

PR #1643, opened 2020-12-11
Fix GPUTreeExplainer ubuntu build

RAMitchell:


JohnZed (approved):
Looks good. I think we'll still need to generalize the search for cuda build tools a bit more in the future. The std::size change looks correct to me.

RAMitchell:
Superseded by #1646

PR #1645, opened 2020-12-11
Tabular Benchmark Example

maggiewu19:
- Modified benchmark init file to accommodate new benchmark API 
- Created framework file to reflect new perturbation file 
- Created perturbation file to accommodate various input data (will further modify with new text API)


slundberg:
Thanks @maggiewu19 ! Could you look into the reason for the test failures? Thanks.

maggiewu19:
Updates will be in another PR instead!

PR #1646, opened 2020-12-11
Generalize CUDA compilation in setup.py

JohnZed:
Two related changes:
 * Replaces usage of std::size, which requires C++17 support not available everywhere
 * Finding CUDA can be a little wonky depending on how it was installed. This approach will search CUDAHOME in addition to CUDA_PATH and will fall back to /usr/local/cuda if it can't find nvcc. This helps building within conda env and is generally inspired by what @benfred's `implicit` does.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,21 +56,62 @@ class build_ext(_build_ext):
         print("numpy.get_include()", numpy.get_include())
         self.include_dirs.append(numpy.get_include())
 
+def find_in_path(name, path):
+    "Find a file in a search path"
+    # adapted from https://github.com/benfred/implicit/blob/master/cuda_setup.py, which is
+    # in turn adapted from:
+    # http://code.activestate.com/recipes/52224-find-a-file-given-a-search-path/
+    for dir in path.split(os.pathsep):
+        binpath = os.path.join(dir, name)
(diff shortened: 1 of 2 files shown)


JohnZed:
CC @RAMitchell

RAMitchell (approved):

mirekphd:
This is spot on, exactly what was needed!:) I will give it a try tomorrow. 

Please don't miss the other issue I reported [there](https://github.com/slundberg/shap/issues/1650#issuecomment-744031612) (the use of `--extended-lambda` which seems to be unavailable for older CUDA versions (like 10.1 which we have in production servers), but a simple replacement (conditional on CUDA version check?) with its alias `--expt-extended-lambda` seems to work), In `nvidia/cuda` CUDA 11.x containers there was no such problem, so it can be easily missed if running latest CUDA versions.

RAMitchell:
@mirekphd I have changed to `--expt-extended-lambda`. Really appreciate your early feedback, feel free to ping me if issues crop up.

slundberg:
Excellent! This looks good to me. The tests failed because of some unrelated plot tests (that are now fixed).

JohnZed:
Thanks! I was hoping this hadn't broken plotting ;)

PR #1653, opened 2020-12-14
Image-To-Text Visualization [RECREATED]

vivekchettiar:
This is initial implementation of the Image-To-Text Visualization, some details:

Can be called by: shap.plots.image_to_text(shap_values) (This is separated from the shap.plots.image() API for now, to differentiate multioutput with text output - this can be changed later).
Allows for Zooming and Setting Opacity of the SHAP Overlay.


slundberg:
Merging this now since it fixes several issues. will iterate more on the API afterwards.

Great work!

PR #1657, opened 2020-12-15
handle categorical features colors

salvomcl:
Small fix in order to handle categorical features mapped as integer in `shap.summary_plot`.
For example features defined as
``` python
pd.Series([0, 1, 2]).astype("category")
```

### DIFF
--- a/shap/plots/_beeswarm.py
+++ b/shap/plots/_beeswarm.py
@@ -80,10 +80,13 @@ def beeswarm(shap_values, max_display=10, order=Explanation.abs.mean(0),
             color = colors.blue_rgb
     color = convert_color(color)
 
+    idx2cat = None
     # convert from a DataFrame or other types
     if str(type(features)) == "<class 'pandas.core.frame.DataFrame'>":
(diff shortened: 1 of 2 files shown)


slundberg:
Thanks! The unit test failures are from now fixed dependency changes.

My overall question is if all categorical features should be set to gray always? It seems like at least for binary features it would be useful to know there are multiple groups, even if the ordering is not meaningful. Also some categorical features have a meaningful ordering as well...

Thoughts?

salvomcl:
Hi @slundberg! Thanks for your answer. 
I faced this problem working with "not ordinal" categorical features with more than 2 levels and this PR should handle this issue.
Regarding binary features, it's useful to color them. In that case, I expect that binary features are coded as  `int`/`float`  so this PR doesn't change the actual behavior of the code.
Same thoughts are valid for "ordinal" categorical features, in that case, the features should be coded as `int`/`float` according to the order of the categorical feature.

slundberg:
Okay, I think I buy that (sorry for the delay was out on travel).

PR #1660, opened 2020-12-16
SHAP Serialization [ Permutation Explainer + Tabular Masker]

vivekchettiar:
Initial implementation of serialization in SHAP. Particularly for Permutation Explainer and Tabular Masker


slundberg:
Looks like it was failing because cloudpickle is used but not listed as a dependency in setup.py. So I added it and lets see if the tests now pass.

PR #1670, opened 2020-12-18
Multi-input Benchmark with MaskedModel Support

maggiewu19:
- Adapted benchmark utility to MaskedModel 
- Adapted to new API with teacherforcing 
- Tested against tabular / text / image-multiclass 
- Original score function remains within perturbation file but can be deprecated / deleted once MaskedModel is in complete support


slundberg:
Great!

PR #1679, opened 2020-12-21
Image Captioning using Azure Cognitive Services

anusham1990:


shap/utils/image.py
```diff
@@ -0,0 +1,117 @@
...
+    Note: Resizing is done for following purposes:
+    1. Azure Cognitive Service does not run for large image sizes. File size should not be greater than 4MB and should have atleast 50x50 size. 
+    2. Reducing image size speedens the SHAP explanation process.
+    """
+    image = load_image(path_to_image)
+    
+    # checking if either of (pixel_size, pixel_size) dimension is greater than 500. 
+    reshaped_path = None
+    x = 500 if image.shape[0] > 500 else image.shape[0]
+    y = 500 if image.shape[1] > 500 else image.shape[1]
```
slundberg, line 80: Here were should not "warp" the image to be square, but rather just ensure that the largest dimension is 500px, while retaining the original aspect ratio...

PR #1686, opened 2020-12-22
Enabling identification of econml trees and generalized random forests

vsyrgkanis:
Enabling identification of econml trees and generalized random forests by TreeExplainer and handled identically to sklearn.tree._tree.Tree, sklearn.tree.DecisionTreeRegressor, sklearn.ensemble.RandomForestRegressor


slundberg:
Thanks! Looks good to me. If you think there are any possible differences from sklearn then unit tests specific to econML are also welcome.

PR #1704, opened 2020-12-30
Updated to use tqdm.auto so progress bars are prettier in notebooks

tommct:
The use of tqdm.auto requires tqdm >= 4.27.0, which was released 2018-10-15.

### DIFF
--- a/shap/utils/_show_progress.py
+++ b/shap/utils/_show_progress.py
@@ -1,5 +1,5 @@
 import time
-import tqdm
+from tqdm.auto import tqdm
 
 
 class ShowProgress():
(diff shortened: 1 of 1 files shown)


ChristianMurphy (changes requested):
setup.py should also be updated to account for the minimum version.
https://github.com/slundberg/shap/blob/6e04be3f4a9c7d61ada301366e4273a096e5c1f5/setup.py#L225

slundberg:
Thanks! I do recall using auto and it causing problems at some point in the past for some people. Perhaps it was the lack of a version constraint? (thanks @ChristianMurphy )

If you update setup.py as @ChristianMurphy suggested I am for trying auto again.

PR #1712, opened 2021-01-05
fix: check clustering attribute type before calling partition tree shuffle

ChristianMurphy:
some maskers, such as text and composite, have a clustering attribute, but attribute is not a numpy ndarray representing a partition tree as partition tree shuffle expects.
this change ensures that if clustering is not the expected numpy ndarray type, it will gracefully fallback to non-partitioned shuffle.

this is what partition shuffle expects
https://github.com/slundberg/shap/blob/6e04be3f4a9c7d61ada301366e4273a096e5c1f5/shap/utils/_clustering.py#L17-L30

this is what text provides:
https://github.com/slundberg/shap/blob/6e04be3f4a9c7d61ada301366e4273a096e5c1f5/shap/maskers/_text.py#L163-L168

and what composite provides:
https://github.com/slundberg/shap/blob/6e04be3f4a9c7d61ada301366e4273a096e5c1f5/shap/maskers/_composite.py#L20-L50

### DIFF
--- a/shap/explainers/_permutation.py
+++ b/shap/explainers/_permutation.py
@@ -69,7 +69,7 @@ class Permutation(Explainer):
             for _ in range(npermutations):
 
                 # shuffle the indexes so we get a random permutation ordering
-                if getattr(self.masker, "clustering", None) is not None:
+                if isinstance(getattr(self.masker, "clustering", None), np.ndarray):
                     # [TODO] This is shuffle does not work when inds is not a complete set of integers from 0 to M TODO: still true?
                     #assert len(inds) == len(fm), "Need to support partition shuffle when not all the inds vary!!"
                     partition_tree_shuffle(inds, inds_mask, self.masker.clustering)
--- a/shap/explainers/_permutation.py
+++ b/shap/explainers/_permutation.py
@@ -56,6 +56,16 @@ class Permutation(Explainer):
         if max_evals == "auto":
             max_evals = 10 * 2 * len(fm)
 
+        # compute any custom clustering for this row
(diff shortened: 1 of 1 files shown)


slundberg:
Thanks! I think this might be something we don't want to just "fall back" for though, since if the masker has a (non-None) clustering attribute it means it needs to be respected. The Composite masker still needs some work, but do you mind sharing the case where this fails with the text masker? Thanks again.

ChristianMurphy:
> do you mind sharing the case where this fails with the text masker?

Sure, I encountered it using a HuggingFace transformer with `PermutationExplainer`
Roughly following https://shap.readthedocs.io/en/latest/example_notebooks/api_examples/plots/text.html

```python
from sentence_transformers import SentenceTransformers # wrapper around HuggingFace transformers
from sklearn.neural_network import MPLClassifer
from shap import PermutationExplainer


model = SentenceTransformer("quora-distilbert-base")
classifier = MPLClassifer()

# train classifier on encoded sentences

def f(x):
    return classifier.predict_proba(model.encode(x))

explainer = PermutationExplainer(f, model.tokenizer)
shap_values = explainer(values_to_explain)
```

slundberg:
Thanks for the example. I see now that the Permutation explainer did not handle maskers where the clustering is dynamic (it changes for each sample (like text). I just added some support for that to this PR. See if it works for you.

ChristianMurphy:
Apologies for the delay in responding.
Yes, trying this out locally a900937  works with the `tokenizer` swimmingly.
Thanks @slundberg!

PR #1737, opened 2021-01-13
Note on future work for benchmarking

maggiewu19:
Added a note in the benchmark notebook regarding future work


slundberg:
Thanks!

PR #1765, opened 2021-01-21
Teacher forcing api update

ryserrao:
This PR has the following:
1. API update (TeacherForcing call function expects (X, Y) as previously compared to (masked_X, X) )
2. OutputComposite Masker
3. Model inference function which handles inferencing for PyTorch and Tensorflow as compared to previously subclassing for each
4. Batched scoring and inferencing
5. Updated notebooks and tests


slundberg:
Thanks @ryserrao ! I think the errors are from an issue where the saving is was not merged into master, but I am not sure about that so I am running it again. Are there outstanding issues with this I should take a look at?

shap/maskers/_output_composite.py
```diff
@@ -0,0 +1,41 @@
...
+        masked_X = self.masker(mask, *args)
+        y = self.model(*args)
+        # wrap model output
+        if not isinstance(y, tuple):
+            y = (y,)
+        # wrap masked input
+        if not isinstance(masked_X, tuple):
+            masked_X = (masked_X,)
+        return masked_X + y
+
```
vivekchettiar: We'd need to add save() and load() functions to this, it will be pretty similar to FixedComposite masker:
This should allow tests to pass, please add tests for this masker if possible

```
    def save(self, out_file, *args):
        super(OutputComposite, self).save(out_file)
        pickle.dump(type(self.masker), out_file)
        self.masker.save(out_file)
        self.model.save(out_file)

    @classmethod
    def load(cls, in_file):
        masker_type = pickle.load(in_file)
        if not masker_type == cls:
            print("Warning: Saved masker type not same as the one that's attempting to be loaded. Saved masker type: ", masker_type)
        return OutputComposite._load(in_file)

    @classmethod
    def _load(cls, in_file):
        masker_type = pickle.load(in_file)
        masker = masker_type.load(in_file)
        model= masker_type.load(in_file)
        outputcomposite_masker = OutputComposite(masker, model)
        return outputcomposite_masker '
```

ryserrao: Thanks @vivekchettiar! Added the load and save functions but seems like "model= masker_type.load(in_file)" runs into an error as it encodes the model instead of the Tokenizer. 
@slundberg there are no outstanding issues with the current PR but noticed that test_gradient.py::test_pytorch_mnist_cnn test is failing (which is unrelated to changes in the PR) and when we run explainer on multiple sentences, the text plot visualization runs into error (arising because of the list of lists having varying sized output sentences).

slundberg: Thanks @ryserrao and @vivekchettiar ! I'll look into the plot error

slundberg: It turned out there were issues with the saving and loading. Specifically models needed to be supported. So I added that and now all the tests pass on my local box. Once they pass on travis then I will merge this :)

shap/_serializable.py
```diff
@@ -0,0 +1,179 @@
...
+        return self
+
+    def __exit__(self, exception_type, exception_value, traceback):
+        # confirm the block end token
+        for _ in range(100):
+            end_token = pickle.load(self.in_stream)
+            if end_token == "END_BLOCK___":
+                return
+        raise ValueError(
+            f"The data block end token wsa not found for the block {self.block_name}."
```
vivekchettiar, line 168: nit: typo
```suggestion
            f"The data block end token was not found for the block {self.block_name}."
```

shap/_serializable.py
```diff
@@ -0,0 +1,179 @@
...
+        )
+
+    def load(self, name, decoder=None):
+        """ Load a data item from the current input stream.
+        """
+        # confirm the block name
+        loaded_name = pickle.load(self.in_stream)
+        print("loaded_name", loaded_name)
```
vivekchettiar, line 177: Can we remove this line?
```suggestion

```

PR #1782, opened 2021-01-26
Fix of tf.python.framework.ops.EagerTensor in _kernel.py

kosinal:
Fixed problem when model_null of type tf.python.framework.ops.EagerTensor.


slundberg:
Thanks! We can't assume tensor flow is always available to import though, so I moved this over to using a safe_isinstance check instead and we will see if things are more happy now :)

PR #1789, opened 2021-01-28
Add support for textual entailment

jsu27:
Textual entailment: given a premise sentence and hypothesis sentence, does the premise imply the hypothesis

Ultimately, we want to pass the input in as 1 argument. The task requires 2 arguments, a premise and hypothesis. To process these, we use the tokenizer to encode the premise and hypothesis then decode the result. This gives the premise and hypothesis concatenated by separator token(s), a single string.

**Code changes:** 
- added logic for preventing the masking of sep tokens (because we want masked sentences to always maintain sep tokens to inform the tokenizer that these represent 2 sentences)
- added logic in group merging function (`merge_score`) to take in optional special tokens (in our case, sep tokens), and ensure that groups with adjacent special tokens are merged last. For this task, this ensures the partition tree's 1st subtree is the 1st sentence and 2nd subtree is the 2nd sentence. The motivation is that we want to explore interaction between tokens within each sentence, though future experimentation can explore interactions between both sentences.
- added demo notebook demonstrating how to run SHAP for a textual entailment model. Preprocessing of the input and model is involved.

**Potential Issues** 
- _Inconsistent Encoding of Masked Sentences:_ Given an original sentence, and given a masked version of the sentence where certain tokens are replaced by masks, it would be ideal to see that their tokenizer encodings have the same shape and that encoded tokens correspond to each other.

We observed that the tokenizer, when encoding then decoding certain unusual masked sentences, adds 1 or more token(s) than the original sentence's encoding shape.

Example:
 ```
original_sentence = 'A boy is jumping on skateboard in the middle of a red bridge.</s></s>The boy skates down the sidewalk.'
# len 27
original_tokens = ['<s>', 'A', ' boy', ' is', ' jumping', ' on', ' skate', 'board', ' in', ' the', ' middle', ' of', ' a', ' red', ' bridge', '.', '</s>', '</s>', 'The', ' boy', ' sk', 'ates', ' down', ' the', ' sidewalk', '.', '</s>']
```
- Bad masked sentence 1
```
masked_sentence1 = 'A boy is jumping on skateboard in the middle of a red bridge.</s></s><mask> <mask> <mask> ates down the sidewalk.' 
# len 28
masked_tokens1 = ['<s>', 'A', ' boy', ' is', ' jumping', ' on', ' skate', 'board', ' in', ' the', ' middle', ' of', ' a', ' red', ' bridge', '.', '</s>', '</s>', '<mask>', '<mask>', '<mask>', ' at', 'es', ' down', ' the', ' sidewalk', '.', '</s>'] 
```
The **'ates'** in the masked sentence was tokenized into 2 parts, 'at' and 'es'. We would expect 'ates' to remain the same. 

- Bad masked sentence 2: 
```
masked_sentence2 =  '<mask> <mask> <mask> <mask> <mask> <mask> <mask> <mask> <mask> <mask> <mask> <mask> <mask> <mask> <mask> </s></s>The boy skates down the sidewalk.
'
# len 28
masked_tokens2 = ['<s>', '<mask>', '<mask>', '<mask>', '<mask>', '<mask>', '<mask>', '<mask>', '<mask>', '<mask>', '<mask>', '<mask>', '<mask>', '<mask>', '<mask>', '<mask>', ' ', '</s>', '</s>', 'The', ' boy', ' sk', 'ates', ' down', ' the', ' sidewalk', '.', '</s>']
```
There is an unnecessary space ' ' token added after the masks and before the separator.

We avoid the issue in our demo notebook by running model evaluations on masked sentences individually instead of passing masked sentences in batches, potentially losing efficiency. Note that this is a tokenizer issue that varies by tokenizer, though we have seen this behavior in at least 2.

**Further comments: Potential Future Work** 
- _Improving Clustering Method:_ Currently, the clustering method to build the partition tree (specifically `merge_score` in `_text.py`) uses heuristics to make decisions about when to combine groups. It would be great to determine if there are more well-founded and better-performing methods.
- _Exploring Interaction Effects between words in both sentences:_ We currently split the partition tree so that the 1st sentence is 1 subtree and 2nd sentence is another so that words in both don't interact with each other. It would be helpful to see if allowing words of both sentences to interact with each other, as they do for humans to understand connections between 2 sentences, improves results.
- _Supporting More Models:_ We used Facebook's BART model trained on mnli in our example, but it would be useful to test and/or extend support for other textual entailment models, such as BERT. Since different tokenizers use different special tokens and tokenizing schemes (such as BERT's [CLS] token), there may be issues at first.

### CODE REVIEW

File: shap/maskers/_text.py [hunk 1]
```diff
@@ -106,7 +105,9 @@ def __call__(self, mask, s):
             out = self.post_process_sentencepiece_tokenizer_output(out)
         # replace sequence of spaces with a single space and strip beginning and end spaces
         out = re.sub(r"[\s]+"," ",out).strip()
-
+        # delete extra spaces around separator tokens
+        if self.tokenizer.sep_token != None:
+            out = re.sub("[\s]*" + self.tokenizer.sep_token + "[\s]*", self.tokenizer.sep_token, out)
```
rebeccaj1211: If my sentence is `"hi </s> </s> bye"`, wouldn't this change it to be `"hi</s></s>bye"` where there's no spaces at all? Is this the intended functionality?

shap/maskers/_text.py
```diff
@@ -152,7 +153,8 @@ def token_segments(self, s):
             token_ids = self.tokenizer.encode_plus(s)['input_ids']
             tokens = self.tokenizer.convert_ids_to_tokens(token_ids)
             special_tokens_mask = self.tokenizer.get_special_tokens_mask(token_ids, already_has_special_tokens = True)
-            tokens = [tokens[i] if special_tokens_mask[i] == 0 else '' for i in range(len(special_tokens_mask))]
+            # avoid masking sep tokens, but still mask bos and eos tokens
```
rebeccaj1211:
```suggestion
            # avoid masking separator tokens, but still mask beginning of sentence and end of sentence tokens
```

shap/maskers/_text.py
```diff
@@ -77,7 +75,8 @@ def __call__(self, mask, s):
                 out = []
                 is_previous_appended_token_mask_token = False
                 for i in range(len(mask)):
-                    if mask[i]:
+                    # mask ignores sep tokens and keeps them unmasked
```
rebeccaj1211:
```suggestion
                    # mask ignores separator tokens and keeps them unmasked
```

sometimes you use sep, sometimes  you use separator in documentation. best to stay consistent.

shap/maskers/_text.py
```diff
@@ -152,7 +153,8 @@ def token_segments(self, s):
             token_ids = self.tokenizer.encode_plus(s)['input_ids']
             tokens = self.tokenizer.convert_ids_to_tokens(token_ids)
             special_tokens_mask = self.tokenizer.get_special_tokens_mask(token_ids, already_has_special_tokens = True)
-            tokens = [tokens[i] if special_tokens_mask[i] == 0 else '' for i in range(len(special_tokens_mask))]
+            # avoid masking sep tokens, but still mask bos and eos tokens
+            tokens = [tokens[i] if (tokens[i] == self.tokenizer.sep_token and i > 0 and i < len(special_tokens_mask) - 1 or special_tokens_mask[i] == 0) else '' for i in range(len(special_tokens_mask))]
```
rebeccaj1211: use parenthesis to group your and/or logic

shap/maskers/_text.py
```diff
@@ -165,8 +167,8 @@ def token_segments(self, s):
     def clustering(self, s):
         self._update_s_cache(s)
         decoded_x = [self.tokenizer.decode([v]) for v in self._tokenized_s]
-        pt = partition_tree(decoded_x)
-        #self._mark_uninvertable(pt)
+        pt = partition_tree(decoded_x, [self.tokenizer.sep_token])
+        # self._mark_uninvertable(pt) # unused because restricts meaningful perturbations
```
rebeccaj1211: Check to make sure removing this won't cause any breaking changes, and then feel free to delete. Since it was commented out in master, chances of it breaking things are low but it's good to double check everything builds/runs

shap/maskers/_text.py
```diff
@@ -229,36 +231,16 @@ def invariants(self, s):
             invariants[:self.keep_prefix] = True
         if self.keep_suffix > 0:
             invariants[-self.keep_suffix:] = True
-
+        # mark sep tokens as invariant
```
rebeccaj1211:
```suggestion
        # mark separator tokens as invariant
```

shap/maskers/_text.py
```diff
@@ -302,8 +284,13 @@ def __add__(self, o):
...
         return len(self.g)
 
-def merge_score(group1, group2):
+# special_tokens: tokens (such as separator tokens) that should be grouped last
+def merge_score(group1, group2, special_tokens=None):
     score = 0
+    # ensures special tokens are combined last, so 1st subtree is 1st sentence and 2nd subtree is 2nd sentence
+    if special_tokens is not None:
+        if group1[-1].s in special_tokens and group2[0].s in special_tokens:
+            score -= 10000
```
rebeccaj1211: do we want to hardcode this score? Are we sure that this value will work for all cases?

notebooks/text_examples/text_entailment/text_entailment_demo.ipynb
```diff
@@ -0,0 +1,784 @@
...
+   "source": [
+    "# Multi-Input Text Explanation: Textual Entailment with Facebook BART\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "packed-capitol",
+   "metadata": {},
+   "source": [
+    "This notebook demonstrates how to get explanations for the output of the Facebook BART model trained on the mnli dataset and used for textual entailment. \n",
```
rebeccaj1211:
```suggestion
    "This notebook demonstrates how to get explanations for the output of the Facebook BART model trained on the SNLI dataset and used for textual entailment. \n",
```

notebooks/text_examples/text_entailment/text_entailment_demo.ipynb
```diff
@@ -0,0 +1,784 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "A boy is jumping on skateboard in the middle of a red bridge.</s></s>The boy skates down the sidewalk.\n"
+     ]
+    }
+   ],
+   "source": [
+    "# encode then decode premise, hypothesis to get concatenated sentences\n",
+    "encoded = tokenizer(premise, hypothesis)['input_ids'][1:-1] # ignore the start and end tokens, since tokenizer will naturally add them\n",
```
rebeccaj1211: Hm, is this why there was a difference between your code and Max's code? You ignore the tokens here?

rebeccaj1211 (commented):
Great job implementing this scenario and good documentation throughout!

shap/maskers/_text.py
maxtell5: My intuition is that the specific value does not matter. Rather, we should be concerned with the relative value compared to the other score adjustments. This is at least 10x larger than any other adjustmennt so it seems likely that it will properly enforce that the two sentences will always be merged last. However, maybe we could consider making it larger if that is a concern?

shap/maskers/_text.py
jsu27: Yes, this is because we found the tokenizer added extra tokens when encoding then decoding, if the sentence contained separators </s> with extra space around them.

shap/maskers/_text.py
jsu27: This was originally left commented out in master, but I can remove it.

shap/maskers/_text.py
```diff
@@ -229,36 +231,16 @@ def invariants(self, s):
...
+        for i in range(len(self._tokenized_s)):
+            if self._tokenized_s[i] == self.tokenizer.sep_token_id:
+                invariants[i] = True
         return invariants.reshape(1,-1)
 
     def feature_names(self, s):
         self._update_s_cache(s)
         return [[self.tokenizer.decode([v]) for v in self._tokenized_s]]
 
-    def save(self, out_file, *args):
```
rebeccaj1211: add this back in

slundberg:
Thanks!!

PR #1794, opened 2021-01-31
fix: correlation dependent linear shap

zduey:
This PR fixes two bugs in `shap.explainers._linear.Linear` related to  #1795 

1. `shap.maskers.Independent` was always used regardless even when `feature_perturbation="correlation_dependent"` was provided in the constructor.
2. `explain_row(...)` was returning a numpy array for the `correlation_dependent` feature perturbation method instead of the expected dictionary. This error only surfaced after fixing the first issue.


slundberg:
Looks good, thanks for the catch! The recent API update missed that change. I made a small tweak and will merge once the tests pass.

PR #1819, opened 2021-02-15
Add display_data ability to numeric vars

zacharyvealey:
Hi @slundberg,

Thanks so much for maintaining such an amazing package!!

This PR adds display_data functionality for numeric features in scatter plots for variables that have been scaled or transformed. In the current implementation, adding untransformed values to shap_values.display_data produced:

image
image

￼
The code in this PR uses the display_data, if available, for the bottom axis, and adds a top axis with the values used in the input:
￼
image

Other small fixes include adding cmap inputs into the in-function dependence_legacy calls, as well as specifying the fontdict key directly in set_xticklabels which removes a matplotlib warning.

I look forward to hearing any thoughts/suggestion!

### DIFF
--- a/shap/plots/_scatter.py
+++ b/shap/plots/_scatter.py
@@ -20,23 +20,18 @@ def scatter(shap_values, color="#1E88E5", hist=True, axis_color="#333333", cmap=
             dot_size=16, x_jitter="auto", alpha=1, title=None, xmin=None, xmax=None, ymin=None, ymax=None,
             ax=None, show=True):
     """ Create a SHAP dependence scatter plot, colored by an interaction feature.
-
     Plots the value of the feature on the x-axis and the SHAP value of the same feature
     on the y-axis. This shows how the model depends on the given feature, and is like a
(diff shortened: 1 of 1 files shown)


nilslacroix:
Produces this error when features for X axis are categorical (strings):
```
# Postcode is categorical/string values
shap.plots.scatter(shap_scaled[:,"Postcode"], color=shap_values[:, "GeoX"], hist=True)

Error:
File ~\miniconda3\envs\Master\lib\site-packages\shap\plots\_scatter.py:144, in scatter(shap_values, color, hist, axis_color, cmap, dot_size, x_jitter, alpha, title, xmin, xmax, ymin, ymax, ax, show)
    142 min_dist = np.inf
    143 for i in range(1,len(vals)):
--> 144     d = vals[i] - vals[i-1]
    145     if d > 1e-8 and d < min_dist:
    146         min_dist = d

TypeError: unsupported operand type(s) for -: 'str' and 'str'
```

PR #1830, opened 2021-02-19
Reduce memory usage for interaction values in large trees.

morriskurz:
Hey everyone,

while looking at the source code I noticed a small optimization in memory usage for the calculation of the interaction values in C++. 

The array for the unique features used in calculating the interaction values in tree_shap.h allocates max_nodes integers per tree to save the unique features used in a tree. For big trees, this is too much, as the number of overall features is an upper bound for the number of unique features used in any tree.

Hope that this will help, thank you for your fantastic work.

Best,
Morris


slundberg:
Excellent. Thanks for this! It LGTM so I will merge.

PR #1841, opened 2021-02-25
Add  tutorial for Multiple Regression SHAP Values

coryroyce:
Created a tutorial for plotting SHAP values with the Kernel Explorer since it was a problem I was trying to solve but didn't find existing examples.


coryroyce:
I saw that the continuous-integration/travis-ci/pr failed, but this is just a tutorial for the Kernel Explainer with Multiple Regression. If there is some other requirements I need to add in order to have this added to the main project, please let me know.

PR #1849, opened 2021-03-03
fix typo (casual inference -> causal inference)

i-shuhei:
fixed this typo issue(https://github.com/slundberg/shap/issues/1848).

- changed from 'casual inference' into 'causal inference' in a docstrings


slundberg:
Thanks!

PR #1889, opened 2021-03-19
fix main_effects switch in Explainer base class

sunshinejnjn:
Fixed main_effects switch which was accidentally overwritten by variant initialization in the Explainer base class


ibrahim3802:
when i use  shap benchmark than I have got an error like that 
AttributeError                            Traceback (most recent call last)
[<ipython-input-11-f73099868ce2>](https://localhost:8080/#) in <module>
      1 results = {}
      2 
----> 3 smasker = shap.benchmark.ExplanationError(
      4     masker, model.predict, X_eval
      5 )

AttributeError: module 'shap' has no attribute 'benchmark'

PR #1949, opened 2021-04-20
Critical Fix: backward permutation of PermutationExplainer unused

### LINKED ISSUE #1948: PermutationExplainer is currently inefficient (does not use backward permutation evaluation)
## Issue 
The PermutationExplainer masks a permutation of the features in both the forward and backward direction but uses only the output of the forward direction. 

## Impact
The PermutationExplainer evaluates the given model on both the forward and backward permutations but the output of the backward permutation is never used yielding an evaluation of the model of (up to) twice the amount of data needed and a potentially drastic slowdown of the model agnostic explainer.

## Solution
To fix this, we need only change a few lines of code to include the result from the backward permutation when estimating the SHAP values. The resulting SHAP value estimates are better as we include effectively per cycle an additional approximation of the SHAP values from the background permutation (whereas prior to this change we would include the forward permutation twice).

## Detailed Explanation
Let’s say we run PermutationExplainer for a model with three features (X is the training data):

`shap_values = PermutationExplainer(model.predict, X)(X).values`

Let’s say the first permutation of the features is the original order, i.e., [1, 2, 3].

The model will be evaluated with the following masking of the three features:
```
[
    [False, False, False],
    [True, False, False],
    [True, True, False],
    [True, True, True],
    [False, True, True],
    [False, False, True],
    [False, False, False]
]
```
This is reflected in the model outputs, i.e., the shape from outputs below is (7,):

https://github.com/slundberg/shap/blob/9411b68e8057a6c6f3621765b89b24d82bee13d4/shap/explainers/_permutation.py#L99

The original code updates the SHAP values in the following way:

https://github.com/slundberg/shap/blob/9411b68e8057a6c6f3621765b89b24d82bee13d4/shap/explainers/_permutation.py#L104-L108

The len(inds) is 3 such that the values in outputs[3+1:] are never used. Effectively, the code just adds the approximation of the SHAP values from the forward pass twice. To include the backward permutation, we need only include the last values in outputs in the second loop. We can confirm with the masking that is evaluated. The backwards permutation is:

```
[
    [False, False, False],
    [False, False, True],
    [False, True, True],
    [True, True, True]
]
```

Or, using the outputs of the model:

`outputs[len(fm):][::-1]`

From this, we can approximate the SHAP values for the reversed (i.e., backward) permutation of the reversed indices, inds[::-1].

Thus, the code to update the SHAP value estimates should be:

```
forwards, backwards = outputs[:len(fm)+1], outputs[len(fm):][::-1]
for i, ind in enumerate(inds):
    row_values[ind] += forwards[i + 1] - forwards[i]
for i, ind in enumerate(inds[::-1]):
    row_values[ind] += backwards[i + 1] - backwards[i]
```

Of course, we could just take differences here (but I personally prefer the more expressive code):

```
forwards, backwards = outputs[:len(fm)+1], outputs[len(fm):][::-1]
row_values[inds] += np.diff(forwards)
row_values[inds[::-1]] += np.diff(backwards)
```

sander-sn:
PermutationExplainer evaluates the model on both the forward and backward permutations but the output of the backward permutation is never used yielding an evaluation of the model of (up to) twice the amount of data needed and a potentially drastic slowdown of the model agnostic explainer. This fix includes the backward permutation in the SHAP value estimates.

Fixes #1948


slundberg:
Sorry I missed this! I have been heads down a bit over the past few months on another project. This is crazy that it got missed before, but indeed you are right. The only change I made to your code was to base the slicing on the length of the `inds` array instead of the full length of the model input, that way it still works when we have non-varying indexes we are skipping.

To prevent this problem in the future I also added (and will push after this merge) a new unit test for the Permutation explainer that ensures we exactly compute Shapley values for second order functions. This exactness only holds if we correctly use the forward and reverse permutations.

I'll merge once travis finishes.

Thanks again, and sorry to have just left this here so long.

sander-sn:
Thanks for addressing this @slundberg! 👍

PR #2017, opened 2021-05-27
Automatic wheel building

PrimozGodec:
#### Issue
I noticed that macOS and Linux wheels are missing on PyPI which is a problem for users without compilers installed on the computer. I the discussion with @slundberg, I offered to propose the solution to build wheels automatically.

#### Solution
This PR enables automatic wheel building with GH Actions. The solution builds wheels and prepares Python tarball for upload to PyPI.

The procedure with this solution is:
1. Bump version
2. Create a new tag and push it to the repository
3. Go to Releases and draft a new release from the tag. It will trigger GH Actions to start building wheels
4. After around 7 minutes find wheels under Actions. Open the action with the name Release x (x is release version) and find two zips under artifacts. Download them unpack and upload to PyPI. The first zip will contain Python tarball and the second wheels for all OSs.

If you agree with this solution merge it to the repository and create a new release as soon as possible that wheels will be available for all operating systems. I am available for any assistence.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -1,0 +1,58 @@
+name: Build wheels
+
+on:
+  release:
+    types: [published]
+
+  # Enable manual run
+  workflow_dispatch:
+
+jobs:
+  build_wheels:
+    name: Build wheels on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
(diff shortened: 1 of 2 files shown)


slundberg:
Sorry to leave this dangling so long! Finally got everything cleaned up for a new release and am merging this :)

Thanks again!

PrimozGodec:
Thanks for considering it. If you will have any issue with using this please let me know

PR #2020, opened 2021-05-31
DOC: fix several doc issues and missing API documentation

EdwinWenink:
- Add missing API entries in the docs for explainers, plots, maskers, models and utils
- The benchmark page was completely empty. Fixed.
- Removed a reference to a non-existing section of api_examples
- Some docstrings referred to non-existing examples. Removed.
- Added missing text examples to toc


thatlittleboy:

PR #2041, opened 2021-06-11
More explicit error handling

### LINKED ISSUE #2033: Raising broad exceptions
There are many places in this package where a bare `Exception` is raised ([example](https://github.com/slundberg/shap/blob/d0b4d59f96adc5d067586c0dd4f7f2326532c47a/shap/explainers/_tree.py#L977)). This is generally considered bad practice and it forces users of the package to do the same in their own code.

One idea is to creating some custom exceptions. Alternatively, you could use some more descriptive built in exceptions.

Apologies if this has been raised before, I did a quick search and couldn't find a similar issue.

collinb9:
* Add an `exceptions` module with one custom exception: `UnsupportedModelError`
* Replace each instance of a bare `Exception` with a more explicit exception, either `ValueError`, `RuntimeError`, `TypeError` or `UnsupportedModelError`

Any exceptions raised in a commented out block of code were left alone

closes https://github.com/slundberg/shap/issues/2033

### DIFF
--- a/shap/__init__.py
+++ b/shap/__init__.py
@@ -68,6 +68,7 @@ else:
 from . import datasets
 from . import utils
 from . import links
+from . import exceptions
 
 from .actions._optimizer import ActionOptimizer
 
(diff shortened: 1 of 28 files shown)


slundberg:
Thank you so much for this! For some reason it got marked "read" and I didn't see it. Now there is also another PR that adds even a bit more detail than this one, so I am going to merge that one and leave my thanks here. (and also so avoid the merge conflicts)

PR #2067, opened 2021-07-01
Changed the background of the header image in the README to be compatible with dark themes

gialmisi:
The image in the header file is unreadable if one is using a dark theme like the one available on GitHub. See example of the issue [here](https://imgur.com/6BU07bc).


slundberg:
Thanks! I am going to merge this and then tweak to use an SVG for easy switching later.

PR #2084, opened 2021-07-13
Allow custom tokenizer in Explainer through text masker

nsorros:


nsorros:
# Rationale
I would like to be able to pass a custom tokenizer to Explainer through text masker. At the moment only tokenisers from the transformers library are accepted. I would like to be able to do something like this:
```
custom_tokenizer = CustomTokenizer()
masker = shap.maskers.Text(custom_tokenizer)

explainer = shap.Explainer(f, masker, output_names=labels)
shap_values = explainer(texts[0:2])
```

The tokenizer that the user will pass would need to implement some basic methods, the API I am envisioning the user would need to implement is
```
class CustomTokenizer():
	def encode(self, x):
                # encodes string x to integers
		pass

	def decode(self, x):
                # decodes integers to string
		pass

	def tokenize(self, x):
                # tokenises string x to tokens
		pass
```
I think this is a rather general API but keen to hear suggestions about things that might be missing.

I suggest that the default behaviour of the text masker in case the user does not pass a transformer tokenizer is to expect a custom tokenizer from the user and thus check that the above methods are implemented before calling them and warn the user.

# Use case

This comes up if you have implemented a custom model say a neural net and want to leverage the same API presented in the text examples. It also builds on the ability to be able to pass a custom function to make predictions so similarly the user should be able to pass a tokenizer of their choice.

An alternative could be to use DeepExplainer but in my case this was not working due to incompatible tensorflow versions. Another alternative would be to implement a custom masker but I am not sure how that would work with text to allow `shap.plots.text` visualisations. Keen to hear thoughts on other ways I can achieve the same thing.

# Working example

Here is a working example based on https://shap.readthedocs.io/en/latest/example_notebooks/text_examples/sentiment_analysis/Emotion%20Classification%20Explanation%20Demo.html that wraps the transformers tokenizer to a custom tokenizer to demonstrate the feature
```
import copy
import pandas as pd
import numpy as np
from transformers import AutoTokenizer, AutoModelForSequenceClassification
import shap
import scipy as sp
from datasets import load_dataset
import torch


dataset  = load_dataset("emotion", split = "train")
data = pd.DataFrame({'text':dataset['text'],'emotion':dataset['label']})

bert_tokenizer = AutoTokenizer.from_pretrained("nateraw/bert-base-uncased-emotion",use_fast=True)
model = AutoModelForSequenceClassification.from_pretrained("nateraw/bert-base-uncased-emotion")

id2label = model.config.id2label
label2id = model.config.label2id
labels = sorted(label2id, key=label2id.get)

class Tokenizer():
	def __init__(self, tokenizer):
		self.tokenizer = tokenizer

	def encode(self, x):
		return self.tokenizer.encode(x)[1:-1]

	def decode(self, x):
		return self.tokenizer.decode(x)

	def tokenize(self, x):
		return self.tokenizer.tokenize(x)

def f(x):
    tv = torch.tensor([bert_tokenizer.encode(v, padding='max_length', max_length=128,truncation=True) for v in x])
    attention_mask = (tv!=0).type(torch.int64)
    outputs = model(tv,attention_mask=attention_mask)[0].detach().cpu().numpy()
    scores = (np.exp(outputs).T / np.exp(outputs).sum(-1)).T
    val = sp.special.logit(scores)
    return val

custom_tokenizer = Tokenizer(bert_tokenizer)
masker = shap.maskers.Text(custom_tokenizer)

explainer = shap.Explainer(f, masker, output_names=labels)
shap_values = explainer(data['text'][0:2])
```

nsorros:
keen to hear your thoughts @slundberg and happy to add tests and examples or tweak the implementation in a way that it serves more users or a more general use case

nsorros:
Is it possible to get your thoughts on this PR @slundberg

slundberg:
Yes! Sorry for the delay. I have been looking through it, and I like the idea (others have also asked for this). I will likely make some changes before merging, but will try and finish that soon.

nsorros:
> Yes! Sorry for the delay. I have been looking through it, and I like the idea (others have also asked for this). I will likely make some changes before merging, but will try and finish that soon.

I m happy to help if there is anything I can do

nsorros:
Closing at the suggestion has been mostly implemented. The main difference with the example I gave is that an extra `__call__` needs to be defined that wraps the encodings with `input_ids` so the following custom tokeniser works.

```
class Tokenizer():
	def __init__(self, tokenizer):
		self.tokenizer = tokenizer

	def encode(self, x):
		return self.tokenizer.encode(x)[1:-1]

	def decode(self, x):
		return self.tokenizer.decode(x)

	def tokenize(self, x):
		return self.tokenizer.tokenize(x)

       def __call__(self, x):
               return {"input_ids": self.encode(x)}
```

PR #2088, opened 2021-07-14
Add show arg to text plot and return html

nsorros:
Similar to other plots, I add a `show` parameter in `shap.plots.text` to return the HTML so that we can for example embed it in a streamlit app.


nsorros:
Since 860893e3c23e0c84f77a74c8771e703ac9af8032 introduced a `display` argument instead of a `show` that I originally suggested, the only thing this PR introduces is a doctoring explaining it @slundberg

slundberg:
Thanks!

M007-lab:
It doesn't work actually it is on github code but not via " pip install " even if it is same version !

PR #2091, opened 2021-07-14
Fix #2087

### LINKED ISSUE #2087: Unused model evaluations in Permutation explainer.
Hi, 
I am really enjoying the SHAP library, and for a while, my curiosity has driven me to analyse and understand its source code.

Currently I have been studying the Permutation explainer  and I think I found out a little problem. At the following lines https://github.com/slundberg/shap/blob/master/shap/explainers/_permutation.py#L104-L108

It appears that the same code is repeated twice. From what I understand, the intent was to have the first for-loop compute marginal contributions of the features when going forward (the first half of the rows of the array "outputs"),
and a second for-loop to compute the marginal contributions when going backward (second half of the rows of the array "outputs"). The way the code is currently implemented, the second half of the rows of "outputs" is never used. This could be sub-optimal because "outputs" stores the results of multiple model evaluations and some of them are never used when computing the Shapley values.

I just wanted to make sure that this is actually an issue and that I am not misunderstanding the code.
Thank you very much.

gablabc:
Hi, 
I am proposing a very simple fix for the issue #2087. Now, all model evaluations stored in `outputs` are used to compute 
the Permutation Shapley values, and not just the first half.

### DIFF
--- a/shap/explainers/_permutation.py
+++ b/shap/explainers/_permutation.py
@@ -102,10 +102,15 @@ class Permutation(Explainer):
                     row_values = np.zeros((len(fm),) + outputs.shape[1:])
 
                 # update our SHAP value estimates
-                for i,ind in enumerate(inds):
+                i = 0
+                # forward
+                for ind in inds:
                     row_values[ind] += outputs[i+1] - outputs[i]
-                for i,ind in enumerate(inds):
-                    row_values[ind] += outputs[i+1] - outputs[i]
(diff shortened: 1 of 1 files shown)


sander-sn:
See #1949

slundberg:
Thanks @gablabc! I merged @sander-sn 's solution and also updated it with your approach of skipping the enumerates.

PR #2097, opened 2021-07-18
use display_data in plots.force if available

### LINKED ISSUE #2096: plots.force ignores display_data of Explanation instances
If the data contains categorical features, we would like to use the `display_data` attribute of the `Explanation` instance for the plot. 
However, in `plots.force`, the information of `dislapy_data` will not be loaded unless you use the `display_features` argument.
image

c56pony:
Fix #2096 

Changed `plots.force` to use `display_data` in preference to `data` if it exists.


slundberg:
Great. Thanks! Just remerging with the latest master and then I will merge here.

PR #2110, opened 2021-07-27
Return matplotlib plot when show False

nsorros:
The idea of this PR is to return the matplotib plot when show False so the user
can furher modify or save it. This could extend to all plots that have a `show`
param

To be honest this is what show=False advertises it does so I might as well be 
missing something

### DIFF
--- a/shap/plots/_bar.py
+++ b/shap/plots/_bar.py
@@ -315,6 +315,8 @@ def bar(shap_values, max_display=10, order=Explanation.abs, clustering=None, clu
     
     if show:
         pl.show()
+    else:
+        return pl
 
 
 


shap/plots/_bar.py
```diff
@@ -315,6 +315,8 @@ def bar(shap_values, max_display=10, order=Explanation.abs, clustering=None, clu
     
     if show:
         pl.show()
+    else:
+        pl
```
slundberg: what the intention to return something here?

nsorros: yes, sorry, I had fixed and forgot to push. I did a force push after rebasing on top of the latest master

nsorros: Also happy to add to other plots as well

nsorros:
Closing as it might not be relevant anymore. The intention was to be able to manipulate the plot and embed in an app like streamlit. Kind of similar to how `text` has a `display` argument

PR #2115, opened 2021-07-30
Return figure in waterfall plot if show is false

jwbron:
Updates waterfall plot to be consistent with the [documentation](https://github.com/slundberg/shap/blob/cb73243c98f94fbb77180a719c780156d0bddfbe/shap/plots/_waterfall.py#L37) and with [force_matplotlib](https://github.com/slundberg/shap/blob/cb73243c98f94fbb77180a719c780156d0bddfbe/shap/plots/_force_matplotlib.py#L409)


slundberg:
Thanks! just rerunning the tests after an unrelated travis fix

PR #2123, opened 2021-08-05
Being NaN-aware in calculation of means and their respective variances

Pseudomanifold:
First of all: thanks for this great package! In our application, we are calculating Shapley values of time series of variable lengths and use NaNs to put them in a single fixed-length tensor. This causes problems in downstream Shapley value calculations because any NaN keeps getting propagated.

The proposed fix addresses this by using `np.nanmean` and `np.nanvar`, respectively. I would suggest that this behaviour become the default one (or a configurable one) because most users will probably prefer the behaviour of said functions over the ones that are not aware of NaNs.

The disadvantage is that it might mask some problems, but I find this risk to be negligible because the calculation only involves the mean and variance, thus only changing the normalisation.


CloseChoice:
@Pseudomanifold thanks for the PR and I am very sorry for the really late reply. I am going over old PRs and check which ones we can close, and which we can get merged. I think this is a sensible enhancement, having a test for this would be great though. Would you mind creating a timeseries example with NaN padding?

Pseudomanifold:
No worries, but I don't think I have the capacity to extend this now, so please feel free to ignore for the time being...

PR #2154, opened 2021-08-27
Update README.md

Amir-P:
The `plots.force` seems to be updated and now requires `base_value` and only values of the shap values.


slundberg:
Thanks! Sorry I was out for a while and let this PR sit. When I tried the readme it works as current written. If you see an issue here still let me know and I'll reopen.

PR #2155, opened 2021-08-28
Reinforcement Learning Example

dynamik1703:
Hello everyone,
Reinforcement learning is becoming more widely used. The agent's learned decisions are not easy to interpret and therefore SHAP provides an excellent method for analysis.

Here I have prepared a notebook as an example. 

I am looking forward to your feedback.

Kind regards
Roman


rfali:
@dynamik1703 I think application of SHAP to an RL method is very useful. While you have given some example, it is not clear how your approach can be applied to a trained RL model that can be fed as an input to the SHAP DeepExplainer object?

PR #2195, opened 2021-09-24
Fix typo (from KernalExplainer to KernelExplainer)

deepyaman:


slundberg:
Thanks! Sorry I was out for a while and let the PR sit.

PR #2200, opened 2021-09-28
Corrected typos on README

TNanukem:
Covariance Matrix was misspelled and there was a not where it should be a no


TNanukem:
It seems that the appveyor is breaking on every single pull request. This PR only changes the README so I don't think it breaked anything. Anyone has more information?

slundberg:
Thanks! And sorry I have been out for while, and let this PR sit around. The CI issues should be resolved now that it is switched over to GitHub actions.

PR #2223, opened 2021-10-13
Change _record_gradient to record_gradient

maxispeicher:
`tensorflow` changed `tensorflow.python.eager.backprop._record_gradient` to `tensorflow.python.eager.backprop.record_gradient` (https://github.com/tensorflow/tensorflow/commit/6c3ab6a07c86f1d6b73aa80d870d7a1e7d042316). Therefore the attribute in `shap` needs to be changed aswell.


slundberg:
Thanks! Sorry I have been out for a while and just let this PR sit around (and others). This is a great bug to fix, and there was also another later PR that had some try catch to support old and new versions of TF. So I merged that and will close this one (with much thanks).

PR #2228, opened 2021-10-16
Fix usage of pyplot when show=False

Ceglowa:
Fix usage of pyplot when show=False.


slundberg:
Thanks! I am sorry for being out for so long and letting this just sit here. I merged a different fix for the same error already, but I noted that you also seem to have some sigmoid logic added in here as well. Could you explain the motivation for that? Thanks again!

slundberg:
Closing this for now. But if there is something with the sigmoid that you think needs incorporated just say so. Thanks!

PR #2219, opened 2021-10-12
Fix waterfall plotting

LetiP:
The waterfall plot imports and uses `matplotlib` as `pl`, but 4 lines use it as `plt`, causing errors.
This commit fixes the issue.

PR #2233, opened 2021-10-22
Fix: Bad matplotlib.pyplot import

banderlog:
`import matplotlib.pyplot as pl`  stated while `plt.ioff()` used in few lines.  
This causes error when calling `shap.plots.waterfall(shap_values[0], show=False)`

### DIFF
--- a/shap/plots/_waterfall.py
+++ b/shap/plots/_waterfall.py
@@ -40,7 +40,7 @@ def waterfall(shap_values, max_display=10, show=True):
     
     # Turn off interactive plot
     if show is False:
-        plt.ioff()
+        pl.ioff()
     
 
     base_values = shap_values.base_values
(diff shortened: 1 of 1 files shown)


kylezeberlein:
I am also running into this issue. Merging in this PR would be a huge help to me and my work!

kevin1kevin1k:
I also have this issue
seems to be a duplicate of #2219

banderlog:
It looks like nobody will give a flying fuck for this merge request before it will pass auto checks.

But the single check it can't pass is a test coverage check.
It wants tests, that will check the wrong library import, which should be checked with a linter (flake8, mypy, etc) before pushing it to master.

So, writing a test for this is idiocy, IMHO.

There are few similar merge requests, maybe someone will write tests to pass this check.

slundberg:
@banderlog thanks for this! Sorry I have been out for a while after paternity leave and left a bunch of PRs dangling. I did merge some that fix this finally. Will be out in the next release, hopefully soon.

Issue #2240, opened 2021-10-27
`beeswarm` corrupts `Explanation` object


ravwojdyla:
Reproduction:

```python
import xgboost
import shap

X, y = shap.datasets.boston()
model = xgboost.XGBRegressor().fit(X, y)
explainer = shap.Explainer(model)
shap_values = explainer(X)
```

First see the valid summary:

```python
shap.summary_plot(shap_values, max_display=10)
```
![image](https://user-images.githubusercontent.com/1419010/139154104-630ca326-d297-4b10-8418-669a1205253a.png)

Now use beeswarm with truncated list:

```python
shap.plots.beeswarm(shap_values, max_display=2)
```
![image](https://user-images.githubusercontent.com/1419010/139154283-de7c6e4f-6073-47c7-8c1e-cbd6f2475af8.png)

Now the summary graph is corrupted:

```python
shap.summary_plot(shap_values, max_display=10)
```
![image](https://user-images.githubusercontent.com/1419010/139154248-60a74e68-0d41-4772-af41-e7853776e541.png)

Notice how the 2nd row (sum of features) corrupted the summary plot ^

AlexTo:
I can confirm that beeswarm function changes the `shap.values` array.

js850:
I have also noticed this.  A little debugging shows it's happening in these lines in _beeswarm.py

```python
    # see how many individual (vs. grouped at the end) features we are plotting
    if num_features < len(values[0]):
        num_cut = np.sum([len(orig_inds[feature_order[i]]) for i in range(num_features-1, len(values[0]))])
        values[:,feature_order[num_features-1]] = np.sum([values[:,feature_order[i]] for i in range(num_features-1, len(values[0]))], 0)
```

A workaround is to use `copy.deepcopy(shap_values)` before passing them into `beeswarm`

thatlittleboy:
Closing, also duplicate of #1970

PR #2241, opened 2021-10-27
Fix beeswarm plot, avoid corrupting the explanation object

ravwojdyla:
Re: #2240 

Changes:
 * beeswarm only works with Explanation object, make that explicit, otherwise we get:
> UnboundLocalError: local variable 'values' referenced before assignment
 * make a copy of the values from the Explanation, to avoid corrupting them later on
 * remove unused `base_values`


slundberg:
Thank you! Sorry I have been out for a long time, and am just now taking care of this. I'll merge once the tests pass

PR #2242, opened 2021-10-27
Allow to control the width of the heatmap plot

ravwojdyla:
Reason: default size might be a bit too small for a larger datasets.


slundberg:
Excellent, thank you! Sorry I have been out for a while and let this sit. Will merge once the new CI tests pass.

Issue #1262, opened 2020-06-10
IndexError: string index out of range


ChenTaHung:
Hello, I am new to `shap` package. 
When I was plotting the waterfall_plot, an error came out. 

Here is my codes 
`shap.waterfall_plot(explainer.expected_value, shap_values[0,:18], catestX_Shap.iloc[0, :18], show=True, max_display=18)`

And the Error was shown like this 
> IndexError: string index out of range

But, when I select the shap_values from `shap_values[0, :1 ]` to `shap_values[0,:17]` and my data `catestX_Shap.iloc[0, :1]` to `catestX_Shap.iloc[0, :17]`, the plot works.

Only when I select the eighteenth variable, the error comes out.

But when I jump over the variable, it works again.

`shap.waterfall_plot(explainer.expected_value, shap_values[0,[17,19,20]], catestX_Shap.iloc[0, [17,19,20]], show=True)`

![image](https://user-images.githubusercontent.com/49839507/84248088-84f09180-ab3b-11ea-911b-82599deae71f.png)

Below is my eighteenth variable 
`catestX_Shap.iloc[:,18]`

>index values 
0        0
1        0
2        0
3        0
4        0
        ..
48308    0
48309    0
48310    0
48311    0
48312    0

For the contents in the variable 

`catestX_Shap.iloc[:,18].value_counts()`

> 0                    47869
1                      234
4                      117
3                       51
5                       28
2                       12
Others or Missing        2

Thank you for your reply.

luiscrouch-brex:
Hey I just ran into a similar error and the culprit was a categorical feature (was using catboost) that had some blank entries (i.e., feature value was `''`).

I think the sanitation of the feature values may be overzealous. Or a `try` `except` framework could be used here:

```
/databricks/python/lib/python3.7/site-packages/shap/plots/_waterfall.py in waterfall_legacy(expected_value, shap_values, features, feature_names, max_display, show)
    411             yticklabels[rng[i]] = feature_names[order[i]]
    412         else:
--> 413             yticklabels[rng[i]] = format_value(features[order[i]], "%0.03f") + " = " + feature_names[order[i]]
    414 
    415     # add a last grouped feature to represent the impact of all the features we didn't show

/databricks/python/lib/python3.7/site-packages/shap/utils/_general.py in format_value(s, format_str)
    232         s = format_str % s
    233     s = re.sub(r'\.?0+$', '', s)
--> 234     if s[0] == "-":
[...]

PR #2247, opened 2021-10-29
Check the type of the feature before formatting

Thuener:
This is a tentative to solve issue #1262


slundberg:
Thanks! Sorry I have been out for a while. I made some small tweaks and will merge once the tests pass.

PR #2249, opened 2021-11-02
- Fix issue 2248

javirandor:
Including "deep" algorithm in `shap.Explainer`


slundberg:
Thank you! And sorry I was out for a while and let this PR just sit around. Merging now.

Issue #2230, opened 2021-10-21
Publishing automated wheels on PyPI


jklaise:
With the automated wheel building workflow which gives wheels for a range of systems (see Artifacs: https://github.com/slundberg/shap/actions/runs/1364867953) it would be very useful to include wheel uploading into the release process on PyPI. I guess the current process to publish is using the `deploy` job via Travis: https://github.com/slundberg/shap/blob/429fb3e0ac2ef179f1cff7e1a64b4f7b26f41eb5/.travis.yml#L27-L36
This could be extended to upload the artifacts generated after manually triggering the Github action.

This would be extremely useful to get pre-built wheels on Linux and MacOS by just running `pip install shap` avoiding the build process. I see that `cibuildwheels` makes `manylinux` wheels by default, so it would also resolve #1383.

dashtagger:
Hi can anyone guide me on how exactly do I build the linux version of shap? I don't understand how to use this travis file.

Else if any one can direct me to a manylinux distribution whl file of shap it would be much appreciated.

jklaise:
@dashtagger you can download pre-built wheels of version 0.40 on here (under artifacts - bdist_files): https://github.com/slundberg/shap/actions/runs/1364867953 . It's just that you have to manually download them as they're not published on PyPI (which is why I opened this issue).

Issue #1577, opened 2020-11-11
a python-3.9 wheel ?


stonebig:
there is "only" Python-3.6 3.7 3.8 wheels on pypi

Bougeant:
At the moment, `shap` requires `numba` which is expected to be compatible with Python 3.9 in mid-December (https://github.com/numba/numba/issues/6345)

Bougeant:
numba and llvmlite should now have support for Python 3.9

SvenSchiffner:
@slundberg Can we do anything to help with a wheel release for python 3.9?

jonnor:
This should now be fixed, thanks to the above PR. This issue can be closed. CC @slundberg

PR #2263, opened 2021-11-12
[ci] add action to publish build wheels to pypi

quantumtec:
@slundberg This is a draft PR how the building of binary wheels could be automated for the PyPi repo using standard Github actions similar to those used on many other OSS projects instead of using travis or manually.

This would resolve open issues like #2230 and #1577.

Building upon the wheel building it:
- Runs only on releases
- Uses the [latest cibuildwheel](https://github.com/pypa/cibuildwheel/releases/tag/v2.2.2) for wide compatibility with different platforms (win, mac, linux)
- Musl is currently disabled because of numba
- Uses the recommended way for publishing wheels on PyPi via [twine](https://github.com/pypa/twine)
- Checks the wheels before upload
- Instead of the travis encrypted secret, use ENV variables for the PyPi repo access

Still required (by a **reviewer**):
- Final check of the generated wheels and formats
- **Set the correct secret PYPI_TOKEN (otherwise the final step fails)**
- Possible integration with testing / appveyor
- Version upgrade and confidence in publishing the wheels automatically for a release


slundberg:
Hey! Thank you so much for doing this. And apologies for letting it sit so long! (was out for a while after paternity leave)

This looks great. I just went through the effort to migrate off of travis and start using GitHub actions to keep everything clean (still need to setup windows/Mac there). So this makes a lot sense. I will look over this more carefully before the next release and try and give it try for 0.41

PR #2282, opened 2021-11-22
Fixed ModuleNotFoundError, matplotlib imported as pl while code written as plt

nevinbaiju:
In shap/plots/_waterfall.py, matplotlib is imported as "pl" while in 4 lines "plt" is used which throws a ModuleNotFoundError.
Fixed the typo from "plt" to "pl" which resolves this issue.

### DIFF
--- a/shap/plots/_waterfall.py
+++ b/shap/plots/_waterfall.py
@@ -40,7 +40,7 @@ def waterfall(shap_values, max_display=10, show=True):
     
     # Turn off interactive plot
     if show is False:
-        plt.ioff()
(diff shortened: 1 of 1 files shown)


ghost:
This is a breaking bug that is stopping me from going on with my work. Good job!

nevinbaiju:
@giostefa Glad, I could help. :D
Do check out #2283, we were discussing some workaround for this as well.

slundberg:
thanks! Sorry I was out for a while and this lingered. It has been fixed!

PR #2290, opened 2021-11-25
SHAP on converted models from onnx-tensorflow

cwmeijer:
We want to run SHAP on models that were converted from onnx using [onnx-tensorflow](https://github.com/onnx/onnx-tensorflow). Using this draft PR, we hope to be able to test against the CI setup from the original slundberg/shap repo.


slundberg:
Hey! Thanks and sorry for letting this sit so long (been out after paternity leave for a while). What are next steps here? (I see this is still in a draft form)

PR #2291, opened 2021-11-26
Update _kernel.py for memory optimization

Qingtian-Zou:
add garbage collector to kernel explainer


slundberg:
Thanks! sorry I have been out for a while and am just reviewing. Could you give some context on what caused the need for this? I am a bit concerned that calling the GC for each instance might make some things slower.

Qingtian-Zou:
The problem arises from the following code that I ran on my workstation with 24 GB of video RAM:
`data_summary = shap.kmeans(trainX, 100)`
`explainer_autoencoder = shap.KernelExplainer(compute_loss, data_summary)`
`shap_values = explainer_autoencoder.shap_values(testX,nsamples=2000)`
For reference, each data sample of trainX and textX contains 1800 features, the datatype of which are both float64, and the output of compute_loss is within 0 to 1.

Every time I ran the code above, I can see my VRAM nearly ran out after 1 iteration, and crashed during the second iteration because of memory exhaustion. This is why I call GC every time. I agree that calling GC too frequently might have performance issue, but I think a "memory-efficient" mode can be necessary at some point. Maybe adding a configuration variable to the Explainer instance might be a better solution.

slundberg:
Thanks for the details! Sounds good, I made it an option.

PR #2294, opened 2021-11-28
`plt` is not defined. `pl` is. Already done in https://github.com/slundberg/shap/pull/2265, remove this PR

ghost:


juanitorduz:
I was about to open this PR as well :) . This is breaking the `shap.plots.waterfall` plot.

juanitorduz:
I also see there are couple of places in this file where `plt` is used instead of `pl`

slundberg:
Thanks! Sorry I have been out, fixed now.

PR #2296, opened 2021-11-30
Changing incorrect plt imports to pl

machur:


machur:
Duplicate of https://github.com/slundberg/shap/pull/2294 and https://github.com/slundberg/shap/pull/2265

PR #2302, opened 2021-12-03
Fix the color bar can't show in jupyter notebook problem

wuyuyu1024:
Hi, I'm newbie here. And I found the colorbars doesn't show in my jupyter notebook. I made some small change and they shows normally now in work. I'm not very sure if it helps. Thanks for your work!


slundberg:
Thanks! This is interesting, it must have come from some change in a new version of matplotlib. Sorry for the slow review, been out for a while after paternity leave. Will merge this after the new CI tests finish.

PR #2305, opened 2021-12-06
Update _waterfall.py

conorosully:
Unable to set show=False for waterfall plots. The issue is that matplotlib.pyplot is imported as pl and not plt.


slundberg:
Thanks so much! Sorry for slow review, was out after parental leave for a while. I merged an earlier fix of the same painful problem.

Issue #2261, opened 2021-11-11
Waterfall with show=False results in error plt not known.


sorenwacker:
Seems there is an import missing somewhere. The waterfall plot works, but not with `show=False`

```

shap.plots.waterfall(shap_values[0], show=False)


NameError                                 Traceback (most recent call last)
/tmp/ipykernel_3533873/1075320206.py in <module>
----> 1 shap.plots.waterfall(shap_values[0], show=False)
      2 savefig('waterfall.pdf')

/bulk/LSARP/envs/conda/pycaret/lib/python3.9/site-packages/shap/plots/_waterfall.py in waterfall(shap_values, max_display, show)
     41     # Turn off interactive plot
     42     if show is False:
---> 43         plt.ioff()
     44 
     45 
    NameError: name 'plt' is not defined

```

kejlnz:
I have not tried it myself, but it seems it is just a typo:
Replacing all `plt` occurrences with `pl` in `_waterfall.py` should do the trick.

sorenwacker:
Maybe ‘pyplot’ was imported as ‘pl’ instead of ‘plt’. Do you want me to create a pull request if that is the case? I am not sure what code standards you use. I would rather import it as ‘plt’ as it is commonly done.

js850:
I hit this issue also, so I created a Pull Request to fix it as suggested by @kejlnz . #2342

js850:
There are about 10 duplicate Pull Requests open all trying to fix this issue.  It would be wonderful if one of the maintainers would approve and merge one of them :)  @slundberg fyi

![image](https://user-images.githubusercontent.com/1587384/147825332-94192ac1-a3dd-46ec-933a-0aeca7c2cc33.png)

rglynn:
@slundberg  can we please get this approved? It's been open for months now with multiple PRs trying to fix it.

bello7777:
yes pls , same here it work fine locally, but on the cloud its 
Traceback (most recent call last):

 


  File "/home/appuser/venv/lib/python3.7/site-packages/shap/plots/_waterfall.py", line 337, in waterfall_legacy

    plt.ioff()

NameError: name 'plt' is not defined

H2CO3:
[...]

PR #2313, opened 2021-12-10
fix pl variable name when using show=False

giahung24:
NameError: name 'plt' is not defined when calling shap.plots.waterfall(show=False)


areeh:
this fixes #2261 but there are more in the same file (ctrl+f plt)

PR #2314, opened 2021-12-12
Fixed typo in example and also updated the command for plotting

abs428:
Summary of changes -

1. The docs refer to AGE when it should actually refer to RM
2. The way to call plot dependence plots is `shap.partial_dependence_plot`


slundberg:
Great thanks! Sorry for slow review, was out after parental leave for a while.

PR #2320, opened 2021-12-14
waterfall plot: use display_data if exists

yuuuxt:
not sure if this is a complete fix and if it breaks anything. Please review, thanks!


shap/_explanation.py
```diff
@@ -323,9 +323,10 @@ def __getitem__(self, item):
             pos += 1
 
             # skip over Ellipsis
-            if t == Ellipsis:
-                pos += len(self.shape) - len(item)
-                continue
+            if type(t) not in [np.ndarray, pd.Series]:
```
slundberg: Can you elaborate on scenarios where this change is needed? Thanks!

slundberg:
Thanks for the fix! Sorry I was slow on the review (was out for a while after paternity leave). See the one comment I left.

shap/_explanation.py
yuuuxt: oh, that was for #2325 , just showing my naive "fix" for that issue. Now it's reverted, so it's clean for this PR now.

yuuuxt:
> Thanks for the fix! Sorry I was slow on the review (was out for a while after paternity leave). See the one comment I left.

Fixed that (replied above).

And congrats! :)

PR #2336, opened 2021-12-26
[plots] fix waterfall_plot matplotlib import

partiallycomplex:
Matplotlib was imported as pl but used as plt


Shawe82:
Any chance to get that in? It fixes an annoying bug.

slundberg:
Thanks for the fix! Sorry I was slow on the review (was out for a while after paternity leave). Fixed this with an earlier PR, but this was much appreciated!

PR #2342, opened 2021-12-31
Fix bug in waterfall with show=False

js850:
This fixes bug `NameError: name 'plt' is not defined` reported in #2261


slundberg:
Thanks for the fix! Sorry I was slow on the review (was out for a while after paternity leave). I got several PRs on this while I was out so I just merged one. Thanks though!

PR #2348, opened 2022-01-07
GPUTreeShap updates

RAMitchell:
Update GPUTreeShap to latest. Fixes compile error with cuda >= 11.4.

Also fixes in error for multiclass xgboost models where the number of trees is incorrectly deduced. This error applies to both CPU and GPU code.


setup.py
```diff
@@ -109,7 +109,8 @@ def compile_cuda_module(host_args):
     cuda_home, nvcc = get_cuda_path()
 
     print("NVCC ==> ", nvcc)
-    arch_flags = "-arch=sm_60 " + \
+    arch_flags = "-arch=sm_37 " + \
```
imatiach-msft, line 112: Is there a disadvantage to adding more architectures here?  Is it possible to add many more architectures - or perhaps a better way may be to also allow the user to configure this somehow to pass their architecture type, or have some way to automatically detect/figure out the architecture based on the current machine?

imatiach-msft (approved):

shap/cext/gpu_treeshap.h
```diff
@@ -20,6 +20,8 @@
 #include <thrust/iterator/discard_iterator.h>
 #include <thrust/logical.h>
 #include <thrust/reduce.h>
+#include <thrust/host_vector.h>
+#include <cub/cub.cuh>
```
imatiach-msft, line 25: it looks like this include breaks older versions of cuda

slundberg:
@RAMitchell do you think the XGBoost unit test errors are at related to this PR? I can start digging but it seems possibly connected to using CUDA.

RAMitchell:
@slundberg sorry for being slow. I think it should be fixed now. Xgboost multiclass in shap's cext has been broken until now - the number of boosting iterations was being passed to the C extension and not the number of trees. The additivity checks were passing because the prediction was also wrong. This issue only applied in cases where the c extension is called and not where shap calls xgboost natively.

slundberg:
@RAMitchell Now it is my turn to be slow... Ah! makes sense, good to fix. I am rerunning this on the new CI setup and will merge after that.

PR #2353, opened 2022-01-10
correct plt to pl

ssrosa:
line 43 calls plt which has been imported at the top as pl. Throws an error if you try to call the function when show = False

### DIFF
--- a/shap/plots/_waterfall.py
+++ b/shap/plots/_waterfall.py
@@ -40,7 +40,7 @@ def waterfall(shap_values, max_display=10, show=True):
     
     # Turn off interactive plot
     if show is False:
-        plt.ioff()
+        pl.ioff()
     
 
     base_values = shap_values.base_values


Demetrio92:
https://github.com/slundberg/shap/blob/26a4b21e70fbe19cf53e8e9f50d902b4899c1abb/shap/plots/_waterfall.py#L297  also needs the same fix

gcoter:
Here as well: https://github.com/slundberg/shap/blob/26a4b21e70fbe19cf53e8e9f50d902b4899c1abb/shap/plots/_waterfall.py#L337
https://github.com/slundberg/shap/blob/26a4b21e70fbe19cf53e8e9f50d902b4899c1abb/shap/plots/_waterfall.py#L592

slundberg:
Thanks for the fix! Sorry I was slow on the review (was out for a while after paternity leave). This got fixed in a larger change that just switched over the whole file to `plt`...so I decided to merge there. Thanks though!

ssrosa:
Thank you! Congratulations on the family enlargement

Steven Rosa
Product Owner, Machine Learning Engineering
FDG Analytics & Insights – Data Products
[A picture containing text  Description automatically generated]

From: Scott Lundberg ***@***.***>
Reply-To: slundberg/shap ***@***.***>
Date: Tuesday, March 22, 2022 at 5:59 PM
To: slundberg/shap ***@***.***>
Cc: Steven Rosa ***@***.***>, Author ***@***.***>
Subject: Re: [slundberg/shap] correct plt to pl (PR #2353)


Thanks for the fix! Sorry I was slow on the review (was out for a while after paternity leave). This got fixed in a larger change that just switched over the whole file to plt...so I decided to merge there. Thanks though!

PR #2354, opened 2022-01-11
add dependece_plot 'ymin' and 'ymax' args

jckkvs:
Added y_min and y_max arguments to dependence_plot.

When comparing between explanatory variables using dependence_plot, it is difficult to evaluate correctly if the y-axis range is different.
In most cases, the explanatory variables are standardized, so setting ymin(-0.75 ~ -1) and ymax(+0.75 ~ +1) is not a big problem. 

Case where ymin and ymax are not set.
Looks the same.

image

Case to set ymin and ymax (-0.75, 0.75).
The effect of explanatory variables becomes clear.
image


slundberg:
looks good thanks! Sorry I was slow on the review (was out for a while after paternity leave).

Issue #2368, opened 2022-01-27
explanation._old_format() with PermutationExplainer


ngupta20:
https://github.com/slundberg/shap/blob/46b3800b31df04745416da27c71b216f91d61775/shap/explainers/_permutation.py#L208

I believe this causes an error `AttributeError: 'Explanation' object has no attribute '_old_format'` no matter what.

It believe the fix is just:
`return explanation`

ngupta20:
#2369

fuxianlei:
> https://github.com/slundberg/shap/blob/46b3800b31df04745416da27c71b216f91d61775/shap/explainers/_permutation.py#L208
> 
> I believe this causes an error `AttributeError: 'Explanation' object has no attribute '_old_format'` no matter what.
> 
> It believe the fix is just: `return explanation`

Hi, thank you for providing the solution on this issue. 
However, may I check how do I remove ._old_format() from my side? Should I add in some functions in my code or change the version of shap?
Thank you in advance for your help.

ngupta20:
You're thinking along the correct lines @fuxianlei 

1. You can use the forked repo tagged above and build it as a dependency using your python environment provider.
2. If your design allows for it, you can also just set a dynamic override of the PermutationExplainer.shap_values method before it gets used in your code.

It seems this problem has persisted through many versions, so changing versions is probably not the way to go.

fuxianlei:
> You're thinking along the correct lines @fuxianlei
> 
> 1. You can use the forked repo tagged above and build it as a dependency using your python environment provider.
> 2. If your design allows for it, you can also just set a dynamic override of the PermutationExplainer.shap_values method before it gets used in your code.
> 
> It seems this problem has persisted through many versions, so changing versions is probably not the way to go.

Hi Gupta @ngupta20, 
Thank you very much for the suggestions. I managed to resolve this problem by override the PermutationExplainer.shap_values with a modified function

juancq:
Any progress on fixing this issue? 

I also resolved it in a hacky way, which is not ideal: 

Replace:
```
shap_values = explainer.shap_values(X)
```
With:
```
 shap_values = explainer(X, max_evals=10*X.shape[1], main_effects=False)
```
where 10 is the default value of the argument npermutations.  See [shap/shap/explainers/_permutation.py](https://github.com/slundberg/shap/blob/46b3800b31df04745416da27c71b216f91d61775/shap/explainers/_permutation.py#L208)

ngupta20:
[...]

PR #2369, opened 2022-01-27
Remove ._old_format() from PermutationsExplainer

ngupta20:
#2368


slundberg:
Thanks! And sorry for the slow review. (I was out after paternity leave for a while) This is not a recommended API anymore, so I added a depreciation noticed and updated the return value to match the old api (which is just the numpy array of values and not a full explanation object).

ngupta20:
Thanks @slundberg, that does it!

PR #2371, opened 2022-01-30
Removed duplicate entry of the Adult income dataset

adityasaini70:
There are two entries of the Adult income dataset in the API reference at https://shap.readthedocs.io/en/latest/api.html#shap.datasets.adult, which this PR reduces to one.


slundberg:
Thanks!

PR #2411, opened 2022-02-25
tentatively fix the text plot display bug

ycouble:
This PR tries to fix the bug described in this issue https://github.com/slundberg/shap/issues/2388


slundberg:
Thanks! I'll review this once the (now-fixed CI tests run again)

slundberg:
Looks good. I changed the code to just compute the text every time and then call display or not once at the end. Otherwise, the functionality should be the same. will merge once the tests pass again.

Thanks again and sorry for the slow review. (was out for a while after paternity leave)

PR #2428, opened 2022-03-12
Fix returning waterfall plot

jkarolczak:


slundberg:
Thanks! I think this might be a moot point given another pr that made all the plt and pl mentions consistent. But if this is still fixing a bug I am failing to see please let me know and I'll reopen it :)

PR #2444, opened 2022-03-19
Fixed Matplotlib.pyplot import as plt uniformly

manuelmathewsm:
`matplotlib.pyplot` has been imported as ´pl´ and has been used later on in four locations as ´plt´ and 59 locations as ´pl´. Fixed it to ´´plt´ on all locations.


slundberg:
Thanks!

manuelmathewsm:
You're welcome. Thank you for developing the package...

PR #2461, opened 2022-03-26
Explainer algorithm argument should not mention `kernel`, `sampling`, `deep`, or `gradient` in docstring as these values are not supported

navdeep-G:
* In `shap` version `0.40.0`, I get the following error when I pass in `kernel`, `sampling`, `deep`, or `gradient` as the algorithm to use in the Shapley calculation:

```
>               raise Exception("Unknown algorithm type passed: %s!" % algorithm)
E               Exception: Unknown algorithm type passed: {insert above mentioned algorithms}
```

* I checked the docs and noticed there was a mismatch between the implementation and documentation

### DIFF
--- a/shap/explainers/_explainer.py
+++ b/shap/explainers/_explainer.py
@@ -53,7 +53,7 @@ class Explainer(Serializable):
             units. For more details on how link functions work see any overview of link functions for generalized
             linear models.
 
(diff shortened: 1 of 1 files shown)


slundberg:
Thanks!

Issue #1268, opened 2020-06-14
KernelExplainer ValueError with inf


williamwongys:
In kernel.py, `solve()`, 

```python
eyAdj_aug = np.hstack((eyAdj, eyAdj - (self.link.f(self.fx[dim]) - self.link.f(self.fnull[dim]))))
```

when `self.fx[dim]` is 1
`self.link.f(1)` returns inf

resulting in the following error:

```
<class 'tuple'>: (<class 'ValueError'>, ValueError("Input contains NaN, infinity or a value too large for dtype('float64').",), <traceback object at 0x000001F116D80D48>)
```

`self.fx` is first set to [1,0] in explain():

```python
if self.keep_index:
            model_out = self.model.f(instance.convert_to_df())
else:
            model_out = self.model.f(instance.x) # returns [[1,0]]
self.fx = model_out[0] # stores [1,0]
```

karinazad:
I have the same issue for KernelExplainer and one model:

in `kernel.py` in `solve()` function (lines 227-233)

```
  if self.keep_index:
  model_out = self.model.f(instance.convert_to_df())   # model_out = [nan]
  else:
  model_out = self.model.f(instance.x) 
  self.fx = model_out[0] 
```

model_out is [nan] which results in the same error:
`ValueError("Input contains NaN, infinity or a value too large for dtype('float64')`

Issue #577, opened 2019-05-07
KernelExplainer for Multiclass XGBoost model - returns NaN Value Error


pogoetic:
Hi - I love SHAP! So I have read that TreeExplainer does not currently handle Mutliclass XGBoost models with a 'probability' output type but thats what I'm trying to generate, so i'm trying to use SHAP with KernelExplainer. I have 3 classes and my input data for SHAP is a 100 row sample with no inf, -inf, or NaN. Ideally, I would see the SHAP plots add up to a probability of 1.0. 

However, I get this error: 

`ValueError: Input contains NaN, infinity or a value too large for dtype('float64').`

Here is a code snippet i'm working with: 
```
import shap

## kernel shap sends data as numpy array which has no column names, so we fix it
def xgb_predict(data_asarray):
    data_asframe =  pd.DataFrame(data_asarray, columns=X_train.columns)
    return model.predict(data_asframe)

explainer = shap.KernelExplainer(xgb_predict, link='logit', data=X_test.iloc[:100])
shap_values = explainer.shap_values(X_test.iloc[1])
shap.initjs()
shap.force_plot(explainer.expected_value[0], shap_values[0][0], X_test.iloc[1])
```

What am I doing wrong here? Is it the same issue as with TreeExplainer where the logit transform is not yet implemented in SHAP? Thank you!

dp-dpo:
I get exactly the same error when using a KernelExplainer regardless of what link I use. Looks like something doesn't pass the infinity check in sklearn.
```
Traceback (most recent call last):
  File "C:\Users\user\AppData\Local\Continuum\anaconda3\envs\pycharm\lib\site-packages\IPython\core\interactiveshell.py", line 2963, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-707-1235e1045d56>", line 1, in <module>
    shap_values = explainer.shap_values(train_shap)
  File "C:\Users\user\AppData\Local\Continuum\anaconda3\envs\pycharm\lib\site-packages\shap\explainers\kernel.py", line 209, in shap_values
    explanations.append(self.explain(data, **kwargs))
  File "C:\Users\user\AppData\Local\Continuum\anaconda3\envs\pycharm\lib\site-packages\shap\explainers\kernel.py", line 405, in explain
    vphi, vphi_var = self.solve(self.nsamples / self.max_samples, d)
[...]

PR #2473, opened 2022-04-02
Fix shrinking to -inf and blowing up to inf of mask_aug and eyAdj_aug of KernelExplainer

aaronwtr:
This behavior occurs when the dimensionality of the feature vectors being fed into the KernelExplainer is very large. L1 regularization consequently runs into issues when trying to convert the -inf values to floats. This fixes the issue I ran into myself as well as the following issues most probably: #577 and #1268 .

### DIFF
--- a/shap/explainers/_kernel.py
+++ b/shap/explainers/_kernel.py
@@ -550,6 +550,10 @@ class Kernel(Explainer):
             eyAdj_aug = np.hstack((eyAdj, eyAdj - (self.link.f(self.fx[dim]) - self.link.f(self.fnull[dim]))))
             eyAdj_aug *= w_sqrt_aug
             mask_aug = np.transpose(w_sqrt_aug * np.transpose(np.vstack((self.maskMatrix, self.maskMatrix - 1))))
+            mask_aug[np.isinf(mask_aug)] = 0
+            mask_aug[np.isneginf(mask_aug)] = 0
+            eyAdj_aug[np.isinf(eyAdj_aug)] = 0
+            eyAdj_aug[np.isneginf(eyAdj_aug)] = 0
             #var_norms = np.array([np.linalg.norm(mask_aug[:, i]) for i in range(mask_aug.shape[1])])
 
             # select a fixed number of top features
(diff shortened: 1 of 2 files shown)


slundberg:
Thanks. Can you give a bit more detail on why setting to zero is the right answer here? I can also walk back through the method, but thought I would start by asking your rational :)

aaronwtr:
Hi @slundberg, my reasoning was more practical than based on solid theory but I can share some of my thinking so that we hopefully will be able to figure out what's causing this. 

What I did was trace back to what features the -infs I was encountering pointed and I noticed that these should not be that important for the prediction I was trying to explain. So I wanted to see whether mapping those particular infs to zero would result in an explanation that still did prioritize features important to the prediction and I found that it did. I stopped debugging there since this was sufficient to serve my purpose of using the KernelExplainer but it still feels a bit unsatisfying not understanding what was causing this. 

Some additional notables that might be useful: the feature vectors I'm explaining are pretty large (~3,000 features) and to obtain robust Shapley value approximations I have found that I needed to crank up the nsamples hyperparameter quite a bit (currently at 75,000). My intuition was that the infinities occurred because of one of these two facts but I was not able to identify precisely where and how this would be happening besides finding back the -infs in mask_aug and eyAdj_aug.

Looking forward to your take on why this might be happening.

slundberg:
Thanks! You might also try using the Permutation explainer since it has been optimized more and might scale better.

I could not reproduce the Inf problems (see the code I tried with below). Perhaps it is because I don't have enough RAM to do 75k samples.

```
import xgboost
import shap
import numpy as np

### train XGBoost model
X = np.random.randn(100, 3000)
y = np.random.randn(100)
model = xgboost.XGBRegressor().fit(X, y)

### compute SHAP values
shap_values = explainer.shap_values(X[:2], nsamples=15000)
```

aaronwtr:
I was searching myself for particular samples in my dataset where this occurred and couldn't immediately find it as well. I remember that a while back I was trying to obtain a global explanation (explanation dataset of 1000 samples) and then at some sample the process halted with the Inf error. So this issue seems to occur quite sparsely in a very particular scenario in my case. If I run into the issue again down the line I'll report back here. 

Thanks for all your work on this module, it has been crucial to my Master's thesis and I'm sure it is for the ML research community in general as well.

PR #2483, opened 2022-04-06
Improving summary plot by dynamically adjusting the plot's subparams

aaronwtr:
Adding matplotlib's pl.tight_layout() method makes sure that the image gets rendered properly. This is particularly useful when your features have long names.


slundberg:
Thanks!

PR #2489, opened 2022-04-08
Update deep_tf.py

Ashishbodla:


slundberg:
Thanks! I think that whole chunk needs to just go, it was not functional

PR #2492, opened 2022-04-08
Fix cohort plot error

joshzwiebel:
fix an error caused in the cohort bar plot stemming from ambiguous Boolean definition of array
Issue described in forum below.
https://github.com/slundberg/shap/issues/2406

This is my first contribution to the library so please let me know if I can improve this pull request!


slundberg:
Thanks! I updated the code is use `is` instead of np.any, and will merge when tests pass.

PR #2500, opened 2022-04-14
Add seed parameter to get reproducible results (default None)

GiovannaNicora:
add the possibility to set a seed when using Explainer to get reproducible shap values


slundberg:
Thanks!

PR #2507, opened 2022-04-20
Fixes type check: adds np.ndarray to list check.

songololo:
An error is currently raised when passing `feature_names` as a `numpy` array.

Adds `np.ndarray` as valid type.


slundberg:
Thanks!

PR #2527, opened 2022-04-30
Warning/throw if matplotlib is not available

klieret:
So far, no meaningful warnings were issued for the following:

## 1. Importing shap and directly accessing plots submodule

The following will throw a ModuleNotFound error about the `plots` submodule if `matplotlib` is not available and work fine otherwise

```
import shap
shap.plots 
```

This PR substitutes a `plots` object that will throw warnings once attributes are accessed, if `matplotlib` is unavailable.

## 2. Directly importing shap.plots 

The following will throw a rather cryptic exception rather than stating that matplotlib isn't available

```python3
import shap.plots 
```

This PR simply tries to import `matplotlib` as soon as `plots` is imported and throws if the former isn't available.

## Impact

The missing warnings might be (partially) responsible for confusion in https://github.com/slundberg/shap/issues/1975


slundberg:
Thanks! This looks good, and I'll merge once the tests pass. (and sorry for the slow review, was stuck out away for a while)

PR #2548, opened 2022-05-15
Cast second argument of round() to int for compatibility with Numpy version < 1.19

zeshengli:
Got the following exception when running shap analysis: 
```
  File "nlp_disaster_classification.py", line 78, in <module>
    explain = explainer.explain(
  File "/home/zeshengl/tryout-ai/tryout/src/tryout/explainer/explainer.py", line 150, in explain
    r = m.explain(e, data=data, real_label=label, gui=gui)
  File "/home/zeshengl/tryout-ai/tryout/src/tryout/nlp/classifier.py", line 416, in explain
    vis_res = explainer._visualize_text_explaination(
  File "/home/zeshengl/tryout-ai/tryout/src/tryout/explainer/explainer.py", line 136, in _visualize_text_explaination
    html = shap.plots.text(shap_vis_input, display=False)
  File "/home/zeshengl/miniconda3/envs/tryout/lib/python3.8/site-packages/shap/plots/_text.py", line 262, in text
    out += svg_force_plot(values, shap_values.base_values, shap_values.base_values + values.sum(), encoded_tokens, uuid, xmin, xmax, output_name)
  File "/home/zeshengl/miniconda3/envs/tryout/lib/python3.8/site-packages/shap/plots/_text.py", line 452, in svg_force_plot
    xcenter = round((xmax + xmin) / 2, round(1-np.log10(xmax - xmin + 1e-8)))
TypeError: integer argument expected, got float
```

It was caused by the second argument passed to the outer `round()` function not being an int. According to NumPy's release note here: https://numpy.org/doc/stable/release/1.19.0-notes.html#change-output-of-round-on-scalars-to-be-consistent-with-python, the output of `round()` when the input is a Numpy object is not guaranteed to be an int. In this case, we called `round()` on a numpy.float object and the return type is still numpy.float. When we pass this object as the second argument to the outer `round()` function call, the exception is thrown. 

So to be compatible with an earlier version of NumPy and extra safe, cast the result of the inner `round()` to int before passing it as the second argument to the outer `round()` call.


slundberg:
Thanks! Will merge when the tests finish.

PR #2555, opened 2022-05-18
fix heatmap and HistGradientBoostingClassifier tree tests

imatiach-msft:
1.) Fix failing gated build on the heatmap tests due to changes in the validation images
2.) Fix failing HistGradientBoostingClassifier tree test due to new scikit-learn release 1.1.0 for python 3.8 only.  Now even for binary classification case self.base_offset is a numpy array, but it only contains a single value.


slundberg:
Thanks! This LGTM

PR #2569, opened 2022-05-31
Bump add numpy 1.22 support since numba 0.55.2  release

q-aaronzolnailucas:


imatiach-msft:
I think this is related to https://github.com/slundberg/shap/pull/2573 .  My thinking is that we shouldn't pin numpy or numba, but instead the user should install compatible versions, since numba follows some specific ranges of numpy.

rmehyde:
Thanks @imatiach-msft. Are you proposing dropping the numpy version requirement altogether? i.e. changes lines 227-228 to
```python
        setup_requires=['numpy'],
        install_requires=['numpy', 'scipy', 'scikit-learn', 'pandas', 'tqdm>4.25.0',
```

Happy to push up a PR if so

rmehyde:
Thanks @imatiach-msft. Are you proposing dropping the numpy version requirement altogether? i.e. changes lines 227-228 to
```python
        setup_requires=['numpy'],
        install_requires=['numpy, 'scipy', 'scikit-learn', 'pandas', 'tqdm>4.25.0',
```

Happy to push up a PR if so

imatiach-msft:
If the only reason numpy is pinned is for numba, then I don't think it makes sense to pin numpy. If there is a different reason for why numpy was pinned, then I think it's fine to keep the pin.  However, numpy and numba should be pinned by the upstream package that depends on shap and I would argue that they should not be pinned by shap itself.

rmehyde:
Thanks @imatiach-msft, I've created #2582 to remove versioning from the numpy dependency.

slundberg:
Thanks all! I agree with @imatiach-msft, this was just a temporary patch to get around a numba/numpy conflict that was not really SHAP's responsibility. I'll close this PR and merge the version less one.

PR #2573, opened 2022-06-01
Update numba version

bchen1116:
With `shap==0.40.0`, various versions of `numba` installations give errors during or after installation:

`numba==0.1` gives this when trying to import:
Pasted Graphic 12

`numba<0.45.0` gives this when trying to install:
Pasted Graphic 13

`numba<=0.48.0` has a successful installation and but failed import:
Pasted Graphic 14
image

`numba>=0.49.0` works successfully!
image

Updating the `numba` allowed versions to avoid these failures
These were all run with py3.8 in a fresh environment with `shap==0.40.0`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -226,7 +226,7 @@ def run_setup(with_binary, test_xgboost, test_lightgbm, test_catboost, test_spar
         cmdclass={'build_ext': build_ext},
         setup_requires=['numpy<1.22'], # <1.22 is because of numba
         install_requires=['numpy<1.22', 'scipy', 'scikit-learn', 'pandas', 'tqdm>4.25.0', # numpy versions are for numba
-                          'packaging>20.9', 'slicer==0.0.7', 'numba>=0.45.0', 'cloudpickle'],
+                          'packaging>20.9', 'slicer==0.0.7', 'numba>=0.46.0', 'cloudpickle'],
         extras_require=extras_require,
         ext_modules=ext_modules,
         classifiers=[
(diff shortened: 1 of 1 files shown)


imatiach-msft:
@bchen1116 numba and numpy have to be updated in sync together, so I think this is a weird change.  For newer versions of numpy, newer versions of numba need to be used.  Similarly, for older versions of numpy, older versions of numba need to be used.  I think pinning it like this (and also pinning numpy) is the wrong way, especially with such a high pin, it will mean that projects that depend on older numpy won't be able to use the latest shap.  Instead, upstream projects that use shap should pin numpy and numba together to compatible versions, and shap should try to minimize pinning as much as possible.

imatiach-msft:
Also, notice how in your successful install you have numpy 1.22.4 installed, but actually in the shap setup.py numpy is pinned to <1.22!  So now you are changing shap to pin to two incompatible ranges of versions of numpy and numba.  I think the better fix is to remove both numpy and numba pins in shap, and the person who installs shap needs to install the right compatible versions, otherwise with such incompatible versions specified this will turn into [dependency hell](https://en.wikipedia.org/wiki/Dependency_hell).

image

imatiach-msft (changes requested):
the pinned numba and numpy versions are incompatible now, numpy<1.22 is incompatible with numba>=0.49.0

bchen1116:
@imatiach-msft I see what you mean, thanks for the response and input. Closing this out

PR #2575, opened 2022-06-03
Added cmap parameter to image_plot

tlabarta:
Goal is to make adaptations of shap_value colors possible. Depending on the use case, it can be necessary to adjust the red_transparent_blue color scheme to a different one. **Therefore I added cmap as input parameter and kept the current colormap as default.**

As input parameter, either predefined colormaps from matplotlib can be used (see here https://matplotlib.org/3.5.0/tutorials/colors/colormaps.html) by specifying a string, e.g. cmap='PuOr', or self defined colormaps can be added.

**This pull request would close issue #2576** 

PS: This is my first pull request ever, so if something is missing or unclear I would be more than happy about your feedback! :-)

### DIFF
--- a/shap/plots/_image.py
+++ b/shap/plots/_image.py
@@ -5,6 +5,7 @@ import json
 from typing import Optional
 
 import numpy as np
+from matplotlib.colors import Colormap
 
 from shap._explanation import Explanation
 from ..utils import ordinal_str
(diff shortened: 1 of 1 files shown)


slundberg:
Looks good! I move the arg later so as to not change the positional argument order for anyone who might depend on it. Once the tests pass we should be good to merge.

PR #2580, opened 2022-06-06
Refactoring Exceptions

alexisdrakopoulos:
SHAP currently calls [Exception](https://docs.python.org/3/library/exceptions.html#Exception) a lot throughout the code base. This makes it tough to handle exceptions internally as one has to match to messages to infer why an exception was raised.

For example; I want to try to build an explainer object and raise an error in cases where a tree explainer (or similar) won't work:
```python
try:
    explainer = shap.Explainer(mod)
except Exception as e:
    if "The passed model is not callable and cannot be analyzed directly with the given masker!" in str(e):
        # do something
    else:
        raise e
```
Since I want to avoid catching all the other Exceptions that might be raised I have to match against the message. This is brittle and ugly in my opinion, I would much rather have something like:
```python
try:
    explainer = shap.Explainer(mod)
except InvalidMaskerError:
    # do something
```

I am opening this draft PR which replaces some Exception calls with "more" appropriate ones including custom ones. As to not waste time in case this is not desired I wrote this very quickly and hence I'm sure some exceptions aren't the ideal ones to raise.

What do we think of doing it this way? If there's a positive reception to the idea I'll do it properly and open this as a PR.


slundberg:
Thanks @alexisdrakopoulos, I like the idea! I agree that handling exceptions via string matching is not ideal, and there is clear room to improvement here. What you have looks like the right direction, so go for it. Once you have a PR ready I expect it will be a simple merge.

alexisdrakopoulos:
There are three tests for which I need to update the exception catching.

alexisdrakopoulos:
@slundberg can we try running tests?

shap/explainers/_explainer.py
```diff
@@ -1,4 +1,5 @@
 import copy
+from sqlite3 import NotSupportedError
```
slundberg: is this needed?

slundberg:
Just looked it over and it all looks good to me, except for one line I commented on. After you get a chance to look at that I think this is ready to merge. Thanks!

shap/explainers/_explainer.py
alexisdrakopoulos: Accidental import! pesky auto importers :) good catch

slundberg (approved):

Issue #2571, opened 2022-05-31
[Action Requested] Erroneous dependency on Numpy<1.22 for Numba when Numba now allows Numpy==1.22


ikekilinc:
The current version of `master` includes a dependency on `Numpy<1.22` since the master branch of Numba used to allow for any Numpy version, while being untested and uncompiled into PyPi wheels for all versions>=1.22 and even breaking under certain edge cases for those versions.

[https://github.com/slundberg/shap/blob/690245c6ab043edf40cfce3d8438a62e29ab599f/setup.py#L227](https://github.com/slundberg/shap/blob/690245c6ab043edf40cfce3d8438a62e29ab599f/setup.py#L227)

While this decision made sense at the time, [Numba has recently published a new release (v0.55.2) that now works with `Numpy==1.22`.](https://github.com/numba/numba/issues/8025) However, the current version of SHAP has not yet been updated to allow for this version upgrade.

Could we please update `setup.py` on master to now accept `Numpy==1.22`?

q-aaronzolnailucas:
Made a PR for this: #2569

PR #2582, opened 2022-06-06
Remove restriction on numpy version

rmehyde:
This PR removes the restrictions on Numpy versions. Per discussions in PRs #2573 and #2569 as well as issue #2571, numba and numpy dependency synchronization should come from upstream rather than this library.

@imatiach-msft please let me know whether this meets your expectations and approve test workflows if so. Thanks!


imatiach-msft (approved):
LGTM!

rmehyde:
Awesome, thanks so much @imatiach-msft! Please let me know if there are any further steps needed from me for merge and release.

Not sure if there's a regular process but it looks like the master branch is continuously updated and releases are cut from it every few months? I'd love to be able to use the newer numpy and numba versions soon. Thanks again!

imatiach-msft:
@rmehyde yes, @slundberg is the author of shap and will approve and merge the PR when ready, but the PR looks good to me.  Yes, shap package is released every couple of months to pypi.

rmehyde:
Makes sense, thanks again for all the help @imatiach-msft!

slundberg:
Thanks @rmehyde! This was a hack because at one point without it a simple pipit install would fail. Happy to remove it now once the tests finish.

PR #2586, opened 2022-06-09
Improve kernel explainer not_equal method

WeichenXu123:
Improve Kernel explainer not_equal method:

The method might raise error on some types such as `timestamp` type.
I improve it in this PR to make it support all kinds of input data types.


WeichenXu123:
Gentle ping @slundberg Thanks!

shap/explainers/_kernel.py
```diff
@@ -397,9 +397,12 @@ def explain(self, incoming_instance, **kwargs):
...
     def not_equal(i, j):
-        if isinstance(i, str) or isinstance(j, str):
-            return 0 if i == j else 1
-        return 0 if np.isclose(i, j, equal_nan=True) else 1
+        try:
+            if isinstance(i, str) or isinstance(j, str):
+                return 0 if i == j else 1
+            return 0 if np.isclose(i, j, equal_nan=True) else 1
+        except Exception:
+            return int(i != j)
```
imatiach-msft: Shouldn't you have an explicit case for timestamp types instead of a try/except?  eg check isinstance(i, datetime) or similar for timestamp and then return i != j?
Also, can you add a test for this?

imatiach-msft:
@WeichenXu123 I approved you to run tests, started the tests and added a comment.  Can you add a test for this case at https://github.com/slundberg/shap/blob/master/tests/explainers/test_kernel.py?

shap/explainers/_kernel.py
WeichenXu123: @imatiach-msft I would like to support all kinds of types besides the `np.isclose` supported types and string type.
I can improve code to remove try...except.

imatiach-msft: this looks much better!

imatiach-msft (approved):

WeichenXu123:
@imatiach-msft @slundberg Could you help approve the CI run and then merge the PR when CI passed ? Thanks!

slundberg:
Thanks! I just approved the run.

PR #2595, opened 2022-06-16
remove unused imports

alexisdrakopoulos:
There are a few unused imports (though it is complicated as some seem unused but are used):
I ran the following to find them:
```
flake8 --select F401
```

Note to try to save time I ran [autoflake](https://github.com/PyCQA/autoflake) which also removes some unused variables. I'm not sure how dangerous this is!


thatlittleboy:

Issue #2528, opened 2022-04-30
Future Warning from sklearn for KernelExplainer concerning LassoLarsIC


c-feldmann:
Calling the method `shap_values` from the  `KernelExplainer` gives the following warning:
```
The default of 'normalize' will be set to False in version 1.2 and deprecated in version 1.4.

If you wish to scale the data, use Pipeline with a StandardScaler in a preprocessing stage. To reproduce the previous behavior:

from sklearn.pipeline import make_pipeline

model = make_pipeline(StandardScaler(with_mean=False), LassoLarsIC())

If you wish to pass a sample_weight parameter, you need to pass it as a fit parameter to each step of the pipeline as follows:

kwargs = {s[0] + '__sample_weight': sample_weight for s in model.steps}
model.fit(X, y, **kwargs)

Set parameter alpha to: original_alpha * np.sqrt(n_samples). 
```
Versions:
python: 3.9.12 (conda)
sklearn: 1.0.1
shap: 0.40.0
OS: Ubuntu 18.04.1 LTS

Aboubacar-cmr:
Oui du coup quelqu'un sais c'est quoi la solution?

c-feldmann:
@Aboubacar-cmr is this translation correct?
"Yes, so does anyone know what the solution is?"
Then the answer is, at least from my site: No, i simply silenced the warning:

```python
with warnings.catch_warnings():
    warnings.filterwarnings("ignore")
    shap_values = model_explainer.shap_values(X_train)
```

Aboubacar-cmr:
@c-feldmann merci beaucoup.
Il n'existe pas d'autre solution sans pour au tant ignoré l'erreur?

c-feldmann:
@Aboubacar-cmr 
Could you phrase your questions in English, please?
Assumed translation: "Is there no other solution without ignoring the error?"

As I stated before, no, there is no solution I'm aware of.

Aboubacar-cmr:
Okay, Thanks @c-feldmann

alvaromiguelo:
Define the parameter normalize to False and you should get rid of the warning.

courtarro:
> Define the parameter normalize to False and you should get rid of the warning.

Where?

PR #2641, opened 2022-08-03
Fix Issue #2528

from-nowhere:
Fix a deprecation warning in `sklearn.linear_model.LassoLarsIC`.

The deprecation concerns the `normalize` keyword, which is currently equal to `True` by default. In order to replicate the current behaviour _after_ deprecation, we follow sklearn's [documentation on LassoLarsIC](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LassoLarsIC.html) and do the following:

1. Take care of normalization with a `sklearn.preprocessing.StandardScaler`
2. Explicitly pass `normalize=False` to the `LassoLarsIC`

From the resulting pipeline, the Lasso object can be recovered with `model[1]`.

### DIFF
--- a/shap/explainers/_kernel.py
+++ b/shap/explainers/_kernel.py
@@ -12,6 +12,8 @@ import itertools
 import warnings
 import gc
 from sklearn.linear_model import LassoLarsIC, Lasso, lars_path
+from sklearn.pipeline import make_pipeline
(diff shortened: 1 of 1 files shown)


TomPham97 (approved):

thatlittleboy:

PR #2649, opened 2022-08-18
Fixed IPython deprecation warning

boomanaiden154:
When using IPython versions greater than 7.14, there is a deprecation warning thrown indicating that all imports from `IPython.core.display` should be switched to `IPython.display`. This PR makes that fix.


thatlittleboy:

PR #2655, opened 2022-08-21
Fix Install dependencies stage in tests pipeline

alexander-pv:
Hi, I noticed that the test pipleine fails before running pytest because of dependency issues.
This PR is aimed to fix dependency conflicts for all python versions in `run_tests.yml`.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -25,6 +25,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
+          python -m pip install -U setuptools
           python setup.py develop
           pip install .[test]
           pip install tensorflow xgboost matplotlib ipython codecov torch torchvision pylint sentencepiece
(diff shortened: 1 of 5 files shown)


alexander-pv:
It looks like `test_heatmap.py` contains outdated tests. I think that it may be also a dependency issue 🤔.

thatlittleboy:
Hi @alexander-pv , I'm currently maintaining a fork over at dsgibbons/shap, and making a similar change to `setup.py` at the moment (dsgibbons#32). Just curious, why you went for manually specifying the numpy versions rather than relying on `oldest-supported-numpy` in the build dependency? (if you can still recall the reason why 😅)

alexander-pv:
Hi, @thatlittleboy, as far as I remember there was an issue in the "Install dependencies" step in tests. I think I got stuck with it at that time and for some reason I couldn’t fix it with `oldest-supported-numpy`.
Thanks for your work! That's great to see that this `shap` fork is actively maintained!  I will be glad to contribute at least with further `causalml` library support.

thatlittleboy:
@alexander-pv  Awesome, contributions very welcome. https://github.com/dsgibbons/shap
There's still plenty to do, but I'm positive we'll get there.

thatlittleboy:

PR #2697, opened 2022-09-21
Fix shap.summary_plot to work with matplotlib 3.6.0

### LINKED ISSUE #2687: summary plot not working with matplotlib v3.6.0
shap v0.41.0 is throwing an error when used in conjunction with matplotlib v3.6.0. It works with versions v3.5.3 and below.

image
image

Am not sure how to resolve the issue other than fixing the older version.

jklaise:
Fixes #2687. Have tested this manually by calling `shap.summary_plot(...)` to produce expected behaviour.

Note that there may be additional places where this is needed, e.g. for other plots, here's a grep of all the occurences of `pl.colorbar` in the codebase:

```
plots/_decision.py
151:        # place the colorbar
154:        cb = pl.colorbar(m, ticks=[0, 1], orientation="horizontal", cax=ax_cb)

plots/_monitoring.py
78:    cb = pl.colorbar()

plots/_embedding.py
68:    cb = pl.colorbar()

plots/_image.py
169:    cb = fig.colorbar(im, ax=np.ravel(axes).tolist(), label="SHAP value", orientation="horizontal",

plots/_heatmap.py
118:    #pl.colorbar()
137:        cb = pl.colorbar(m, ticks=[min(vmin,-vmax), max(-vmin,vmax)], aspect=1000, fraction=0.0090, pad=0.10,

plots/_violin.py
460:        cb = pl.colorbar(m, ticks=[0, 1], aspect=1000)

plots/_beeswarm.py
365:        cb = pl.colorbar(m, ax=pl.gca(), ticks=[0, 1], aspect=80)
865:        cb = pl.colorbar(m, ax=pl.gca(), ticks=[0, 1], aspect=80)

plots/_scatter.py
346:            cb = pl.colorbar(p, ticks=tick_positions, ax=ax, aspect=80)
349:            cb = pl.colorbar(p, ax=ax, aspect=80)
700:            cb = pl.colorbar(p, ticks=tick_positions, ax=ax, aspect=80)
703:            cb = pl.colorbar(p, ax=ax, aspect=80)

```


connortann:
Thank you for the PR!

May I suggest also adding the same fix to a few other places in `shap/plots`:
- `_violin.py:L310`
- `_monitoring.py:L73`
- `_embedding.py:L63`

I think the other uses of colorbar are ok, as they already provide the `ax` or `cax` argument.

connortann:
PS - it would be great to validate the fix with the unit tests on CI with the problematic version. I'd suggest pushing a commit to temporarily pin matplotlib in pyproject.toml line 39:

```toml
[project.optional-dependencies]
plots = ["matplotlib==3.6.0", "ipython"]
```

(Nb we can ignore the test on python 3.7, which isn't compatible with matplotlib 3.6.0).

Once the other tests pass, we can then remove the version pin.

jklaise:
FYI `3.6.1` removes the error and makes it a deprecation warning: https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.6.1.html (however, I have not been able to see the deprecation warning they mention).

That being said, it's still worth adding explicit axes for people running `3.6.0` and also if the feature does get deprecated eventually.

Btw with the explicit axes we still get a warning `No data for colormapping provided via 'c'. Parameters 'vmin', 'vmax' will be ignored`, not sure at the moment if this is something we can ignore.

connortann:
I _think_ that the `No data for colormapping provided` warning is unrelated to the colobar issue. In any case, I'd like to poractively address all of the warnings in the test suite, before any of them become breaking changes. I created a separate issue to track them on #3066 , so we can tick them off over time.

jklaise:
@connortann `embedding_plot` doesn't raise errors due to the colorbar, I think it's because it's called without arguments whilst in `summary_plot` the colorbar is created from a colormap.

The same should apply for `monitoring_plot`, but it seems this method doesn't actually work (and there are no uses of it in the docs or tests).

connortann:
That's rather concerning, that we have no tests for `monitoring_plot`! Well, let's leave that for another day then.

If you'd kindly remove the temporary version pin, I think this PR is good to go!

connortann (approved):

kabartay:
Downgrading to `0.41.0` leads to this issue: https://github.com/shap/shap/issues/2909

PR #2769, opened 2022-11-17
Update _kernel.py

11301858:
Replacing np.int, which is deprecated, with int, as encouraged by the Python warning message.


thatlittleboy:

PR #2781, opened 2022-11-30
fix background to avoid missed leaves

slundberg:
Was causing some random unit test failures.


thatlittleboy:

PR #2792, opened 2022-12-10
Update models.py

nagarindkx:
keras is no longer supported, please use tf.keras instead. Your TensorFlow version is newer than 2.4.0 and so graph support has been removed in eager mode and some static graphs may not be supported. See PR #1483 for discussion.


thatlittleboy:

PR #2799, opened 2022-12-13
add brackets to the shape argument of reshape when calling as np.resh…

SomeUserName1:
…ape(a, (x,y)).

When calling `np.reshape(a, shape)`, the shape argument needs to be in brackets, otherwise the second dimension of the shape argument is interpreted as thrid argument which is actually order (i.e. specifying how the array is indexed).

```
File "/opt/miniconda3/envs/xai/lib/python3.10/site-packages/shap/plots/_scatter.py", line 559, in dependence_legacy
    shap_values = np.reshape(shap_values, len(shap_values), 1)         # here we are missing brackets
  File "<__array_function__ internals>", line 180, in reshape
  File "/opt/miniconda3/envs/xai/lib/python3.10/site-packages/numpy/core/fromnumeric.py", line 298, in reshape
    return _wrapfunc(a, 'reshape', newshape, order=order)                # Causing newshape to be a single int and the second dimension gets interpreted as order
  File "/opt/miniconda3/envs/xai/lib/python3.10/site-packages/numpy/core/fromnumeric.py", line 54, in _wrapfunc
    return _wrapit(obj, method, *args, **kwds)
  File "/opt/miniconda3/envs/xai/lib/python3.10/site-packages/numpy/core/fromnumeric.py", line 43, in _wrapit
    result = getattr(asarray(obj), method)(*args, **kwds)
TypeError: order must be str, not int                                                 # Causing this error

```

The fix is to add brackets:
shap_values = np.reshape(shap_values, (len(shap_values), 1))


connortann (approved):
LGTM!

PR #2800, opened 2022-12-13
move shap_values() to __call__ to comply to the usual explainer API

SomeUserName1:
This just moves the code from shap_values(self, X, **kwargs) to a new method __call__(self, X, **kwargs) and calls __call__ from shap_values such that the class is usable as the usual Explainers in order to keep the inheritance structure usable (i.e. to treat all explainers the same without ifs for special treatment).

### DIFF
--- a/shap/explainers/_kernel.py
+++ b/shap/explainers/_kernel.py
@@ -105,8 +105,7 @@ class Kernel(Explainer):
         else:
             self.D = self.fnull.shape[0]
 
-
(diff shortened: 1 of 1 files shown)


SomeUserName1:
Probably this is unneccessary as KernelExplainer is not yet bound into the Explainer class as a case in the selection part of the code. API compliance will probably be enforced when everything is integrated.

Would also be nice, if you could allow the model to fit pandas DataFrames.

PR #2812, opened 2022-12-20
Pytorch DeepExplainer register_full_backward_hook fix

helgehr:
I got the following warning for computing the `shap_values` with a pytorch model:
```
Using a non-full backward hook when the forward contains multiple autograd Nodes is deprecated and will be removed in future versions. This hook will be missing some grad_input. Please use register_full_backward_hook to get the documented behavior.
```
Changing `register_backward_hook` to the `register_full_backward_hook` method fixed this warning and should now receive all `grad_input`.


connortann:
This was subsequently implemented on #3259

PR #2830, opened 2023-01-11
Replace warnings.DeprecationWarning with just DeprecationWarning

colorlace:
As per [this](https://github.com/slundberg/shap/issues/2668) issue, an error is thrown:

`AttributeError: module 'warnings' has no attribute 'DeprecationWarning'`

because `DeprecationWarning` should be referenced as a standalone class rather than as `warnings.DeprecationWarning`.


thatlittleboy:

PR #2839, opened 2023-01-20
Correct xlim in force_matplotlib in cases where the signs of force are all the same

zaburo-ch:
For `shap.force_plot(matplotlib=True)`, if we specify `shap_values` that all have the same sign, the xlim looks wrong. This PR fixes that.

## Minimal reproducible example

```python
import shap
import numpy as np

np.random.seed(0)
base = 100
contribution = np.r_[-np.random.rand(5)]
names = [f"minus_{i}" for i in range(5)]

shap.force_plot(
    base,
    contribution,
    names,
    matplotlib=True,
    show=False,
)
plt.show()

np.random.seed(0)
base = 100
contribution = np.r_[np.random.rand(5)]
names = [f"plus_{i}" for i in range(5)]

shap.force_plot(
    base,
    contribution,
    names,
    matplotlib=True,
    show=False,
)
plt.show()
```

image
image

## Results after correction

image
image


thatlittleboy:
Hi there @zaburo-ch , would you still be interested in following up with this PR? I think we're pretty much there :)

We recently merged many bug fixes, warnings and test fixes from our community fork (see #2944 ).
Can you 1) rebase your changes on top of the latest master, and 2) add tests for this particular case so we don't accidentally regress this behaviour in the future. The test case in your PR description would suffice.

dsgibbons:
I've added some tests. I couldn't replicate the image from @zaburo-ch, but it is still an improvement from before.

thatlittleboy:
> I couldn't replicate the image from @zaburo-ch

@dsgibbons yeah , the bbox seems to be wrong when using the Agg/osx matplotlib backend. Running the same code on Jupyter notebook shows the correct expected output.

image

dsgibbons:
Is this OK to merge @thatlittleboy?

dsgibbons:
@connortann can I get a quick review on this? It would be good to tick this off the roadmap.

connortann (approved):
LGTM!

connortann:
@dsgibbons it looks like the tests are failing due to an issue accessing the tiny translation model from hugging face that we use for integration tests. I'm traveling at the moment - would you be able to take a look?

This PR looks good in itself, so I've approved it ready for when we fix this latest CI issue.

dsgibbons:
@connortann apologies I merged this before I saw your follow-up message. Yes, I can take a look at the CI issue tomorrow.

PR #2848, opened 2023-01-30
ENH: introduce vmax parameter in image plot

sd3ntato:
To keep scale of color-coding consistent during multiple explanations on the same image.

For example, I have been working on an app where you compute explanation for the same image at different times during training, to see how feature importance changes during training (in a particular setting where this make sense: Continual Learning. I am visualizing catastrophic forgetting) But if scale of the color-coding changes at every plot, the sequence of explanation doen't make sense anymore (visually)


connortann:
Thanks for the PR! Please would you add an entry to the docstring to describe the new parameter?

Apologies that no-one has reviewed until now, as we've been working through a big backlog of issues.

connortann (approved):
Looks great, thanks for the PR!

Apologies for the long delay in merging this PR; we've been steadily working through a number of issues relating to the plots and are very short of maintainers. So, your PR is highly appreciated.

I fixed up some merge confliects, and whilst I was at it I renamed the parameter to `vmax` for consistency with `plt.imshow`, and added an entry into the docstring.

Issue #2844, opened 2023-01-28
Replace references to deprecated Boston housing prices In ReadMe


LipeiranNJU:
boston housing prices in ReadMe has replaced by California housing Price. But in readme document, with explaining tree model, the sample also uses boston housing price. I think it may be confused for some greeners when use it.

thatlittleboy:
Closing this as duplicate of #2322

Issue #2322, opened 2021-12-14
Replace references to deprecated Boston housing prices dataset with another dataset


skylerwharton:
As noted in [sklearn's documenation](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_boston.html), the Boston housing dataset is being deprecated due to a significant ethical concern: 
>**Warning:** The Boston housing prices dataset has an ethical problem: as investigated in [[1]](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_boston.html#rec2f484fdebe-1), the authors of this dataset engineered a non-invertible variable “B” assuming that racial self-segregation had a positive impact on house prices [[2]](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_boston.html#rec2f484fdebe-2). Furthermore the goal of the research that led to the creation of this dataset was to study the impact of air quality but it did not give adequate demonstration of the validity of this assumption.
>
>The scikit-learn maintainers therefore strongly discourage the use of this dataset unless the purpose of the code is to study and educate about ethical issues in data science and machine learning.

Could uses of this dataset within the `shap` documentation (for example, on the ["An introduction to explainable AI with Shapley values" page](https://shap.readthedocs.io/en/latest/example_notebooks/overviews/An%20introduction%20to%20explainable%20AI%20with%20Shapley%20values.html#An-introduction-to-explainable-AI-with-Shapley-values)) be replaced with uses of another dataset? For example, sklearn suggests that either the California housing dataset or the Ames housing dataset are reasonable alternative datasets [[source](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_boston.html)].

Thank you!

shanewalsh1123:
I can have a look at this.

znacer:
To help deal with this issue, i listed files where "shap.datasets.boston" occurs : 
- notebooks:
  - [x] notebooks/tabular_examples/model_agnostic/Simple Boston Demo.ipynb
  - [x] notebooks/tabular_examples/tree_based_models/tree_shap_paper/Tree SHAP in Python.ipynb
  - [x] notebooks/tabular_examples/tree_based_models/Catboost tutorial.ipynb
  - [x] notebooks/tabular_examples/tree_based_models/Example of loading a custom tree model into SHAP.ipynb
  - [x] notebooks/tabular_examples/tree_based_models/Python Version of Tree SHAP.ipynb
[...]

PR #2873, opened 2023-03-08
Replace sample dataset on readme

serkef:
`load_boston` has been removed from scikit-learn since version 1.2. Replacing it with California housing sample dataset

Solves #2844 , #2322

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,9 +32,10 @@ While SHAP can explain the output of any machine learning model, we have develop
 ```python
 import xgboost
 import shap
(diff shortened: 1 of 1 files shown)


thatlittleboy:
Hi @serkef , thanks for the PR. However, as it stands, the PR is a little incomplete. We would need to update all of the prose and figures in the README.md as well.

connortann (changes requested):
As per the above comment, please also update the relevant plot images.

thatlittleboy:
Closing as Boston has been removed from README since #3200

PR #2877, opened 2023-03-15
Update _monitoring.py to prevent failure of monitoring_plot when sample size is <50.

bhanudaysharma:
If the sample size is <50, the 'inc' becomes greater than or equal to len(ys)-inc. As a result, the subsequent 'for loop' does not run. Hence, the list pvals remains empty. As a result, np.min(pvals) fails resulting in the failure of the entire monitoring_plot function. So, I corrected this error by adding a check before the for looop. If (inc < len(ys)-inc), the t-test is skipped and a Warning is produced to prevent the failure of function of monitoring_plot.

Also, added 'inc' in the function arguments to allow users to choose 'inc' value at runtime.  Kept the default value of 'inc' as 50 for backward compatibility.

### DIFF
--- a/shap/plots/_monitoring.py
+++ b/shap/plots/_monitoring.py
@@ -17,7 +17,7 @@ def truncate_text(text, max_len):
     else:
         return text
 
-def monitoring(ind, shap_values, features, feature_names=None, show=True):
+def monitoring(ind, shap_values, features, feature_names=None, show=True, inc=50):
     """ Create a SHAP monitoring plot.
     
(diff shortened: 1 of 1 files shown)


CloseChoice:
Thanks for the PR @bhanudaysharma, we appreciate your effort. I'll close this for now due to lack of activity, feel free to come back to this anytime and reopen if you want to follow this up. IMO this needs a test that reproduces the issue on master and is fixed by this PR.

PR #2890, opened 2023-03-31
Fixed Visual Studio 2022 support for Nvidia GPU support (NVCC Compilation Error Fix)

prabathbr:
NVCC compilation in Windows with Visual Studio 2022 is not working with current `setup.py` file and compilation fails with below error. Therefore, GPU bases models like `shap.explainers.GPUTree` can't be used without CUDA support.  
 
Error during compilation:   
`C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v11.5\include\crt/host_config.h(160): fatal error C1189: #error:  -- unsupported Microsoft Visual Studio version! Only the versions between 2017 and 2019 (inclusive) are supported! The nvcc flag '-allow-unsupported-compiler' can be used to override this version check; however, using an unsupported host compiler may cause compilation failure or incorrect run time execution. Use at your own risk.`

This pull request has modified NVCC command by adding `-allow-unsupported-compiler` to override limitation up to Visual Studio 2019.

`nvcc_command = "-allow-unsupported-compiler shap/cext/_cext_gpu.cu -lib -o {} -Xcompiler {} -I{} " \ `


thatlittleboy:

PR #2912, opened 2023-05-04
[BUGFIX] yticklabels in `plots.waterfall()`

dwolfeu:
This allows for cases where the type of `features[order[i]]` is incompatible with string addition, e.g. boolean type.


thatlittleboy:

PR #2915, opened 2023-05-08
Fix GradientExplainer when tabular data is passed, instead of image

renero:
When tabular data is used, instead of images, the attribute `self.features` raise an exception because the shape of the first element (`data[0]`) is not a tuple, but a single element, with the number of feature in it.

An additional check is introduced to allow for tabular data, with 2D tensors.


connortann:
I believe this was fixed by #3153 , so I shall close this PR.

PR #2931, opened 2023-05-22
Update _tree.py

parviz11:
Replacing np.int with int because np.int is deprecated in numpy >=1.20

shap requires numba==0.57.0 and numba requires 1.21 <= numpy < 1.25


Vourer:
You could also replace `np.bool` with `bool` since it is deprecated in numpy >=1.20, too.

parviz11:
Added a new pull request with both `np.int` and `np.bool` modified to `int` and `bool`, respectively.

PR #2939, opened 2023-06-01
Update _tree.py

parviz11:
Replacing np.int and np.bool with int and bool respectively, because both are deprecated in numpy >=1.20

shap requires numba==0.57.0 and numba requires 1.21 <= numpy < 1.25


thatlittleboy:

PR #2950, opened 2023-06-03
Add GH Action for precommit hooks

connortann:
Interim measure to ensure precommit hooks are run on CI.

In future, we should plan to remove this action when the GH Check is implemented via pre-commit.ci rather than GH Actions. In the meantime, the priority is to ensure the Check is enabled one way or another.


thatlittleboy (approved):
LGTM

fake-paws-dog-clapping

PR #2966, opened 2023-06-04
Add __call__ method to KernelExplainer

dwolfeu:
Currently the `KernelExplainer` class inherits its `__call__` method from the superclass `Explainer`. This is a problem for `KernelExplainer` because it does not have a `masker` attribute, e.g. [this SO question](https://stackoverflow.com/questions/75137492/attributeerror-kernel-object-has-no-attribute-masker-during-training-of-svc). I based the new `__call__` method on that of the `TreeExplainer`.

This is a port of https://github.com/dsgibbons/shap/pull/107.

### DIFF
--- a/shap/explainers/_kernel.py
+++ b/shap/explainers/_kernel.py
@@ -150,7 +150,13 @@ class Kernel(Explainer):
         else:
             ev_tiled = np.tile(self.expected_value, v.shape[0])
 
-        return Explanation(v, base_values=ev_tiled, data=X, feature_names=feature_names, compute_time=time.time() - start_time)
+        return Explanation(
+            v,
+            base_values=ev_tiled,
+            data=X,
+            feature_names=feature_names,
+            compute_time=time.time() - start_time,
+        )
 
     def shap_values(self, X, **kwargs):
         """ Estimate the SHAP values for a set of samples.
--- a/shap/explainers/_kernel.py
+++ b/shap/explainers/_kernel.py
@@ -32,10 +32,6 @@ from ._explainer import Explainer
 from .._explanation import Explanation
 import time
 
-# Suppress the sklearn warning
-# 'UserWarning: X does not have valid feature names, but <model> was fitted with feature names'
-warnings.filterwarnings('ignore', module='sklearn')
-
 log = logging.getLogger('shap')
 
 
@@ -136,7 +132,6 @@ class Kernel(Explainer):
 
         if safe_isinstance(X, "pandas.core.frame.DataFrame"):
             feature_names = list(X.columns)
-            X = X.values
         else:
             feature_names = getattr(self, "data_feature_names", None)
        
--- a/shap/explainers/_kernel.py
+++ b/shap/explainers/_kernel.py
@@ -18,11 +18,18 @@ from tqdm.auto import tqdm
 
 from .._explanation import Explanation
 from ..utils import safe_isinstance
-from ..utils._legacy import (DenseData, IdentityLink, SparseData,
-                             convert_to_data, convert_to_instance,
-                             convert_to_instance_with_index, convert_to_link,
-                             convert_to_model, match_instance_to_data,
-                             match_model_to_data)
+from ..utils._legacy import (
+    DenseData,
+    IdentityLink,
+    SparseData,
+    convert_to_data,
+    convert_to_instance,
+    convert_to_instance_with_index,
+    convert_to_link,
+    convert_to_model,
+    match_instance_to_data,
+    match_model_to_data,
+)
 from ._explainer import Explainer
 
 log = logging.getLogger('shap')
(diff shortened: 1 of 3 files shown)


shap/explainers/_kernel.py
```diff
@@ -17,6 +17,12 @@
 from sklearn.preprocessing import StandardScaler
 from tqdm.auto import tqdm
 from ._explainer import Explainer
+from .._explanation import Explanation
+import time
+
+# Suppress the sklearn warning
+# 'UserWarning: X does not have valid feature names, but <model> was fitted with feature names'
+warnings.filterwarnings('ignore', module='sklearn')
```
thatlittleboy: Seems like a strange thing to filter, just curious where this is coming from? and whether it is possible to fix rather than just ignoring it?

dwolfeu: Yes, it is a hack. The error is printed repeatedly as the code iterates over the rows of data, so avoiding it is preferable, but not essential. The issues arises if the model is trained on a Pandas dataframe. The shap code only takes the values (line 128: `X = X.values`), so sklearn throws this warning. See [this SO question](https://stackoverflow.com/questions/69326639/sklearn-warning-valid-feature-names-in-version-1-0). I looked into trying to fix the error, rather than just ignoring it, but a lot of methods for Numpy arrays are used and unweaving all of that would require considerable effort.

shap/explainers/_kernel.py
```diff
@@ -108,6 +119,27 @@ def __init__(self, model, data, link=IdentityLink(), **kwargs):
         else:
             self.D = self.fnull.shape[0]
 
+    def __call__(self, X):
+
+        start_time = time.time()
+
+        if safe_isinstance(X, "pandas.core.frame.DataFrame"):
+            feature_names = list(X.columns)
+            X = X.values
```
thatlittleboy: let's not extract out `.values` here. `shap_values()` accepts a pandas DataFrame as input.

shap/explainers/_kernel.py
thatlittleboy: I see. I do think that it's too invasive to filter all sklearn warnings on behalf of the end-user though. Can we scope the warning to only the location where sklearn throws it?  Rather than doing it globally? Note that since the .py files are executed on import, anybody who imports shap will get sklearn warnings filtered out, which is definitely not a good thing.

thatlittleboy (changes requested):
Overall looks good. Can you add at least 1 unit test utilizing this new functionality? Right now the diff coverage is pretty bad. It'll just incur technical debt for us in the future if we don't enforce this now, hope you understand.

shap/explainers/_kernel.py
dwolfeu: Done

shap/explainers/_kernel.py
dwolfeu: Thank you for pushing on this. I think I have found a much better solution, namely by setting the attribute `feature_names_in_` to `None` (which if non-empty causes sklearn to raise the warning). I make a copy so not to modify the original model.

dwolfeu:
> Overall looks good. Can you add at least 1 unit test utilizing this new functionality? Right now the diff coverage is pretty bad. It'll just incur technical debt for us in the future if we don't enforce this now, hope you understand.

I have added a test, but it is failing. I'm not sure why, as it works for me locally. I will take a look.

EDIT: The test is now successful.

dwolfeu:
@thatlittleboy Could you look at the modifications?

shap/explainers/_kernel.py
```diff
@@ -73,7 +68,12 @@ class Kernel(Explainer):
     See :ref:`Kernel Explainer Examples <kernel_explainer_examples>`
     """
 
-    def __init__(self, model, data, link=IdentityLink(), **kwargs):
+    def __init__(self, model, data, feature_names=None, link=IdentityLink(), **kwargs):
```
thatlittleboy, line 83: Can you add docstring for this new param `feature_names`?

shap/utils/_legacy.py
```diff
@@ -96,9 +98,18 @@ def __init__(self, f, out_names):
...
     if isinstance(val, Model):
-        return val
+        out = val
     else:
-        return Model(val, None)
+        out = Model(val, None)
+
+    # Fix for the sklearn warning
+    # 'X does not have valid feature names, but <model> was fitted with feature names'
+    out = copy.deepcopy(out)
```
thatlittleboy: I don't think we need to copy all the time. How about shifting this into the if-block?

thatlittleboy (changes requested):
Sorry for the delayed review, was busy at work. The `linting` tests are failing, please also address it, thanks!

dwolfeu:
> Sorry for the delayed review, was busy at work. The `linting` tests are failing, please also address it, thanks!

No worries! That's a situation with which I am not unfamiliar. I have made the requested changes.

thatlittleboy (approved):
cat-standing-thankyou

dwolfeu:
@thatlittleboy Many thanks for your help with this. The code is much better for your suggestions.

PR #2968, opened 2023-06-04
chore: re-enable ruff F811 and fix violations

thatlittleboy:
F811 ruff rule is "Redefinition of unused variable".

One thing to note is that `_maple.py` is a piece of vendored code, so I've opted to exclude ruff linting for it (to avoid making changes to it).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,3 +14,6 @@ ignore = [
 [tool.ruff.per-file-ignores]
 # Unused imports often intentional in top-level __init__.py 
 "__init__.py" = ["F401"]
+# Don't apply linting/formatting to vendored code
+"shap/explainers/other/_maple.py" = ["ALL"]
+"shap/plots/colors/_colorconv.py" = ["ALL"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,9 +7,10 @@ addopts = "--mpl"
 testpaths = ["tests"]
 
 [tool.ruff]
-select = ["F"]
-ignore = [
+select = [
+    "F", # Pyflakes
 ]
+ignore = []
 
 [tool.ruff.per-file-ignores]
 # Unused imports often intentional in top-level __init__.py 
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -7,13 +7,14 @@ addopts = "--mpl"
...
 [tool.ruff]
-# Aim to get all pyflakes logical errors ("F") passing
-select = ["F"]
-# For now, deselect rules that do not pass
-ignore = [
-  "F811", # Redefinition of unused variable
+select = [
+    "F", # Pyflakes
 ]
+ignore = []
```
connortann: Nit: suggest removing this redundant line

pyproject.toml
```diff
@@ -7,13 +7,14 @@ addopts = "--mpl"
...
 
 [tool.ruff.per-file-ignores]
 # Unused imports often intentional in top-level __init__.py 
 "__init__.py" = ["F401"]
+# Don't apply linting/formatting to vendored code
+"shap/explainers/other/_maple.py" = ["ALL"]
+"shap/plots/colors/_colorconv.py" = ["ALL"]
```
connortann, line 19: An alternative implementation could be to add `# ruff: noqa` to the modules in question. I have a slight preference for that approach, as the unusual configuration is more clear in the relevant context

connortann (approved):
Some minor nitpick suggestions, which I'll leave to you to decide on & then merge. Otherwise LGTM, approved!

pyproject.toml
thatlittleboy: Hmm, I'm fine with either, but are you also suggesting to do so for the `__init__.py`'s? We should stick to one way regardless. Otherwise it's harder to reason about what rules are applied to what files.

connortann: I think the F401 exception makes more sense in this place, as it applies as a general rule to each file named `__init.py__`. Whereas, these new exceptions feels feel more file-specific.

In any case, it's a minor stylistic choice so feel free to proceed as you wish (and merge).

PR #2943, opened 2023-06-02
Merge all changes from SHAP Community Fork

connortann:
Meta-PR to merge all changes from the dsgibbons [SHAP Community Fork](https://github.com/dsgibbons/shap).
Pending ongoing discussion: https://github.com/slundberg/shap/discussions/2942

Meta-issue: resolves #2944 .

## Related issues that will be fixed

Issues that are fixed on the community fork, and will be closed by this PR: [list is sorted numerically]

- Closes #1081 
- Closes #2144
- Related to #2322 
- Closes #2471 
- Closes #2528 
- Closes #2668 
- Closes #2706 
- Closes #2738
- Closes #2771
- Closes #2831
- Closes #2837 
- Closes #2866 
- Closes #2878 
- Closes #2904 
- Closes #2909
- Closes #2911

## Duplicated PRs

PRs here at `slundberg/shap` that have already been covered by the commits over at `dsgibbons/shap`, thus can be closed once this PR is merged. [list is sorted numerically]

- Duplicates #1802
- Duplicates #1890 
- Duplicates #1931 
- Duplicates #2020 
- Duplicates #2048 
- Duplicates #2166 
- Duplicates #2595
- Duplicates #2641
- Duplicates #2649 
- Duplicates #2655
- Duplicates #2672 
- Duplicates #2679
- Duplicates #2726 
- Duplicates #2769
- Duplicates #2781
- Duplicates #2792
- Duplicates #2830 
- Duplicates #2885 
- Duplicates #2890 
- Duplicates #2912 
- Duplicates #2924 
- Duplicates #2932
- Duplicates #2939

Plus many further fixes, as per the closed issues and PRs on the Community Fork.

PR #2969, opened 2023-06-04
Numpy version in conflict with bool

utkarshsaraf19:
AttributeError: module 'numpy' has no attribute 'bool'. `np.bool` was a deprecated alias for the builtin `bool`. To avoid this error in existing code, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here. The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations


connortann:
I believe the issue you describe with warnings from `np.bool` has been fixed on master by #2943. Closing as a duplicate.

PR #2971, opened 2023-06-04
Handling tf import issue in pyspark.

skamdar:
While Calling Gradient Explainer's shap_values method inside a pyspark UDF, it gives NoneType object has no attribute executing_eagerly. This happens because the global tf variable used inside this method is not initialized on worker node since import tf happens within init of _TFGradient class, which is executed on master only once. Same thing happens with torch as well. This behaviour is observed for picklable tf models.

Example Codebase:

```
from pyspark.sql.types import StructType, StructField, FloatType, StringType, ArrayType
import pyspark.sql.functions as F

import shap

explainer = shap.GradientExplainer(model, background_data)

def calculate_shap_udf(x, params):
    x = np.array([x])
    shap_values = explainer.shap_values(x, **params)
    return shap_values[0].tolist()

shap_udf = F.udf(lambda x: calculate_shap_udf(x, params_dict), ArrayType(FloatType()))

data_df = data_df.withColumn("shap_values", shap_udf(F.col("features")))

```


thatlittleboy:
Just curious, does it work if you `import tensorflow as tf` in your globals? I'm assuming `spark-submit` should capture the globals and load it into the workers...

```python
from pyspark.sql.types import StructType, StructField, FloatType, StringType, ArrayType
import pyspark.sql.functions as F

import shap
import tensorflow as tf  # <-------------

explainer = shap.GradientExplainer(model, background_data)

def calculate_shap_udf(x, params):
    x = np.array([x])
    shap_values = explainer.shap_values(x, **params)
    return shap_values[0].tolist()

shap_udf = F.udf(lambda x: calculate_shap_udf(x, params_dict), ArrayType(FloatType()))

data_df = data_df.withColumn("shap_values", shap_udf(F.col("features")))

```

It would be great if we had a solution that didn't involve importing these large packages at the top-level, though. `shap` imports are already slow enough as it is. 😅

skamdar:
@thatlittleboy  It does not work by import tensorflow in user module since tf is not attached to _gradient module in that case. I will create a new pull request outlining a different solution than the existing one.

PR #2973, opened 2023-06-04
Chore: Enable builtin pre-commit checks

connortann:
Enables a selection of builtin pre-commit checks, which are fast to run and prevent some common issues.

Seeing as we have already enabled pre-commit, this seems like a reasonably unintrusive ci change.


.pre-commit-config.yaml
```diff
@@ -3,3 +3,10 @@ repos:
   rev: v0.0.265
   hooks:
   - id: ruff
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v2.3.0
+  hooks:
+  - id: check-added-large-files
+  - id: mixed-line-ending
```
thatlittleboy: 1. Could you run `pre-commit autoupdate` to update the versions? I think pre-commit-hooks is currently at v4.4.0.
2. I propose adding `check-merge-conflict` and `end-of-file-fixer` as well.

thatlittleboy (changes requested):
Always up for enforcing stringent ci/code quality checks :)

.pre-commit-config.yaml
connortann: Good suggestions, done. The end-of-file fixer adds a fair number of fixes as part of this PR.

connortann:
I note that the codecov patch target is missed by this PR. I feel comfortable in this instance, as the changes are purely are made by the automatic `end-of-file-fixer` check and have no functional consequences. However, it does highlight the gaps in the test suite that we can hope to close over time.

thatlittleboy (approved):
trump-showing-lgtm

PR #2977, opened 2023-06-05
chore: bump minimum version of tqdm

### LINKED ISSUE #714: No module tqdm.auto found here
I believe this is the query being used with in shap package and it gave me this error 

<from tqdm.auto import tqdm >

I installed tqdm package but that didn't help

thatlittleboy:
Just a minor change. We have a few files in `shap` that rely on `import tqdm.auto`, which only came into `tqdm` after version `4.26.0`. See the repo: https://github.com/tqdm/tqdm/blob/v4.27.0/auto/__init__.py. (v.s. tag==v4.26.0)

Shouldn't be a major issue for most folks regardless, since it's such an old version.

Closes #714

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -212,7 +212,7 @@ def run_setup(
             'shap.actions', 'shap.models'
         ],
         package_data={'shap': ['plots/resources/*', 'cext/tree_shap.h']},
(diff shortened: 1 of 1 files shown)


connortann (approved):
LGTM

PR #2983, opened 2023-06-10
Handling tf import issue in pyspark.

skamdar:
While Calling Gradient Explainer's shap_values method inside a pyspark UDF, it gives NoneType object has no attribute executing_eagerly. This happens because the global tf variable used inside this method is not initialized on worker node since import tf happens within init of _TFGradient class, which is executed on master only once. Same thing happens with torch as well. This behaviour is observed for picklable tf models.

Example Codebase:

```
from pyspark.sql.types import StructType, StructField, FloatType, StringType, ArrayType
import pyspark.sql.functions as F

import shap

explainer = shap.GradientExplainer(model, background_data)

def calculate_shap_udf(x, params):
    x = np.array([x])
    shap_values = explainer.shap_values(x, **params)
    return shap_values[0].tolist()

shap_udf = F.udf(lambda x: calculate_shap_udf(x, params_dict), ArrayType(FloatType()))

data_df = data_df.withColumn("shap_values", shap_udf(F.col("features")))

```
As part of the change we have added import statements within shap_values method while declaring tf and keras as global variables. Same declaration for these variables is done in other methods as well so that global copy of these variables are used by respective methods. The import statement is executed once only, subsequent calls will be just sys.modules dict lookups by import machinery. In case of distributed scenario the import will be done once per python worker.


skamdar:
@thatlittleboy FYI

thatlittleboy (approved):
LGTM

spongebob-patrick-hooray

PR #2986, opened 2023-06-13
Introduce pyupgrade to remove py2 and < py3.6 related code

thatlittleboy:
## Overview

* Added UP ruff checks during pre-commit and set minimum python version in `setup.py`. **Note**: I ran both [asottile's pyupgrade](https://github.com/asottile/pyupgrade/) and ruff's UP rule locally just to make sure we cover our bases.
* Ran `ruff --fix`. This removed `__future__` imports and switched to using `f-string`'s for the most part. I visually inspected all the fixes and everything looks fine.
* Removed the try-except block for `urrlib.request`, this module is present in stdlib since py3.
* I also removed pylint and flake8 related config. We don't use these tools, and should be translated to corresponding ruff config, if necessary.

## Checklist


connortann (approved):
LGTM

Issue #2979, opened 2023-06-06
Meta-issue: preparing for a release


connortann:
## Overview

It has been over a year since the last release. A number of bugs and deprecations have been implemented on the community-led [SHAP Community Fork](https://github.com/dsgibbons/shap), which was recently merged in #2943. It would be great to make these fixes available on PyPI. This meta-issue tracks progress towards making a release.

## Context

So far the repository has been administered by @slundberg. Now that we have a small but growing pool of maintainers, we could take this opportunity to review the release process to ensure it is documented, repeatable and appropriately automated.

## TODO

(Maintainers: feel free to edit)

### Critical path

- [x] Define steps to make a release
- [x] Review setup of PyPI secrets & permissions
- [x] Update CHANGELOG.md
  - #2990
- [x] Update build action to run tests against wheel
  - #2995
  - #3015
- [x] Fix failing tests on built wheel
  - #3012
  - #3030 
  - #3032
  - #3020
  - #3029
  - #3048
- [x] Bump version in `__init__.py` and `CHANGELOG.md`
   - #3058 
- [x] Dry-run release with https://test.pypi.org/
  - #3014
- [x] Fix failing build with numpy `1.24`
  - #3063
  - #3064
- [x] Publish release `v0.42.0` to pypi
- [x] Publish release `v0.42.0` to conda feedstock
   - https://github.com/conda-forge/shap-feedstock/pull/57

### Longer term

- [ ] Agree a release strategy
   - [ ] Document somewhere (in a CONTRIBUTING.md ?)
   - [ ] Define who can trigger a release, and when
- [x] Migration repo to `SHAP` GH Org (Pending discussion)
- [x] Review packaging configuration
  - #3022
  - #3019
- [x] Review the GH Actions workflow for publishing to PyPI
   - [x] Consider running unit tests from the _built_ packages, not a local install
   - [X] Consider updating build logic to a modern [PEP-517](https://peps.python.org/pep-0517/)-compliant setup. EDIT: Build dependencies are already defined in pyproject.toml and picked up by cibuildwheel, no change needed.
   - [x] Deploy to TestPyPi on CI when wheels are built

@dsgibbons @thatlittleboy @slundberg  what do you think?

thatlittleboy:
On release strategy, I think we can keep it flexible initially. If one of us thinks it's time for a release, we can raise it in our (private) team discussions, once we have migrated to the `shap` organisation.

[...]

PR #2995, opened 2023-06-16
Update wheel building

PrimozGodec:
## Overview

While testing wheel building, I realized there were a few script updates need to be made. In this PR, I upgraded the script according to changes last year (after the script was updated last time). I ensured that all `cp` wheels in the Numpy package are built. I skipped some exotic Linux wheels that also don't exist for Numpy since they take a long to build, and if the user needs to build Numpy wheels themselves, he can also build SHAP wheels. 

The updated workflow doesn't run in the PR since it only runs on new releases (when manually triggered). You can see the resulting builds (and wheels) here: https://github.com/PrimozGodec/shap/actions/runs/5287884834.

## Checklist

- [ ] Closes #xxxx  

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -13,25 +13,22 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-20.04, windows-2019, macOS-10.15]
+        os: [ubuntu-20.04, windows-2019, macOS-11]
 
     steps:
       - name: Check out the repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
-      # Used to host cibuildwheel
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
-
-      - name: Install cibuildwheel
-        run: python -m pip install cibuildwheel==2.2.2
+      - name: Set up QEMU
+        if: runner.os == 'Linux'
+        uses: docker/setup-qemu-action@v2
+        with:
+          platforms: all
 
       - name: Build wheels
-        run: python -m cibuildwheel --output-dir wheelhouse
-        env:
-          CIBW_SKIP: "cp27-* cp35-* pp* *-musllinux_*"
+        uses: pypa/cibuildwheel@v2.13.1
 
-      - uses: actions/upload-artifact@v2
+      - uses: actions/upload-artifact@v3
         with:
           path: ./wheelhouse/*.whl
           name: bdist_files
@@ -41,17 +38,17 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - name: Check out the repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
 
       - name: Build sdist (pep517)
         run: |
-          python -m pip install pep517
-          python -m pep517.build -s .
+          python -m pip install build
+          python -m build --sdist
       - name: Upload sdist
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
         with:
           name: sdist_files
           path: dist/*.tar.gz
@@ -60,21 +57,22 @@ jobs:
     name: Publish wheels on pypi
     needs: [build_wheels, build_sdist]
     runs-on: ubuntu-latest
+    if: github.event_name == 'release' && github.event.action == 'created'
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Set up Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
 
       - name: install requirements
-        run: pip install twine==3.4.2
+        run: pip install twine==4.0.2
 
-      - uses: actions/download-artifact@v2
+      - uses: actions/download-artifact@v3
         with:
           name: bdist_files
           path: dist
 
-      - uses: actions/download-artifact@v2
+      - uses: actions/download-artifact@v3
         with:
           name: sdist_files
           path: dist
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-20.04, windows-2019, macOS-11]
+        os: [ubuntu-22.04, windows-2022, macOS-12]
 
     steps:
       - name: Check out the repo
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,6 +24,8 @@ target-version = "py37"
 [tool.cibuildwheel]
 skip = ["cp36-*", "pp*", "*-musllinux_*", "*-manylinux_i686", "*_ppc64le", "*_s390x"]
 build-verbosity = 2
+test-command = "pytest {project}/tests"
+test-extras = ["test", "plots"]
 
 [tool.cibuildwheel.macos]
 archs = ["x86_64", "arm64"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,3 +20,13 @@ target-version = "py37"
 # Don't apply linting/formatting to vendored code
 "shap/explainers/other/_maple.py" = ["ALL"]
 "shap/plots/colors/_colorconv.py" = ["ALL"]
+
+[tool.cibuildwheel]
+skip = ["cp36-*", "pp*", "*-musllinux_*", "*-manylinux_i686", "*_ppc64le", "*_s390x"]
+build-verbosity = 2
+
+[tool.cibuildwheel.macos]
+archs = ["x86_64", "arm64"]
+
+[tool.cibuildwheel.linux]
+archs = ["auto", "aarch64"]


connortann:
Thanks for the PR, this looks great. May I suggest adding a ENV parameter to the `cibuildwheel` action to run the test suite against the installed wheel?

From the [docs](https://cibuildwheel.readthedocs.io/en/stable/options/#testing), I think this may work:
```
CIBW_TEST_COMMAND: pytest {project}/tests
CIBW_TEST_EXTRAS: "test,plots"
```

connortann:
Thanks for updating the versions of actions such as `checkout` and `upload-artifact`. Suggest doing the same for the final stage `publish_wheels`.

Tagging related meta-issue: #2979

PrimozGodec:
> Thanks for updating the versions of actions such as checkout and upload-artifact. Suggest doing the same for the final stage publish_wheels.

Done. 

I changed this PR to draft until I ensure all tests run successfully. Working on it.

.github/workflows/build_wheels.yml
```diff
@@ -61,20 +65,20 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Set up Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
 
       - name: install requirements
         run: pip install twine==3.4.2
```
connortann: Nitpick: could update twine to `4.0.2`. The main breaking change from v3 to v4 was dropping support for python 3.6.

connortann:
Another minor suggestion: the docs for cibuildwheel recommend putting configuration statically in `pyproject.toml` rather than as environment variables, for convenience of running the build locally.

.github/workflows/build_wheels.yml
```diff
@@ -11,27 +11,31 @@ jobs:
...
       - name: Build wheels
-        run: python -m cibuildwheel --output-dir wheelhouse
+        uses: pypa/cibuildwheel@v2.13.1
         env:
-          CIBW_SKIP: "cp27-* cp35-* pp* *-musllinux_*"
+          CIBW_SKIP: "cp36-* cp37-* pp* *-musllinux_* *-manylinux_i686 *_ppc64le *_s390x"
+          CIBW_ARCHS_MACOS: x86_64 arm64
+          CIBW_ARCHS_LINUX: auto aarch64
+          CIBW_TEST_COMMAND: pytest {project}/tests
+          CIBW_TEST_EXTRAS: "test,plots"
```
connortann: Another suggestion: the docs for cibuildwheel suggest putting most fo the configuration in `pyproject.toml`, for ease of running the build locally.

.github/workflows/build_wheels.yml
```diff
@@ -11,27 +11,32 @@ jobs:
   build_wheels:
     name: Build wheels on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
+    continue-on-error: true
```
thatlittleboy: May I know why we need this `continue-on-error` setting? What kinds of errors have you encountered before during this build step?

I'm just thinking that if there are errors, there might be something wrong with the build/upload and we should abort instead of continuing.

.github/workflows/build_wheels.yml
```diff
@@ -11,27 +11,32 @@ jobs:
...
     strategy:
       matrix:
-        os: [ubuntu-20.04, windows-2019, macOS-10.15]
+        os: [ubuntu-20.04, windows-2019, macOS-11]
```
thatlittleboy: should we upgrade ubuntu and windows runners as well? https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#choosing-github-hosted-runners

.github/workflows/build_wheels.yml
```diff
@@ -11,27 +11,32 @@ jobs:
...
+          platforms: all
 
       - name: Build wheels
-        run: python -m cibuildwheel --output-dir wheelhouse
+        uses: pypa/cibuildwheel@v2.13.1
         env:
-          CIBW_SKIP: "cp27-* cp35-* pp* *-musllinux_*"
-
-      - uses: actions/upload-artifact@v2
+          CIBW_SKIP: "cp36-* cp37-* pp* *-musllinux_* *-manylinux_i686 *_ppc64le *_s390x"
```
thatlittleboy: We're still officially supporting py3.7 for one last version release (see discussion [here](https://github.com/dsgibbons/shap/issues/73)), please remove cp37 from the specs here

.github/workflows/build_wheels.yml
```diff
@@ -51,7 +56,7 @@ jobs:
           python -m pip install pep517
           python -m pep517.build -s .
```
thatlittleboy: Running this locally gives

```
pep517.build is deprecated. Consider switching to https://pypi.org/project/build/
```

We can do this instead (need to pip install beforehand)
```
python -m build --sdist
```

.github/workflows/build_wheels.yml
```diff
@@ -41,7 +46,7 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - name: Check out the repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v2
```
thatlittleboy: while we're at it, let's update the setup-python action to v4 as well.

thatlittleboy (changes requested):

PrimozGodec:
I will consider all comments. Currently, I am trying to make wheel tests run. Does anyone know the reason for the errors I am getting currently? https://github.com/PrimozGodec/shap/actions/runs/5308945958/jobs/9608991484

I am getting some circular import, but on the other hand, it seems that it doesn't find the `cext` module. I checked the wheels, and this module is built into the wheels.

.github/workflows/build_wheels.yml
```diff
@@ -61,20 +66,20 @@ jobs:
     needs: [build_wheels, build_sdist]
```
connortann, line 58: Please add a condition to the `publish_wheels` job so that it only runs on a tagged commit. That way, we can safely use the "manual dispatch" trigger to do a trigger the build and diagnose any issues.

I would suggest:
```yaml
    if: github.event_name == 'release' && github.event.action == 'created'
```

.github/workflows/build_wheels.yml
connortann: Agreed, I think we should remove this setting.

In this [example run](https://github.com/PrimozGodec/shap/actions/runs/5308945958/jobs/9609416817), the publish-to-pypi job runs even though the cibuildwheel tests fail, which is not desirable.

PrimozGodec: I totally agree and will remove it later. It is just temporary for debugging that the build is not cancelled when one of the other builds fails. For example, when Windows build fails, MacOS build continues.

connortann:
I haven't seen that particular issue before. Nevertheless, this just highlights how useful this PR will be and how important it is we test the wheel before publishing.

Unexpected message in the logs here: https://github.com/PrimozGodec/shap/actions/runs/5308945958/jobs/9608991484#step:4:1241

```
  ---------------------------- Captured stdout setup -----------------------------
  C extension was not built during install!
```

thatlittleboy:
> I am getting some circular import, but on the other hand, it seems that it doesn't find the cext module.

Yea, It seems the C-extension wasn't built properly during the build step.

> I checked the wheels, and this module is built into the wheels.

@PrimozGodec How did you check this? Is there a way to download the artifact from your ci build? I think we should expect to see a shared object (`.so`) file if you inspect the whl file.

image

If the `.so` file exists inside the wheel and the importerror still occurs, then I'm stumped.


EDIT: And maybe turning on build verbosity just for the purposes of debugging might be useful: https://cibuildwheel.readthedocs.io/en/stable/options/#build-verbosity

PrimozGodec:
@thatlittleboy, I downloaded and opened Wheel. I will retake a look. To get the resulting wheels, I need to disable tests so that the wheels build completely. Then they are present in artefacts.

PrimozGodec:
Screenshot 2023-06-19 at 21 07 32

This is what I fid in the `cp38-win_amd64` wheel that fils the test first. For me, it looks normal, or do I miss something? I will try to do some prints before the tests if possible.

And this is in `cp38-manylinux_x86_64`:
Screenshot 2023-06-19 at 21 11 10

connortann:
To keep things moving, I think it would helpful to merge in this PR to master (even with the build failing), then debug and fix the failing tests in a new PR.

This PR will add the safety check of preventing a broken release, so for that reason alone I'm keen to merge it. It will also enable the other maintainers to trigger the build on master in the main repo, to diagnose issues and access artifacts (the wheels).

So, may I suggest addressing the threads above and re-enabling tests, so we can get this one merged in?

PrimozGodec:
@connortann, let's do it as you proposed. I can fix all tasks that are not connected to tests tomorrow morning (Europe time) and make PR ready to merge. Then we can continue to work on tests in separate PR.

thatlittleboy:
@PrimozGodec Yea looks about right. That's really strange.

> This PR will add the safety check of preventing a broken release, so for that reason alone I'm keen to merge it. It will also enable the other maintainers to trigger the build on master in the main repo, to diagnose issues and access artifacts (the wheels).

@connortann Is this safe? If the tests suddenly pass, it would trigger the PyPI publish (before we're ready to publish anything), no? Or did I misunderstand?

PrimozGodec:
> @connortann Is this safe? If the tests suddenly pass, it would trigger the PyPI publish (before we're ready to publish anything), no? Or did I misunderstand?

It is safe. There is an if clause, `if: github.event_name == 'release' && github.event.action == 'created'` that trigger publishing only on release. So only when this job is triggered by making a new release on Gibut. Also, wheel building is not triggered on every commit on the master or pull request.

PrimozGodec:
I considered all comments and prepared this PR for merge without tests, as discussed. Here is also the job log: https://github.com/PrimozGodec/shap/actions/runs/5319350266/jobs/9631743378

PrimozGodec:
Ok, I forgot to update the runner's os versions. So let's see if builds pass: https://github.com/PrimozGodec/shap/actions/runs/5319511730.

Edit: it looks good.

connortann:
Thanks @PrimozGodec ! As a final change, would you kindly re-enable the (currently failing) cibuildwheel tests? I think you can also mark it "Ready for Review".

@thatlittleboy happy to merge? Due to [this addition](https://github.com/slundberg/shap/pull/2995/files#r1233761592), the publish step will not take place unless we make a release. We'll probably also want to add the ability to to dry-runs to testPyPI, but that can be a separate PR.

PrimozGodec:
@connortann sorry I misunderstood you. I wanted to add test commands in a separate PR. Now tests are enabled.

connortann (approved):
LGTM, great stuff! @thatlittleboy , are you happy to approve too?

thatlittleboy (approved):

thatlittleboy:
Looks good, thanks for the explanation. I've set the py3.9 tests to run again, feel free to merge when its done running

PR #3001, opened 2023-06-17
fix: UnboundLocalError with bar plots using dictionary input

### LINKED ISSUE #3000: BUG: shap.plots.bar with cohorts dict throws UnboundLocalError
### Bug report checklist

- [X] I have confirmed this bug exists on the [latest version](https://github.com/slundberg/shap/releases) of shap.
- [X] I have confirmed this bug exists on the master branch of shap.

```python
import shap
import numpy as np

shap.plots.bar({
    "t1": shap.Explanation(
        values=np.random.randn(20, 5),
        base_values=0.5,
    ),
    "t2": shap.Explanation(
        values=np.random.randn(20, 5),
        base_values=0.5,
    ),
})
```

### Issue Description

Run the above code snippet, an UnboundLocalError is thrown

image

### Expected Behavior

No error is thrown

### Installed Versions

The image above was with shap 0.41.0, but issue is reproducible with master (f77a5ae5c6fefa9a353cdd5422e05b4e565904a3) as well.

thatlittleboy:
## Overview

* Fixes the UnboundLocalError by assigning the `cohorts` variable when the input is a dictionary (previously, only an assertion was checked with no binding of the `cohorts` variable).
* Added a test for ensuring that dictionary input to `shap.plots.bar` works as expected (just a basic plot).
* Added some additional tests to ensure the input type to `shap.plots.bar` is either an Explanation, a Cohorts, or a dictionary of Explanation objects with the same number of feature columns in each Explanation object.

## Checklist

- [X] Closes #3000  


shap/plots/_bar.py
```diff
@@ -54,17 +55,35 @@ def bar(shap_values, max_display=10, order=Explanation.abs, clustering=None, clu
...
-            cohort_exps[i] = cohort_exps[i].abs.mean(0)
-        assert isinstance(cohort_exps[i], Explanation), "The shap_values paramemter must be a Explanation object, Cohorts object, or dictionary of Explanation objects!"
-        assert cohort_exps[i].shape == cohort_exps[0].shape, "When passing several Explanation objects they must all have the same shape!"
+    for i, exp in enumerate(cohort_exps):
+        if not isinstance(exp, Explanation):
+            emsg = (
+                "The shap_values argument must be an Explanation object, Cohorts "
+                "object, or dictionary of Explanation objects!"
+            )
+            raise TypeError(emsg)
```
thatlittleboy, line 76: Note that the order of checks has switched. I check for Explanation object first, because otherwise some invalid inputs won't have the `shape`  attribute (used in the check on the following line).

shap/plots/_bar.py
```diff
@@ -54,17 +55,35 @@ def bar(shap_values, max_display=10, order=Explanation.abs, clustering=None, clu
...
         cohorts = shap_values.cohorts
+    elif isinstance(shap_values, dict):
+        cohorts = shap_values
     else:
-        assert isinstance(shap_values, dict), "You must pass an Explanation object, Cohorts object, or dictionary to bar plot!"
+        emsg = (
+            "The shap_values argument must be an Explanation object, Cohorts "
+            "object, or dictionary of Explanation objects!"
+        )
+        raise TypeError(emsg)
```
connortann, line 65: Same comment as other PR - suggest remove the redundant intermediate variable `emsg`

thatlittleboy: https://beta.ruff.rs/docs/rules/raw-string-in-exception/#rule-selection

The reason why I choose to do so is explained here ^

connortann: Thanks for the link. I didn't find the given argument persuasive: it feels like an ugly hack to try to modify the builtin traceback formatting.

The "flake8-errmsg" plugin doesn't seem widely used, and other libs (e.g. pandas) seem to implement exceptions directly rather than with this workaround. Have you seen this pattern for exceptions applied elsewhere in other projects?

thatlittleboy: > Have you seen this pattern for exceptions applied elsewhere in other projects?

Not really.

What's the argument for _not_ using this pattern, other than the fact that it saves us one line of code?

To me, the benefit of using the `emsg` pattern is quite tangible (you don't show the error message "twice" to the user). Whereas the alternative (saving 1 LOC) is just a matter of code style.

ps: I wouldn't say this as a workaround/hack, since both solutions don't even result in the same effect.

connortann: > What's the argument for not using this pattern, other than the fact that it saves us one line of code?

Yeah it's simply that, just saving a line of code and keep things minimalistic and simple where possible. Maybe @dsgibbons will cast a tie-breaking vote.

For now let's "agree to disagree" and move on as it's only a style issue. I'll approve the PR :)

connortann (approved):
LGTM, nice one!

shap/plots/_bar.py
thatlittleboy: Thanks :)

PR #3008, opened 2023-06-19
Move tokensizer to shared fixture

### LINKED ISSUE #3002: Huggingface timeout error in CI
Once in a while we will encounter a `TimeoutError` to Huggingface servers in our CI's, e.g.

Logs: https://github.com/slundberg/shap/actions/runs/5260258100/jobs/9506925432
```
FAILED tests/explainers/test_partition.py::test_translation - requests.exceptions.ReadTimeout: HTTPSConnectionPool(host='huggingface.co', port=443): Read timed out. (read timeout=10.0)
```

or 

Logs: https://github.com/slundberg/shap/actions/runs/5296501329/jobs/9587659229?pr=3001
```
FAILED tests/maskers/test_fixed_composite.py::test_serialization_fixedcomposite_masker - requests.exceptions.ReadTimeout: HTTPSConnectionPool(host='huggingface.co', port=443): Read timed out. (read timeout=10.0)
```

It's probably because we are running multiple tests that download from HuggingFace at once. I suppose we could try fixing this with a fixture, but we must ensure that the fixture returns a fresh pretrained model everytime.

---

Or it could just be HuggingFace's servers being down for a while. But I think the issue still stands, we could _probably_ optimize our tests a little so we don't make too many requests to HuggingFace in our CI

connortann:
## Overview

- Refactors the test suite to use a fixture for the tokenizer, rather than a regular python function
- Sets the fixure to have `scope="session"`, which should hopefully fix the flaky CI issue with Huggingface

## Checklist

- [X] Closes #3002 
- N/A Added entry to `CHANGELOG.md` (if changes will affect users)

It would be helpful to discuss what level of scoping is appropriate for the `Tokenizer` fixture: in particular whether there is a risk of side-effects between tests if the same tokenizer is re-used.

### DIFF
--- a/tests/explainers/common.py
+++ b/tests/explainers/common.py
@@ -24,22 +24,6 @@ def basic_xgboost_scenario(max_samples=None, dataset=shap.datasets.adult):
 
     return model, X
 
-def basic_translation_scenario():
-    """ Create a basic transformers translation model and tokenizer.
-    """
-    AutoTokenizer = pytest.importorskip("transformers").AutoTokenizer
-    AutoModelForSeq2SeqLM = pytest.importorskip("transformers").AutoModelForSeq2SeqLM
-
-    tokenizer = AutoTokenizer.from_pretrained("Helsinki-NLP/opus-mt-en-es")
-    model = AutoModelForSeq2SeqLM.from_pretrained("Helsinki-NLP/opus-mt-en-es")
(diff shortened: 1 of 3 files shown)


thatlittleboy (approved):
tom-hanks-laptop-lgtm

PR #3009, opened 2023-06-19
Make sure to include all cuda and C sources in sdist

### LINKED ISSUE #1834: CUDA build using PyPI releases not working
Following #1650 I was able to build `shap` with the cuda extension from source by just running `python setup.py bdist_wheel` (with cuda 10.1, no environment variables set). However, instead of going through the process of cloning the repo, building the wheel manually and then installing it, I wanted to make `pip` do it automatically. Forcing a wheel build and logging errors with the following command:

```
pip install shap --no-cache-dir -vvv
```

results in the following error and the cuda extension not built as a result:

```
Running setup.py (path:/tmp/pip-install-__dq7b2a/shap/setup.py) egg_info for package shap
    Created temporary directory: /tmp/pip-pip-egg-info-k3cm5emn
    Running command python setup.py egg_info
    gcc: error: shap/cext/_cext_gpu.cu: No such file or directory
    gcc: warning: [34m~@~X-x c++[34m~@~Y after last input file has no effect
    gcc: fatal error: no input files
    compilation terminated.
    NVCC ==>  /usr/local/cuda-10.1/bin/nvcc
    Compiling cuda extension, calling nvcc with arguments:
    ['/usr/local/cuda-10.1/bin/nvcc', 'shap/cext/_cext_gpu.cu', '-lib', '-o', 'build/lib_cext_gpu.a', '-Xcompiler', '-fPIC', '-I/home/janis/miniconda3/envs/tf2/include/python3.8', '--std', 'c++14', '--expt-extended-lambda', '--expt-relaxed-constexpr', '-arch=sm_60', '-gencode=arch=compute_70,code=sm_70', '-gencode=arch=compute_75,code=sm_75', '-gencode=arch=compute_75,code=compute_75']
    Error building cuda module: CalledProcessError(1, ['/usr/local/cuda-10.1/bin/nvcc', 'shap/cext/_cext_gpu.cu', '-lib', '-o', 'build/lib_cext_gpu.a', '-Xcompiler', '-fPIC', '-I/home/janis/miniconda3/envs/tf2/include/python3.8', '--std', 'c++14', '--expt-extended-lambda', '--expt-relaxed-constexpr', '-arch=sm_60', '-gencode=arch=compute_70,code=sm_70', '-gencode=arch=compute_75,code=sm_75', '-gencode=arch=compute_75,code=compute_75'])
    WARNING: Could not compile cuda extensions
```

As you can see the build process can't find the extension. I have a suspicion this may be because the extensions are not part of the `MANIFEST.in` which is now outdated after extensions where moved under `shap/cext`.

jklaise:
## Overview

Currently GPU support is missing for both pre-built wheels (understandable due to needing `nvcc` and a GPU to compile these), but also for the `sdist` uploaded to PyPI (this is an oversight as the sources should be distributed in the `sdist`. This PR fixes that by bringing back the `MANIFEST.in` file specifying to include sources under `shap/cext`.

I have tested that `python -m build` does pick up the files in `MANIFEST.in` and package them into an `sdist`. Furthermore, installing from the `sdist` on a machine with `nvcc` and a GPU compiles the sources and results in a working installation of `shap.explainers.GPUTree`.

For more detailed troubleshooting and info please refer to this comment: https://github.com/slundberg/shap/issues/1834#issuecomment-1597125774

## Checklist

- [x] Closes #1834  
- [x] Added entry to `CHANGELOG.md` (if changes will affect users)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   ([dsgibbons#29](https://github.com/dsgibbons/shap/pull/29) by @dsgibbons,
    [dsgibbons#20](https://github.com/dsgibbons/shap/pull/20) by @simonangerbauer,
    [dsgibbons#24](https://github.com/dsgibbons/shap/pull/24) by @connortann).
+- Include CUDA GPU C extension files in the source distribution
+  ([#3009](https://github.com/slundberg/shap/pull/3009) by @jklaise).
 - Fixed installation of package via setuptools
   ([dsgibbons#51](https://github.com/dsgibbons/shap/pull/51) by @thatlittleboy).
 - Introduced a minimal set of `ruff` linting
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include shap/cext/*
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,2 @@
 include shap/cext/*
+include shap/plots/resources/*
--- a/setup.py
+++ b/setup.py
@@ -212,7 +212,6 @@ def run_setup(
             'shap.plots', 'shap.plots.colors', 'shap.benchmark', 'shap.maskers', 'shap.utils',
             'shap.actions', 'shap.models'
         ],
-        package_data={'shap': ['plots/resources/*', 'cext/tree_shap.h']},
         install_requires=['numpy', 'scipy', 'scikit-learn', 'pandas', 'tqdm>=4.27.0',
                           'packaging>20.9', 'slicer==0.0.7', 'numba', 'cloudpickle'],
         extras_require=extras_require,


jklaise:
Noting that this likely belongs with https://github.com/slundberg/shap/pull/2995.

thatlittleboy (commented):
Changes look good to me.

One thing I'll comment on is that if we're gonna use a MANIFEST.in, let's just include all the package data we need in here. And remove the corresponding entries in setup.py. I don't want this information spread out across 2 files.

jklaise:
Makse sense to me, will push a commit.

Perhaps it's worthwhile also checking if we can use `pyproject.toml` to specify package data to avoid introducing `MANIFEST.in`. I don't think it's possible without fully switching from `setup.py`.

Alternatively, we could just extend `package_data` in `setup.py`?

thatlittleboy:
> Perhaps it's worthwhile also checking if we can use pyproject.toml to specify package data to avoid introducing MANIFEST.in

I think we can (it'll just be analogous to what we have currently in `setup.py` for package data). But since we can't ever migrate fully from `setup.py` to `pyrpoject.toml` because of the C extension support, I would'nt really bother with this. 

> Alternatively, we could just extend package_data in setup.py?

Yep that works for me too.

jklaise:
Have gone for `MANIFEST.in` as it (imo) has some benefits:
 - https://packaging.python.org/en/latest/guides/using-manifest-in/ mentions using it
 - ~can use globs in `MANIFEST.in` but not in `package_data` in `setup.py` making the packaging more robust~
 - presence of `MANIFEST.in` makes it a bit more obvious what is included in the package

thatlittleboy (approved):
LGTM, will merge once tests pass.

husky-head-through-door-flap-thankyou

thatlittleboy:
Sorry, I just added a CHANGELOG entry. will merge in after this.

connortann:
@thatlittleboy  I think this PR may have inadvertently introduced a bug. I don't think `MANIFEST.in` is sufficient to include those files in a wheel, if they are not included with `package_data` in setup.py.

jklaise:
Apologies, you're correct @connortann. Python packaging has a habit of tripping me up. I've checked that it should be enough to just have `package_data` and we can remove `MANIFEST.in` - the files end up in both the `sdist` and the wheel.

PR #3014, opened 2023-06-21
Add publish step to TestPyPI

connortann:
## Overview

Updates the `build_wheel` workflow to support publishing to TestPyPi:
- Makes it run on all workflow events.
- Use a GH action rather than twine update, for simplicity
- Restrict the main PyPI publish to only run on actual releases.

~~This requires a `TEST_PYPI_TOKEN`, which may not exist currently as a repository secret.~~

Following the PyPI docs, I would prefer to use Trusted Publishing instead of tokens, which can be set up as explainer here:

https://docs.pypi.org/trusted-publishers/

Supports #2979

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -60,18 +60,7 @@ jobs:
     name: Publish wheels on pypi
     needs: [build_wheels, build_sdist]
     runs-on: ubuntu-latest
-    if: github.event_name == 'release' && github.event.action == 'created'
     steps:
-      - uses: actions/checkout@v3
-
-      - name: Set up Python 3.11
-        uses: actions/setup-python@v4
-        with:
-          python-version: '3.11'
-
-      - name: install requirements
-        run: pip install twine==4.0.2
-
       - uses: actions/download-artifact@v3
         with:
           name: bdist_files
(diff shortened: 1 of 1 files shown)


connortann:
@dsgibbons @thatlittleboy @slundberg  I don't currently have access to this project on PyPi or TestPyPi. Please could a maintainer add a token for TestPyPI as a repo secret?

Or, even better would be to configure [Trusted Publishing ](https://docs.pypi.org/trusted-publishers/), which would improve security and remove the need for repo Tokens.

dsgibbons:
I didn't know about Trusted Publishers - I'm guessing we would need to do that for both PyPI servers? I'll chase up @slundberg if he doesn't see this.

connortann:
Now that we have the token configured, @thatlittleboy I've triggered a `build_wheels` workflow run for this action here:
https://github.com/slundberg/shap/actions/runs/5437584891

If the tests pass (which may take several hours on all platforms), we should see the wheels published here for the current given version `0.41.0`:
<https://test.pypi.org/project/shap>

So would like to confirm that all occurs smoothly before completing this PR.

EDIT: workflow deployed to test PyPi as expected.

I would also like to tentatively propose that we only run the _core_ tests in cibuildwheel, rather than the full suite. That is, in `pyproject.toml` to change  `[tool.cibuildwheel]: test-extras` from `test` to `test-core`. My thinking is:

1. Reduces the `build_wheels` workflow time from hours to probably minutes
1. Reduces risk that the workflow fails due to flaky failures when we push a tagged commit (which would be really annoying and disrupt the release process)
1. We run full tests on all platforms on PRs now
1. The workflow is still an improvement, as all previous releases didn't test the wheels with cibuildwheel on CI at all AFAIK

What do you think?

connortann:
@dsgibbons I'd be grateful for a review on this if you get time today, one of the last few things before we can make a release :)

thatlittleboy (approved):

thatlittleboy:
I'm agreeable with running only core tests before the build/release. We could always switch back to full tests later if needed.

Let's wait to see if @dsgibbons ' is also on board with this

connortann:
I agree, let's put the propsal to change the build tests on the backburner for now. It's probably better to err on the side of caution for our first release, given how many build issues we identified with these tests.

PR #3015, opened 2023-06-21
Wheel build - Skip tests for cp37 and cp38-macosx_x86_64

### LINKED ISSUE #3012: BUG: Tests fail when running from built wheel using cibuildwheel
### Issue Description

The test suite does not fail when installling from the built package. Example [failing run](https://github.com/PrimozGodec/shap/actions/runs/5308945958/jobs/9608991484).

This was uncovered in #2995 , which updated the release action and added a test of the built wheel. This will need to be fixed for #2979.

Run locally:

```bash
pip install cibuildwheel
cibuildwheel --platform linux
```

Or to run on CI, trigger the `build_wheels` GitHub Action. See previous runs [here](https://github.com/slundberg/shap/actions/workflows/build_wheels.yml).

The issue also arises if the `--editable / -e` flag is removed from the pip install step on CI, as seen in #3017

### Expected Behavior

Build should pass including pytest suite.

### Bug report checklist

PrimozGodec:
## Overview

Fix tests in the wheel build

- Skip cp37: since many packages that Shap depend on do not provide cp37 wheel anymore. They would need to be built before tests can run. One of the packages is Catboost which would need Ninja to build; also, if Ninja is installed, it takes a long time to build, so I suggest skipping cp37.
- Skip cp38-macosx_x86_64: In the MacOS and Python 3.8 case, pip will not download the universal Catboost wheel but will take .tar.gz and try to build it. It is likely due to: https://github.com/catboost/catboost/issues/2371

## Checklist

- [X] Related: #3012 
 -  EDITED: may not fully close this issue


connortann (changes requested):
This looks good, thanks for the PR!

Just one request, please add a comment in the YAML file to explain the rationale behind the skipped tests, that the catboost wheel is not available. You could reference this PR number for more info.

FYI I edited the description to change "closes #3012" to "related: #3012", as I don't think this PR by itself will fully close that issue.

connortann (approved):

pyproject.toml
```diff
@@ -24,8 +24,9 @@ target-version = "py37"
 [tool.cibuildwheel]
 skip = ["cp36-*", "pp*", "*-musllinux_*", "*-manylinux_i686", "*_ppc64le", "*_s390x"]
 build-verbosity = 2
-test-command = "pytest {project}/tests"
+test-command = "pytest -v {project}/tests"
 test-extras = ["test", "plots"]
+test-skip = "cp37-* cp38-macosx_x86_64 *-win32"  # temporary: -win32
```
connortann: Please remove `win32` - I don't think it should be in this list of environments to skip?

connortann (changes requested):
@PrimozGodec  I noticed one other change to address, please see the thread above.

pyproject.toml
PrimozGodec: The plan is not to have `win32` skipped, but since it fails in a different way, I skipped it temporarily to focus on other failures. Will remove it before we merge PR.

PrimozGodec:
I changed the PR to draft since I want to test what work and what doesn't will change it to ready later

connortann:
I would like to get the main changed merged in soon, so we can keep progressing towards a release. So, if the win32 build issue is not something that can be resolved quickly, I'd suggest removing that from the scope of this PR and raising a separate issue.

.github/workflows/build_wheels.yml
```diff
@@ -12,6 +12,7 @@ jobs:
     name: Build wheels on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
```
PrimozGodec, line 15: I added this for easier debugging. That one build does not finish because of another. E.g. the Windows build does not finish because of failing MacOS. The whole building will still fail at the end, and no upload will happen.

PrimozGodec:
> I would like to get the main changed merged in soon, so we can keep progressing towards a release. So, if the win32 build issue is not something that can be resolved quickly, I'd suggest removing that from the scope of this PR and raising a separate issue.

Good. I made a PR ready to merge and will continue with debugging in a separate PR after https://github.com/slundberg/shap/pull/3021 and https://github.com/slundberg/shap/issues/3020 are merged.

connortann (approved):
Thanks, LGTM! I merged in master to update the branch, so will merge once the tests pass.

PR #3017, opened 2023-06-21
FIX: Test against a full install, not an editable install

### LINKED ISSUE #3020: CI: unit tests should run against a full install, not an editable install
### Issue Description

The CI unit tests do not test the built library; they use a local "editable" install. This means that problems with the build/packaging are likely to be missed. This is concerning, as it could lead us to unknowingly publish a broken wheel.

Issue https://github.com/slundberg/shap/issues/3012 is an example of this: the tests pass on CI, but fail when running with cibuildwheel. This change aims to prevent that happening in future.

Another recent example is #3021 - a packaging error was inadvertently introduced in a previous PR, which broke the wheels and was not caught on CI.

Related #2979

### Ensuring the correct library is picked up

The fix is not as simple as removing the `--editable` flag from `pip install`, as the wrong library can be imported on sys.path. The library is installed to `/lib/python3.11/site-packages`, but when shap is imported the library incorrectly imported from the current working directory `/home/runner/work/shap/shap/shap/__init__.py`. More info in this recent [CI run](https://github.com/slundberg/shap/actions/runs/5337284077/jobs/9673122016?pr=3017#step:6:68) .

A quick fix is to replace `python -m pytest` with `pytest`, as that avoids adding the current working directory to sys.path. However, this is not particularly robust.

A full fix would be migrating to a `src/` layout. From the [pytest docs](https://docs.pytest.org/en/7.2.x/explanation/goodpractices.html#tests-outside-application-code):

> Generally, but especially if you use the default import mode prepend, it is **strongly suggested** to use a src layout. Here, your application root package resides in a sub-directory of your root, i.e. src/mypkg/ instead of mypkg.
> 
> This layout prevents a lot of common pitfalls and has many benefits, which are better explained in this excellent [blog post](https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure%3E) by Ionel Cristian Mărieș.

### Bug report checklist

connortann:
## Overview

Fixes #3020.

- Removes the `--editable / -e` flag from the `pip install` command during CI, to trigger a full wheel build & install.
- Changes the pytest invocation from to avoid adding the current working directory to sys.path
- Changes the pytest import mode from `prepend` to `append`

This is a rather temporary fix to an important problem, so I would advocate for switching to a `src` layout as strongly recommended by pytest.

Supports #2979.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -61,7 +61,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           # Use "eager" update strategy in case cached dependencies are outdated
-          pip install --upgrade --upgrade-strategy eager -e '.[test,plots]'
+          pip install --upgrade --upgrade-strategy eager '.[test,plots]'
       - name: Test with pytest
         run: |
           python -m pytest --durations=20 \
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -61,10 +61,12 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           # Use "eager" update strategy in case cached dependencies are outdated
+          # Using regular install NOT editable install: see GH #3020
           pip install --upgrade --upgrade-strategy eager '.[test,plots]'
       - name: Test with pytest
         run: |
-          python -m pytest --durations=20 \
+          # Using "pytest" over "python -m pytest" to avoid adding working directory to sys.path
+          pytest --durations=20 \
           --cov=shap --cov-report=xml --cov-report=term-missing \
           --mpl-generate-summary=html --mpl-results-path=./mpl-results
       - name: Upload mpl test report
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -65,10 +65,13 @@ jobs:
           pip install --upgrade --upgrade-strategy eager '.[test,plots]'
       - name: Test with pytest
         run: |
-          # Using "pytest" over "python -m pytest" to avoid adding working directory to sys.path
+          # Ensure we avoid adding current working directory to sys.path:
+          # - Use "pytest" over "python -m pytest"
+          # - Use "append" import mode rather than default "prepend"
           pytest --durations=20 \
           --cov=shap --cov-report=xml --cov-report=term-missing \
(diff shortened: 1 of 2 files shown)


PrimozGodec:
I could reproduce the error locally, removing the development installation and installing with `pip install .`. For me, the case was when I ran import (`python -c "import shap;print(shap.__file__)"`) from the project dir, it imported the development version, but if I moved somewhere else, it imported site-packages versions, which looks ok (when in package dir there shap dir exists, and it has priority against the installed version).

For test according to [this](https://stackoverflow.com/questions/19530849/force-py-test-to-use-installed-version-of-module) the solution is either: 
1. to remove `__init__.py` from the root of tests dir.
2. move tests to the module and call them from the module with https://docs.pytest.org/en/7.1.x/example/pythoncollection.html#interpreting-cmdline-arguments-as-python-packages

For now, I suggest the first solution. Can you try this in this PR?

connortann:
I don't think the issue in that stackoverflow thread is related to the issue here, but lets try deleting `test/__init__.py` anyway to confirm. EDIT: the issue remains. As per [this job](https://github.com/slundberg/shap/actions/runs/5338745593/jobs/9676527023?pr=3017) the wrong library is still imported.

If the `shap` directory is in the current working directory, and if that directory is on `sys.path`, then that's always going to take precedence over what is installed in `..../sys-modules`. 

From the [pytest docs](https://docs.pytest.org/en/7.2.x/explanation/goodpractices.html#tests-outside-application-code):

> Generally, but especially if you use the default import mode prepend, it is **strongly suggested** to use a src layout. Here, your application root package resides in a sub-directory of your root, i.e. src/mypkg/ instead of mypkg.
> 
> This layout prevents a lot of common pitfalls and has many benefits, which are better explained in this excellent [blog post](https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure%3E) by Ionel Cristian Mărieș.

connortann:
The correct library is being picked up now the pytest invocation has changed. I expect that this PR will pass once #3021 is merged.

dsgibbons:
LGTM

dsgibbons (approved):

connortann:
@dsgibbons I think this PR may have inadvertently broken code coverage measurements. I think this should be fixable by "combining paths" in the coverage configuration. Attempting a fix on #3033

PR #3018, opened 2023-06-21
Chore: Fix mixed line endings

connortann:
## Overview

Some files in this repo use an unepected mix of LF and CRLF line endings, probably due to unusual git settings on windows machines. This is noticable in other PRs, where there appears to be a large superfluous diff.

This PR renormalises all files to be the normal default, as specified in `.gitattributes`. As per the usual convention, files will be checked-out in a system-dependent way, but commited in a system-agnosic manner.

Created with the command `git add --renormalize .`


dsgibbons:
LGTM.

PR #3019, opened 2023-06-21
Migrate to a src layout

### LINKED ISSUE #3020: CI: unit tests should run against a full install, not an editable install

connortann:
## Overview

Closes #3020.

Tentative work-in-progress, exploring migrating to a `/src` layout as explained in issue above.

## Checklist


PrimozGodec:
I don't like the solution with src since it doesn't look nice (and almost no python modules use it). I think the solution in https://github.com/slundberg/shap/pull/3017#issuecomment-1601613640 would be enough to solve all our problems.

connortann:
The "src" pattern is advocated for rather strongly in the [pytest docs](https://docs.pytest.org/en/7.2.x/explanation/goodpractices.html#tests-outside-application-code). There is a good discussion in the python [packaging guide](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/). It's used in several other repos, for example [Streamlit](https://github.com/streamlit/streamlit), [pytest](https://github.com/pytest-dev/pytest), [hypothesis](https://github.com/HypothesisWorks/hypothesis/tree/master) and I'm sure many others.

I think that avoiding nefarious packaging problems is a much higher priority than stylistic neatness, so if the `src` layout leads to more reliable testing and packaging then I think it's at least something worth contemplating.

Personally I don't think it's particularly ugly in any way; but, the big diff in this change is a downside. I'll explore other options as you say, and I suggest we keep this on the backburner for now for consideration.

jklaise:
@connortann you can add a `.git-blame-ignore-revs` file with hashes of commits to ignore when running `git blame` either locally or through the Github UI, it's a good solution for doing things like these and mitigating issues with large diffs when looking back.
https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

connortann:
Yes indeed, in fact that was raised in an earlier thread in https://github.com/slundberg/shap/pull/3006#discussion_r1236752351 .

I'll add this in a future PR, as we have had several style & refactor PRs recently.

EDIT: opened a PR #3027

dsgibbons:
I don't think a `src` layout is ugly. It seems to be the standard when you read about packaging on the Python website anyway. If we think this will prevent problems in the future, then I'm on board with it.

connortann:
Closing as not planned for now, as we have got the testing & packaging working adequately at least for now.

That said, I'd be keen to come back to this if we encounter any more subtle build issues. For example - at present, our test suite in cibuildwheel will fail unless pytest is run with a non-default `import_mode` option... 🫤

PR #3021, opened 2023-06-21
CHORE: remove redundant MANIFEST.in

connortann:
## Overview

Removes the redundant manifest file. Other changes originally in this PR were superceeded by #3022, as packaging is now handled by `[tool.setuptools.package-data]` in `pyproject.toml`.

Previous description:

<details>

Supports #3012 and #2979.

- Reinstates the `package_data` argument in setup.py, partially reverting #3009
- Removes MANIFEST.in
- Broadens the package data to `cext/*` rather than just `cext/tree_shap.h`

From the [packaging docs](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#manifest-in):

> Note MANIFEST.in does not affect binary distributions such as wheels.

Once this is merged, #3017 will prevent similar issues from occuring on CI in future.

</details>

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,0 @@
-include shap/cext/*
-include shap/plots/resources/*


setup.py
```diff
@@ -212,6 +212,7 @@ def run_setup(
             'shap.plots', 'shap.plots.colors', 'shap.benchmark', 'shap.maskers', 'shap.utils',
             'shap.actions', 'shap.models'
         ],
+        package_data={'shap': ['plots/resources/*', 'cext/tree_shap.h']},
```
jklaise: Change to `'cext/*'` to include the missing files?

jklaise:
I think we can also remove the `MANIFEST.in` file as testing locally `package_data` was enough to make sure files end up in both the wheel and `sdist`.

setup.py
connortann: Agreed, done

connortann:
Agreed, in fact already pushed a commit

thatlittleboy (approved):
Approving, feel free to merge in the 2 PRs whichever order is best

Issue #2607, opened 2022-06-30
Gradient Explainer Issue


jrudar:
Hello,

I am using the latest version of 'shap', specifically the Gradient Explainer class for a project using TensorFlow. I believe I found a bug in this bit of code on line 72 of _gradient.py but I am not sure:

self.features = list(range(data[0].shape[1]))

I believe it should be:

self.features = list(range(data[0].shape[0]))

Joe

Lalalaashen:
having the same prob today, i think it should be shape[0] rather than shape[1]

ockaro:
Hi there! I also face the same issue. Should be a rather easy fix by just replacing the line by
`self.features = list(range(data.shape[1]))`

dsgibbons:
Can you please raise a PR if you think this is the case @ockaro?

ockaro:
Done! https://github.com/slundberg/shap/pull/3025

PR #3025, opened 2023-06-22
Fix getting number of features for Gradient Explainer

ockaro:
## Overview

Fixes getting the number of features when instantiating the gradient explainer with a data type different than pandas Dataframe.


connortann:
Thanks for the PR. Please would you add a unit test that covers the bug which this PR addressses?

dsgibbons:
Tests failing. I agree; we need a decent unit test for #2607 before proposing a fix.

connortann:
I believe this was covered by #3153, so I shall close this PR.

PR #3028, opened 2023-06-22
FIX: Update cibuildwheel config with supported architectures and import-mode

connortann:
### Overview

Closes most of #3012 

- Edits the cibuildwheel config to specify the precise set of supported architectures, copying the set used by sklearn.
- Changes the test invocation to fix an issue where the wrong library was imported, and `cext` is missing.

Example [wheel build run](https://github.com/slundberg/shap/actions/runs/5351527272)

### Import issue

This took me many hours to debug. More info in https://github.com/slundberg/shap/issues/3012#issuecomment-1603448067

### Architectures

Previously a very broad array of architectures were attempted by cibuildwheel. The full list of is given [here](https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip).

Many of these are not well supported by the wider scientific python ecosystem, and wheels are not available for dependencies. The shap wheel tests fail because the dependencies cannot be built from source easily. For example, `scipy` publishes [wheels](https://pypi.org/project/scipy/#files) for win64 but not win32, as explained in this [thread](https://github.com/scikit-learn/scikit-learn/blob/main/.github/workflows/wheels.yml).

I propose creating wheels for the same set of architectures as used by scikit-learn, which is given in their build matrix [here](https://github.com/scikit-learn/scikit-learn/blob/main/.github/workflows/wheels.yml).

I note that several recent releases of shap only included wheels for a small subset of architectures. SHAP 0.39 only [published wheels](https://pypi.org/project/shap/0.39.0/#files) for `win_amd64`.

Another option could be to publish wheels for a wide range of architectures, but disable testing for them. This seems rather risky to be, as I think we should only publish what we can test. In any case, the source distribution is available as an alternative for esoteric scenarios.

## Checklist

- [ ] Confirm wheel workflow passes


connortann:
The `build_wheel` job seems to be passing on Linux, but not yet on Window or Macosx due to what looks like issues with 3rd party libs used in the tests. A few options come to mind:

- We could only publish wheels for Linux for now
- We could skip testing on Windows & MacOSX for now
- We could add the other platforms to the Checks that run on PRs, and get the full set of tests passing.

@dsgibbons  @thatlittleboy  any thoughts?

thatlittleboy:
hi @connortann I'm on board with only testing and releasing the same ABIs/archs as scikit-learn, it seems like a pragmatic decision to make.

Shall we add a comment in `pyproject.toml` with a reference to this PR, so that our future selfs remember to follow sklearn?

I'll lean more towards ensuring build tests pass & release wheels for Windows/MacOS.

connortann:
> Shall we add a comment in pyproject.toml with a reference to this PR, so that our future selfs remember to follow sklearn?

Good idea, I'll push a commit.

> I'll lean more towards ensuring build tests pass & release wheels for Windows/MacOS.

Likewise, it seems better to get it working on CI rather than have to do painful debugging on each release. I implemented a wider set of test jobs in: #3029 

This reproduces the Segmentation Fault from lightgbm on MacOS that we see in cibuildwheel.

thatlittleboy (approved):
I'll approve. @dsgibbons how about you?

dsgibbons (approved):

dsgibbons:
Appreciate the effort @connortann - this looked like an annoying issue to resolve!

PR #3033, opened 2023-06-23
FIX: update code coverage configuration for installed package

connortann:
## Changes

- Fixes the code coverage configuration
- Removes the coverage terminal output
- Adds a test to check that pytest has picked up the correct version of SHAP, to be skipped when running locally.

image

## Context

The code coverage config was inadvertently broken when the tests were configured to run against the installed package rather than the local project. After a lot of trial and error the fix seems to be two parts:
- specifying `coverage.run:source_pckg` to ensure coverage measures the package, not the directory
- specifying `coverage.paths:combine` to relate the package to the local directory.

Without sounding like a broken record, this is another concern which would be much more simple and less error prone with a `src` layout.

TODO:

- [x] Ensure coverage is > 0%
- [x] Get codecov upload working

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -44,19 +44,20 @@ jobs:
           - os: windows-latest
             python-version: "3.11"
             extras: "test"
-            name: windows-3.11
+            name: 3.11-windows
           - os: macos-latest
             python-version: "3.11"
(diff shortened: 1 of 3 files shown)


thatlittleboy (approved):
LGTM, sorry for the delay

seal-of-approval-seal

Issue #2960, opened 2023-06-03
Random floating-point errors in GitHub Actions


dsgibbons:
Copied from https://github.com/dsgibbons/shap/issues/98

We have seen GitHub Actions randomly fail occasionally due to floating-point precision issues. See, for example:

- https://github.com/dsgibbons/shap/actions/runs/5085485586/attempts/1 (EDIT: py3.8, `tests/explainers/test_deep.py::test_pytorch_mnist_cnn`)
- https://github.com/dsgibbons/shap/actions/runs/5135384415/attempts/1 (EDIT: py3.11, `tests/explainers/test_deep.py::test_pytorch_single_output`)

In both cases, the tests passed after being re-run with no source code changes. We should consider either increasing some of the floating-point tolerances, or investigate why we are seeing such large random deviations in `test_deep.py`.

## List of flaky tests

Here's a list of all the flaky tests we know of so far (feel free to edit). 

- [ ] `test_tf_keras_linear` with seed `256`
- [ ] `test_single_tree_compare_with_kernel_shap` with seed `434`
- [ ] `test_single_tree_nonlinear_transformations` with seed `116`
- [ ] `test_xgboost_classifier_independent_margin` with seed `364`
- [ ] `test_gradient.py::test_pytorch_mnist_cnn` with seed `24`
- [ ] `test_deep.py::test_pytorch_mnist_cnn` with seed `212`
- [ ] `test_pytorch_single_output`
- [ ] `test_xgboost_direct` with seed `28`
- [ ] `test_several_trees` with seed `411`
- [ ] `test_shape_values_linear_many_features` with seed `94` (**this test seems particular egregious, it's failing 59 out of the first 100 seeds**)

dsgibbons:
Another example, this time from `test_tree.py`: https://github.com/slundberg/shap/actions/runs/5163671345/attempts/1

(EDIT: py3.8, `tests/explainers/test_tree.py::test_xgboost_direct`)

thatlittleboy:
I came across a tip that is helpful in debugging this, for those interested in tackling this. https://calmcode.io/pytest-tricks/flaky-tests.html

For the `tests/explainers/test_tree.py::test_xgboost_direct` test, I managed to reproduce the failing test for seed=28.

```python
def test_xgboost_direct(i=28):
    np.random.seed(i)  # <----------------------------------
    xgboost = pytest.importorskip('xgboost')

    N = 100
    M = 4
    X = np.random.randn(N, M)
[...]

PR #3040, opened 2023-06-24
fix: heatmap plot colorbar and internal refactoring

thatlittleboy:
## Overview

* Fixed the colorbar aspect ratio in `shap.plots.heatmap`, see:
image
* Internally, I've switched to an `ax`-based implementation instead of `plt.imshow()`, `plt.barh()` etc.; primarily so we don't have to do `pl.gca()` all over the place, and also the axes is needed to pass into `plt.colorbar()` to avoid warnings about 
  ```
     Unable to determine Axes to steal space for Colorbar. Using gca(), but will raise in the future. Either provide the *cax* argument to use as the Axes for the Colorbar, provide the *ax* argument to steal space from it, or add *mappable* to an Axes.
  ```
* The keen eye might notice that in the baseline images above, the tick directions have swapped from "in" to "out" -> note that the tick directions being "in" in the original baseline image is just a result of `pytest-mpl` using the "classic" matplotlib style; anybody with a more recent matplotlib version should already have the tick directions being "out" even prior to this PR. I only made this explicit in-code so that the baseline images generated by `pytest-mpl` reflects what most users would get by default ("out" tick directions).
* Render the `f(x)` text in MathTex rather than plain text to match with the force plots.
* Added comments and grouped similar code (e.g. the tick/axes manipulation code) together for better code readability.

## Checklist

- Closes #xxxx  


connortann (approved):
LGTM! There are several notices from Codecov about patches not covered by tests, as the code coverage is not currently being reported correctly. I'll see if I can look in to that tomorrow.

thatlittleboy:
Failing tests are from pytorch mnist cnn, irrelevant to the changes in this PR. See #2960

PR #3048, opened 2023-06-27
CHORE: Explicit random seed for tests

connortann:
Improves the use of randomness in the test suite for reproducibility, and to mitigate the occurence of flaky tests.

Key changes are in `conftest.py`:

- Adds a fixture to provide a changing random seed for tests. If a test fails, the random seed will be printed by pytest.
- Use local `RandomState` in each test rather than the global random state.
- Adds a CLI argument to fix the random state, for easy local reproduction of any failures.
- Resets the _global_ random state before all tests to zero.

Should help address #2960 , improving the reproducibility of any test failures.

### DIFF
--- a/tests/explainers/test_gradient.py
+++ b/tests/explainers/test_gradient.py
@@ -13,7 +13,7 @@ def test_tf_keras_mnist_cnn(random_seed):
 
     tf = pytest.importorskip('tensorflow')
 
-    rng = np.random.default_rng(seed=random_seed)
+    rs = np.random.RandomState(random_seed)
     tf.compat.v1.random.set_random_seed(random_seed)
 
     from tensorflow.compat.v1 import ConfigProto, InteractiveSession
@@ -43,11 +43,10 @@ def test_tf_keras_mnist_cnn(random_seed):
 
     # the data, split between train and test sets
     #(x_train, y_train), (x_test, y_test) = tf.keras.datasets.mnist.load_data()
-    n_samples=200
-    x_train = rng.standard_normal(size=(n_samples, 28, 28))
-    y_train = rng.integers(low=0, high=9, size=n_samples)
-    x_test = rng.standard_normal(size=(n_samples, 28, 28))
-    y_test = rng.integers(low=0, high=9, size=n_samples)
+    x_train = rs.randn(200, 28, 28)
+    y_train = rs.randint(0, 9, 200)
+    x_test = rs.randn(200, 28, 28)
+    y_test = rs.randint(0, 9, 200)
 
     if K.image_data_format() == 'channels_first':
         x_train = x_train.reshape(x_train.shape[0], 1, img_rows, img_cols)
@@ -94,7 +93,7 @@ def test_tf_keras_mnist_cnn(random_seed):
     )
 
     # explain by passing the tensorflow inputs and outputs
-    inds = rng.choice(x_train.shape[0], 20, replace=False)
+    inds = rs.choice(x_train.shape[0], 20, replace=False)
     e = shap.GradientExplainer((model.layers[0].input, model.layers[-1].input), x_train[inds, :, :])
     shap_values = e.shap_values(x_test[:1], nsamples=2000)
 
@@ -113,7 +112,7 @@ def test_pytorch_mnist_cnn(random_seed):
 
     torch = pytest.importorskip('torch')
     torch.manual_seed(random_seed)
-    rng = np.random.default_rng(seed=random_seed)
+    rs = np.random.RandomState(random_seed)
 
     from torch import nn
     from torch.nn import functional as F
@@ -207,7 +206,7 @@ def test_pytorch_mnist_cnn(random_seed):
         train(model, device, train_loader, optimizer, 1)
 
         next_x, _ = next(iter(train_loader))
-        inds = rng.choice(next_x.shape[0], 3, replace=False)
+        inds = rs.choice(next_x.shape[0], 3, replace=False)
         if interim:
             e = shap.GradientExplainer((model, model.conv1), next_x[inds, :, :, :])
         else:
--- a/tests/explainers/test_gradient.py
+++ b/tests/explainers/test_gradient.py
@@ -7,9 +7,11 @@ import shap
 
 # pylint: disable=import-error, import-outside-toplevel, no-name-in-module, import-error
 
-def test_tf_keras_mnist_cnn():
+def test_tf_keras_mnist_cnn(random_seed):
     """ This is the basic mnist cnn example from keras.
     """
+    rng = np.random.default_rng(seed=random_seed)
+
     tf = pytest.importorskip('tensorflow')
     from tensorflow.compat.v1 import ConfigProto, InteractiveSession
     from tensorflow.keras import backend as K
@@ -39,7 +41,6 @@ def test_tf_keras_mnist_cnn():
     # the data, split between train and test sets
     #(x_train, y_train), (x_test, y_test) = tf.keras.datasets.mnist.load_data()
     n_samples=200
-    rng = np.random.default_rng(seed=0)
     x_train = rng.standard_normal(size=(n_samples, 28, 28))
     y_train = rng.integers(low=0, high=9, size=n_samples)
     x_test = rng.standard_normal(size=(n_samples, 28, 28))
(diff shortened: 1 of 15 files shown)


thatlittleboy:
Thanks @connortann for the PR. I think setting the seeds is a good idea in general.

I do have one concern (which I raised in #2960 as well), which is that the nature of the failures that we are encountering is not of the `random() > 0.5` kind, but are mostly additivity checks failing.

Right now, it's not entirely clear to me _why_ this is happening, so it's particularly concerning because it could mean there are some edge cases where our algorithms are not considering => thus resulting in additivity failures. (I'm leaning towards this conclusion because there are many open issues here talking about this very problem.)

In some ways, it's a good thing that we're getting the occasional errors => it's actually giving us examples upon which we can debug.

connortann:
I completely agree. So, I suggest we don't close issue #2960 , but keep it open until we can diagnose the root cause. However, in the meantime it's probably preferable for this issue not to affect other unrelated PRs.

Perhaps setting the random seed will also be helpful for diagnosing the issue: we could determine a value of the random seed that reliably causes the tests to fail.

thatlittleboy:
But the only way for us to produce these failures at the moment is via the tests running in CI (where we don't fix the random seed), right?

If we fix the random seed now, then it would make it harder for us to identify which tests we need to focus our attention on (to fix the additivity issues)

> in the meantime it's probably preferable for this issue not to affect other unrelated PRs.

The impact isn't that bad, we just need to re-run the CI. It's annoying, but I see the aforementioned problem as a bigger problem than the annoyance from having to re-run CI.

connortann:
> But the only way for us to produce these failures at the moment is via the tests running in CI (where we don't fix the random seed), right?

I don't think that's quite right: if you wanted to work on the additivity issue, you could create a PR that sets the random seed to a number that _does_ reliably cause a failure. That's an improvement on the current state, as you will be able to determine when the issue is actually fixed.

thatlittleboy:
That's not quite what I meant. Let me rephrase: are we able to confidently list down all of the **existing** tests that fail additivity checks for particular random seeds, and also the random seed that generates the failure? If we can't, I don't think #2960 should be closed.

So far, I've listed one test (the xgboost test) with one random seed in the original issue to reproduce the problem.
Are there any more? And what seeds cause the additivity failures in those?

My point is that we don't have full answers to the above question. And the only way to get the answer is to leave the tests running as they are in CI. (how else would we know which tests are "flaky"?) Unless we introduce the hypothesis library into our testing.

connortann:
I see what you mean, I'm with you.

I'll rejig this PR with a slightly different aim then, to ensure that each test with randomness accepts a random seed, and the the seed is printed to the pytest logs if the test fails.

connortann:
I had a go implementing the above. Hopefully that is the best of both worlds: a different seed will be used in each run by default, but it will be easy to fix the seed to reproduce a given failure.

My suggestion is to use the new `random_seed` fixture:

```python
def test_foobar(random_seed):
    assert False
```

If the test fails, the seed will be printed by pytest:

```
tests/explainers/test_deep.py F                                 [100%]

============================== FAILURES ===============================
_____________________________ test_foobar _____________________________

random_seed = 736

    def test_foobar(random_seed):
>       assert False
E       assert False

tests/explainers/test_deep.py:622: AssertionError
======================= short test summary info =======================
FAILED tests/explainers/test_deep.py::test_foobar - assert False
================== 1 failed, 261 deselected in 2.56s ==================
```

connortann:
Found a flaky failure: `test_tf_keras_linear`, with 
random_seed = 896

tests/explainers/test_gradient.py
```diff
@@ -7,9 +7,11 @@
 
 # pylint: disable=import-error, import-outside-toplevel, no-name-in-module, import-error
 
-def test_tf_keras_mnist_cnn():
+def test_tf_keras_mnist_cnn(random_seed):
     """ This is the basic mnist cnn example from keras.
     """
+    rng = np.random.default_rng(seed=random_seed)
+
     tf = pytest.importorskip('tensorflow')
```
thatlittleboy: Should we also set the random seed for tf for good measure?

```python
tf.compat.v1.random.set_random_seed(random_seed)
```

Same goes for the tensorflow tests in `test_deep.py` as well.

thatlittleboy (changes requested):
EDIT: This approach I can get on board with :)

--- 

There are a few more test files that I think we should also cover in this PR. (I just did a grep for `np.random.seed` in the project)

- `tests/explainers/test_linear.py`
- `tests/explainers/test_tree.py`
- and a couple more tests in `tests/explainers/test_kernel.py` (like `test_linear()` in that file)

thatlittleboy:
> Found a flaky failure: `test_tf_keras_linear`, with random_seed = 896

It failed again with `random_seed = 823`, so there's clearly something off here about the implementation. We'll need to look into this at some point.

tests/explainers/test_gradient.py
connortann: Good shout

connortann:
> There are a few more test files that I think we should also cover in this PR

I noticed in `test_linear.py` that the _global_ random seed was reset, but randomness doesn't seem to be used explicitly in the test. I think adding fuzzing here wouldn't really make sense, as it isn't clear what if anything is being fuzzed.

However, it's probably wise to set the global seed explicitly for reproducibility, as the implicit default expectation is that unit tests are deterministic and reproducible.  I added a `global_random_seed` fixture to handle this, which has `autouse=true`.

Then, for tests that explicitly wish to use fuzzing, the `random_seed` fixture is used with a new numpy random Generator. That should make the application of fuzzing explicit and obvious to future readers of the tests.

connortann:
I made a few further updates from having examined some failures:

- Suggesting use of `RandomState` rather than `default_rng()`, as it has stricter compatibility guarantees between versions and platforms
- Added more pytorch, tensorflow and xgboost random state seeds
- Did not change tests for plotting functions. Both the test data generation and the plotting function itself use the global random state, so changing the way the test data is generated will lead to a different output image.
- Added a pytest CLI argument to set the random state, for ease of local debugging. Example call: 
   ```bash
   pytest -k my_function --random-state 123
   ```
- Pinned a passing seed for the tests that we've already identified as being flaky, as tracked in #2960 . Once we've identified that a given test is flaky, I don't see any _further_ benefit in having the test fail on other unrelated PRs.

I've re-run the tests a few times, I'll keep noting any flaky issues in #2960.

connortann:
FYI I've re-run the suite a few times to try to identify other flaky failures & seeds. Things seem to be passing consistently now, run 4 sets (of 8 parallel runs) without a failure 🎉

tests/explainers/test_linear.py
```diff
@@ -130,15 +122,15 @@ def test_perfect_colinear():
...
 def test_shape_values_linear_many_features():
-    Ridge = pytest.importorskip('sklearn.linear_model').Ridge
 
-    np.random.seed(0)
+    Ridge = pytest.importorskip('sklearn.linear_model').Ridge
 
     coef = np.array([1, 2]).T
 
+    rs = np.random.RandomState(0)
     # generate linear data
```
thatlittleboy, line 131: @connortann Was this accidentally missed out or it was set to seed=0 explicitly? I was expecting we would use the `random_seed` fixture here.

(If it was set to 0 explicitly, then we'll need the comment here to refer back to #2960 as well)

thatlittleboy (approved):
Just one last clarification. I'll pre-approve since it's a minor one. Thanks for the good work!

domino-effect-self-high-five

tests/explainers/test_linear.py
connortann: Ah yes, good spot. That was indeed a flaky test; I've updated the list in the issue accordingly.

connortann: I've resisted the urge to do any debugging or fixing of the underlying failures in this PR, as the diff is already very large! Hoping we can address that after the next release.

thatlittleboy: Just came back to this and realized this comment wasn't addressed (I suppose I wasn't clear in my comment). I'll raise a PR. :) 
EDIT: #3073

connortann: Ah, apologies if I misunderstood you - I had updated the list of flaky tests in #2960, but not done any "fixing" of the tests other than pinning the seed here to something that passes.

PR #3050, opened 2023-06-27
CHORE: Fixup pycodestyle issues

connortann:
Refactoring PR, fixes up most of the `"E"` pycodestyle issues. No functional changes.

For safely, all fixes implemented via pre-commit Ruff, apart from bare "Except" clauses which were fixed with a regex substitution.


thatlittleboy (approved):
nicholas-cage-letsgo

PR #3058, opened 2023-07-04
Bump version

connortann:
## Overview

Bumps version ready for release

Alternative idea to consider for future releases (I'll take a look tomorrow): #3059

### DIFF
--- a/shap/__init__.py
+++ b/shap/__init__.py
@@ -1,7 +1,7 @@
 # flake8: noqa
 import warnings
 
-__version__ = "0.41.0"
+__version__ = "0.42.0"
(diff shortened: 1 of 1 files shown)


connortann:
@thatlittleboy  I was looking in to `setuptools-scm` as a way of managing the version number (as per #3059 ), which would have a few benefits. But, it would take a bit of care to implement: for example, to ensure compatibility with readthedocs.com and cibuildwheel.

So, I propose for today that we keep things simple and just bump the version number this way.

slundberg (approved):
Thanks again!

PR #3062, opened 2023-07-05
API: Multi-output summary plot legend

101AlexMartin:
## Overview

Added flag to allow the printing of the mean SHAP values in the legend of a multi-output summary bar plot. The motivation behind this change is to easily compare rankings when different chunks of data are used to get the SHAP values. Comparing visually the overall ranking in a multi-output problem was quite challenging, so this pull request solves the problem.

## Checklist

- [ ] Closes #xxxx  
- [x] Added entry to `CHANGELOG.md` (if changes will affect users)


thatlittleboy:
Hi @101AlexMartin , thanks for the PR!

A couple of things before we kick off the PR review:

- Please fill up the checklist in the PR description.
- As this is a new feature, please add tests to ensure the desired functionality is implemented properly.
- I intend to deprecate `summary_legacy` at some point in the future. So I'll also suggest adding this param into `_bar.py` as well.

101AlexMartin:
Hello! I implemented your proposed changes :)
As for the depreciation of `summary_legacy` and the addition of the parameter to `_bar.py` as well. If I'm not mistaken, `_bar.py` is not ready yet to display graphs for multiouput models. Thus, I do not see the addition of this parameter feasible.

thatlittleboy:
Hi @101AlexMartin , next steps would be to make sure the tests pass :)

In particular for plotting tests, we are using [pytest-matplotlib](https://github.com/matplotlib/pytest-mpl). You will need to generate the baseline images and commit that into the repository.

This way we can ensure that the function always generates the same-looking plot.

101AlexMartin:
I added the tests and the baseline figures :)
Also merged with master branch

shap/plots/_beeswarm.py
```diff
@@ -441,6 +441,7 @@ def summary_legacy(shap_values, features=None, feature_names=None, max_display=N
                  class_inds=None,
                  color_bar_label=labels["FEATURE_VALUE"],
                  cmap=colors.red_blue,
+                 print_values_legend_multioutput=False,
```
thatlittleboy: May I suggest to rename this as `show_values_in_legend`? I think we can leave `multioutput` out of the param name.

Then we can make clear it's for multioutput only in the docstring.

tests/plots/baseline/test_random_multi_class_summary_legend.png
thatlittleboy, line 1: If you could help me to understand the motivation for doing this a little better:, the SHAP values on the legend are the mean values across all features for each class, right?

And having these values in the legend helps you to compare visually against the individual features within the same class?

shap/plots/_beeswarm.py
101AlexMartin: Done :)

tests/plots/baseline/test_random_multi_class_summary_legend.png
101AlexMartin: Certainly! Your interpretation of the numbers is correct. Regarding the added value: it lets you have a quantitative comparison of the SHAP values among output classes. For instance, when the legend is ordered by which class is most impacted by changes in the features (default behavior, contained in `class_inds`), without numbers showed in the legend, it becomes impossible to know if class1 appearing in the legend has a mean value much bigger than the second class appearing in the legend or if, otherwise, they are very close. For instance in a problem where some classes are more impacted by some features and some others classes are more impacted by different features, without these numbers, it is impossible to see if the global impact on one class is quantitatively much bigger than for other classes.

101AlexMartin:
What is the status of this then? 
Let me know if you need any extra explanation :)

thatlittleboy:
@101AlexMartin Sorry, let me get back to you over the weekend. Bit busy at work this week.

shap/plots/_beeswarm.py
```diff
@@ -880,11 +885,31 @@ def color(i):
             class_inds = np.argsort([-np.abs(shap_values[i]).mean() for i in range(len(shap_values))])
         elif class_inds == "original":
             class_inds = range(len(shap_values))
+
+        if show_values_in_legend:
+            # Get the smallest decimal place of the first significant digit
+            # to print on the legend. The legend will print ('n_decimal' + 1)
+            # decimal places.
+            # Set to 1 if the smallest number is bigger than 1.
+            smallest_shap = np.abs(np.min(np.abs(shap_values).mean((1, 2))))
```
thatlittleboy: I think the outermost `np.abs` isn't necessary

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 <!--Changes from new PRs should be put in this section-->
 
+### Added
+- Added `print_values_legend_multioutput` parameter to `summary_legacy`
```
thatlittleboy: remember to change the option name here as well

shap/plots/_beeswarm.py
```diff
@@ -880,11 +885,31 @@ def color(i):
...
+                        smallest_shap
+                    )
+                ))
+
         for i, ind in enumerate(class_inds):
             global_shap_values = np.abs(shap_values[ind]).mean(0)
+            if show_values_in_legend:
+                label = f'{class_names[ind]}: {np.round(np.mean(global_shap_values),(n_decimals+1))}'
+            else:
+                label = class_names[ind]
```
thatlittleboy: In terms of the formatting of the legend, I just want to note here that the cohort bar plots are already using `cohort_name [cohort_size]`

image

---

And in `decision_plot` we have `group_name (shap_value)` kind of label formatting.

image

Can we change the label formatting to use parenthesis so we conform more to the same "look" as decision plot?

We might have to relook into adjusting how the labels look in the future, when we eventually add multi-output support to `shap.plots.bar`, though.

thatlittleboy (changes requested):
Hi @101AlexMartin Sorry this review took so long! This should be the last round, I just have a couple more suggestions for you to make.

101AlexMartin:
Done :)

thatlittleboy:
Just a heads up, the 3.11 tests are failing, not anything to do with your PR. Probably due to the 4.0 lightgbm release. #3092 

This PR looks good to me, I'll come back to this once we get our builds running again. Hang tight!

thatlittleboy (approved):
LGTM

dog-teaches-baby-crawling

PR #3073, opened 2023-07-08
Fix seed for flaky tests

thatlittleboy:
## Overview

Some follow ups from #3048 and this review comment: https://github.com/slundberg/shap/pull/3048#discussion_r1252233290 in particular.

- Identified two more flaky tests (and so, fixed the seed here)
    - `test_single_tree_compare_with_kernel_shap`, logs [here](https://github.com/slundberg/shap/actions/runs/5493438025/jobs/10011608579?pr=3072).
    - `test_gradient.py::test_pytorch_mnist_cnn`, logs [here](https://github.com/slundberg/shap/actions/runs/5493769307/jobs/10012166317?pr=3074).
- <s>I explicitly added the `random_seed` fixture and the FIXME messages so we remember to come back to this eventually.</s>
- Also, I updated the list over at (https://github.com/slundberg/shap/issues/2960#issuecomment-1618007543) with some more seeds for some of the tests.

## Checklist

- [X] Related to #2960  


tests/explainers/test_deep.py
```diff
@@ -240,7 +240,7 @@ def test_tf_keras_imdb_lstm(random_seed):
     assert np.allclose(sums, diff, atol=1e-02), "Sum of SHAP values does not match difference!"
 
 
-def test_pytorch_mnist_cnn():
+def test_pytorch_mnist_cnn(random_seed):
```
connortann: I don't follow the purpose of this change. Why include the fixture here when it's not used?

connortann (changes requested):
I think we should either use the `random_seed` fixture (for fuzz testing) or hard-code a fixed seed, but not both. I don't think we should provide a fixture that isn't used.

tests/explainers/test_deep.py
thatlittleboy: It's not doing anything here except as an indicator (to us) that the "normal" state of things is to have `random_seed` as a fixture.

And at the moment we have a temporary fix (indicated by FIXME + setting random seed = 0 as a code block a couple lines down).  

The reason why I'm having `random_seed` in the test fn params:  
Suppose we've already fixed the bug, we can then just remove this code block without having to worry about forgetting to set `random_seed` as a fixture. Hence, it's more about minimizing the changes that we need to make potentially weeks/months down the line when we eventually fix the bug.

Since the fixture itself doesn't add much time, if at all, to the test, can we just leave this in?

connortann: > It's not doing anything here except as an indicator

I think this is the underlying point, we want to make sure the implementation is clear to other developers who might not have read this discussion. I agree with the intent to provide context & indicate our plans, but I think in this case an explicit FIXME comment with a reference to GH issue is a clearer way of providing context.

Including the fixture has two minor downsides:
- A test failure will lead pytest to print out a misleading irrelevant value e.g. `random_seed=42`.
- The "redefinition of an unused variable" is typically a code smell and a common [linting error](https://www.flake8rules.com/rules/F811.html).

> [...] the "normal" state of things is to have random_seed as a fixture

I'm actually quite undecided on this point: I'll share some thoughts on #2960 .

> Since the fixture itself doesn't add much time

Yeah agreed, I don't think we should worry about any negligible performance difference.

-------

This is a relatively minor issue, so to avoid bikeshedding I'll approve the PR and let you make the decision. In any case, we'll probably come back to these tests soon to address the FIXMEs.

connortann (approved):
Approving with suggestion

tests/explainers/test_deep.py
thatlittleboy: > A test failure will lead pytest to print out a misleading irrelevant value e.g. random_seed=42.

Yea good point. Not that the tests will fail, but I take your point.

PR #3076, opened 2023-07-09
Fix benchmark top-level import

### LINKED ISSUE #3075: BUG: Cannot import `shap.benchmark` with shap==0.42.0
### Issue Description

When I tried to import `shap.benchmark` with shap==0.42.0, I got the following ImportError.

```
ImportError: cannot import name 'BenchmarkResult' from partially initialized module 'shap.benchmark' (most likely due to a circular import) (/usr/local/lib/python3.10/site-packages/shap/benchmark/__init__.py)
```

It seems that the import order changed by [this commit](https://github.com/slundberg/shap/commit/6be5b996cc7f048bf04875a8dc7775b2233dbeb4) causes the above error, which was resolved when I modified the order to import first the `BenchmarkResult`.

I have confirmed that it works fine in 0.41.0.

```python
import shap.benchmark
```

### Expected Behavior

We can import `shap.benchmark` without the above error.

### Bug report checklist

### Installed Versions

0.42.0

thatlittleboy:
## Overview

- Changed the import of the files within `shap/benchmark/` to be direct imports instead of importing from the `__init__.py` file.

I've tried `import shap.benchmark` locally and can verify the reported issue in #3075 is fixed.

### Note 

I thought about fixing our tests alongside this PR.  
Right now, the `tests/benchmark/*.py` aren't being collected by pytest because the filenames are not prefixed by `test_`. But the tests themselves are using a somewhat broken API, e.g. `framework` needs to be imported in `benchmark/__init__.py`, and the `SequentialPerturbation()` doesn't have a `model_score` method, etc. 

The fixes there are non-trivial and possibly involve public API changes, so I think we should leave that to a separate PR.

## Checklist

- [X] Closes #3075 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## Unreleased
 <!--Changes from new PRs should be put in this section-->
 
+### Fixed
+
+- Fixed circular import issues with `shap.benchmark`
+  ([#3076](https://github.com/slundberg/shap/pull/3076) by @thatlittleboy).
+
 ## 0.42.0 - 2023-07-05
 
 This release incorporates many changes that were originally contributed by the SHAP
--- a/shap/benchmark/_compute.py
+++ b/shap/benchmark/_compute.py
@@ -1,4 +1,4 @@
-from . import BenchmarkResult
+from ._result import BenchmarkResult
 
 
 class ComputeTime():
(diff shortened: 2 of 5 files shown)


connortann:
An optional suggestion, at a pinch it could be useful to add a very minimal test to confirm the import works, such as:

```python
# tests/benchmark/test_import.py
def test_import():
    import shap.benchmark
```

As you say, it would be great to get the other benchmark tests working in future.

connortann (approved):
LGTM, with one suggestion. Over to you to merge

Issue #2044, opened 2021-06-15
Tree Explainer bug: unable to handle trees with a single leaf


brunofacca:
I was getting that `'TreeEnsemble' object has no attribute 'values'` error (which already has [its own issue](https://github.com/slundberg/shap/issues/480)] with LightGBM 3.2.1 and SHAP 0.39.0, without using any categorical features, only floats.

I've found that [this try/catch statement](https://github.com/slundberg/shap/blob/d0b4d59f96adc5d067586c0dd4f7f2326532c47a/shap/explainers/_tree.py#L921) is also catching exceptions that are not related to categorical splits. I've removed the `try` statement and got a more informative error message: `KeyError: 'leaf_index'` at [line 1243 of _tree.py](https://github.com/slundberg/shap/blob/master/shap/explainers/_tree.py#L1243)

The problem is that the code is unable to handle trees that have a single leaf, where the output of `model.booster_.dump_model()["tree_info"]` is like `{'tree_index': 0, 'num_leaves': 1, 'num_cat': 0, 'shrinkage': 1, 'tree_structure': {'leaf_value': 0.4591098358151824}}`. Notice that the dict does not have the `leaf_index` key.

Do you think that replacing [these lines](https://github.com/slundberg/shap/blob/d0b4d59f96adc5d067586c0dd4f7f2326532c47a/shap/explainers/_tree.py#L1243-L1253) with the following would work? I've tried it, it successfully avoids the error but I have no idea of how it will affect the end result (the SHAP values):

```
leaf_index = vertex['leaf_index'] if 'leaf_index' in vertex else 0
leaf_count = vertex['leaf_count'] if 'leaf_count' in vertex else 0
self.children_left[leaf_index+num_parents] = -1
self.children_right[leaf_index+num_parents] = -1
self.children_default[leaf_index+num_parents] = -1
self.features[leaf_index+num_parents] = -1
self.children_left[leaf_index+num_parents] = -1
self.children_right[leaf_index+num_parents] = -1
self.children_default[leaf_index+num_parents] = -1
self.features[leaf_index+num_parents] = -1
self.thresholds[leaf_index+num_parents] = -1
self.values[leaf_index+num_parents] = [vertex['leaf_value']]
self.node_sample_weight[leaf_index+num_parents] = leaf_count
```

Thank you.

PR #3077, opened 2023-07-09
Refactor SingleTree

thatlittleboy:
## Overview

I've recently been looking into fixing the bug reported in #2044. And in doing so, I just did some refactoring to make the code (hopefully) a little cleaner and easier to understand.  
And then it should be easier to review the PR that contains the actual bug fix (coming soon).

The commits are intentionally atomic so they can be reviewed one at a time. 

- Added docstrings for `SingleTree` class, to list down the available attributes on that class + what they individually mean. Just for our collective understanding only. It's not rendered into the public API documentation.
- Added an *intentionally failing* test that demonstrates the problem in #2044. The fix will come in the next PR.
    - The other reason why I added a test is to hopefully convince that there wasn't any bugs introduced as a result of this refactor.
- There were a lot of repetitive dictionary lookups in the original implementation. I saved the lookup value into a local variable and reused it, so the code is more concise and should technically run just a tad faster (I didn't bother benchmarking).
- Refactored the if-else logic for `if vertex['split_index'] not in visited:` to use the [guard clause pattern](https://sourcery.ai/blog/explaining-refactorings-3/) (saves us 1 level of indentation).

## Checklist

- [X] Related to #2044  

### DIFF
--- a/tests/explainers/test_tree.py
+++ b/tests/explainers/test_tree.py
@@ -11,6 +11,7 @@ import sklearn
 import sklearn.pipeline
 
 import shap
+from shap.explainers._tree import SingleTree
 from shap.utils._exceptions import InvalidModelError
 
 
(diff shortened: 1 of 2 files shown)


tests/explainers/test_tree.py
```diff
@@ -771,6 +772,57 @@ def test_xgboost_buffer_strip():
...
+    """Tests for the SingleTree class."""
+
+    def test_singletree_lightgbm_basic(self):
+        """A basic test for checking that a LightGBM dump_model() dictionary
+        is parsed properly into a SingleTree object.
+        """
+
+        # Stump (only root node) tree
+        # FIXME: this test should NOT throw a KeyError, see #2044
+        with pytest.raises(KeyError, match="leaf_index"):
```
connortann, line 785: Would pytest's [xfail](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail) be more appropriate here than a `pytest.raises(Error)`?

Example use:

```python
@pytest.mark.xfail(reason="known issue, see #2044")
def test_singletree_lightgbm_basic(self):
    ...
```

connortann:
I think there may be further opporunity to refactor. The `__init__` method is very long, as it instantiates the object from multiple different sources. As a consequence, it's not particuarly modular.

One alternative pattern could be to use a series of class constructor methods. What do you think? In pseudo-code, something like:

```python
class SingleTree:
    def __init__(self, children_left, children_right, children_default):
        self.children_left = children_left
        self.children_right = children_right
        self.children_default = children_default

    @classmethod
    def from_treelike(cls, tree):
        if safe_isinstance(tree, "sklearn.tree._tree.Tree"):
            return cls.from_sklearn(tree)

        elif type(tree) is dict and "features" in tree:
            return cls.from_dict(tree)

        elif safe_isinstance(tree, "pyspark.foo.bar"):
            return cls.from_pyspark(tree)

    @classmethod
    def from_sklearn(cls, tree):
        ...

    @classmethod
    def from_lightgbm(cls, tree):
        ...

    @classmethod
    def from_xgboost_json(cls, tree):
        ...

    @classmethod
    def from_pyspark(cls, tree):
        ...
```

Nb. this would be a significant change, so if you agree then this is probably best kept for a future PR.

connortann (approved):
LGTM, approved with one suggestion about pytest.

I appreciate the effort that you went to creating clearly named modular commits, that did indeed make it straightforward to understand - thanks!

thatlittleboy:
> I think there may be further opporunity to refactor. The `__init__` method is very long, as it instantiates the object from multiple different sources. As a consequence, it's not particuarly modular.
> 
> One alternative pattern could be to use a series of class constructor methods. What do you think? In pseudo-code, something like:
> 
> ```python
> class SingleTree:
>     def __init__(self, children_left, children_right, children_default):
>         self.children_left = children_left
>         self.children_right = children_right
>         self.children_default = children_default
> 
>     @classmethod
>     def from_treelike(cls, tree):
>         if safe_isinstance(tree, "sklearn.tree._tree.Tree"):
>             return cls.from_sklearn(tree)
> 
>         elif type(tree) is dict and "features" in tree:
>             return cls.from_dict(tree)
> 
>         elif safe_isinstance(tree, "pyspark.foo.bar"):
>             return cls.from_pyspark(tree)
> 
>     @classmethod
>     def from_sklearn(cls, tree):
>         ...
> 
>     @classmethod
>     def from_lightgbm(cls, tree):
>         ...
> 
>     @classmethod
>     def from_xgboost_json(cls, tree):
>         ...
> 
>     @classmethod
>     def from_pyspark(cls, tree):
>         ...
> ```
> 
> Nb. this would be a significant change, so if you agree then this is probably best kept for a future PR.

I like this idea very much :) Thanks!

Yea , I'll keep this in the backburner for now, will revisit this after the bug fix.

tests/explainers/test_tree.py
thatlittleboy: `xfail` is a good shout. will keep this in mind for future. but since I've already implemented the fix and the fix-PR will be raised soon, I think we can let this slide :p

Issue #3061, opened 2023-07-05
ENH: Wheels for aarch64


asyarif93:
### Problem Description

Please make wheels for linux arm. context, i'm trying to add shap inside airflow container. i'm using m1 with docker
preferably for python 3.8
```
#0 46.07 Building wheels for collected packages: shap, crcmod, docopt
#0 46.07   Building wheel for shap (pyproject.toml): started
#0 46.60   Building wheel for shap (pyproject.toml): finished with status 'error'
#0 46.62   Building wheel for crcmod (setup.py): started
#0 46.62   error: subprocess-exited-with-error
#0 46.62   
#0 46.62   × Building wheel for shap (pyproject.toml) did not run successfully.
#0 46.62   │ exit code: 1
#0 46.62   ╰─> [123 lines of output]
#0 46.62       The nvcc binary could not be located in your $PATH. Either  add it to your path, or set $CUDAHOME to enable CUDA
#0 46.62       Error building cuda module: TypeError('cannot unpack non-iterable NoneType object')
#0 46.62       WARNING: Could not compile cuda extensions
#0 46.62       running bdist_wheel
#0 46.62       running build
#0 46.62       running build_py
#0 46.62       creating build
#0 46.62       creating build/lib.linux-aarch64-cpython-38
#0 46.62       creating build/lib.linux-aarch64-cpython-38/shap
#0 46.62       copying shap/datasets.py -> build/lib.linux-aarch64-cpython-38/shap
#0 46.62       copying shap/links.py -> build/lib.linux-aarch64-cpython-38/shap
#0 46.62       copying shap/_serializable.py -> build/lib.linux-aarch64-cpython-38/shap
#0 46.62       copying shap/_explanation.py -> build/lib.linux-aarch64-cpython-38/shap
#0 46.62       copying shap/__init__.py -> build/lib.linux-aarch64-cpython-38/shap
#0 46.62       creating build/lib.linux-aarch64-cpython-38/shap/explainers
#0 46.62       copying shap/explainers/_gradient.py -> build/lib.linux-aarch64-cpython-38/shap/explainers
#0 46.62       copying shap/explainers/tf_utils.py -> build/lib.linux-aarch64-cpython-38/shap/explainers
#0 46.62       copying shap/explainers/_exact.py -> build/lib.linux-aarch64-cpython-38/shap/explainers
#0 46.62       copying shap/explainers/_explainer.py -> build/lib.linux-aarch64-cpython-38/shap/explainers
[...]

PR #3078, opened 2023-07-10
Build macos arm64 wheels

PrimozGodec:
## Overview

SHAP does not provide arm64 wheels for MacOS, which is essential since most of modern Mac computers use this architecture. With this PR, I am enabling the building of amr64 wheels. 

I would also be very happy if arm64 wheels would be added to PyPi as soon as possible.

I ran the building process on the branch, and it successfully build arm64 wheels: https://github.com/PrimozGodec/shap/actions/runs/5504208255

## Checklist

- [ ] Closes #xxxx  

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -137,4 +137,4 @@ archs = ["x86_64"]
 archs = ["AMD64"]
 
 [tool.cibuildwheel.macos]
-archs = ["x86_64"]
+archs = ["x86_64", "arm64"]


connortann:
Thanks for the PR! Yes, it would be great to get a broader range of wheels built. To get last week's release out the door, we took a pragmatic decision last week to build with a relatively restricted set of wheels (mirroring the set chosen in `scikit-learn`), accepting it wasn't as comprehensive as we'd like.

## Patch release

As you mention you're eager to have wheels on PyPI as soon as possible, maybe we can make a patch release `0.42.1` with a broader range of wheels. `linux: aarch64` wheels have also been requested on #3061. What do you think @thatlittleboy ? 

Nb. I'm testing things on a separate branch `maint/add-archs`, to ensure we don't include other changes on `master` that aren't in the release notes.

## Testing on MacOS

One snag I noticed is that cibuildwheel can't currently run tests on arm64:

> While arm64 wheels can be built on x86_64, they cannot be tested. The ability to test the arm64 wheels will be added in a future release of cibuildwheel, once Apple Silicon CI runners are widely available.

I don't know if there's anything we can do about this, so may have to just take the chance. @PrimozGodec , are you by any chance able to run the tests locally on your machine against the wheel on `arm64`?

thatlittleboy:
The way I see it is that for those on arm64 macs, (in the absence of the appropriate CI runners) it's either we release an untested wheel or None at all, and I think the former is a better scenario for us to be in.
It's the same situation with our cuda c extension module actually. 😅

A patch release sounds good. We need one anyway because of #3075.

connortann:
OK, I'll work on a patch release. Created a milestone and a new meta-issue #3082 .

I'll come back to this PR shortly once I've done some related testing :)

PrimozGodec:
@connortann, thank you for taking quick action. 

> I don't know if we can do anything about this, so we may have to just take the chance. @PrimozGodec , are you, by any chance, able to run the tests locally on your machine against the wheel on arm64?

I think we cannot do anything about it. As I know cibuildwheel plans to enable that in the future. Also, numpy/scipy don't test wheels on arm64. I tested the wheel locally, and the tests passed.

>  it's either we release an untested wheel or None at all

I think it is better to provide a wheel and the community report if anything is wrong. Having wheels makes life easier for people without compilers on the computer.

connortann (approved):
Looks good!

PR #3080, opened 2023-07-10
MAINT: Add codecov.yml, specify `after_n_builds`

### LINKED ISSUE #3065: MAINT: wait for multiple builds before sending CodeCov notifications
At present, most PRs recieve an initial notification that the PR will reduce coverage by 17%. This is because not all the CI runs have completed, and only the "core" tests have been reported.

This could be a bit confusing and disconcerting for new contributors. We should enable the [notifyafter_n_builds](https://docs.codecov.com/docs/codecovyml-reference#codecovnotifyafter_n_builds) setting, for example to `2`.

On a broader note, we may as well add a codecov.yml to the repo, so the setting can be managed as code.

connortann:
Closes #3065

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,9 @@
+codecov:
+  notify:
+    after_n_builds: 2
+    wait_for_ci: yes
+coverage:
+  range: 70..90
+comment:
+  require_changes: true
+  after_n_builds: 2
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,9 +1,9 @@
 codecov:
   notify:
-    after_n_builds: 2
+    after_n_builds: 7
     wait_for_ci: yes
 coverage:
   range: 70..90
 comment:
   require_changes: true
-  after_n_builds: 2
+  after_n_builds: 7
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -37,6 +37,7 @@ jobs:
     strategy:
       matrix:
         os: ["ubuntu-latest"]
+        # The total number of matrix jobs should match codecov.yml `after_n_builds`.
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
         extras: ["test"]
         include:


.github/codecov.yml
```diff
@@ -0,0 +1,9 @@
+codecov:
+  notify:
+    after_n_builds: 7
```
thatlittleboy, line 3: I'm wondering if it is necessary for this number to be so high? I'm presuming the main intention is to avoid the first spurious coverage comment from the core tests. So 2 or 3 should work just fine (the coverage variance from different OS x python version shouldn't be that big).

Put another way, it's more disastrous for the number to be set too high than low, since the latter is just an annoyance. The former might result in no coverage notification at all, right?

connortann: I think there is indeed quite an annoyance if `after_n_builds` < `total_num_builds`.

Ideally, we'd like the bot to only comment if the coverage has actually changed. This cuts down on redundant notifications, and keep the comments meaningful.

However, if `after_n_builds` is too low, CodeCov will post a message about a small diff, which is then edited alter. Some builds have slight differences: for example, I think there there are 2 lines that are only covered on MacOS. So we get the situation that all PRs get an initial notification of e.g. 2 lines less coverage. Ideally, the bot would just wait until it had all the results before deciding whether to comment at all.

I don't think there is any disadvantage on having `after_n_builds` set number "too high", is there? Is there anything "disaterous" that would happen? At present there are 7 test jobs, so a passing PR must upload 7 coverage reports.

(Aside: actually there are 8 jobs currently, but we're removing python 3.7 soon)

thatlittleboy: >  don't think there is any disadvantage on having after_n_builds set number "too high", is there?

By "too high" I mean if we have 7 jobs and we had `after_n_builds` set to 8. In that case, I presume there won't be any notifications, right?  
Assuming the above is true, then the main disadvantage is that we have to remember to change this number whenever we drop support for a python version (without at the same time adding the support for a new python version). Otherwise we risk not getting notified of coverage at all. And I'm not in favour of such strong coupling in our infra configuration, especially if they are spread out across multiple files.

What do you think about setting it to a number like 5 as a middleground? I don't think the likelihood is high that we will drop the number of CI builds below 5 (since python releases are yearly now, so even if we only test on Linux, we should have 5 CI jobs at least).

thatlittleboy: Actually, how about this as an idea:

https://hynek.me/articles/ditch-codecov-python/

We add a new job after all the tests are complete, combine the reports and upload to codecov?

connortann: >  ... the main disadvantage is that we have to remember to change this number whenever we drop support for a python version

True. I don't think this is particularly burdensome: it would only need to be updated maybe once a year or so if that. I added a comment to the `run_tests` matrix to remind us to keep it in sync. 

> What do you think about setting it to a number like 5 as a middleground?

I did try that at first, but I found it still led to a lot of false positive notifications, for example if the MacOS tests happen to finish last. So, the most robust solution seems to be waiting until _all_ the runs have passed.

> https://hynek.me/articles/ditch-codecov-python/

Interesting blog, thanks for sharing. I don't think we want to do what they do and fully ditch CodeCov: it's still got some great features, like an easily navigable website, badges, PR comments, PR statuses and so on.

On the other hand, combining reports & uploading after they complete as you suggest could work, as a workaround. The trade-off is then about whether adding this extra job makes things overall more or less complex than just the `n_builds` setting.

May I suggest we start with this simpler approach (the `.codecov.yml` config`) for now ? Then, welcome to open a PR to simplify the uploading part.

thatlittleboy: Sure, let's go with the simpler option for now.

PR #3083, opened 2023-07-10
Build wheels in parallel, add aarch64

connortann:
This PR should prepare us for a patch release `0.42.1`. I've bumped the version number and changelog accordingly.

Changes:

- Modifies our wheel job to run in parallel, which is must faster and makes the jobs more understandable.
- Changes the cibuildwheel dependencies from `test` to `test-core`, to only run the limited test set on the built wheels.
- Adds `aarch64` on linux, closes #3061
- Skips tests on emulated architectures (which are very slow)

Example [wheel build workflow run](https://github.com/slundberg/shap/actions/runs/5540996538) (passes).

Example job matrix:

image

PR #3087, opened 2023-07-12
MAINT: Add wheels for musllinux arch

### LINKED ISSUE #3089: ENH: Build wheels for musllinux
### Problem Description

We could publish [PEP-656](https://www.python.org/dev/peps/pep-0656/) musilinux wheels would be useful for use in docker containers (e.g. the Alpine image).

Suggested by @PrimozGodec .

The ease of doing this hinges somewhat on our dependencies:
- Numpy supports musllinux as of `1.25.1`, which is python 3.9+ (https://github.com/numpy/numpy/issues/20089)
- scikit learn doesn't currently ship with musllinux wheels

It may be that we have to wait until we've dropped python 3.8; a timeline for that is in #3057

PrimozGodec:
## Overview

Compared with Scipy, SHAP will offer all wheels except musllinux in the next releases. With this PR, I am also trying to add musllinux wheels that are also available for numpy/scipy. Musllinux wheels are mostly used for containers.

Currently, I am testing wheels building here https://github.com/PrimozGodec/shap/actions/runs/5533568616, but having musllinux wheels is not as necessary as other wheels, so it is not necessary to have them in next patch release.

## Checklist

- [x] Closes #3089

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -139,7 +139,11 @@ source_pkgs = ["shap"]
 combine = ["shap", "*/site-packages/shap"]
 
 [tool.cibuildwheel]
-skip = ["pp*", "*-musllinux_*"]
+# Restrict the set of builds to mirror the wheels available in scipy. See #3028
+# skip *-musllinux_aarch64 since numpy doesn't provid those wheels
+# skip cp38-musllinux_x86_64 since numpy never provided cp38 musllinux wheels
+#  they introduced musllinux in 1.25 when they already dropped cp38
+skip = ["pp*", "*-musllinux_aarch64", "cp38-musllinux_x86_64"]
 build-verbosity = 2
 # Change import-mode to ensure we test against installed package, not local project
 test-command = "pytest -v {project}/tests --import-mode=append"
@@ -147,7 +151,8 @@ test-extras = ["test-core", "plots"]
 # skip tests on cp38-macosx_x86_64 because of https://github.com/catboost/catboost/issues/2371
 # skip tests on emulated architectures, as they are very slow
 # skip tests on *-macosx_arm64 , as cibuildwheel does not support tests on arm64 (yet)
-test-skip = "cp38-macosx_x86_64 *-*linux_{aarch64,ppc64le,s390x} *-macosx_arm64"
+# skip tests on *-musllinux*" since llvmlite and numba do not provide musllinux wheels
+test-skip = "cp38-macosx_x86_64 *-*linux_{aarch64,ppc64le,s390x} *-macosx_arm64 *-musllinux*"
 
 [tool.cibuildwheel.linux]
 archs = ["x86_64", "aarch64"]


connortann:
Tagging relevant issue: #3089

connortann:
@PrimozGodec when prototyping on this PR, it might be helpful to merge in the `maint/wheel-parallel` branch from #3083 (which will hopefully be merged soon), as the builds will run a lot faster in parallel.

PrimozGodec:
After some work on this PR, I think we have two options:
1. Provide untested wheels
2. Wait with musllinux wheels until Numba provides musllinux wheels.

Reason: Wheels build successfully, but the testing part fails since it needs to install Numba and llvmlite, which do not provide musllinux wheels. Building Numba and llvmlite fail since building llvmlite needs building LLVM. Building this package is complicated (look here https://llvmlite.readthedocs.io/en/latest/admin-guide/install.html#things-to-fix-it).

I think it is better to ask (or help) llvmlite and Numba maintainers to provide wheels, and then we continue with work here or add tests (if we decide on option 1).

@connortann, what do you think? Is option 1 acceptable for you, or should we wait completely?

In both cases, I would ask Numba and llvmlite maintainers for wheels. As I already said, musllinux wheels are not as important since users must build many other packages anyway (including Numba and llvmlite, which seems more difficult than Shap).

connortann:
It's a good question. I think I am in favour of option (1) on balance, but I'm open to other views.

In the short term, it probably won't matter whether we provide musllinux wheels or not, given that most users are unlikely to be able to use them without numba support.

Longer term though, musllinux wheels may be helpful as adoption of musllinux increases. Besides, I can't think of any downsides of releasing the weels.

So, why don't we add wheels for musllinux, and skip the tests (with a comment explaining that the numba doesn't have wheels yet)

PrimozGodec:
@connortann agree with you. I skipped tests and now building is successful https://github.com/PrimozGodec/shap/actions/runs/5677105162/job/15384883519

I think it is ready to merge.

pyproject.toml
```diff
@@ -139,15 +139,17 @@ source_pkgs = ["shap"]
 combine = ["shap", "*/site-packages/shap"]
 
 [tool.cibuildwheel]
-skip = ["pp*", "*-musllinux_*"]
+# Restrict the set of builds to mirror the wheels available in scikit-learn. See #3028
+skip = ["pp*", "*-musllinux_aarch64", "cp38-musllinux_x86_64"]
```
connortann, line 146: What is the motivation behind the two new specific skips here?

Also, the comment about mirroring scikit-learn is probably no longer relevant, as we're now choosing to build a wider range of wheels.

PrimozGodec: I added comments that describe skips here. `*-musllinux_aarch64` is not provided since scipy/numpy doesn't provide them. We would need to build numpy in the wheel-building process to compile SHAP. `cp38-musllinux_x86_64` is not provided since numpy never had cp38 (musllinux) wheels. They introduced musllinux wheels after dropping cp38. The later can be removed when SHAP drop Python 3.8, which can be safely done in my opinion and according to NEP https://numpy.org/neps/nep-0029-deprecation_policy.html.

PrimozGodec: > Also, the comment about mirroring scikit-learn is probably no longer relevant, as we're now choosing to build a wider range of wheels.

It is true. I changed the comment from scikit to scipy. As I see in #3028 it was already discussed that we provide wheels that scipy has (Numpy has everything that scipy has +win32 which was not included as discussed in #3028).

connortann (approved):
Thanks again for the PR! LGTM.

mpierrau:
Hi! Have you reached out to the numba team about whether MUSL wheels will be available for their package any time soon? :)

mpierrau:
FYI it seems that `scikit-learn` also won't start releasing MUSL wheels any time soon: https://github.com/scikit-learn/scikit-learn/issues/27004#issuecomment-1966926217

Edit (from comment):
> EDIT: note that the maintainers might reconsider that decision (not to provide musllinux wheels on pypi.org) based on expressed user feedback e.g. next year.

PR #3090, opened 2023-07-12
FIX: remove ipython import warning from top-level shap import

### LINKED ISSUE #3086: IPython warning
### Issue Description

Is there any reason why the IPython warning could not be lazily imported into the function that requires it? This isn't a bug, more of an annoyance to see the warning when the function that requires it isn't used.

`shap/plots/_image.py`
Function: `image_to_text`

```python
import shap
```

### Expected Behavior

No `IPython could not be loaded` warning in the terminal.

### Bug report checklist

### Installed Versions

0.42

connortann:
Fixes #3086

Removes the warning raised in `_image.py` if Ipython is not installed, and instead lazily raises an `ImportError` if the relevant function is called.

Also raises a more informative error in `_text.py`, where IPython's `HTML` and `display` functions may not be defined if not imported.

### DIFF
--- a/shap/plots/_image.py
+++ b/shap/plots/_image.py
@@ -1,20 +1,20 @@
 import json
 import random
 import string
-import warnings
 from typing import Optional
 
 import matplotlib.pyplot as pl
 import numpy as np
 from matplotlib.colors import Colormap
 
-from .._explanation import Explanation
-from ..utils import ordinal_str
-
 try:
     from IPython.display import HTML, display
+    have_ipython = True
 except ImportError:
-    warnings.warn("IPython could not be loaded!")
+    have_ipython = False
+
+from .._explanation import Explanation
+from ..utils import ordinal_str
 from ..utils._legacy import kmeans
 from . import colors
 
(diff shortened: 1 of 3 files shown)


shap/plots/_image.py
```diff
@@ -190,6 +190,8 @@ def image_to_text(shap_values):
         for each sample
 
     """
+    if not have_ipython:
+        raise ImportError("IPython is required for this function but could not be imported")
```
LukeHankey:
```suggestion
        raise ImportError("IPython is required for this function but could not be imported. Run `pip install ipython` to fix this.")
```

This last part you added in `_text.py` is more informative and gives direction to the user.

LukeHankey:
There is also a similar error message at `shap/__init__.py:22` for matplotlib. Should this be changed from a ValueError to an ImportError?

shap/plots/_image.py
connortann: Good idea

LukeHankey (approved):

connortann:
I also propose changing the import warning in `shap.__init__.py` to a full ImportError. If the plotting functions are called and do not work, an explicit exception seems appropriate.

LukeHankey:
> I also propose changing the import warning in `shap.__init__.py` to a full ImportError. If the plotting functions are called and do not work, an explicit exception seems appropriate.

In that case, you won't need to import warnings in `shap.__init__.py` then

PR #3091, opened 2023-07-13
ENH: Explicitly define public API in modules

connortann:
## Overview

Changes:

- Adds `__all__` to the various submodule `__init__.py`s, to define the public API
- Sort imports
- Enables ruff linter
  - Removes `noqa` from top level init.py
  - Removes "unused import" exception

## Motivation

According to PEP 8, we should define the public API:

> To better support introspection, modules should explicitly declare the names in their public API using the __all__ attribute.

The `__all__` variable lets type checkers know what is part of the public API, and prevents false positive ruff `unused-import` errors.

### DIFF
--- a/shap/__init__.py
+++ b/shap/__init__.py
@@ -1,22 +1,21 @@
-# flake8: noqa
 
 __version__ = "0.42.1"
 
-from ._explanation import Explanation, Cohorts
-
 # explainers
-from .explainers._explainer import Explainer
-from .explainers._kernel import Kernel as KernelExplainer
-from .explainers._sampling import Sampling as SamplingExplainer
-from .explainers._tree import Tree as TreeExplainer
-from .explainers._gpu_tree import GPUTree as GPUTreeExplainer
+
+from ._explanation import Cohorts, Explanation
+from .explainers import other
+from .explainers._additive import Additive as AdditiveExplainer
 from .explainers._deep import Deep as DeepExplainer
+from .explainers._explainer import Explainer
+from .explainers._gpu_tree import GPUTree as GPUTreeExplainer
 from .explainers._gradient import Gradient as GradientExplainer
+from .explainers._kernel import Kernel as KernelExplainer
 from .explainers._linear import Linear as LinearExplainer
 from .explainers._partition import Partition as PartitionExplainer
 from .explainers._permutation import Permutation as PermutationExplainer
-from .explainers._additive import Additive as AdditiveExplainer
-from .explainers import other
+from .explainers._sampling import Sampling as SamplingExplainer
+from .explainers._tree import Tree as TreeExplainer
 
 _no_matplotlib_warning = "matplotlib is not installed so plotting is not available! Run `pip install matplotlib` " \
                          "to fix this."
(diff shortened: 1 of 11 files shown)


shap/__init__.py
```diff
@@ -1,22 +1,21 @@
-# flake8: noqa
 
 __version__ = "0.42.1"
 
-from ._explanation import Explanation, Cohorts
-
 # explainers
+
+from ._explanation import Cohorts, Explanation
```
thatlittleboy, line 4: small nit, but Explanation and Cohorts aren't explainers, I'll prefer to keep it where it was

shap/__init__.py
```diff
@@ -33,24 +32,26 @@ def __getattribute__(self, item):
...
     from .plots._partial_dependence import partial_dependence as partial_dependence_plot
-    from .plots._bar import bar_legacy as bar_plot
-    from .plots._waterfall import waterfall as waterfall_plot
-    from .plots._group_difference import group_difference as group_difference_plot
+    from .plots._scatter import dependence_legacy as dependence_plot
     from .plots._text import text as text_plot
+    from .plots._waterfall import waterfall as waterfall_plot
 else:
     summary_plot = unsupported
     decision_plot = unsupported
```
thatlittleboy, line 58: could you also rearrange the variables within this else to match the new import order from above?

shap/__init__.py
```diff
@@ -73,13 +74,56 @@ def __getattribute__(self, item):
...
+    "waterfall_plot",
+
+    # Other stuff
+    "datasets",
+    "utils",
+    "links",
+    "ActionOptimizer",
+    "kmeans",
+    "sample",
+    "approximate_interactions",
```
thatlittleboy: nit: also arranging it in the same order as the imports

```suggestion
__all__ = [
    "Cohorts",
    "Explanation",

    # Explainers
    "other",
    "AdditiveExplainer",
    "DeepExplainer",
    "Explainer",
    "GPUTreeExplainer",
    "GradientExplainer",
    "KernelExplainer",
    "LinearExplainer",
    "PartitionExplainer",
    "PermutationExplainer",
    "SamplingExplainer",
    "TreeExplainer",

    # Plots
    "plots",
    "bar_plot",
    "summary_plot",
    "decision_plot",
    "multioutput_decision_plot",
    "embedding_plot",
    "force_plot",
    "getjs",
    "initjs",
    "save_html",
    "group_difference_plot",
    "image_plot",
    "monitoring_plot",
    "partial_dependence_plot",
    "dependence_plot",
    "text_plot",
    "waterfall_plot",

    # Other stuff
    "datasets",
    "links",
    "utils",
    "ActionOptimizer",
    "approximate_interactions",
    "sample",
    "kmeans",
```

thatlittleboy (approved):
Couple of nitpicks on the orderings, but otherwise LGTM

choi-aera-fightformyway-thumbsup

shap/__init__.py
connortann: Would it be better to sort each group in `__all__` in alphabetical order? That could make it easier for users to find a particular function, for example when using autocomplete in Jupyter.

The order that `isort` chooses is pretty arbitrary from the point of view of users, as it's determined by the order of our private submodules. e.g. `dependence_plot` comes later in the list as it happens to be defined in `_scatter.py`.

thatlittleboy: Does the ordering in Jupyter's autocomplete options depend on how we sort `__all__`?
If it does then I agree with you.

Otherwise, I think the `__all__` ordering is more for maintainers than users. It's easier to edit both the imports and the `__all__` list in tandem when they are ordered the same way.

connortann: Ah, I tested it and I was mistaken, actually Jupyter just uses alphabetic order. I'll do as you suggest and order `__all__` in accordance with the imports.

PR #3094, opened 2023-07-16
fix: support LightGBM ensemble containing single leaf trees (stump)

### LINKED ISSUE #2044: Tree Explainer bug: unable to handle trees with a single leaf

thatlittleboy:
## Overview

The changes in this PR should make it such that `TreeExplainer` instantiated on a LightGBM ensemble model with >0 Tree with only a single leaf will no longer error out on "KeyError: "leaf_index" (which results in `SingleTree` not having the `.values` attribute set). Bug was initially reported in #2044.

NOTE:

There is still a small kink in the solution, which is that the `node_sample_weight` (== "leaf_count") of the stump's leaf is currently set to 0 instead of number of samples used to train the model (as you would expect).

There's not much we can do to fix it here, because this information is not even present in the `dump_model()` output produced by LightGBM's booster. I've raised an upstream issue + attempting to resolve it upstream https://github.com/microsoft/LightGBM/pull/5964, but that might take some time to get merged.

Regardless, this PR does at least make sure that `TreeExplainer` doesn't error out, so I think it's worth a merge.

## Checklist

- [X] Closes #2044  


shap/explainers/_tree.py
```diff
@@ -1476,7 +1478,12 @@ def buildTree(index, node):
                     self.features[vleaf_idx] = -1
                     self.thresholds[vleaf_idx] = -1
                     self.values[vleaf_idx] = [vertex["leaf_value"]]
-                    self.node_sample_weight[vleaf_idx] = vertex["leaf_count"]
+                    # FIXME: "leaf_count" currently doesn't exist if we have a stump tree.
+                    # We should be technically be assigning the number of samples used to
+                    # train the model as the weight here, but unfortunately this info is
+                    # currently unavailable in `tree`, so we set to 0 first.
```
connortann, line 1484: Will using `0` as a placeholder be problematic if this variable is used as a weight?

connortann (approved):
LGTM!

shap/explainers/_tree.py
thatlittleboy: I believe the functions in `tree_shap.h` have already taken the 0's into account.

Also, (1) the `node-sample_weights` are instantiated as an array of 0's in general, and (2) tests are passing, so we should be good. :)

PR #3096, opened 2023-07-16
fix: pytest imports for isolation forests

thatlittleboy:
## Overview

* The pytest importorskip statements were written incorrectly, resulting in the IsolationForest tests being skipped all the time. This PR fixes that.

## Checklist

- Closes #xxxx  

### DIFF
--- a/tests/explainers/test_tree.py
+++ b/tests/explainers/test_tree.py
@@ -9,6 +9,7 @@ import pandas as pd
 import pytest
 import sklearn
 import sklearn.pipeline
+from sklearn.utils import check_array
 
 import shap
 from shap.explainers._tree import SingleTree
@@ -392,8 +393,6 @@ def _average_path_length(n_samples_leaf):
 
     For use in isolation forest tests.
     """
-    from sklearn.utils import check_array
-
     n_samples_leaf = check_array(n_samples_leaf, ensure_2d=False)
 
     n_samples_leaf_shape = n_samples_leaf.shape
--- a/tests/explainers/test_tree.py
+++ b/tests/explainers/test_tree.py
@@ -386,9 +386,36 @@ def test_catboost_categorical():
 # TODO: Test tree_limit argument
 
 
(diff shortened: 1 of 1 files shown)


tests/explainers/test_tree.py
```diff
@@ -387,8 +387,9 @@ def test_catboost_categorical():
 
 
 def test_isolation_forest():
-    IsolationForest = pytest.importorskip("sklearn.ensemble.IsolationForest")
-    _average_path_length = pytest.importorskip("sklearn.ensemble._iforest._average_path_length")
+    from sklearn.ensemble import IsolationForest
+    from sklearn.ensemble._iforest import _average_path_length
```
connortann: We should be cautious about importing private objects from other libraries: `_iforest` is not part of sklearn's public API and could change or break without notice.

Would it be better to copy the relevant [private function](https://github.com/scikit-learn/scikit-learn/blob/150ff34e85f91a3a19a29dc4e19d50131fbb0251/sklearn/ensemble/_iforest.py#L531) and vendor it here?

connortann (changes requested):

tests/explainers/test_tree.py
thatlittleboy: done

tests/explainers/test_tree.py
```diff
@@ -386,9 +386,37 @@ def test_catboost_categorical():
 # TODO: Test tree_limit argument
 
 
+def _average_path_length(n_samples_leaf):
+    """
+    Vendored from: https://github.com/scikit-learn/scikit-learn/blob/399131c8545cd525724e4bacf553416c512ac82c/sklearn/ensemble/_iforest.py#L531
+
+    For use in isolation forest tests.
+    """
+    from sklearn.utils import check_array
```
connortann: Nit: as a core dependency, this import could follow PEP8 and go at the top.

connortann (approved):
One optional nitpick but otherwise LGTM!

PR #3106, opened 2023-07-20
Fix/update js deps

remborg:
## Overview

Adding Jest and Babel to allow testing javascript code. 
A simple test is checking that the AdditiveForceVisualizer react component HTML renders correctly.

Test can be run with `npm run test`.

I've deleted the files 'javascript/random-explanation.js' and 'javascript/test.js'. They seemed to be old test code that's not working anymore.


.github/workflows/test_js.yml
```diff
@@ -0,0 +1,25 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: lts/Hydrogen
+    - run: npm ci
+      working-directory: ./javascript
+    - run: npm run build
+      working-directory: ./javascript
```
connortann, line 27: May I suggest adding `npm run test` here, to get the JS unit tests running on CI?

connortann (approved):
Thank you remborg, this is fantastically helpful. In particular, it gives us a great foundation to add more tests in future, and evaludate whether dependabot updates cause issues or not.

Issue #2951, opened 2023-06-03
CI Bug: Required Checks block non-python PRs from being completed


connortann:
Some PRs are unable to be merged at present: the python unit tests do not run unless the path filter conditions are met, but the PR cannot be completed as the tests are configured as [required status checks](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches#require-status-checks-before-merging) in the repository settings.

Example PR #2949 :
![image](https://github.com/slundberg/shap/assets/71127464/c5812418-6287-451e-bbe8-3ac21ebf0abf)

This could be resolved by:

1. Changing the branch protection rules, for example to require all Checks _that run_ to pass
2. Just running all the tests on all PRs.

We cannot do (1) without admin-level permissions, so as an interim measure we could implement (2).

PR #3107, opened 2023-07-20
MAINT: Enable CI test skipping by path filter

connortann:
Changes:

- On PRs, skips the python tests if the shap python package has not been changed.
- On push to master, always runs the tests.

Related:  #2951

GitHub Docs: [handling skipped but required checks](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/troubleshooting-required-status-checks#handling-skipped-but-required-checks)

Other examples in the action: <https://github.com/dorny/paths-filter>

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -1,21 +1,12 @@
 name: tests
 
 on:
-  # only run tests and coverage when src code changes
   push:
     branches:
       - master
   pull_request:
     branches:
       - master
-    # Interim measure: run all tests on all PRs, as these Checks
-    # are configured as Required Status Checks. See #2951
-    # paths:
-    #   - "shap/**"
-    #   - "tests/**"
-    #   - "data/**"
-    #   - ".github/workflows/run_tests.yml"
-    #   - "setup.py"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


thatlittleboy:
@connortann How do we feel about removing the Required Status Checks? Then we won't need this workaround.

connortann:
I don't really have a preference either way. I suppose it's nice to have the check "required", as it means it's less likely that the CI pipeline will break due to error or negligence. On the other hand, the basic path filter had a simpler implementation.

At present I don't think either of us have permissions to change the required checks, so this implementation could be a starting point at least. We could revisit it if&when we get the ability to change the Required Checks setting, which we'll need to do anyway whenever python 3.8 is dropped.

thatlittleboy:
I prefer the simplicity of the path filter, tbh. I don't see much of a benefit in keeping the Required checks.

Because even if a merge happened by mistake AND that merge causes one of the "required" tests to fail, we can just easily revert the change.

I think if we're agreeable, let's just get Scott to remove the checks? For something so trivial I don't think an "intermediate" solution is required.

connortann:
Fair, good point. I'll email him..

connortann:
OK, done :)

thatlittleboy (approved):
Nice~

baby-biscuit-expression-change

PR #3108, opened 2023-07-20
Add explicit dependabot config

connortann:
Creates an explict dependabot configuration, as recommended in the GitHub docs.

For now, this just looks as javascript packages. In future, it could be useful if we pin dev dependencies. If we end up with too many PRs, we could configure the `groups` parameter to make dependabot do updates in bulk.

Should help get dependabot working again: since the migration to the SHAP org, the bot cannot rebase old PRs e.g. #2826

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/javascript/"
+    schedule:
+      interval: "weekly"


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/javascript/"
+    schedule:
+      interval: "weekly"
```
thatlittleboy, line 6: is there a longer interval we can set? weekly is way too frequent, IMO. I think monthly is a good compromise, what do you think?

connortann: I agree that `monthly` is probably better for the longer term, as it would reduce the frequency of PRs to merge. Another setting that would help in this regard is the `group` parameter, which means multiple updates can be bunched together in a single dependabot PR.

However, in the short term I'd prefer to keep it a weekly or daily, so we can test it's working as expected. In particular, I'd like to see if the existing dependabot PRs are closed or rebased automatically. Then, once things settle down on the javascript side (and we've fixed the remaining major security alerts!), we could tweak the above two settings to reduce the frequency of updates.

thatlittleboy (approved):
spongebob-patrick-hooray

PR #3074, opened 2023-07-08
chore: Add a Jupyter style guide and supporting infra

thatlittleboy:
## Overview

This PR is in support and born out of PR #3037. The idea is to use this thread to discuss how we should handle Jupyter notebooks so that we can receive better quality contributions for #3036.

Many of the changes here are inspired by https://www.pymc.io/projects/docs/en/stable/contributing/jupyter_style.html.

### Discussions w.r.t. style guide

1. Using sphinx cross-references rather than URLs to refer to docstrings or other notebooks  \[**Status**: Under discussion\]
1. Add pre-commit to enforce notebook styles  \[**Status**: Active\]
    - `ruff` linting in code cells via `ruff`
    - `black` formatting in code cells via `black-jupyter` (instead of `nbqa` so that when we eventually apply black to the code-base, we can automatically ensure the black version applied to both `.py` and `.ipynb` is the same).
    - pygrep check for `watermark` extension (inspired from pymc docs)
1. Add [watermark](https://github.com/rasbt/watermark) cell to the notebooks. (enforced by pre-commit)
2. <s>Using `nb-clean` to clear notebook metadata (TBD)</s> We'll come back to this again if needed.

### Outside the scope of this PR

I propose the following scopes to be **outside** of the initial style guide + the efforts over at #3036 (at least, initially). These should probably be their own issues.

- [ ] Integrating MyST: I propose that this should be a separate effort on its own, and not be lumped together with #3036. It's a pretty large structural change on its own, and detracts from the main objective of #3036, which is to ensure the notebooks are runnable and updated. That said, I do like MyST and would be happy to transition to MyST in the medium term.
- [ ] Executing notebooks in CI/CD rather than committing in the notebook outputs. The idea of this is to reduce merge conflicts and diffs by only committing in the notebook source. Then the notebooks are executed in CI/CD in a controlled environment, before we run sphinx. This makes sure that the notebooks are always up-to-date, and errors if we change the API but forget to change the notebook examples. Again, this is nice-to-have, but I think is outside the scope.

## Checklist

- [X] Supports #3036.  
- [X] All [pre-commit checks](https://pre-commit.com/#install) pass.
- [X] Unit tests added (if fixing a bug or adding a new feature)
- [ ] Added entry to `CHANGELOG.md` (if changes will affect users)

Issue #3036, opened 2023-06-23
[Meta-issue] Notebooks are outdated / non-runnable


thatlittleboy:
## Background

We should also make sure that our documentation is kept up to date.

A scour through the open issues in this repo and also on [StackOverflow](https://stackoverflow.com/questions/tagged/shap) shows that the outdated documentation (or lackthereof) is causing confusion among our users.

Just some examples: 
- https://github.com/slundberg/shap/issues/2786
- https://stackoverflow.com/q/76522186

---

## Plan

The plan here is to thoroughly go through each and every notebook example that we have, to:

- Run the notebook from top-to-bottom and ensure there are no errors.
- Update the prose where necessary to provide better clarity and fix any typos
- Update the code (where necessary and appropriate) to demonstrate up-to-date API & python features, such as:
  - Replacing the deprecated boston dataset
  - Using the `Explanation` API for plotting (e.g. `shap.plots.beeswarm` rather than `shap.summary_plot`)

An example of such a PR: [#3037](https://github.com/slundberg/shap/pull/3037)

## Useful info for contributors

To potential contributors (thank you in advance!), please stick to one notebook per PR when contributing.  
Also, any help in updating the GPU-related notebooks would be very much appreciated.  

We have two linting checks for notebooks, each of which has an "exclude list" that ignores notebooks that haven't been cleaned. When a notebook has been fixed up, it should be **removed from the "exclude" list** in these places:
  - The `run_notebooks_timeout` job in `scripts/run_notebooks_timeouts.py`
  - The `nbcheckorder` job in `.pre-commit-config.yaml`

For more details on how to preview the built documentation, see the [contributing guide](https://shap.readthedocs.io/en/latest/contributing.html#documentation).

## TODO

- notebooks
    - text_examples
        - language_modelling
            - [x] Language Modeling Explanation Demo.ipynb (#3691)
        - text_generation
            - [x] Open Ended GPT2 Text Generation Explanations.ipynb
        - question_answering
            - [x] Explaining a Question Answering Transformers Model.ipynb
        - translation
            - [ ] Machine Translation Explanations.ipynb
        - summarization
            - [ ] Abstractive Summarization Explanation Demo.ipynb
        - text_entailment
            - [ ] Textual Entailment Explanation Demo.ipynb
        - sentiment_analysis
            - [ ] Using custom functions and tokenizers.ipynb
            - [ ] Emotion classification multiclass example.ipynb
[...]

PR #3111, opened 2023-07-21
DOCS: Enable pre-commit checks for notebooks

connortann:
## Overview

Adds two setsof pre-commit checks to run on notebooks.

1) A basic set of checks, that will run on all the notebooks:

- `check-json` to ensure ipynb files are valid JSON
- `nbqa ruff`, initially with a restricted subset of linting rules

2) A more stringent set of checks for notebooks which have been brought up to scratch:

- `nbqa ruff` with the same set of Ruff rules that are applied to the python package
- `black` formatting

This PR also fixes up some small logical errors in a handful of notebooks to get these two hooks passing, and applies `black` to two example notebooks as a starting point to illustrate the second group of pre-commit checks.

Related: #3074, #3036

## Discussion

It would be helpful to enable precommit hooks to help us review PRs relating to documentation. A lot of notebooks are in a messy state, so it's not possible to enable many linting rules without causing a really large diff. So, this PR is intended as a good starting point that we can build on over time.

I envisage that as each notebook is tidied up and brought up to scratch, we can add it to the list of checks in the second group of precommit checks above. This gives us a way to track which notebooks are clean, and ensure that cleaned notebooks are easily "locked-in" to a tidy state.

Related: submitted a PR to ensure that notebooks are picked up by default by the check-json pre commit hook:
https://github.com/pre-commit/identify/pull/401


connortann:
@thatlittleboy do you have any thoughts on what hooks we should start with?

This is still a draft, as I haven't run the hooks on the existing files (yet).

thatlittleboy:
> @thatlittleboy do you have any thoughts on what hooks we should start with?
> 
> This is still a draft, as I haven't run the hooks on the existing files (yet).

I think the two you have at the moment are sufficient, I also had `black` over at #3074, plus a couple more like checking the watermark and ruff linting, but I think those can come later along with #3074.

> This PR will inevitably lead to a large bulk-update of all notebooks

We also have the option to *not* run on all the notebooks in this PR, no?

connortann:
> We also have the option to not run on all the notebooks in this PR, no?

I don't think that's possible: our CI configuration mandates that all pre-commit hooks must pass on all files.

I've updated this PR to be a better starting point, just checking that Ruff passes. EDIT: updated description above with current proposal.

owenlamont:
I don't think you need the nbQA pre-commit anymore (I have gone down that path before) as [ruff now supports running on Jupyter notebooks](https://beta.ruff.rs/docs/usage/#pre-commit) just by tweaking some of its pre-commit CLI options. Likewise, its likely unnecessary to add Black, as the[ ruff formatter is nearly at feature parity with Black](https://github.com/astral-sh/ruff/issues/5828) (this is being very actively worked on) and soon hopefully will be.

owenlamont:
I recently updated ruff and added the auto-fix option in [PR-3208](https://github.com/shap/shap/pull/3208). There was no current issues with the Python modules but if that causes too much pain with the notebooks in this PR feel free to remove that auto fix option.

connortann:
> I don't think you need the nbQA pre-commit anymore (I have gone down that path before) as [ruff now supports running on Jupyter notebooks](https://beta.ruff.rs/docs/usage/#pre-commit) just by tweaking some of its pre-commit CLI options. Likewise, its likely unnecessary to add Black, as the[ ruff formatter is nearly at feature parity with Black](https://github.com/astral-sh/ruff/issues/5828) (this is being very actively worked on) and soon hopefully will be.

It's great to see that ruff can run on notebooks directly. I'll take a look and use that rather than indirectly via nbqa.

Ruff's formatter doesn't seem to be released yet, so to start with I'd advocate with sticking with `black` which is a very mature project. It should be straightforward to switch to ruff's formatter in future, should we so desire.

connortann:
As a reminder to myself for later, here's the blog post announcing stable Jupyter support in Ruff:
<https://astral.sh/blog/ruff-v0.0.285>

owenlamont:
Hi @connortann - I didn't realise the Rust formatter hadn't been released yet. That's fine then - if we can implement this without needing nbQA that'd be great. I definitely would like Black included in that case.

thatlittleboy (approved):
Approving in-principle, I'll leave you to merge when you are ready @connortann 

nicholas-cage-letsgo

connortann:
Thanks @thatlittleboy . I added some better comments, and simplified the pre-commit config as best I could.

In future, we could aim to run `ruff` just once on all files. For now though, I think a second run of Ruff with an "opt-in" selection of cleaned notebooks is the simplest way to go.

One subtlety I noticed is that Ruff's `include` and `exclude` configuration options are ignored when running with pre-commit. So, we must put the selection of files to include or exclude in `.pre-commit-config.yaml`, not in `pyproject.toml`.

PR #3125, opened 2023-07-22
docs: fix the heading levels in some notebooks

thatlittleboy:
## Overview

* Fix heading levels used in some Jupyter notebooks, to prevent them from appearing in the sidebar

Before:

These 3 are actually subheadings in the "SHAP Values for Multi-Output Regression Models" notebook, but because they were of the same heading level as the title of the notebook, it was showing up in the sidebar.

After:


connortann (approved):
LGTM!

PR #3126, opened 2023-07-22
fix: slice `feature_names` properly in Explanation objects with square `.values`

### LINKED ISSUE #2394: Feature Names not set correctly if Feature Set is too small
When defining an explainer with set `feature_names`, the size of the dataset to calculate shap values for `X` cannot be smaller than `feature_names`, for the features in a `waterfall` plot to be displayed correctly.

```
X.shape 
>>> (500, 10)
shap_explainer = shap.Explainer(model, X, feature_names=X.columns)
shap_values = shap_explainer(X.iloc[:5])
shap.waterfall_plot(shap_values[0])  # this will not label the feature names correctly 
```

the reason is that only the array `shap_values` has its attribute `feature_names` set correctly, but not the individual rows, `shap_values[0].feature_names` only contained a single feature name of the list

### LINKED ISSUE #2699: Symmetrical SHAP array shape fails to slice on feature names (e.g., shap_values[:,"feature_1"])
Lets say I have a pandas DataFrame that contains some features and a dependent variable...

df_testing_shap =
<html>
<body>

 index | y_val | feature_1 | feature_2 | feature_3
-- | -- | -- | -- | --
1 | 10.041427 | 47 | 0.434511 | 6
2 | 9.841583 | 50 | 0.452915 | 9
3 | 9.713457 | 34 | 0.564274 | 9

</body>
</html>

Now lets train our model and generate SHAP values (I am using XGBoost, make sure this is installed and imported to replicate the code)...
```python
# Split X and Y variables
df_x = df_testing_shap[["feature_1", "feature_2", "feature_3"]]
df_y = df_testing_shap["y_val"]

# Train model
hp_regressor = xgb.XGBRegressor()
hp_regressor.fit(df_x, df_y)

# Generate SHAP values
explainer = shap.Explainer(hp_regressor)
shap_values = explainer(df_x)

# Print shape info
print(f"df_x shape: {df_x.shape}")
print(f"SHAP shape: {shap_values.shape}")
```
Output...
```
df_x shape: (3, 3)
SHAP shape: (3, 3)
```

Note that the resulting SHAP array is symmetrical (the rows and columns are equal in length). Now if we try to slice this data by one of our feature names we get an error...
```python
shap_values[:,"feature_1"]
```
Output...
```
IndexError                                Traceback (most recent call last)
[<ipython-input-45-ac5b665a088f>](https://localhost:8080/#) in <module>
----> 1 shap_values[:,"feature_1"]

[/usr/local/lib/python3.7/dist-packages/shap/_explanation.py](https://localhost:8080/#) in __getitem__(self, item)
    419         if new_self is None:
    420             new_self = copy.copy(self)
--> 421         new_self._s = new_self._s.__getitem__(item)
    422         new_self.op_history.append({
    423             "name": "__getitem__",

[/usr/local/lib/python3.7/dist-packages/slicer/slicer.py](https://localhost:8080/#) in __getitem__(self, item)
    110                 index_slicer = AtomicSlicer(index_tup, max_dim=1)
    111                 slicer_index = index_slicer[tracked.dim]
--> 112                 sliced_o = tracked[slicer_index]
    113                 sliced_dim = resolve_dim(index_tup, tracked.dim)
    114 

[/usr/local/lib/python3.7/dist-packages/slicer/slicer_internal.py](https://localhost:8080/#) in __getitem__(self, item)
     67 
     68         # Slice according to object type.
---> 69         return UnifiedDataHandler.slice(self.o, index_tup, self.max_dim)
     70 
     71 

[/usr/local/lib/python3.7/dist-packages/slicer/slicer_internal.py](https://localhost:8080/#) in slice(cls, o, index_tup, max_dim)
    581 
    582         is_element, sliced_o, cut = head_slice(o, index_tup, max_dim)
--> 583         out = tail_slice(sliced_o, index_tup[cut:], max_dim - cut, is_element)
    584         return out
    585 

[/usr/local/lib/python3.7/dist-packages/slicer/slicer_internal.py](https://localhost:8080/#) in tail_slice(cls, o, tail_index, max_dim, flatten)
    437                 return AtomicSlicer(o, max_dim=max_dim)[tail_index]
    438         else:
--> 439             inner = [AtomicSlicer(e, max_dim=max_dim)[tail_index] for e in o]
    440             if _safe_isinstance(o, "numpy", "ndarray"):
    441                 import numpy

[/usr/local/lib/python3.7/dist-packages/slicer/slicer_internal.py](https://localhost:8080/#) in <listcomp>(.0)
    437                 return AtomicSlicer(o, max_dim=max_dim)[tail_index]
    438         else:
--> 439             inner = [AtomicSlicer(e, max_dim=max_dim)[tail_index] for e in o]
    440             if _safe_isinstance(o, "numpy", "ndarray"):
    441                 import numpy

[/usr/local/lib/python3.7/dist-packages/slicer/slicer_internal.py](https://localhost:8080/#) in __getitem__(self, item)
     67 
     68         # Slice according to object type.
---> 69         return UnifiedDataHandler.slice(self.o, index_tup, self.max_dim)
     70 
     71 

[/usr/local/lib/python3.7/dist-packages/slicer/slicer_internal.py](https://localhost:8080/#) in slice(cls, o, index_tup, max_dim)
    580         tail_slice = cls.type_map[o_type].tail_slice
    581 
--> 582         is_element, sliced_o, cut = head_slice(o, index_tup, max_dim)
    583         out = tail_slice(sliced_o, index_tup[cut:], max_dim - cut, is_element)
    584         return out

[/usr/local/lib/python3.7/dist-packages/slicer/slicer_internal.py](https://localhost:8080/#) in head_slice(cls, o, index_tup, max_dim)
    416         cut_index = index_tup[:cut]
    417         is_element = any([True if isinstance(x, int) else False for x in cut_index])
--> 418         sliced_o = o[cut_index]
    419 
    420         return is_element, sliced_o, cut

IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices
```

The error seems to be in the `__getitem__` method from the `shap._explanation.Explanation` class. I have tried to monkey patch this class with some debugging code, but without prior knowledge of this method I think it would take quite a while to get to the bottom of this.

As a workaround, it should be easy enough to build a DataFrame containing the SHAP values with columns set to feature names, then I can make whatever selections I want. But I think it would be better to have a proper fix so I can continue to use the built-in plots and what not.

Any ideas on this one? - Thanks in advance!

### LINKED ISSUE #2722: "string index out of range" for waterfall plot
When producing a SHAP waterfall plot, I got the 'string index out of range' error.  I had 7 features in my model, and 7 days of predictions in my test set, and I was attempting to produce a waterfall plot for the first day in the test set. After trying many scenarios, I discovered if I simply removed a row of data in the test set, the waterfall plot generated without error.

On further investigation it looks like if the number of features is the same as the number of rows in the dataset, then the error occurs. I cannot attach my original code but I have replicated it in a simple example with 12 features where the waterfall plot works correctly if the number of rows is greater than or less than the number of features, but errors when the two are the same.

import xgboost
import shap

###  Example of waterfall plot NOT working correctly when number of rows (12) == number of features (12) #############
X,y = shap.datasets.adult()
X = X.iloc[0:12]
y = y[0:12]
model = xgboost.XGBClassifier(eval_metric="logloss").fit(X, y)
shap_values = explainer(X)
print("Length of X:",len(X),"; Length y:",len(y),"; Number of features:",len(shap_values[0].feature_names))
print("\nFeature Names:",shap_values[0].feature_names)
shap.plots.waterfall(shap_values[0])
##### Number of features is incorrectly listed as 3. Feature names are incorrectly listed as age only. String Index out of range error.

Output:
Length of X: 12 ; Length y: 12 ; Number of features: 3
Feature Names: Age
IndexError: string index out of range 

###  Example of waterfall plot working correctly when number of rows (13) > number of features (12) ####################
X,y = shap.datasets.adult()
X = X.iloc[0:13]
y = y[0:13]
model = xgboost.XGBClassifier(eval_metric="logloss").fit(X, y)
shap_values = explainer(X)
print("Length of X:",len(X),"; Length y:",len(y),"; Number of features:",len(shap_values[0].feature_names))
print("\nFeature Names:",shap_values[0].feature_names)
shap.plots.waterfall(shap_values[0])
##### Number of features is correctly listed as 12. Feature names are correctly listed. Plot generates.
Output:
Length of X: 13 ; Length y: 13 ; Number of features: 12
Feature Names: ['Age', 'Workclass', 'Education-Num', 'Marital Status', 'Occupation', 'Relationship', 'Race', 'Sex', 'Capital Gain', 'Capital Loss', 'Hours per week', 'Country']

image

thatlittleboy:
## Overview

* Ensures that we slice `feature_names` **correctly** in Explanation objects with "square" `.values` ("square" := the shap values array has the same number of feature columns as number of samples).
* Add a test to ensure we don't have regressions in the future.

### Explanation

First, some background. For 2D arrays, there is an ambiguity in how to assign the feature names to the slicer index.
E.g. if `feature_names` is a list of 5 elements, say `[a,b,c,d,e]`, and the `shap_values` is a (5,5) array, it's ambiguous
whether the axis=0 or axis=1 in `shap_values` refers to the "feature columns".

Previously, the code assigned `axis=0` with higher priority. This causes problems like #2722 because when users index into the explanation object like `explanation[0]` expecting to get the first sample's shap values / data (e.g. for inputting into waterfall plot),  
`Slicer` would slice this `[0]` into axis=0, which it incorrectly assumes to be the `.feature_names`.
This means that the resulting Explanation object's `.feature_names` would be (`a` == `feature_names[0]`) instead of the full `feature_names[:]` array.

See the test code for an explicit example.

This PR changes such that at least for 2+ dimensional square arrays, we always assume the feature column is on axis=1 instead of axis=0. I believe this to be a more reasonable assumption.
Since most of the time, the 2D shap values arrays are assembled as (# samples, # features).

For non-square arrays, there's no change, since there is no ambiguity which axis the `feature_names` array should refer to (obviously, it would be the axis with the same length as `feature_names`).

## Checklist

- [X] Closes #2699, closes #2722, closes #2394 

### DIFF
--- a/shap/_explanation.py
+++ b/shap/_explanation.py
@@ -118,13 +118,13 @@ class Explanation(metaclass=MetaExplanation):
         if output_names is None and len(self.output_dims) == 1:
             output_names = [f"Output {i}" for i in range(values_shape[self.output_dims[0]])]
 
-        if len(_compute_shape(feature_names)) == 1: # TODOsomeday: should always be an alias once slicer supports per-row aliases
-            if len(values_shape) >= 1 and len(feature_names) == values_shape[0]:
-                feature_names = Alias(list(feature_names), 0)
-            elif len(values_shape) >= 2 and len(feature_names) == values_shape[1]:
+        if len(_compute_shape(feature_names)) == 1:  # TODO: should always be an alias once slicer supports per-row aliases
+            if len(values_shape) >= 2 and len(feature_names) == values_shape[1]:
                 feature_names = Alias(list(feature_names), 1)
+            elif len(values_shape) >= 1 and len(feature_names) == values_shape[0]:
+                feature_names = Alias(list(feature_names), 0)
 
-        if len(_compute_shape(output_names)) == 1: # TODOsomeday: should always be an alias once slicer supports per-row aliases
+        if len(_compute_shape(output_names)) == 1:  # TODO: should always be an alias once slicer supports per-row aliases
             output_names = Alias(list(output_names), self.output_dims[0])
             # if len(values_shape) >= 1 and len(output_names) == values_shape[0]:
             #     output_names = Alias(list(output_names), 0)
--- a/tests/test_explanation.py
+++ b/tests/test_explanation.py
@@ -79,7 +79,7 @@ def test_feature_names_slicing_for_square_arrays(random_seed):
     featnames = list("abcde")
 
     exp = shap.Explanation(
-        # usually this arises as the shap values of N=6 samples, k=5 features
+        # an array of this shape typically arises as the shap values of N=6 samples, k=5 features
         values=rs.rand(6, 5),
         feature_names=featnames,
         output_names=featnames,
@@ -90,7 +90,7 @@ def test_feature_names_slicing_for_square_arrays(random_seed):
     assert column_e.feature_names == "e"
 
     exp = shap.Explanation(
-        # usually this arises as the shap values of N=5 samples, k=5 features
+        # an array of this shape typically arises as the shap values of N=5 samples, k=5 features
         values=rs.rand(5, 5),
         feature_names=featnames,
         output_names=featnames,
--- a/tests/test_explanation.py
+++ b/tests/test_explanation.py
@@ -58,3 +58,40 @@ def test_explanation_hstack_errors(random_seed):
             base_values=np.ones(20) * 0.987,
         )
         _ = base_exp.hstack(exp2)
+
+
+def test_issue_2722(random_seed):
+    """Checks that feature names and output names in Explanations are properly sliced with
+    "square" arrays (N==k).
+
+    For 2D arrays, there is an ambiguity in how to assign the feature names to the slicer index.
+    E.g. if feature_names is a list of 5 elements, and the shap_values is a (5,5) array, it's ambiguous
+    whether the axis=0 or axis=1 refers to the "feature columns".
+
+    This test ensures that we give higher priority to axis=1 for the feature_names (and output_names)
+    for square arrays. Since most of the time, the 2D shap values arrays are assembled as
+    (# samples, # features).
+
+    cf. GH Issue #2722.
+    """
+    rs = np.random.RandomState(random_seed)
+    featnames = list("abcde")
+
+    exp = shap.Explanation(
+        # usually this arises as the shap values of N=5 samples, k=5 features
+        values=rs.rand(5, 5),
+        feature_names=featnames,
+        output_names=featnames,
+    )
+    first_sample = exp[0]
+    # this used to return "a" incorrectly, instead of ["a","b","c","d","e"]
+    assert first_sample.feature_names == first_sample.output_names == featnames
+
+    exp = shap.Explanation(
+        # usually this arises as the shap values of N=6 samples, k=5 features
+        values=rs.rand(6, 5),
+        feature_names=featnames,
+        output_names=featnames,
+    )
+    first_sample = exp[0]
+    assert first_sample.feature_names == first_sample.output_names == featnames
--- a/tests/test_explanation.py
+++ b/tests/test_explanation.py
@@ -60,16 +60,17 @@ def test_explanation_hstack_errors(random_seed):
         _ = base_exp.hstack(exp2)
 
 
-def test_issue_2722(random_seed):
-    """Checks that feature names and output names in Explanations are properly sliced with
-    "square" arrays (N==k).
+def test_square_explanation_feature_names_slicing_issue_2699(random_seed):
+    """Checks that feature names in Explanations are properly sliced with "square"
+    arrays (N==k).
 
-    For 2D arrays, there is an ambiguity in how to assign the feature names to the slicer index.
-    E.g. if feature_names is a list of 5 elements, and the shap_values is a (5,5) array, it's ambiguous
-    whether the axis=0 or axis=1 refers to the "feature columns".
+    For 2D arrays, there is an ambiguity in how to assign the feature names to the
+    slicer index. E.g. if feature_names is a list of 5 elements, and the shap_values is
+    a (5,5) array, it's ambiguous whether the axis=0 or axis=1 refers to the "feature
+    columns".
 
-    This test ensures that we give higher priority to axis=1 for the feature_names (and output_names)
-    for square arrays. Since most of the time, the 2D shap values arrays are assembled as
+    This test ensures that we give higher priority to axis=1 for the feature_names for
+    square arrays. Since most of the time, the 2D shap values arrays are assembled as
     (# samples, # features).
 
     cf. GH Issue #2722, #2699.
--- a/tests/test_explanation.py
+++ b/tests/test_explanation.py
@@ -72,10 +72,21 @@ def test_issue_2722(random_seed):
     for square arrays. Since most of the time, the 2D shap values arrays are assembled as
     (# samples, # features).
 
-    cf. GH Issue #2722.
+    cf. GH Issue #2722, #2699.
     """
     rs = np.random.RandomState(random_seed)
     featnames = list("abcde")
+
+    exp = shap.Explanation(
+        # usually this arises as the shap values of N=6 samples, k=5 features
+        values=rs.rand(6, 5),
+        feature_names=featnames,
+        output_names=featnames,
+    )
+    first_sample = exp[0]
+    assert first_sample.feature_names == first_sample.output_names == featnames
+    column_e = exp[..., "e"]
+    assert column_e.feature_names == "e"
 
     exp = shap.Explanation(
         # usually this arises as the shap values of N=5 samples, k=5 features
@@ -86,12 +97,5 @@ def test_issue_2722(random_seed):
     first_sample = exp[0]
     # this used to return "a" incorrectly, instead of ["a","b","c","d","e"]
     assert first_sample.feature_names == first_sample.output_names == featnames
-
-    exp = shap.Explanation(
-        # usually this arises as the shap values of N=6 samples, k=5 features
-        values=rs.rand(6, 5),
-        feature_names=featnames,
-        output_names=featnames,
-    )
-    first_sample = exp[0]
-    assert first_sample.feature_names == first_sample.output_names == featnames
+    column_e = exp[..., "e"]
+    assert column_e.feature_names == "e"
--- a/tests/test_explanation.py
+++ b/tests/test_explanation.py
@@ -60,7 +60,8 @@ def test_explanation_hstack_errors(random_seed):
         _ = base_exp.hstack(exp2)
 
 
-def test_feature_names_slicing_for_square_arrays(random_seed):
+@pytest.mark.parametrize("N", [4, 5, 6])
+def test_feature_names_slicing_for_square_arrays(random_seed, N):
     """Checks that feature names in Explanations are properly sliced with "square"
     arrays (N==k).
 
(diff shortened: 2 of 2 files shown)


connortann:
I found it a little hard to review this PR at first look. I have a few questions to help my understanding:
### Slicing

> First, some background. For 2D arrays, there is an ambiguity in how to assign the feature names to the slicer index.

The fact there the dimension is inferred from the shape of the data seems a bit concerning in the first place: why is the syntax ambiguous? Would it be preferable to have unambiguous syntax throughout, so intentions don't need to be guessed by matching up the lengths of each dimension?

For example, numpy array slicing such as `array[i, j]` is unambiguous: `i` slices along dimension 0, and `j` along dimension 1. What is the canonical behaviour of slicing an Explanation with a single integer, is that defined behaviour?

### The Explanation class attributes

As the `init` method doesn't have a complete docstring, it's not immediately obvious to me what exactly the various attributes mean and what the allowed types and shapes are. I don't think it's clear from the docs how the class is used or invoked: for example, it's not listed as a return type of the various explainers. Would formalising the inputs of the class and the expected return values of `__getitem__` be a good place to start?

thatlittleboy:
> What is the canonical behaviour of slicing an Explanation with a single integer, is that defined behaviour?

Yes, I would say it's well-defined. It simply goes through all of its internal arrays and picks out the object in the first dimension (axis=0), for all the arrays.

> why is the syntax ambiguous?

It's not the syntax of slicing that is ambiguous, per se. It's the assignment of the feature names to a particular axis. I'll give an example: this is a **perfectly valid** invocation of a `shap.Explanation` object:

```python
import numpy as np, shap

e = shap.Explanation(values=np.arange(25).reshape(5,5), feature_names=list("abcde"))
print(e)

# .values =
# array([[ 0,  1,  2,  3,  4],
#        [ 5,  6,  7,  8,  9],
#        [10, 11, 12, 13, 14],
#        [15, 16, 17, 18, 19],
#        [20, 21, 22, 23, 24]])
```

Note that I only specified an array-like of shape `(5,)` for `feature_names` but a shap values array of `(5,5)`. Which is fair, we usually only have 1 axis that corresponds to the feature names anyway.

Now, what is `e[0]`? As mentioned above, the _behaviour_ itself is well-defined. It reaches in for the first axis and returns you a new Explanation object sliced appropriately.

You would *expect* it to be an Explanation object containing the shap values of `[0, 1, 2, 3, 4]` and a feature names array of `[a, b, c, d, e]`. And what would it take to achieve this? It means that the `['a','b','c','d','e']` array had to be assigned to the axis=1, and "broadcasted" along axis=0.

But right now, the result is an Explanation with shap_values of `[0, 1, 2, 3, 4]` and `e[0].feature_names` is `'a'` (!!). Because Slicer is incorrectly treating `[a, b, c, d, e]` to be along axis=0, and consequently `e[0].feature_names` returns `'a'`.

You could've avoided all of this, if you had done:

```python
e = shap.Explanation(values=np.arange(25).reshape(5,5), feature_names=np.array(list("abcde")).reshape(1,-1))
print(e[0].feature_names)  # ['a' 'b' 'c' 'd' 'e']
```

I.e., make `feature_names` during instantiation the same `ndims` as your `values`, then it would be unambiguous which axis your ` ['a' 'b' 'c' 'd' 'e']` belongs to. I.e, axis=1.
But that is not ergonomic / intuitive to use. So here we are.

---

> Would formalising the inputs of the class and the expected return values of __getitem__ be a good place to start?

I think that's out of scope here, but it can be a separate PR on its own. Definitely over time, the in-code documentation needs to be beefed up.

That said, I think a tutorial on how the `Explanation` object works and slices is due.. It's actually a lot like numpy broadcasting, but it has some weird rules (such as this one w/ the `feature_names`). So it takes some getting used to.

---

If the explanation (heh) above is still not clear, I would highly encourage going through this example with a debugger, running through line by line. And seeing why `'a'` pops out of `e[0].feature_names` instead of `['a' 'b' 'c' 'd' 'e']` as one would expect.

connortann:
I want to take a moment to appreciate your impeccably well-written explanation, which is very highly appreciated!

heh, thanks for the Explanation Explanation

shap/_explanation.py
```diff
@@ -118,13 +118,13 @@ def __init__( # pylint: disable=too-many-arguments
...
 
-        if len(_compute_shape(feature_names)) == 1: # TODOsomeday: should always be an alias once slicer supports per-row aliases
-            if len(values_shape) >= 1 and len(feature_names) == values_shape[0]:
-                feature_names = Alias(list(feature_names), 0)
-            elif len(values_shape) >= 2 and len(feature_names) == values_shape[1]:
+        if len(_compute_shape(feature_names)) == 1:  # TODO: should always be an alias once slicer supports per-row aliases
+            if len(values_shape) >= 2 and len(feature_names) == values_shape[1]:
                 feature_names = Alias(list(feature_names), 1)
+            elif len(values_shape) >= 1 and len(feature_names) == values_shape[0]:
+                feature_names = Alias(list(feature_names), 0)
```
connortann, line 125: Should there be a final `else` clause here: what happens if `len(_compute_shape(feature_names)) == 1` and the neither sub-condition is met, perhaps a ValueError?

tests/test_explanation.py
```diff
@@ -58,3 +58,45 @@ def test_explanation_hstack_errors(random_seed):
             base_values=np.ones(20) * 0.987,
         )
         _ = base_exp.hstack(exp2)
+
+
+def test_square_explanation_feature_names_slicing_issue_2699(random_seed):
```
connortann: Nit: perhaps consider a test name that describes the functionality rather than the GH issue, e.g.
`test_explanation_slicing_for_square_arrays`

tests/test_explanation.py
```diff
@@ -58,3 +58,45 @@ def test_explanation_hstack_errors(random_seed):
...
+    square arrays. Since most of the time, the 2D shap values arrays are assembled as
+    (# samples, # features).
+
+    cf. GH Issue #2722, #2699.
+    """
+    rs = np.random.RandomState(random_seed)
+    featnames = list("abcde")
+
+    exp = shap.Explanation(
+        # usually this arises as the shap values of N=6 samples, k=5 features
```
connortann: Nit-nit : gramatically I wasn't sure what phenomenon is referened by _"usually this arises"_. We're talking about how values of that shape are interpreted, right?

Another optional idea: perhaps consider using `@pytest.mark.parametrize` to run this test for various shapes and input feature names, to see if there are any other odd edge cases?

connortann (approved):
I added some suggestions, all of which are optional so approving. Great work!

shap/_explanation.py
thatlittleboy: To be honest, I'm not sure. Slicer can handle such cases perfectly fine (as in, no errors), 

```python
e = shap.Explanation(
    values=np.random.rand(5, 5),
    feature_names=list("abcd"),  # <----- mismatch in shape
)
e[:, 4].feature_names  # ['a', 'b', 'c', 'd']
```

and I'm not sure whether we would be breaking any other use-cases if we throw a ValueError here. (If we had proper test coverage, then maybe I'll be more comfortable making this change.)

Since the original code didn't have an `else`, I think it's better we leave it as status quo / address this in a separate PR, since the `else` branch doesn't really play a role in the bug fix proposed here.

tests/test_explanation.py
thatlittleboy: Yep, I went with `test_feature_names_slicing_for_square_arrays`.

tests/test_explanation.py
thatlittleboy: > We're talking about how values of that shape are interpreted, right?

yep, I was just trying to convey that a shap values array of shape `(6, 5)` typically arises from having 6 samples, 5 features. I've now modified the phrasing here slightly, hopefully it's a little better.

> perhaps consider using @pytest.mark.parametrize to run this test for various shapes and input feature names

Any particular examples in mind that you would like to test? In the context of the original issue, this test seems sufficient to me, but I suppose we could make the test more general.

connortann: >Any particular examples in mind

I was just observing that the current test seems to be two similar halves, which could maybe be combined in a parameterised test. Something like:

```python
@parameterize(n_samples=[4,5,6])
def test_feature_names_slicing(n_samples):
    featnames = list("abcde")
    exp = shap.Explanation(
        values=np.ones((n_samples, 5)),
        feature_names=featnames,
        output_names=featnames,
    )
    first_sample = exp[0]
    assert first_sample.feature_names == first_sample.output_names == featnames
    column_e = exp[..., "e"]
    assert column_e.feature_names == "e"
```

I'm happy with either style, so will leave it to you to decide which you prefer.

thatlittleboy: Thanks @connortann (sorry for the late reply, finally got around to finishing up this PR), I decided to go with your suggestion in the end.

PR #3128, opened 2023-07-24
MAINT: Fix path filter issue, move linter to seperate workflow

connortann:
## Overview

An unintended side-effect of implementing workflow-level path filters for the unit tests on #3107 is that the precommit linter no longer runs on all PRs.

This PR address that my moving the linter to a seperate workflow, that runs regardless of the path filter.

## Checklist

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,19 @@
+name: lint
+
+on:
+    push:
+      branches:
+        - master
+    pull_request:
+      branches:
+        - master
+
+jobs:
+    precommit-check:
+      runs-on: ubuntu-latest
+      steps:
+      - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


thatlittleboy:
Good catch.

I've been thinking about this, now that we're in the org, we should utilize pre-commit-ci instead of the GH actions.

We get autoupdates to the pre-commit hook versions, autofixes when the PR isn't pre-commit compliant (so we can rmove the reminder to contributors to ensure their pre-commit passes), and it potentially runs faster than GHA because of the caches. What do you think?

connortann:
Agreed, I saw that the pre-commit service is recommended in the github action description. In fact I requested access for the pre-commit.ci app yesterday.

This is just a temporary bug fix until we get that running.

thatlittleboy (approved):
wwe-referee-thumbsup

Issue #3122, opened 2023-07-22
[Meta-issue]: Update major versions of javascript dependencies


connortann:
The JS package dependencies are several years out of date, and we have several **security advisories** and NPM warnings about deprecated and unsupported packages.

Current versions contain breaking changes, and currently break the build. We should update the relevant packages, and then do some thorough manual testing of the new JS bundle.

### CICID prep

- [x] Get build running on CI
- [x] Add jest tests on CI
- [x] Get dependabot updates working

### Updates

- [x] Apply non-breaking minor updates to all packages
- [x] Update webpack
   - #3116 
   - [x] Fix broken `npm run dev` server
- [x] Update D3
   - #3114 
   - #3132
   - #3133
- [x] Update react
   - [x] Remove deprecated/outdated `react-tap-event-plugin`
   - [x] #3161
- [x] Update SSL provider
   - #3102  
- [x] Confirm all JS-related Security advisories are fixed

### Testing & deployment

- [x] Manual testing of all JS plots
- [x] Recreate `bundle.js` and place in python package
   - #3148
- [ ] Reset dependabot update frequency to monthly
- [ ] New release of npm package?

## Manually testing changes

To manually build and test the javascript bundle:

1. Run `npm run build` from the `javascript/` directory to build the bundle
2. Copy `javascript/build/bundle.js` into `shap/plots/resources/`
3. Create an example force plot, as below

```python
# Example plot to test
import xgboost
import shap

shap.initjs()
X,y = shap.datasets.california()
bst = xgboost.train({"learning_rate": 0.01}, xgboost.DMatrix(X, label=y), 100)
explainer = shap.TreeExplainer(bst)
shap_values = explainer.shap_values(X)
shap.force_plot(explainer.expected_value, shap_values[0:100,:], X.iloc[0:100,:])
```

Example output:

![image](https://github.com/shap/shap/assets/71127464/feec65a8-35d7-4e10-a3f0-5cc7fd13f2d2)

PR #3133, opened 2023-07-24
Update deprecated d3.mouse, and update deprecated webpack content-base

### LINKED ISSUE #3132: BUG: 'd3.mouse' object deprecated in d3 v6
## Description of issue

Running `npm run build` in the javascript directory throws a warning:

```
WARNING in ./visualizers/AdditiveForceArrayVisualizer.jsx 164:14-19
export 'mouse' (imported as 'mouse') was not found in 'd3-selection' (possible exports: create, creator, local, matcher, namespace, namespaces, pointer, pointers, select, selectAll, selection, selector, selectorAll, style, window)
```

`d3.mouse` was removed in version 6, according to the [release notes](https://github.com/d3/d3/releases/tag/v6.0.0).

It is used in `AdditiveForceArrayVisualizer`, I think just once.

Related threads:
- https://stackoverflow.com/questions/64270094/export-mouse-imported-as-d3-was-not-found-in-d3
- https://stackoverflow.com/a/64202214

## Possible solutions

- (preferred) Upgrade syntax, replacing d3.mouse with [d3.pointer](https://github.com/d3/d3-selection/blob/main/README.md#pointer)
- Downgrade D3

connortann:
## Overview

Fixes two issues caused by breaking changes in our dependencies:

- Updates a D3 callback in `AdditiveForceArrayVisualizer`
   - Fixes #3132.
   - Switched to `d3.pointer` rather than the deprecated `d3.mouse`.
   - Breaking change was caused by #3114
- Updates the use of `webpack` in dev server 
   - Remove deprecated `--content-base` CLI flag, instead set `static` config value
   - Fixes breaking change caused by #3130

Both issues were not caught by our CI/CD, which is currently very basic. So, I'm interested in any ideas on how we can improve the tests to catch similar issues in future.

Changes were also manually tested, following steps outlined in #3122.

### DIFF
--- a/javascript/package.json
+++ b/javascript/package.json
@@ -9,7 +9,7 @@
   ],
   "scripts": {
     "test": "jest --verbose ./tests",
-    "dev": "export SET NODE_OPTIONS=--openssl-legacy-provider && webpack-dev-server --content-base build/ --open",
+    "dev": "export SET NODE_OPTIONS=--openssl-legacy-provider && webpack-dev-server --open",
     "build": "export SET NODE_OPTIONS=--openssl-legacy-provider && webpack --mode production",
     "pub": "export SET NODE_OPTIONS=--openssl-legacy-provider && npm run build && npm publish"
   },
--- a/javascript/visualizers/AdditiveForceArrayVisualizer.jsx
+++ b/javascript/visualizers/AdditiveForceArrayVisualizer.jsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { select, mouse } from "d3-selection";
+import { select, pointer } from "d3-selection";
 import { scaleLinear, scaleTime } from "d3-scale";
 import { format } from "d3-format";
 import { timeFormat, timeParse } from "d3-time-format";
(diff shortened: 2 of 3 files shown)


connortann:
Pinging @remborg in case he has any feedback on this PR, as our resident JavaScript guru

remborg:
Haha, I'm just helping occasionally and wouldn't consider myself resident.
I usually work with Typescript, this helps catching some issues early but it might be a bit overkill in shap.
I think adding a linter would definitely help catching some of the breaking changes but it's not a silver bullet. I'm thinking about _eslint_ and the plugin _eslint-plugin-import_.

connortann:
@thatlittleboy this is ready for a review.

As a bit of context, I've been tracking the overall state of the JS packages in the meta-issue #3122. I think it's important that we close the reported security vulnerabilities (which you should be able to see [here](https://github.com/shap/shap/security/dependabot)) - almost all of which are addressed now. I plan to update ReAct next, which I think is the last major update needed.

I know you're relatively less comfortable reviewing javascript changes, but even a cursory look I think would be helpful. Once we've updated ReAct, I think that would be the perfect time to do really thorough manual testing of the bundle.

thatlittleboy (approved):
Thanks for the work on this

dog-hopping-low-five

PR #3153, opened 2023-07-29
FIX: 1-d-numpy and pd input for gradient explainer

### LINKED ISSUE #3152: BUG: gradient Explainer pd.DataFrame input failing
### Issue Description

The gradient explainer class accepts `pandas.Dataframe` as a data input for TF models. The dataframe is correctly used for its feature_name columns, but is then passed on as a `pd.Dataframe` to `_TFGradient`, which does not deal with this type properly.
A simple fix would be to convert the Dataframe to its underlying np.array

```python
import shap
    import pandas as pd
    import tensorflow as tf

    feature_names = [f"feature_{i}" for i in range(5)]

    background = pd.DataFrame(np.zeros((10, 5)), columns=feature_names)

    model = tf.keras.Sequential([
        tf.keras.layers.Dense(10, input_shape=(5,), activation='relu'),
        tf.keras.layers.Dense(1, activation='linear')
    ])
    model.compile(optimizer='adam', loss='mse')
    e = shap.GradientExplainer(model, background)
```

### Expected Behavior

The `GradientExplainer` should deal with pd.Dataframes as input as is described.

### Bug report checklist

### Installed Versions

0.42.1

ponseko:
## Overview

Closes #3152, #2607 (#3025)

#3152 is fixed by simply converting a pd.Dataframe to its underlying np.array in `_TFGradient` (after setting the feature names in `Gradient`).

#2607 (#3025) seemed stale, so I figured I'd close that by simply setting the feature_names to None if there is no pd.Dataframe provided. It seems like the labels are set by the plots if the feature names are None, so there is no need to set them to a list of numbers here. 
Additionaly, an extra parameter can be added here for the feature names (like some explainers already do). But I am unsure whether, going forward, you would want to keep that logic in the plots themselves only.

Added two unit tests for simple pd and np 1-d inputs.

Let me know if anything needs changing.

## Checklist

### DIFF
--- a/tests/explainers/test_gradient.py
+++ b/tests/explainers/test_gradient.py
@@ -286,7 +286,6 @@ def test_tf_input(random_seed, input_type):
     background = np.zeros((batch_size, num_features))
     if input_type == "dataframe":
         background = pd.DataFrame(background, columns=feature_names)
-    x1 = np.ones((batch_size, num_features))
 
     model = tf.keras.Sequential([
         tf.keras.layers.Dense(10, input_shape=(num_features,), activation='relu'),
@@ -294,16 +293,11 @@ def test_tf_input(random_seed, input_type):
     ])
     model.compile(optimizer='adam', loss='mse')
 
-    e = shap.GradientExplainer(model, background)
-    shap_values = e.shap_values(x1[:1])
+    explainer = shap.GradientExplainer(model, background)
+    example = np.ones((1, num_features))
+    explanation = explainer(example)
 
-    if input_type == "dataframe":
-        diff = (model.predict(x1[:1]) - model.predict(background.values)).mean(0)
-    elif input_type == "numpy":
-        diff = (model.predict(x1[:1]) - model.predict(background)).mean(0)
-    else:
-        raise NotImplementedError("Input type %s not implemented" % input_type)
-
-    sums = np.array([shap_values[i].sum() for i in range(len(shap_values))])
+    diff = (model.predict(example) - model.predict(background)).mean(0)
+    sums = np.array([values.sum() for values in explanation.values])
     d = np.abs(sums - diff).sum()
     assert d / (np.abs(diff).sum() + 0.01) < 0.1, "Sum of SHAP values does not match difference! %f" % (d / np.abs(diff).sum())
--- a/tests/explainers/test_gradient.py
+++ b/tests/explainers/test_gradient.py
@@ -271,3 +271,62 @@ def test_pytorch_multiple_inputs(random_seed):
     sums = np.array([shap_x1[i].sum() + shap_x2[i].sum() for i in range(len(shap_x1))])
     d = np.abs(sums - diff).sum()
     assert d / (np.abs(diff).sum()+0.01) < 0.1, "Sum of SHAP values does not match difference! %f" % (d / np.abs(diff).sum())
+
+def test_tf_pd_input(random_seed):
+    """ Test tabular (batch_size, features) pd.DataFrame input. """
+    pd = pytest.importorskip('pandas')
+    tf = pytest.importorskip('tensorflow')
+    import pandas as pd
+    import tensorflow as tf
+
+    tf.random.set_seed(random_seed)
+
+    batch_size = 10
+    num_features = 5
+    feature_names = [f"TF_pd_test_feature_{i}" for i in range(num_features)]
+
+    background = np.zeros((batch_size, num_features))
+    background = pd.DataFrame(background, columns=feature_names)
+    x1 = np.ones((batch_size, num_features))
+
+    model = tf.keras.Sequential([
+        tf.keras.layers.Dense(10, input_shape=(num_features,), activation='relu'),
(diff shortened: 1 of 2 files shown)


tests/explainers/test_gradient.py
```diff
@@ -271,3 +271,62 @@ def forward(self, x1, x2):
     sums = np.array([shap_x1[i].sum() + shap_x2[i].sum() for i in range(len(shap_x1))])
     d = np.abs(sums - diff).sum()
     assert d / (np.abs(diff).sum()+0.01) < 0.1, "Sum of SHAP values does not match difference! %f" % (d / np.abs(diff).sum())
+
+def test_tf_pd_input(random_seed):
+    """ Test tabular (batch_size, features) pd.DataFrame input. """
+    pd = pytest.importorskip('pandas')
```
connortann: Pandas is a non-optional dependency of the package, so it can be safely imported in the normal way at the top of the file

tests/explainers/test_gradient.py
```diff
@@ -271,3 +271,62 @@ def forward(self, x1, x2):
...
+    # testing pd input
+    e = shap.GradientExplainer(model, background)
+    shap_values = e.shap_values(x1[:1])
+
+    diff = (model.predict(x1[:1]) - model.predict(background.values)).mean(0)
+    sums = np.array([shap_values[i].sum() for i in range(len(shap_values))])
+    d = np.abs(sums - diff).sum()
+    assert d / (np.abs(diff).sum() + 0.01) < 0.1, "Sum of SHAP values does not match difference! %f" % (d / np.abs(diff).sum())
+
+def test_tf_np_input(random_seed):
```
connortann: The two unit tests have a fair amount of duplicated code, so I would suggest combining them and using pytest's `parameterize` to provide different different model inputs

connortann (changes requested):
Thanks for the PR, this looks like a great improvement!

I've added a couple of minor suggestions for changes to the unit tests.

tests/explainers/test_gradient.py
```diff
@@ -271,3 +271,62 @@ def forward(self, x1, x2):
     sums = np.array([shap_x1[i].sum() + shap_x2[i].sum() for i in range(len(shap_x1))])
     d = np.abs(sums - diff).sum()
     assert d / (np.abs(diff).sum()+0.01) < 0.1, "Sum of SHAP values does not match difference! %f" % (d / np.abs(diff).sum())
+
+def test_tf_pd_input(random_seed):
+    """ Test tabular (batch_size, features) pd.DataFrame input. """
+    pd = pytest.importorskip('pandas')
+    tf = pytest.importorskip('tensorflow')
+    import pandas as pd
+    import tensorflow as tf
```
thatlittleboy: `pytest.importorskip` returns the imported module itself, so there's no need to import again below.

ponseko:
@thatlittleboy @connortann Thank you for the feedback. I've included the requested changes to the unit tests in the latest commit.

tests/explainers/test_gradient.py
```diff
@@ -271,3 +272,38 @@ def forward(self, x1, x2):
...
+
+    e = shap.GradientExplainer(model, background)
+    shap_values = e.shap_values(x1[:1])
+
+    if input_type == "dataframe":
+        diff = (model.predict(x1[:1]) - model.predict(background.values)).mean(0)
+    elif input_type == "numpy":
+        diff = (model.predict(x1[:1]) - model.predict(background)).mean(0)
+    else:
+        raise NotImplementedError("Input type %s not implemented" % input_type)
```
connortann: I think this `if/else` clause is unnecessary here, as this should work for both cases:

```python
diff = (model.predict(x1[:1]) - model.predict(background)).mean(0)
```

tests/explainers/test_gradient.py
```diff
@@ -271,3 +272,38 @@ def forward(self, x1, x2):
...
+        background = pd.DataFrame(background, columns=feature_names)
+    x1 = np.ones((batch_size, num_features))
+
+    model = tf.keras.Sequential([
+        tf.keras.layers.Dense(10, input_shape=(num_features,), activation='relu'),
+        tf.keras.layers.Dense(1, activation='linear')
+    ])
+    model.compile(optimizer='adam', loss='mse')
+
+    e = shap.GradientExplainer(model, background)
```
connortann: This test is in line with the rest of the unit tests in style; but I think we can improve on it a bit. We could use the newer `Explainer` API here, and use some slightly more descriptive variable names. Over time we could update the rest of the tests to match.

How about something like this?

```python
explainer = shap.GradientExplainer(model, background)
example = np.ones((1, num_features))
explanation = explainer(example)
diff = (model.predict(example) - model.predict(background)).mean(0)
sums = np.array([values.sum() for values in explanation.values])
```

tests/explainers/test_gradient.py
ponseko: Oops, I convinced myself the TF model required np input, but the predict function deals with that just fine.

connortann (approved):
This looks great - thanks for making those tweaks.

PR #3157, opened 2023-07-30
ENH: Add heatmap and violin plot to main API

connortann:
Adds `violin_plot` and `heapmap_plot` to the top-level `shap/__init__.py`.

These functions are already exposed publicly in `shap.plots`, so for consistency with the rest of the package I suggest also including them in the top-level namespace.


thatlittleboy (approved):
lgtm!

Issue #3066, opened 2023-07-05
MAINT: address all deprecation warnings in test suite


connortann:
# Context

At present, there are several deprecation warnings in the test suite logs. If left unaddressed, these will turn into future failures.

We've had  two a few cases of breaking issues, which were previously deprecation warnings. For example, a breaking change in numpy 0.24 delayed our most recent release. Let's fix these issues whilst they're still just warnings!

As an aside, it will also be useful to explicitly ignore warnings that are not within our control or that are false positives. That will keep the pytest output concise, and mean we're more likely to spot any new important warnings. This can be done by adding a line to the `filterwarnings` setting in `pyproject.toml`.

### Where are the warnings?

Look at the most [recent test run on the master branch](https://github.com/shap/shap/actions/workflows/run_tests.yml?query=branch%3Amaster), on the job called `ubuntu-latest, 3.12, test`. Example screenshot:

<details>

![Image](https://github.com/user-attachments/assets/8066e0a5-4513-4648-90f2-63914feccea3)

</details>

## TODO

- [x] #3969
- [ ] #3970
- [x] #3971
- [ ] #3972
- [ ] #3973
- [x] #3974
- [x] #3975

## Completed

<details>

- [x] typing.io
- [x] distutils
- [x] sentencepiece
- [x] tf.keras
- [x] xgboost
- [x] LightGBM 
- [x] sklearn (column names)
- [x] Unable to determine axes for colorbar
- [x] default value of `n_init`
- [x] Numba types scheduled for deprecation
- [x] pytorch Using a non-full backward hook
- [x] Pyspark check_additivity
- [x] No data for colormapping

</details>

znacer:
Hi,
While looking for the first sklearn task (```X has feature names, but LinearRegression was fitted without feature names```), I noticed that this warning is due to the function ```convert_to_model``` in ```shap/utils/_legacy.py```.  
The features names are removed to avoid another sklearn warning (```X does not have valid feature names, but <model> was fitted with feature names```).
I can suggest filter this specific warning as it is an intended behavior.

connortann:
I was wondering about that one. I think the warning might be an indication that there is a risk of indexing errors.

[...]

PR #3167, opened 2023-07-31
 MAINT: address Matplotlib deprecation warnings in test suite 

znacer:
## Overview

Closes corresponding tasks #3066  

Removed `vmin` and `vmax` statements in `plt.scatter` for NA values, that are grey anyway.
Changed the statement `plt.tight_layout` to `fig.set_layout_engine("tight")`. This statement is not to be deprecated by matplotlib. Updated the baselines images corresponding.

## Checklist


connortann:
Thanks for the PR! A few questions just to check my understanding.

> Changed the statement plt.tight_layout to fig.set_layout_engine("tight"). This statement is not to be deprecated by matplotlib. 

This caught my eye. I couldn't find a deprecation notice for the original function - did you find something in the matplotlib docs suggesting `plt.tight_layout` will be deprecated?

I notice the baseline images have changed very slightly. What causes that change: does `fig.set_layout_engine("tight")` have different behaviour to `plt.tight_laytout`?

znacer:
To be more precise, I first used ‘fig.set_tight_layout(True)‘ to get rid of the warning. This gave a me a depecation warning (https://matplotlib.org/stable/api/figure_api.html#matplotlib.figure.Figure.set_tight_layout). That is where the ‘fig.set_layout_engine‘ comes from.

znacer:
Concerning the change in the images, I am not sure where it comes from. If you prefer to keep the original images, I suggest to add tolerances to the tests.

connortann (approved):
Thanks again for the PR - collectively we're making great progress fixing all these warnings.

PR #3175, opened 2023-08-01
DOCS: Fix failing ReadTheDocs build, pin Sphinx dependencies

connortann:
I noticed the docs have started failing unexpectedly. It seems the issue is that the theme we use is not compatible with Sphinx v7.

Related:

- https://github.com/readthedocs/readthedocs.org/issues/10279
- https://github.com/readthedocs/sphinx_rtd_theme/issues/1463

As a fix, this pins the sphinx packages dependencies to a recent working version.

For reference, there are suggestions on reproducible builds in the ReadTheDocs [guide to reproducible builds](https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html].

Alternatively, we could consider switching to a Sphinx theme that is apparently better supported, like [furo](https://github.com/pradyunsg/furo)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ docs = [
   "ipython",
   "numpydoc",
   "sphinx_rtd_theme",
-  "sphinx",
+  "sphinx<7",
   "nbsphinx",
   "sphinx_github_changelog",
   "requests",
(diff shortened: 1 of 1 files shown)


connortann:
@thatlittleboy given the docs are currently failing on all PRs, I shall merge this bugfix PR now to fix the build. Hoping we can discuss next week to see if we're on the same page on pinning dependencies 👍

thatlittleboy:
Seems like sphinx 7 is now supported https://github.com/readthedocs/sphinx_rtd_theme/pull/1464

Issue #3156, opened 2023-07-30
ENH: Deprecate and remove legacy functions


connortann:
Many plotting functions have a "legacy" version, designated either in the name or the docstring. We could officially deprecate them, and then remove them.

I believe the "legacy" refers to accepting `shap_values` numpy arrays directly, rather than the newer-style `Explanation` objects.

Deprecation tracker: 

## Guardrails

1. To mitigate the impact on users, all breaking changes will have a deprecation cycle: a DeprecationWarning is added, and then two releases later the function is removed
2. Deprecations and removals are tracked here: #3507
3. Where we raise deprecation warnings, the warning should include instructions on how users can update their code to remove the warning

## Legacy things to remove:

Legacy plotting functions:

- [ ] `bar_legacy`
  - #3739
- [ ] `summary_legacy`
- [ ] `dependence_legacy`
- [ ] `waterfall_legacy`

Suspected unused / dead code:

- [ ] `shap/utils/_masked_models.py:_build_delta_masked_inputs()`
- [ ] `shap/plots/_text.py:text_old()`
- [ ] `shap/explainers/_explainer.py:Explainer._compute_main_effects()`

Possibly related other legacy things:

- [ ] The module `shap.utils._legacy.py`
- [ ] The method `Permutation.shap_values`
- [ ] `Partition.owen()` vs `Partition.owen3()` (?)

arshiaar:
Hi, I'm happy to do this. Would you like to deprecate these first or remove them completely?

connortann:
I think we should use a proper deprecation cycle for anything being remove from the public API, by adding a `DeprecationWarning` for at least one release.

However, internal refactoring that isn't part of the public API doesn't need a deprecation cycle.

thatlittleboy:
I'll just like to leave a note here requesting to leave summary_legacy in (but still go ahead with the deprecation warning), until we replicate all of its functionality. I seem to recall there's some TODO not yet completed.

But I'm in favour of the changes proposed here in general. 👍🏼

connortann:
I'm a little cautious about causing significant disruption to users by deprecating commonly-used functions, so I think we should tread carefully here.

It would be helpful to get an estimate of what fraction of this library's users would be impacted by this change. I suspect that `summary_legacy` might be a very widely-used plot, so removing that would potentially be a very significant breaking change. We could consider using a longer deprecation cycle (e.g. 2 releases?) for any very widely used functions.

arshiaar:
[...]

Issue #3070, opened 2023-07-06
ENH: Consistent axes and figsize parameters for plots


davidshumway:
### Problem Description

shap plots should consistently include axes and figsize parameters.

For example, 
`shap.plots.force(... , figsize=(20, 3), ...)` includes `figsize` but not `ax`.

`shap.plots.bar( ... )` does not include `figsize` or `ax` parameters.

`shap.plots.scatter(..., ax=None, ...)` includes `ax` but not `figsize`.

### Feature request checklist

- [X] I have checked the issue tracker for duplicate issues.
- [x] I'd be interested in making a PR to implement this feature

davidshumway:
Workaround for changing the figsize of `shap.plots.bar`, for example:

```python
shap.plots.bar(shap_values, max_display=8, show=False)
fig = matplotlib.pyplot.gcf()
fig.set_size_inches(2, 2)
plt.show()
```

thatlittleboy:
@davidshumway Hi, thanks for bringing this issue up. There has been discussion on this before in the community fork dsgibbons#5 (at least, partially).

And I'm in favour of exposing the `ax` parameter to all of the plotting functions, in general. Then `figsize` would be made redundant I think -- if the user wanted to control the figsize they can create their own Axes on a figure with the right dimensions before sending it into `shap.plots`.  

If you would like to raise a PR, I'm happy to work this through with you.

In the meantime I'm also refactoring the plotting functions in general (just FYI), but it shouldn't really affect the progress of your PR should you choose to raise it.

connortann:
Having a consistent plotting API would be great. We could take some inspiration from other libraries like seaborn on the API design.

> If the user wanted to control the figsize they can create their own Axes on a figure with the right dimensions before sending it into shap.plots

In this specific case I don't think the user would strictly need to pass the newly created Axes to `shap.plots`, as by default the plotting function would use `plt.gca()`.

I would expect users would be able to use either of matplotlib's [two interfaces](https://matplotlib.org/stable/users/explain/api_interfaces.html): the imperative style or object-oriented style.

```python
# Imperative "pyplot" interface
plt.figure("figure1", figsize=(10,10))
plots.plot_thing(data)

plt.figure("figure2", figsize=(5,5))
plots.plot_thing(data)
```

Or alternatively:

```python
# Explicit object-oriented interface
f1, ax1 = plt.subplots(figsize=(10,10))
f2, ax2 = plt.subplots(figsize=(5,5))

[...]

Issue #2363, opened 2022-01-25
Enabling the matplotlib "ax" keyword argument to plot functions


mrektor:
Hi all,

It would be very useful to provide in the plot API the possibility of passing the ax on which the plot has to be made

flo-ri-an:
Yes, this would greatly simplify arranging complex layouts. Definitely something I would love to see too.

connortann:
This would be a great enhancement. There is some discussion about the same issue on #3070 , so let's use that one to track the issue. I'll close this as a duplicate.

PR #3200, opened 2023-08-21
DOCS: Remove boston dataset from the README

### LINKED ISSUE #8: Replace Boston dataset in examples
Test logs are showing 

```
`load_boston` has been removed from scikit-learn since version 1.2.
```

with an elaborate explanation about how the Boston dataset has some ethical issues.

We should find a replacement dataset to use, and change all existing examples to that.

Partially related to #4 .

dsgibbons:
## Overview

Closes https://github.com/dsgibbons/shap/issues/8 and resolves the issues noted in https://github.com/dsgibbons/shap/issues/8#issuecomment-1545567370

Related to #2322

Updating the README required the following additional changes:

- Use relative paths for README images instead of URLs tied to master.
- HTML plots generated inside a notebook, rendered in Chrome, and then screenshotted. There is no obvious way to convert the HTML straight to an image, as it requires JS - unless @connortann or @thatlittleboy know of a better way to convert `shap` HTML to PNG? I've tried several different HTML to PNG converters, but I keep getting a `shap` error saying JS needs initialising.

## Checklist

- [ ] All [pre-commit checks](https://pre-commit.com/#install) pass. - no, dependabot is not happy?


shap/plots/__init__.py
```diff
@@ -19,6 +21,9 @@
 from ._violin import violin
 from ._waterfall import waterfall
 
+plt.rcParams['figure.constrained_layout.use'] = True
```
dsgibbons, line 24: required to prevent labels from being cut

dsgibbons:
Tests are failing due to calls to `shap.summary_plot`. Something about constrained mode is breaking `shap.summary_plot`. I think `shap.summary_plot` is an outdated API that used to be a catchall term for the various plot types (beeswarm, violin, etc.). Is it worth me fixing the test, or should we remove `shap.summary_plot` as it probably shouldn't be used any more? The new plots in constrained mode all look fine in `mpl`.

dsgibbons:
Never mind, I think I've fixed `shap.summary_plot`. It may still be worth removing `shap.summary_plot` any way to reduce the size of the codebase.

dsgibbons:
> Never mind, I think I've fixed `shap.summary_plot`. It may still be worth removing `shap.summary_plot` any way to reduce the size of the codebase.

Looks like this is a duplicate of #3156

dsgibbons:
I've added a commit to update the mpl image baselines. We can revert this if references to the legacy `summary_plot` are removed.

dsgibbons:
There's a discrepancy between how my setup generates the baselines and GitHub Actions. I'll need to alter my setup so that I produce the same outputs as GitHub Actions.

shap/plots/_bar.py
```diff
@@ -341,7 +341,7 @@ def bar(shap_values, max_display=10, order=Explanation.abs, clustering=None, clu
 
     if show:
         pl.show()
-
+    return pl.gcf()
```
dsgibbons: _waveform.py returns `pl.gcf()` when `show == False`. I've edited the other plotting functions to also return `pl.gcf()`. Is this a bad idea?

shap/plots/__init__.py
```diff
@@ -19,6 +21,9 @@
...
+
+
```
thatlittleboy: I'm not so fully convinced that this is a good idea. 1) In my experience constrained_layout has caused some other issues (I can't quite communicate what off the top of my head, but there are some trade-offs for sure). 2) Global rcparams like these might interfere with the user's own config (?), I'm not really in favour of this.

I noted that you mentioned about some issues with ylabels being cut-off from a change in dataset. That's a little unusual, do you think it's worth to raise an issue reproducing the problem (without setting constrained layout here) and let's investigate/brainstorm if there are alternative ways of solving this?

shap/plots/_bar.py
thatlittleboy: @dsgibbons I think I see the intention, which is that when `show=False`, it means the user is likely to take the plot and customize it further. So returning *something* makes sense to me.

I'm wondering whether we should return the Figure or the Axes. (If this would turn into a larger discussion, I suggest leaving this change out of this PR since it would hinder the PR being merged).

I'm thinking Axes, as I'm hoping our shap plots to be based around Ax rather than Figs in the future. (we can discuss reasons for this in #3070).

dsgibbons: I prefer Ax too. I chose `pl.gcf()` because it is already used in the waterfall diagram. Migrating to Ax sounds like a broader API change - perhaps for 0.44?

shap/plots/__init__.py
dsgibbons: See #3213. I'll leave this in draft until this issue is resolved.

shap/plots/_bar.py
thatlittleboy: @dsgibbons That's true, let's stick to the current api for now then (for consistency). I'll work on the PR to migrate to axes for 0.44.

dsgibbons:
I've removed the changes to SHAP's plotting functionality. I think these changes should be postponed pending the wider plotting API change proposed for 0.44 by https://github.com/shap/shap/issues/2363.

thatlittleboy (approved):
LGTM, @dsgibbons one minor comment, any reason why we're not removing the `boston_xxx.png` files from `docs/artwork`?

<br>
<br>

recursive-thumbsup-in-video-recording

dsgibbons:
> LGTM, @dsgibbons one minor comment, any reason why we're not removing the `boston_xxx.png` files from `docs/artwork`?

Good point, removed!

PR #3209, opened 2023-08-22
Enh: deprecate legacy plots and code

arshiaar:
## Overview

Part of #3156  

* Added deprecation decorators using sklearn deprecated since its an existing dependency. 
* Decorators added to: bar_legacy, summary_legacy, dependence_legacy, waterfall_legacy
* Added unit tests for dependence_legacy, summary_legacy, and waterfall_legacy. 
* Added decorator for kmeans in utils/._legacy.py

Potentially could create own decorator using the python warnings module. 
Did not include unit tests for bar_legacy as function is breaking outside of decorator. 

## Checklist

### DIFF
--- a/shap/plots/_bar.py
+++ b/shap/plots/_bar.py
@@ -373,7 +373,7 @@ def bar(shap_values, max_display=10, order=Explanation.abs, clustering=None, clu
 
 #     return max(left_val, right_val) + 1, max(left_sum, right_sum)
 
-@deprecated("bar_legacy is being deprecated in Version 0.43.0")
+@deprecated("bar_legacy is being deprecated in Version 0.43.0. This will be removed in Version 0.44")
 def bar_legacy(shap_values, features=None, feature_names=None, max_display=None, show=True):
 
     # unwrap pandas series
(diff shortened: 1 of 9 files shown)


dsgibbons:
Maybe update the message to say something like "will be removed in 0.44"?

thatlittleboy (changes requested):
Few suggestions (up for discussion, except the last one):

* Use the phrase "marked for deprecation" instead of "deprecated". I've always found it confusing myself, but maybe it's just me.
* Maybe `shap v0.43.0` instead of `Version 0.43.0`. I'm thinking to specify "shap" because there might be wrapper libraries around shap, and I'm afraid end users will confuse 0.43 for the wrapper library's version rather than shap's.
* Mention what the alternative should be. "Please use `shap.plots.bar` instead" or words to that effect. For summary_legacy, maybe just nudge the user towards using the dedicated shap.plots functions.
* The docstrings of the legacy functions should be updated as well with the deprecation message & reference to the suggested alternative, in a deprecation admonition / warning box.

CloseChoice:
Thanks for the effort, we really appreciate it. Due to lack of activity I'll close this for now, feel free to come back to this anytime and reopen

PR #3211, opened 2023-08-22
Add nbstripout

owenlamont:
## Overview

Adding nbstripout to remove Jupyter outputs, take 2.

## Checklist


owenlamont:
Hrmmm... not sure what the issue with readthedocs and nbsphinx is. Maybe if the notebook cells don't contain output it tries to run them, possibly related [this SO question/answer](https://stackoverflow.com/questions/69759351/error-jupyter-client-kernelspec-nosuchkernel-no-such-kernel-named-python3-occu).

owenlamont:
Maybe this idea was a complete mistake, looks like all the example notebooks, output and all, are incorporated into the [readthedocs site](https://shap.readthedocs.io/en/latest/). If I remove the outputs, either I need to disable nbsphinx from running the cells (and you'll lose the output in the readthedocs)... or I'd need to setup a Python environment with all the dependencies necessary to run all the notebooks and given the heaviness of the dependencies that seems completely impractical.

owenlamont:
I'll leave this open for a moment to see what you think @dsgibbons but I suspect I'll probably just abandon this idea.

PR #3216, opened 2023-08-24
ENH: add dot_size arg to beeswarm

nkasing:
## Overview

Add argument for `dot_size` in beeswarm plot

## Checklist

### DIFF
--- a/shap/plots/_beeswarm.py
+++ b/shap/plots/_beeswarm.py
@@ -27,7 +27,7 @@ from ._utils import (
 # TODO: Add support for hclustering based explanations where we sort the leaf order by magnitude and then show the dendrogram to the left
 def beeswarm(shap_values, max_display=10, order=Explanation.abs.mean(0),
              clustering=None, cluster_threshold=0.5, color=None,
-             axis_color="#333333", alpha=1, show=True, log_scale=False,
+             axis_color="#333333", alpha=1, show=True, dot_size=16, log_scale=False,
              color_bar=True, plot_size="auto", color_bar_label=labels["FEATURE_VALUE"]):
     """Create a SHAP beeswarm plot, colored by feature values when they are provided.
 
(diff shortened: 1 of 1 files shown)


thatlittleboy:
Hi @noah-asing , thanks for the interest in contributing to `shap`! Can you elaborate a little more about your use case (e.g. why is a fixed dot size of 16 not sufficient for you)? If possible, please also provide images of the current beeswarm that you are getting so we can also understand the issue at hand.

With plotting wrappers (like our `shap.plots.xxx` functions), I understand that there will always be customizations that users will require, but on our end I also wish to control the number of arguments that we expose. Or at least ensure that each new one we add in has a use-case that is well-documented (e.g. as an issue or in PRs like these).

nkasing:
Hi @thatlittleboy, there isn't a specific "need" per se, more of a "nice to have".

I thought it might be nice to have more definition in the beeswarm to try to visually separate NaN clusters, high-feature-value clusters and low-feature-value clusters. Reducing the `alpha` value resulted in the NaN clusters essentially disappearing completely underneath purple blobs, and I was hoping reducing the dot size might do the trick.

I saw that the `scatter` plots had a `dot_size` arg, so I didn't think there'd be a reason not to add `dot_size` to the beeswarm.

I didn't really think about controlling exposed arguments, so as of now this PR isn't really necessary (feel free to close).

thatlittleboy:
@noah-asing

> I thought it might be nice to have more definition in the beeswarm to try to visually separate NaN clusters, high-feature-value clusters and low-feature-value clusters. Reducing the alpha value resulted in the NaN clusters essentially disappearing completely underneath purple blobs and I was hoping reducing the dot size might do the trick.

Did it make it better?

---

I'm definitely open to adding the customization option, I'll leave this up for a while longer, if anyone else is interested in this functionality feel free to state your use case!

Thanks for the contribution in any case! :)

connortann:
I shall close this for now as we are thinking of implementing style customisation in a different manner: https://github.com/shap/shap/pull/3377

However, thank you for the PR in any case, as it's helpful to understand what features you need! We'll aim to get a proper customisation API implemented, and any and all contributions are highly appreciated.

PR #3229, opened 2023-08-29
Fix several notebook rendering issue #2743

### LINKED ISSUE #2743: Figures in Documentation are broken links
Your docs and articles are well written and very helpful, thank you! I was reading the following but was having a hard time following it 100% because the figures appear to be broken?  Thanks for your amazing work on this package!

https://shap.readthedocs.io/en/latest/example_notebooks/overviews/Be%20careful%20when%20interpreting%20predictive%20models%20in%20search%20of%20causal%C2%A0insights.html

jiandaoz1:
## Overview

Closes #2743   

## Checklist

- [v] All [pre-commit checks](https://pre-commit.com/#install) pass.
- [v ] Unit tests added (if fixing a bug or adding a new feature)


connortann:
Thanks for the PR! The changes to the docstrings look good.

However, I'm not clear what has changed in the notebook. The "Explain ResNet50 using the Partition explainer" notebook has changed, but the result seems to be identical, and in any case that notebook doesn't seem to be broken in the docs. So what is the purpose of the change to this file in this PR?

I don't think this PR actually fixes the bug you referenced, #2743 . That issue links to the notebook "Be careful when interpreting predictive models..", which isn't edited in this PR.

For any notebooks that _are_ edited, you can enable linting checks on those notebooks by adding the notebook path to `.pre-commit-config.yaml`, which enables the Ruff linter and Black formatter.

connortann:
Another note, I don't think this PR actually fixes the bug you referenced, #2743 . That issue links to the notebook "Be careful when interpreting predictive models..", whereas this PR changes a different notebook "Explain ResNet50 using the Partition explainer".

connortann (changes requested):
Please add the path of the notebook you have changed to the list of clean notebooks in `.pre-commit-config.yaml`

connortann:
Marking this as `stale` for now due as there has been on recent activity. Feel free to comment if you're still interested in working on this PR

PR #3231, opened 2023-08-30
MAINT: Reduce dependabot frequency and batch updates

connortann:
This PR is a therapeutic fix aiming to keep @thatlittleboy 's mental health in good working order.


thatlittleboy (approved):
Thanks！

PR #3235, opened 2023-09-02
DOCS: lint tabular examples/linear models

connortann:
Enables Ruff linting and Black formatting to another batch of notebooks in our docs. A step towards being able to run the same linting rules on the whole codebase.

Whereever possible I've tried to keep the diff minimal, by avoiding any changes to the outputs. However, the "multioutput regression" notebook inevitably has a very large diff, as the current file is in an unusual format from Google Colab, with different JSON indentation and lots of other metadata. So, to review that notebook, I think the most pragmatic solution is to compare the rendered ReadTheDocs pages with the current pages:

<https://shap--3235.org.readthedocs.build/en/3235/tabular_examples.html#linear-models>


thatlittleboy (approved):
@connortann LGTM in general, but I'm a little confused, seems like more than just the set of `Linear Models` notebooks that were modified here. Should the `.pre-commit-config.yaml` file be updated accordingly as well?

tom-hanks-laptop-lgtm

connortann:
> @connortann LGTM in general, but I'm a little confused, seems like more than just the set of `Linear Models` notebooks that were modified here. Should the `.pre-commit-config.yaml` file be updated accordingly as well?

Good spot. I updated the regex pattern to include everything in `tabular_examples`, apart from the `tree_shap_paper` subdirectory.

PR #3236, opened 2023-09-03
refactor: remove assertions in `force` plot, test verify_valid_cmap

thatlittleboy:
## Overview

This PR is part of a larger effort of refactoring `_force.py` and `_force_matplotlib.py`. This PR covers some of the smaller / non-functional changes, so we can focus on the core functionalities in the following PR.

* Removed most `assert` statements for enforcing logic, as it is not good practice. The only exception is the use of assert to test for `IPython` to be installed. (No strong reason for this, if there is a more suitable Exception type suggested, I can change it).
* Added a set of tests for `verify_valid_cmap()` function (defined inside `_force.py`). To ensure no errors are introduced in the refactoring.
* Add docstrings for some functions in `_force.py`.
* Re-use an existing `getjs()` function to read the js bundle (DRY principle).
* Added docstring for `initjs()`, just a quality-of-life improvement.

## Checklist

### DIFF
--- a/shap/plots/_force.py
+++ b/shap/plots/_force.py
@@ -314,14 +314,20 @@ def ensure_not_numpy(x):
     else:
         return x
 
+
 def verify_valid_cmap(cmap):
-    assert (isinstance(cmap, str) or isinstance(cmap, list) or str(type(cmap)).endswith("unicode'>")
-        ),"Plot color map must be string or list! not: " + str(type(cmap))
+    """Checks that cmap is either a str or list of hex colors"""
+    if not (isinstance(cmap, (str, list)) or str(type(cmap)).endswith("unicode'>")):
+        emsg = f"Plot color map must be string or list! Not {type(cmap)}"
+        raise ValueError(emsg)
+
     if isinstance(cmap, list):
-        assert (len(cmap) > 1), "Color map must be at least two colors."
+        if len(cmap) < 2:
+            raise ValueError("Color map must be at least two colors.")
         _rgbstring = re.compile(r'#[a-fA-F0-9]{6}$')
         for color in cmap:
-             assert(bool(_rgbstring.match(color))),"Invalid color found in CMAP."
+            if not _rgbstring.match(color):
+                raise ValueError(f"Invalid color {color} found in cmap")
 
     return cmap
 
--- a/shap/plots/_force.py
+++ b/shap/plots/_force.py
@@ -32,9 +32,11 @@ def force(base_value, shap_values=None, features=None, feature_names=None, out_n
 
     Parameters
     ----------
-    base_value : float
-        This is the reference value that the feature contributions start from.
+    base_value : float or shap.Explanation
+        If a float is passed in, this is the reference value that the feature contributions start from.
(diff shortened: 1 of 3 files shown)


shap/plots/_force.py
```diff
@@ -90,15 +117,19 @@ def force(base_value, shap_values=None, features=None, feature_names=None, out_n
...
-                            "for multi-output models try " \
-                            "shap.force_plot(explainer.expected_value[0], shap_values[0]).")
-
+            emsg = (
+                "In v0.20, force plot now requires the base value as the first parameter! "
+                "Try shap.plots.force(explainer.expected_value, shap_values) or "
+                "for multi-output models try "
+                "shap.plots.force(explainer.expected_value[0], shap_values[0])."
+            )
+            raise Exception(emsg)
```
connortann: Would a TypeError be more appropriate here, and in other instances below?

tests/plots/test_force.py
```diff
@@ -2,9 +2,41 @@
...
+    [
+        ("coolwarm", None),
+        (["#000000", "#ffffff"], None),
+        (777, "Plot color map must be string or list!"),
+        ([], "Color map must be at least two colors"),
+        (["#8834BB"], "Color map must be at least two colors"),
+        (["#883488", "#Gg8888"], r"Invalid color .+ found in cmap"),
+        (["#883488", "#1111119"], r"Invalid color .+ found in cmap"),
+    ],
+    ids=[
```
connortann: A minor suggestion, for readability you can provide the IDs in-line with the test case like so:

```python
from pytest import param

@pytest.mark.parametrize(
    "cmap, emsg",
    [
        param("coolwarm", None, id="valid-str"),
        ...
]

connortann (approved):
LGTM, with a few minor suggestions. Over to you to merge when happy!

connortann:
> Removed most assert statements for enforcing logic, as it is not good practice

I agree asserts aren't good practice for cases where we expect the exception could be reached by the calling code, for example validating bad inputs. I think that's the case in the PR, so I agree with you here 👍

However, personally I think asserts can be helpful in other contexts, so I'd be wary of declaring them Bad Practice overall. Specifically for checking our internal logic, in cases where the assert statement _should_ always pass unless we have any bugs. It can make debugging issues easier and can also serve as documentation of any assumptions we have about the state of any variables.

connortann:
@thatlittleboy it would be good to get this one finished off in time for our next release, if you have time this week to look at the comments above?

shap/plots/_force.py
thatlittleboy: Changed to `TypeError`. I also used `DimensionError` on the only other `Exception` call in `shap/plots/_force.py`.

tests/plots/test_force.py
thatlittleboy: oh nice!, didn't know this option existed

thatlittleboy:
@connortann Thanks for the comments, I made changes accordingly, please take a look when you get back.

connortann:
LGTM!

PR #3238, opened 2023-09-03
MAINT: Fix bad config in dependabot.yml

connortann:
## Summary

This fixes what I think is a typo in our dependabot config.

## Context

In #3231 I added a `groups` parameter, following the example in the GitHub docs:

<https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file>

However, we get [an error ](https://github.com/shap/shap/runs/16403344859):

```
The property '#/' contains additional properties ["groups"] outside of the schema when none are allowed
```

So this PR changes the indentation to be under the `updates` level, not the top level.

## Related:

I _think_ the GitHub docs example has the wrong level of indentation, so I've raised an issue there:
https://github.com/github/docs/issues/27987

Frustratingly, Dependabot only validate the config when pushed to the main branch, which is a known issue: https://github.com/dependabot/dependabot-core/issues/4605


thatlittleboy (approved):
LGTM!

PR #3246, opened 2023-09-05
MAINT: Reduce frequency of pre-commit autoupdates

connortann:
Reduces frequency of auto-updates, following guidance here:
<https://pre-commit.ci/#configuration>

We get one PR per week, which seems relatively excessive at the moment for minor linter changes.


thatlittleboy (approved):
LGTM, thanks! (I'm away from home this week so will be slightly slower in reviewing)

PR #3247, opened 2023-09-05
ENH: Rename explainers to usual long-form names for consistency with docs

### LINKED ISSUE #3244: ENH: Make documentation API names to match with there call names
### Problem Description

Short names are usually use to call SHAP explainers. Those names are defined in the `shap/__init__.py` file : 
```py
from .explainers._additive import Additive as AdditiveExplainer
from .explainers._deep import Deep as DeepExplainer
from .explainers._explainer import Explainer
from .explainers._gpu_tree import GPUTree as GPUTreeExplainer
from .explainers._gradient import Gradient as GradientExplainer
from .explainers._kernel import Kernel as KernelExplainer
from .explainers._linear import Linear as LinearExplainer
from .explainers._partition import Partition as PartitionExplainer
from .explainers._permutation import Permutation as PermutationExplainer
from .explainers._sampling import Sampling as SamplingExplainer
from .explainers._tree import Tree as TreeExplainer
```
In the documentation tree, these names are not mentionned. Making it unclear where is documented the function called by users.

https://shap.readthedocs.io/en/latest/api.html#explanation
https://shap.readthedocs.io/en/latest/api_examples.html#explainers

### Alternative Solutions

I can suggest 2 solutions:
1- add the alias name of a function in the docstring. This would make it more clear and easier when using the search bar.
2- use the alias name in the sidebar tree ?

### Additional Context

I also found that at least `Kernel` and `Gradient` are missing in the documentation.

### Feature request checklist

connortann:
## Overview

Closes #3244 

User-facing changes:
- Updates the docs to use the usual long names such as `TreeExplainer`, rather than the short form `Tree`.
- Adds missing explainers  `Kernel` and `Gradient`.

Implementation changes:
- Update the source code to use the long-form names, rather than the short-form names which are later aliased.

## TODO

- [x] Decide how to handle explainers in `shap.explainers.other` that are not present in the top-level namespace and do not have a "long-form" alias
- [x] Decide our long-term objective for handling these aliases (deprecate the short form? Or just prefer the long form?)

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -25,14 +25,16 @@ explainers
     :toctree: generated/
 
     shap.Explainer
-    shap.explainers.Tree
-    shap.explainers.GPUTree
-    shap.explainers.Linear
-    shap.explainers.Permutation
-    shap.explainers.Partition
-    shap.explainers.Sampling
-    shap.explainers.Additive
-    shap.explainers.Deep
+    shap.TreeExplainer
+    shap.GPUTreeExplainer
+    shap.LinearExplainer
+    shap.PermutationExplainer
+    shap.PartitionExplainer
+    shap.SamplingExplainer
+    shap.AdditiveExplainer
+    shap.DeepExplainer
+    shap.KernelExplainer
+    shap.GradientExplainer
     shap.explainers.Exact
     shap.explainers.other.Coefficient
     shap.explainers.other.Random
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -35,7 +35,7 @@ explainers
     shap.DeepExplainer
     shap.KernelExplainer
     shap.GradientExplainer
-    shap.explainers.Exact
+    shap.ExactExplainer
     shap.explainers.other.Coefficient
     shap.explainers.other.Random
     shap.explainers.other.LimeTabular
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -409,30 +409,27 @@ def setup(app):
 
     # need to assign the names here, otherwise autodoc won't document these classes,
     # and will instead just say 'alias of ...'
-    shap.explainers.Tree.__name__ = "Tree"
-    shap.explainers.Tree.__module__ = "shap.explainers"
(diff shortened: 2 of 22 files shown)


dsgibbons:
As mentioned [here](https://github.com/shap/shap/issues/3244#issuecomment-1716905240), my current opinion is to:
- Prefer long-form (but don't deprecate short-form)
- Create long-form aliases for `shap.explainers.other`

Deprecation of short-form API can be delayed until there is a more pressing requirement to do so.

connortann:
Given we will prefer the "long-form" names and want the documentation to match, perhaps it would be simpler to actually define these classes with the long-form name. That keeps the source code consistent with the docs, meaning we won't need any Sphinx workarounds to get references to work.

I'll try making the change and see how it looks...

connortann:
@dsgibbons , what do you make of this approach?

I haven't (yet) create long-form aliases for `shap.explainers.other` : we could perhaps do that as part of a follow-up PR, as this is already a fairly significant change. Or, if you prefer, I can implement it here.

dsgibbons:
I like your proposal, @connortann. I was going to suggest using long-form names everywhere, but I was worried about backwards compatibility. I had yet to consider just manually including assignments to the short-form. 

> I haven't (yet) create long-form aliases for shap.explainers.other : we could perhaps do that as part of a follow-up PR, as this is already a fairly significant change. Or, if you prefer, I can implement it here.

I think this is a pretty easy PR to read, but yes, perhaps this should be done in a future PR.

dsgibbons (approved):
I like this consistent API that doesn't require any Sphinx tricks. I approve, but I may just leave this PR open for a bit in case @thatlittleboy wants to comment.

connortann:
Huh - I noticed the tests are now failing. I think they passed before - I'll take a look (might be tomorrow). Re-marking as draft for now.

I'm surprised to see these failures, as they don't seem immediately related to the changes.

connortann:
I note that the tests are currently failing on master.  Created a new issue #3254 

This PR seems ready to go, so I'll mark it as "ready" again for when the tests are fixed.

PR #3253, opened 2023-09-13
DOCS: Remove redundant toctree level from API reference

connortann:
Configures Sphinx to not add a toctree entry for each domain object (class, function).

At present, there is a redundant level in the navigation sidebar, which always contains only one item:
image

This PR aims to remove the redundant final level, to make the docs a bit easier to browse:
image


thatlittleboy (approved):
terminator2-thumbsup

PR #3255, opened 2023-09-13
MAINT: Fix failing tests, pin XGBoost in test dependencies to 1.7.6

### LINKED ISSUE #3254: BUG: CI Tests failing since XGBoost 2.0.0 release
The CI tests seem to be failing on master, as of yesterday. This appears to be due to the release of xgboost `2.0.0`

Here's an example of how some of our baseline images have changed:

image

A fix could be to pin the dependency to a previous version, or even better to get the tests passing with the newer version.

connortann:
## Overview

Closes #3254

Pins XGBoost in test dependencies to `1.7.6`.

## Context

The baseline images we have are signifcantly altered in XGBoost 2.0.0. So, pinning the test dependency to the last working version seems like a pragmatic short-term fix to get the tests working again.

Once the tests are fixed, we should consider:

- Can we get the tests passing without needing this pin? #3256
- How to manage test dependencies. Should they all be pinned?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,7 +58,7 @@ test = [
   "pytest",
   "pytest-mpl",
   "pytest-cov",
-  "xgboost",
+  "xgboost==1.7.6",
   "lightgbm",
   "catboost",
   "gpboost",


connortann:
This has successfully fixed the tests. However, let's see if #3257 will fix the issue instead without needing to pin an old version.

connortann:
@dsgibbons , @thatlittleboy , would you kindly take a look and review? It would be good fix the tests, so other PRs are not blocked.

dsgibbons (approved):
LGTM, seems like a sensible fix for now

Issue #2204, opened 2021-10-03
Output of DeepExplainer


guyj-s1:
Hi,
I'm using DeepExplainer with a PyTorch model that outputs probabilities (softmax) for a binary classification task.
I'm wondering about the shap_values, whether or not they are in "probability" or "logit" units. 
Namely, for a sample to be explained, is each feature given a shap_value that is the amount of "probability" needed to get from the expected_value of the class to the model's output for that class? because those two are in "probability space".

I'm asking because: sum(shap_values) + expected_value != model_output (for a given class)

Help would be much appreciated!

Thanks,
Guy

saifullah3396:
I am also facing this on every model I try except for the mnist example given in the docs. I tried Alexnet with th same mnist example and it doesn't seem to satisfy this equation. Although the shap values still stay proportional to the output relatively. (Note: for any networks you're running, make sure relu inplace=False and no layer is repeated (e.g repitition of same relu activation) as these two conditions are not yet considered in shap deepexplainer computations as far as I could see). Also, avoid any kind of functional layer in the forward pass.

Also could you please post your output regarding how much it the equation deviates for each class?

rashecl:
For me, sum(shap_values) + expected_value == model_output when using a shap.DeepExplainer on keras model on a binary classification task. And the shap values appear to be output on a probability scale. I would like to output shap values in a log-odds domain. Is that possible @slundberg?

rabyj:
Thanks @saifullah3396, your comment helped me.

Turns out, in my case, the problem was that my model `forward` outputted probabilities, as so
```Python
def forward(self, x: torch.Tensor):
    """Return probabilities."""
    return F.softmax(self.forward_train(x), dim=1)

def forward_train(self, x):
    """Return logits."""
    return self._pt_model(x) # output of nn.Sequential(*layer_list)
```
By changing my model to output logits, I finally obtained `sum(shap_values) + expected_value == model_output`.

I think this answers the OP question, as `DeepExplainer` seems to expect "raw" output from the model, with no final functional layer like softmax in the forward, from what I understand. 

This issue could be closed now @guyj-s1 ?

PR #3265, opened 2023-09-15
Enh/3263: PyTorchDeep now also allows to check for additivity (activated by default)

### LINKED ISSUE #3263: ENH: Implement check_additivity for DeepExplainer with pytorch
### Problem Description

Many users are confused by the fact that shapley values sometimes are not adding up to the model output:
- https://github.com/shap/shap/issues/2875
- https://github.com/shap/shap/issues/2204
- https://github.com/shap/shap/issues/2785
- https://github.com/shap/shap/issues/2124
- https://github.com/shap/shap/issues/2765

(Partially) this is due to problems like explained in https://github.com/shap/shap/issues/2124#issuecomment-1719253006, https://github.com/shap/shap/issues/2204#issuecomment-933036857 or https://github.com/shap/shap/issues/2204#issuecomment-1526545669.

Fixing all of that probably is not that easy and shap will still be prone to future changes in the `torch` library. It would be nice to see `check_additivity` being implemented for the pytorch DeepExplainer and in case it shows a warning, having a nicely documented help page where users can find out what could actually be going on and providing alternate solutions (e.g. using `GradientExplainer` instead, which apparently does the trick for a couple of users).

Right now the variable is [there](https://github.com/shap/shap/blob/96467db542d519495023d735fa35d4855e25be9b/shap/explainers/_deep/deep_pytorch.py#L134), but unused.

The implementation of this check could be inspired by the [implementation](https://github.com/shap/shap/blob/96467db542d519495023d735fa35d4855e25be9b/shap/explainers/_deep/deep_tf.py#L320) in the tensorflow-part.

### Alternative Solutions

_No response_

### Additional Context

_No response_

### Feature request checklist

noxthot:
## Overview

Closes #3263 

- PyTorchDeep now also allows to check for additivity (sum of shapley values should equal model output)
- Activate this check by default

## Checklist

### DIFF
--- a/shap/explainers/_deep/__init__.py
+++ b/shap/explainers/_deep/__init__.py
@@ -86,6 +86,7 @@ class Deep(Explainer):
             self.explainer = PyTorchDeep(model, data)
 
         self.expected_value = self.explainer.expected_value
+        self.explainer.framework = framework
 
     def shap_values(self, X, ranked_outputs=None, output_rank_order='max', check_additivity=True):
         """ Return approximate SHAP values for the model applied to the data given by X.
--- a/shap/explainers/_deep/deep_pytorch.py
+++ b/shap/explainers/_deep/deep_pytorch.py
@@ -4,6 +4,7 @@ import numpy as np
 from packaging import version
 
 from .._explainer import Explainer
+from .deep_utils import _check_additivity
 
 torch = None
 
@@ -131,8 +132,7 @@ class PyTorchDeep(Explainer):
                 grads.append(grad)
             return grads
 
-    def shap_values(self, X, ranked_outputs=None, output_rank_order="max", check_additivity=False):
-
+    def shap_values(self, X, ranked_outputs=None, output_rank_order="max", check_additivity=True):
         # X ~ self.model_input
         # X_data ~ self.data
 
@@ -144,6 +144,8 @@ class PyTorchDeep(Explainer):
             assert isinstance(X, list), "Expected a list of model inputs!"
 
         X = [x.detach().to(self.device) for x in X]
+
+        model_output_values = None
 
         if ranked_outputs is not None and self.multi_output:
             with torch.no_grad():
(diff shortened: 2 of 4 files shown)


connortann:
Thanks for the PR, this looks like a good change!

I think we should consider where the best place to make this check is. In an ideal world, we would have sufficient unit tests checking for additivity that that this check never fails at run time. However, that's probably not realistic as we can't account for every possible input model.

Regarding default values, it looks like the default is actually defined in the the `DeepExplainer` class, and the defaults in `PyTorchDeep` have no effect (I think). In any case, it looks like the parameter was previously "True" by default, but had no effect in Pytorch.

May I suggest adding a unit test in `test_deep.py` to run this additivity check on an example model(s)?

noxthot:
Thanks for your fast feedback.

> I think we should consider where the best place to make this check is. In an ideal world, we would have sufficient unit tests checking for additivity that that this check never fails at run time. However, that's probably not realistic as we can't account for every possible input model.

I agree that this should not fail in runtime, but I am afraid that this check makes sense during runtime because:
- we can not account for every possible change and/or new feature in `torch` library
- problems like this: https://github.com/shap/shap/issues/2204#issuecomment-933036857

> Regarding default values, it looks like the default is actually defined in the the `DeepExplainer` class, and the defaults in `PyTorchDeep` have no effect (I think).

I guess you are right because this seems to be the usual entry point:
```
e = shap.DeepExplainer(...)
shap_values = e.shap_values(...)
```
but as far as I understand, you could also execute
```
shap_values = e.explainer.shap_values(...)
```
where it would take the default values within the `PyTorchDeep` or `TFDeep`. Imho it's better to have a single point of defining defaults for optional values to avoid confusion and ease maintenance and it seems like `__init__.py` is a good place for defining the defaults.

> In any case, it looks like the parameter was previously "True" by default, but had no effect in Pytorch.

Agreed

> May I suggest adding a unit test in `test_deep.py` to run this additivity check on an example model(s)?

Not sure if I understand correctly, but the additivity check is done by default due to the `check_additivity = True` by default and the tests are already covering the `shap_values()` method (which then also calls `check_additivity`).

connortann:
> tests are already covering the shap_values() method (which then also calls check_additivity).

Ah, indeed you are right. Well then, I think this PR is good to go!

(Apologies for the delayed response, I'm traveling this week)

connortann (approved):
LGTM, thanks again for the PR.

noxthot:
Thanks. I unfortunately just found a bug moments after you merged the request -> #3280

PR #3234, opened 2023-09-01
DOCS: Lint tabular_examples/tree based models

connortann:
Applies ruff and black to a batch of tabular example notebooks.

I have separated "Ruff" and "Black" as separate commits, for ease of reviewing.

I've tried to avoid changing any notebook metadata other than the code, to keep the diff as small as possible.

PR #3275, opened 2023-09-24
DOCS: clean up league of legends notebook

CloseChoice:
## Overview

works towards #3036.

Fix the League of Legends Win Prediction with XGBoost notebook

## Checklist

Note: Am not quite sure if I added the precommit hook correctly, it seems like the notebook is not tested, maybe I need to escape the whitespaces?


.pre-commit-config.yaml
```diff
@@ -37,6 +37,7 @@ repos:
                 notebooks/api_examples/.*.ipynb|
                 notebooks/benchmarks/.*.ipynb|
                 notebooks/genomic_examples/.*.ipynb|
+                notebooks/tabular_examples/tree_based_models/League of Legends Win Prediction with XGBoost.ipynb|
```
connortann: This list is a regular expression, so I don't think the spaces will match correctly here. You could use a shorter pattern instead:

```suggestion
                notebooks/tabular_examples/tree_based_models/League.*.ipynb|
```

This is just a temporary measure whilst some notebooks are not yet clean.

connortann:
Thanks for the PR, it would be great to clean up more of these notebooks.

As per my other comment on your other PR, may I suggest:
1. Fix the warning messages, which might be a case of just updating your 3rd party dependencies, and
2. Keep the cell execution numbers

thatlittleboy:
As mentioned in your other PR, this is potentially duplicated in #3234 (which was recently approved, I'll leave Connor to merge that in).
As usual, if there are other changes you would like to propose on top of that (linting + formatting), feel free to rebase on top of master once #3234 is merged in. Thanks!

connortann:
The code cells in this PR have been cleaned up nicely, so I'd like to get this version merged (overwriting whatever is in master).

However, I noticed there seems to be something off with the SHAP values.

The Shap plots look rather different in this new version. Here are some screenshots, with the existing docs on the left and this PR on the right-hand side:

image

image

The `# of inhibitor kills` parameter has a much more significant effect, and there are many features with zeros. It might be that the model is underfitted. Maybe the original plots were actually made with a different model than is written?

I think the original SHAP value plots are more appealing and a better demonstration of SHAP. May I suggest tweaking the params of the XGBoost model or leaving them as defaults, to see if we can restore SHAP values that resemble the original version a bit more closely?

connortann (changes requested):
The code cleanup looks great; however, as per my comment above, can you try to get the SHAP values looking similar to the original, perhaps by tweaking the model?

connortann:
I pushed a merge commit to fix the merge conflict, keeping the notebook as it is on this branch.

The pre-commit change is no longer required as this notebook is now linted on master.
EDIT: actually this notebook is missed by pre-commit due to a regex mistake, which will be fixed by #3298 .

connortann:
From testing locally, it looks like these XGBoost parameters lead to a more reasonable model that have better-looking Shapley plots. May I suggest using these instead?

```python
params = {
    "objective": "binary:logistic",
    "base_score": np.mean(yt),
    "eval_metric": "logloss",
}
model = xgb.train(
    params,
    dt,
    num_boost_round=10,
    evals=[(dt, "train"), (dv, "valid")],
    early_stopping_rounds=5,
    verbose_eval=25,
)
```

This leads to a "Gold earned per min" plot like this:

image

CloseChoice:
@connortann Thanks for the review and the suggested code. I reran the notebook with that and it looks way similar to the homepage version. Think this is ready for merge now

connortann (approved):
Look good, thanks again for the PR! It's great to have you on board as a contributor.

PR #3282, opened 2023-09-25
DOCS: Refactor nb/census income xgboost

CloseChoice:
## Overview

Works towards #3036 

 - clean up Census XGBoost notebook. There was not much to do, just remove some unnecessary commands and one typo

## Checklist


connortann:
Thanks for the PR! A few comments from an initial look:

1. It looks like there are several warnings in the new rendered notebook, which should be addressed. I think these are related to 3rd party libraries like pandas- downgrading to pandas 1.x may help. 
2. On a minor style note, may I suggest keeping the cell execution numbers? I think it looks visually nicer, and confirms to us maintainers that the notebook was run sequentially top-to-bottom in one run.

CloseChoice:
Thanks for the review. I updated the notebook and fixed most of the warnings. There is still the "invalid value encountered in divide" error, but I cannot fix this with downgrading pandas.

connortann (approved):
Looks great, thanks for the PR!

PR #3259, opened 2023-09-14
fix(deep_pytorch): replace deprecated register_backward_hook() by reg…

noxthot:
…ister_full_backward_hook() and remove special handling for MaxPool1d (was a workaround prior register_full_backward_hook() existed) #2124

## Overview

Closes #2124 

Description of the changes proposed in this pull request:
- Uses `register_full_backward_hook` instead of deprecated `register_backward_hook` in torch deep explainer.
- Removes workaround code for maxpool1d which is no longer needed since `register_full_backward_hook` returns the expected results.

## Checklist

- [x] All [pre-commit checks](https://pre-commit.com/#install) pass.
- [x] Unit tests added (if fixing a bug or adding a new feature)

PR #3283, opened 2023-09-26
MAINT: Extend pytorch tests to include CUDA

### LINKED ISSUE #3279: Run test_deep.py regarding torch also on cuda devices
### Problem Description

Currently bugs are not captured when something goes wrong when SHAP is used with torch on gpu.

The tests currently only run on `cpu`.

### Alternative Solutions

_No response_

### Additional Context

_No response_

### Feature request checklist

noxthot:
## Overview

Closes #3279  

- deep_test extended such that torch-related functions are also tested on cuda

## Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,7 +10,6 @@ except ImportError:
 
 import numpy as np
 import pytest
-import torch
 
 
 def pytest_addoption(parser):
@@ -70,5 +69,7 @@ def global_random_seed():
 def torch_devices_to_test():
     """Looks whether cuda is available. If so, torch-related tests are also tested on gpu.
     """
+
+    torch = pytest.importorskip('torch')
 
     return ["cuda", "cpu"] if torch.cuda.is_available() else ["cpu"]
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -63,4 +63,3 @@ def global_random_seed():
     `np.random.RandomState` rather than use the global numpy random state.
     """
     np.random.seed(0)
-    
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,6 +10,7 @@ except ImportError:
 
 import numpy as np
 import pytest
+import torch
 
 
 def pytest_addoption(parser):
@@ -63,3 +64,11 @@ def global_random_seed():
     `np.random.RandomState` rather than use the global numpy random state.
     """
     np.random.seed(0)
+
+
+@pytest.fixture()
+def torch_devices_to_test():
+    """Looks whether cuda is available. If so, torch-related tests are also tested on gpu.
+    """
+
+    return ["cuda", "cpu"] if torch.cuda.is_available() else ["cpu"]
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -63,13 +63,4 @@ def global_random_seed():
     `np.random.RandomState` rather than use the global numpy random state.
     """
     np.random.seed(0)
-
-
-@pytest.fixture()
-def torch_devices_to_test():
-    """Looks whether cuda is available. If so, torch-related tests are also tested on gpu.
-    """
-
-    torch = pytest.importorskip('torch')
-
-    return ["cuda", "cpu"] if torch.cuda.is_available() else ["cpu"]
+    
--- a/tests/explainers/test_deep.py
+++ b/tests/explainers/test_deep.py
@@ -261,15 +261,16 @@ def _torch_cuda_available():
     return False
 
 
-@pytest.mark.parametrize(
-        "torch_device", 
-        [
-            "cpu", 
-            pytest.param(
-                "cuda", marks=pytest.mark.skipif(not _torch_cuda_available(), reason="cuda unavailable (with torch)")
-            ),
-        ],
-)
+TORCH_DEVICES = [
+                    "cpu", 
+                    pytest.param(
+                        "cuda", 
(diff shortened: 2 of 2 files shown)


noxthot:
These tests will fail due to a bug regarding (i guess) MaxPool1D in current master of shap. That bug is fixed by https://github.com/shap/shap/pull/3259

So tests should succeed, once https://github.com/shap/shap/pull/3259 is merged.

noxthot:
And I probably need guidance regarding:
```
FAILED tests/explainers/test_deep.py::test_pytorch_mnist_cnn - AssertionError: Torch not compiled with CUDA enabled
FAILED tests/explainers/test_deep.py::test_pytorch_custom_nested_models - AssertionError: Torch not compiled with CUDA enabled
FAILED tests/explainers/test_deep.py::test_pytorch_single_output - AssertionError: Torch not compiled with CUDA enabled
```

Some tests currently also fail due to another issue regarding hugging face models.

connortann:
Fundamentally I don't think what you are trying to achieve is possible on GitHub hosted runners, as they do not have GPUs. 

<https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources>

connortann:
One option could be to include these new test cases using `pytest.mark.parameterize`, but skip if cuda is not available. That would make it easier to run the test on a local machine.

However, personally I think this would be more trouble than it's worth, as the tests would not be run with any regularity in a systematic way.

connortann:
Ah, apologies I hit "close" by mistake. On mobile today, and hit the wrong button!

noxthot:
> Fundamentally I don't think what you are trying to achieve is possible on GitHub hosted runners, as they do not have GPUs.
> 
> https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

Damn. I expected it to not work on Github runners but hoped for the best. In my commit [334def3](https://github.com/shap/shap/pull/3283/commits/334def33a8050e39f3ec731bdc3e1a2538e77ff5) I decided to use torch to check whether CUDA is available.
So locally - having cuda - more tests are performed than on Github. Think this is helpful since the bug i introduced in `check_additivity` when being used with CUDA, would not have slipped through that easily then. Of course this is no guarantee since it depends on the hardware of the developer, but I would say, it is better than nothing.

In case you do not agree, I would suggest to fixate the tests to `["cpu"]` in `conftest.py` since I also improved the tests themselves by making them capable of being run on gpu and I removed a little bit of obsolete code. So in case anyone wants to test `cuda` locally, there is a single place where the device can be added.

connortann:
I think your approach is sensible. It will indeed make it easier to run the tests locally with CUDA.

I have one suggestion for the implementation. Rather than have the fixture return a list of devices, the fixture could be parameterised and return just a single device.

This would have a few advantages:
- removes the need for each test to use a "for" loop
- each sub-test will automatically get a descriptive name
- the value of the fixture would be displayed if the test fails without the need for print statements.

It is possible to use `skipif` with `parameterize`, as in this example:
<https://doc.pytest.org/en/latest/how-to/skipping.html#skip-xfail-with-parametrize>

noxthot:
> I think your approach is sensible. It will indeed make it easier to run the tests locally with CUDA.
> 
> I have one suggestion for the implementation. Rather than have the fixture return a list of devices, the fixture could be parameterised and return just a single device.
> 
> This would have a few advantages:
> 
>     * removes the need for each test to use a "for" loop
> 
>     * each sub-test will automatically get a descriptive name
> 
>     * the value of the fixture would be displayed if the test fails without the need for print statements.
> 
> 
> It is possible to use `skipif` with `parameterize`, as in this example: https://doc.pytest.org/en/latest/how-to/skipping.html#skip-xfail-with-parametrize

Agreed. I also did not quite like the `for`-loop approach, but sometimes the hacky solution is so tempting.    : )

I refactored such that it now should use `skipif` and `parameterize` and also reverted some of the altered strings. Let's see what the tests say.

Thanks for your suggestions.

connortann:
Thanks for making those changes! Looking good.

I think we could reduce duplication a little. The torch device parameters could be defined just once as a tuple. Also, to generate a Cartesian product of multiple inputs, one can stack multiple `pytest.parameterize` decorators.

So in pseudo-code, the tests could be parameterised like this:

```python
TORCH_DEVICES = ("cpu", pytest.param("cuda", marks=...))


@pytest.mark.parametrize("device", TORCH_DEVICES)
def test_foo(device):
    ...

@pytest.mark.parametrize("device", TORCH_DEVICES)
@pytest.mark.parametrize("thing", [True, False])
def test_bar(device, thing):
    ...
```

noxthot:
Good point.

Still need to get familiar with `pytest`. Looks a lot better now (see [ed76d5c](https://github.com/shap/shap/pull/3283/commits/ed76d5ccf68f584e33fdb380b8461920941f8327) and [ea4c787](https://github.com/shap/shap/pull/3283/commits/ea4c78799500713d71ae5332ce3c75352035d0cf))

Thanks!

connortann:
The pytest parameterisation looks clean, nice work.

Regarding the content of the test files: do we still need to nest the logic within an inner function such as `run_test()` ?

The inner functions were previously used to enable a loop over multiple options, such as `interim=False` and `interim=True`. However, this is now cleanly handled by pytest `parameterize`, so I think the inner functions aren't needed now.

noxthot:
True that - there you go.

connortann (approved):
Looking good. Thanks for your work on this!

PR #3046, opened 2023-06-26
CHORE: Use faster test translation scenario, cut CI time by ~5mins

connortann:
Supports #3045

### Overview

Changes the model used in the translation scenario to amuch smaller one that will run much faster:
<https://huggingface.co/mesolitica/finetune-translation-t5-super-super-tiny-standard-bahasa-cased>

### Timings

The change seems to save ~5 min on Linux, and 7+ min on MacOS.

On Linux GH runner python 3.11, test timings before:

```
80.04s call     tests/explainers/test_partition.py::test_translation
76.44s call     tests/explainers/test_partition.py::test_translation_auto
76.27s call     tests/explainers/test_partition.py::test_translation_algorithm_arg
74.24s call     tests/explainers/test_partition.py::test_serialization
73.27s call     tests/explainers/test_partition.py::test_serialization_custom_model_save
69.42s call     tests/explainers/test_partition.py::test_serialization_no_model_or_masker
```

Test timings after:

```
22.75s call     tests/explainers/test_partition.py::test_translation
<19s   call     tests/explainers/test_partition.py::test_translation_auto
<19s   call     tests/explainers/test_partition.py::test_translation_algorithm_arg
20.38s call     tests/explainers/test_partition.py::test_serialization
20.70s call     tests/explainers/test_partition.py::test_serialization_custom_model_save
19.80s call     tests/explainers/test_partition.py::test_serialization_no_model_or_masker
```

Overall that's `328 seconds` faster on python 3.11 🎉

Timings vary between python versions and platforms, so the overall average speedup may differ.

### Note about protobuf

This new model require that we use `protobuf<=3.20.x`, or otherwise a TypeError is thrown. There is a related thread on stackoverflow [here](https://stackoverflow.com/q/72441758).

Here is the full traceback:

<details>

```
____________ ERROR at setup of test_serialization_custom_model_save ____________

    @pytest.mark.skipif(sys.platform == 'win32', reason="Integer division bug in HuggingFace on Windows")
    @pytest.fixture(scope="session")
    def basic_translation_scenario():
        """ Create a basic transformers translation model and tokenizer.
        """
        AutoTokenizer = pytest.importorskip("transformers").AutoTokenizer
        AutoModelForSeq2SeqLM = pytest.importorskip("transformers").AutoModelForSeq2SeqLM
    
        # Use a very small model, for speed
        name = "mesolitica/finetune-translation-t5-super-super-tiny-standard-bahasa-cased"
[...]

PR #3286, opened 2023-09-26
MAINT: Revert transformer model in tests to previous larger model

connortann:
This fixes the test suite, which has started failing recently due to a broken HuggingFace dependency.

#3046 switched the model used in the test suite to a "tiny" model, which reduced CI time by a full 5 minutes. However, it looks like the "tiny" model has been removed from HuggingFace.

This PR restores the previous, larger model. This will increase CI times by ~5mins, but should get the tests passing. So, once tests are fixed, we can find a better smaller model for the tests. Or, we could look for an alternative implementation that is faster and doesn't depend on external resources like HuggingFace.

### DIFF
--- a/tests/explainers/conftest.py
+++ b/tests/explainers/conftest.py
@@ -13,7 +13,8 @@ def basic_translation_scenario():
 
     # Use a *tiny* tokenizer model, to keep tests running as fast as possible.
     # Nb. At time of writing, this pretrained model requires "protobuf==3.20.3".
-    name = "mesolitica/finetune-translation-t5-super-super-tiny-standard-bahasa-cased"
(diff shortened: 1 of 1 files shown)


connortann:
@dsgibbons or @thatlittleboy I'd be grateful for a review on this if you get a minute. Keen to get this fixed so other PRs can be unblocked.

We've had a number of CI failures due to changing external dependencies, which highlights that the test suite is not a pure function of the code and is not reproducible. Over time, it would be great to fix the root causes of these issues so that the test suite cannot unexpectedly break on master.

This probably merits a deeper discussion; for now let's just get the tests passing once more!

thatlittleboy:
oopsies, sorry, forgot to submit my approval. but LGTM

PR #3262, opened 2023-09-14
WIP: Fix ngboost shapley value estimation for col_sample < 1

connortann:
This adds a test to reproduce the issue in #3261.

As of yet this is just a test, I have not found a fix..

PR #3294, opened 2023-09-30
Fix/ngboost col sample lt 1

### LINKED ISSUE #3261: BUG: test_ngboost fails with col_sample < 1
### Issue Description

The is a bug in shapley values for NGBoost when col_sample < 1. 
It seems that shapley values excludes some features when col_sample <1 though the the excluded feature is included in trees and has features importance.

image

model.feature_importances_

array([[7.51157818e-01, 2.93312779e-02, 2.81802693e-02, 0.00000000e+00,
        0.00000000e+00, 1.90468841e-01, 7.64777502e-04, 9.70170090e-05],
       [5.36576961e-01, 9.40885514e-02, 1.57119892e-03, 2.60296105e-02,
        3.21665665e-02, 2.49221611e-01, 4.92869891e-02, 1.10585114e-02]])

```python
# %%
import shap 
import ngboost as ngb

def test_ngboost():

    X, y = shap.datasets.california(n_points=500)
    model = ngb.NGBRegressor(n_estimators=20, col_sample = 0.9).fit(X, y)
    predicted = model.predict(X)

    # explain the model's predictions using SHAP values
    explainer = shap.TreeExplainer(model, model_output=0)
    
    explanation = explainer(X)
    explanation.base_values = explanation.base_values.reshape(-1)
    # check the properties of Explanation object
    assert explanation.values.shape == (*X.shape,)
    assert explanation.base_values.shape == (len(X),)

    # check that SHAP values sum to model output
    assert (
        np.abs(explanation.values.sum(1) + explanation.base_values - predicted).max()
        < 1e-5
    )

shap.summary_plot(explanation, max_display=30, plot_type="bar")
model.feature_importances_
```

### Traceback

```shell
AssertionError                            Traceback (most recent call last)
Cell In[16], line 1
----> 1 test_ngboost()

\NGBoost_Example\Claim_Severity_ngboost.py in line 14, in test_ngboost()
     64 assert explanation.base_values.shape == (len(X),)
     65 # check that SHAP values sum to model output
     66 assert (
---> 67     np.abs(explanation.values.sum(1) + explanation.base_values - predicted).max()
     68     < 1e-5
     69 )

AssertionError:
```

### Expected Behavior

_No response_

### Bug report checklist

### Installed Versions

0.42.1

CloseChoice:
## Overview

Closes #3261  

This PR only solves the problem if we have just 1 estimator. So for this script:
```python
import shap 
import ngboost as ngb
import numpy as np

X, y = shap.datasets.california(n_points=500)
model = ngb.NGBRegressor(n_estimators=1, col_sample = 0.9).fit(X, y)

predicted = model.predict(X)

# explain the model's predictions using SHAP values
explainer = shap.TreeExplainer(model, model_output=0)

explanation = explainer(X_predict_explain)

# check that SHAP values sum to model output
assert (
    np.abs(explanation.values.sum(1) + explanation.base_values - predicted).max()
    < 1e-5
)
```

## Checklist

Here is a breakdown of the issue: If we use ngboost with `col_sample < 1` some columns of X get dropped, the rest is reordered, see [here](https://github.com/stanfordmlgroup/ngboost/blob/master/ngboost/ngboost.py#L147-L151). Unfortunately, this might be different for each estimator. Which means, we would need to rewrite the `shap_values` function or `predict` function, not quite sure though.

### DIFF
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -1129,7 +1129,25 @@ class TreeEnsemble:
             self.internal_dtype = shap_trees[0].tree_.value.dtype.type
             self.input_dtype = np.float32
             scaling = - model.learning_rate * np.array(model.scalings) # output is weighted average of trees
-            self.trees = [SingleTree(e.tree_, scaling=s, data=data, data_missing=data_missing) for e,s in zip(shap_trees,scaling)]
+            # ngboost reorders the features, so we need to map them back to the original order
+            missing_col_idxs = [[i for i in range(model.n_features) if i not in col_idx] for col_idx in model.col_idxs]
+            feature_mapping = [{i: col_idx for i, col_idx in enumerate(list(col_idxs) + missing_col_idx)}
+                               for col_idxs, missing_col_idx in zip(model.col_idxs, missing_col_idxs)]
+            self.trees = []
+            for idx, shap_tree in enumerate(shap_trees):
+                tree_ = shap_tree.tree_
+                values = tree_.value.reshape(tree_.value.shape[0], tree_.value.shape[1] * tree_.value.shape[2])
+                values = values * scaling[idx]
+                tree = {
+                    "children_left": tree_.children_left.astype(np.int32),
+                    "children_right": tree_.children_right.astype(np.int32),
+                    "children_default": tree_.children_left,
+                    "features": np.array([feature_mapping[idx].get(i, i) for i in tree_.feature]),
+                    "thresholds": tree_.threshold.astype(np.float64),
+                    "values": values,
+                    "node_sample_weight": tree_.weighted_n_node_samples.astype(np.float64)
+                }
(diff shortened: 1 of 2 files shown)


thatlittleboy:
Thanks for the PR @CloseChoice , I've not yet taken a close look at this, but can you add your script as a test that reproduces the error so we make sure we don't regress next time?

Maybe Connor's PR #3262 might be of inspiration as well.

CloseChoice:
@thatlittleboy so I found a better fix that also fixes the problem for `n` estimators. The problem was, that due to the reordering of columns in ngboost, the predictions of the ngboost model do not match the predictions that we get from the treeensemble model, that we directly construct from the ngboost trees. I fixed this by reordering the features correctly. I also added a test that checks that the predictions are always identical

connortann (approved):
This is a great PR - thanks for the fix and the test!

PR #3297, opened 2023-10-03
MAINT: Add release note template

connortann:
Adds a template file for easier auto-generation of release notes according to category.

Docs for reference: <https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes>

Nb. The automatically generated release notes do not have to be perfect, as we can edit them after the first-pass has been generated. So, I think it is better to err on the side of producing too much information, as we can then easily strip out anything superfluous manually and remove any unused sections.

Example screenshot (Prior to manual edit):

image

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -6,6 +6,7 @@ changelog:
       - ci
     authors:
       - dependabot
+      - pre-commit-ci
   categories:
     - title: Added
       labels:
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -2,21 +2,25 @@ changelog:
   exclude:
     labels:
       - skip-changelog
-      - documentation
-      - ci
     authors:
       - dependabot
       - pre-commit-ci
   categories:
+    - title: Breaking changes
+      labels:
+        - BREAKING
     - title: Added
       labels:
         - enhancement
-    - title: Changed
-      labels:
-        - BREAKING
     - title: Fixed
       labels:
         - bug
+    - title: Documentation
+      labels:
+        - documentation
+    - title: Maintenance
+      labels:
+        - ci
     - title: Other Changes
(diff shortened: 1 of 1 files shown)


.github/release.yml
```diff
@@ -0,0 +1,22 @@
+changelog:
+  exclude:
+    labels:
+      - skip-changelog
+      - documentation
+      - ci
```
thatlittleboy: In the spirit of producing more information, should we leave these two tags out? I'm thinking of several scenarios where we maybe make some critical improvements to the CI / build process, and it's worth documenting those down as well.

Including the current documentation efforts (linting and what-not), those PRs get tagged as "Documentation", but I think it's better to keep them in the changelog (condensed as 1 "Change" of course, not as individual items).

.github/release.yml
```diff
@@ -0,0 +1,22 @@
...
+    authors:
+      - dependabot
+      - pre-commit-ci
+  categories:
+    - title: Added
+      labels:
+        - enhancement
+    - title: Changed
+      labels:
+        - BREAKING
```
thatlittleboy: I think `BREAKING` should be a category of its own, but maybe "Breaking changes" is a more suitable name compared to "Changed", and it should be the first category / at the very top. Higher than "Added".

.github/release.yml
connortann: Good point. We could move these to a new sections "documentation" and "maintenance", which we can then manually scan and either delete entirely, or cherry-pick anything notable.

In most releases I expect we will probably remove these sections, as they're unlikely to be useful for our users.

thatlittleboy (approved):
Looks good, excited to give this a try for 0.43!

terminator2-thumbsup

PR #3300, opened 2023-10-03
DOCS: Apply Ruff and Black to image_examples

connortann:
Applying linting to another batch of notebooks: ruff and black. No functional changes, other than fixing `import *` imports.

As usual, using pre-commit & a text editor to make the changes, in order to keep the diff as small as possible. Two notebooks have a large diff not viewable in GitHub, so for these I think we have to compare the ReadTheDocs views.

Nb. the "Front Page DeepExplainer MNIST Example" contains some errors in the outputs, so is in need of a cleanup. That's out of scope for this PR though, which just applies the linters.


thatlittleboy (approved):
LGTM, feel free to merge this in after 0.43 is released.

seal-of-approval-seal

connortann:
I'm happy to go ahead now: as this won't affect the released package, the timing won't really matter.

PR #3257, opened 2023-09-13
MAINT: Get tests passing with XGBoost 2.0 in test suite

connortann:
Closes #3256

Addressing the breaking changes detailed in:
<https://github.com/dmlc/xgboost/releases/tag/v2.0.0>

Issue #2984, opened 2023-06-12
xgboost ntree_limit deprecation warning


adnene-guessoum:
Hi,

Working on removing the boston dataset from the README.md (following https://github.com/slundberg/shap/pull/2873 ), I encountered this deprecation warning : 

**ntree_limit is deprecated, use `iteration_range` or model slicing instead.**

Here is the test example that triggered it: 

``` 
import xgboost
import shap

X, y = shap.datasets.california()                              
model = xgboost.XGBRegressor().fit(X, y)  

explainer = shap.Explainer(model)                                   
shap_values = explainer(X)
              
shap.plots.waterfall(shap_values[0])
```

This warning is due to ntree_limit being deprecated ( see : https://xgboost.readthedocs.io/en/stable/python/python_api.html ).

A fix for this is to use iteration_range instead of ntree_limit in _tree.py (and elsewhere). for instance, this :

https://github.com/slundberg/shap/blob/c24ff9ac72a2a21fc83f5097ffccf50f5429875d/shap/explainers/_tree.py#L347

would be : 

```try:                                                           
    phi = self.model.original_model.predict(                   
        X, iteration_range=(0, tree_limit), pred_contribs=True,
        approx_contribs=approximate, validate_features=False   
    )                                                          
```

This works in the previous case. If this is not an outrageous solution for some reasons, I would be happy to open a PR.

Let me know.

thatlittleboy:
Hey @adnene-guessoum , thanks for raising this issue. I don't think that what you proposed is an unreasonable fix.

I'll just note that the `iteration_range` parameter is introduced in xgboost 1.4 ([released Apr 11, 2021](https://github.com/dmlc/xgboost/releases/tag/v1.4.0)), for the benefit of any future users stumbling across this PR/issue.

A PR would be very much appreciated!

adnene-guessoum:
thanks for the reply @thatlittleboy . I am not entirely sure about the code coverage and the ci process. if you want to have a look.

PR #3307, opened 2023-10-04
Fix tests for xgboost 2.0

### LINKED ISSUE #3256: ENH: Test against XGBoost 2.0
### Problem Description

The tests are failing with XGBoost 2.0, as per #3254 . A short-term fix could be to just pin the previous version.

This issue however is to get our test suite passing against the latest version, which I think would be a preferable long-term solution:
<https://github.com/dmlc/xgboost/releases/tag/v2.0.0>

### Feature request checklist

CloseChoice:
## Overview

Closes #3256  

- recreated failing images
- updated xgboost dependency

## Checklist


connortann:
What is the underlying reason that the baseline plots have changed? I don't see anything else in the XGBoost release notes that suggested the fitted models will be different, so I would have thought we should be able to reproduce the previous plots pretty much exactly.

It might just be that the low-level implementation has changed and that the fitted model is now just stochastically different. If that's the case and XGBoost models are not very reproducible between versions, we could consider switching the model used to test the plots to something a bit more stable and deterministic, like a ridge regression.

In any case, it would be helpful to establish the root cause of the change in the plots, to confirm that nothing in SHAP has broken.

CarloLepelaars:
The fitted models have changed in XGBoost 2.0. The most important change being the removal of the `ntree_limit` argument. 

The error we get is `TypeError: Booster.predict() got an unexpected keyword argument: 'ntree_limit'`. I think removal of passing this argument will fix the main issues.

In this XGBoost test it is mentioned that this specifically breaks SHAP functionality:
https://github.com/dmlc/xgboost/blob/4d7a187cb0ebc2194abc1efe0aaee6f06af937c5/tests/python/test_with_shap.py#L16

connortann:
It looks like `ntree_limit` is _not_ used in shap's `TreeExplainer` in `_tree.py`, so I don't think that is the issue. (The parameter is indeed used in `pytree.py`, but that's a reference implementation not exposed in the shap package and not used in the shap tests as far as I can see.)

`ntree_limit` was removed in the main treeshap code by #2984

CarloLepelaars:
Thank you for looking at how we can resolve this!

Hmm, for my particular application that is the issue. I'm using `shap.explainers.Tree`, which is exposed in the `shap` package. Can't imagine I'm the only one that uses `shap.explainers`.

Indeed `ntree_limit` is used here:
https://github.com/shap/shap/blob/36556dd4ad3caac13c7967d9506da6a34ef6bacf/shap/explainers/pytree.py#L178

connortann:
Like I said, I don't think the file `pytree.py` is exposed. The object `shap.explainers.Tree` is an alias for `TreeExplainer`, which is defined here:

https://github.com/shap/shap/blob/a8efcc84cb4426ec03d49378422794aff1b7ce19/shap/explainers/__init__.py#L24

The `TreeExplainer` is defined in `_tree.py`, which does not use the `ntree_limit` parameter.

CarloLepelaars:
> Like I said, I don't think the file `pytree.py` is exposed. The object `shap.explainers.Tree` is an alias for `TreeExplainer`, which is defined here:
> 
> https://github.com/shap/shap/blob/a8efcc84cb4426ec03d49378422794aff1b7ce19/shap/explainers/__init__.py#L24
> 
> The `TreeExplainer` is defined in `_tree.py`, which does not use the `ntree_limit` parameter.

Aha, good point! You're right. I'll check again where exactly the issue occurred for me.

connortann:
Closing as this has now been addressed by #3257 . Thanks again to @CloseChoice and @CarloLepelaars for helping get this one sorted.

PR #3314, opened 2023-10-06
FIX: Xgboost sklearn model parameters that should be passed to xgboost.DMatrix

### LINKED ISSUE #3313: BUG: SHAP does not take into account certain key parameters from the xgboost sklearn models that can affect predictions
### Issue Description
The TreeEnsemble class responsible for automatically identifying and loading in the passed model does not take into account important parameters set at the class level on a sklearn compatible xgboost model. This is because these parameters are passed to the DMatrix object by the xgboost library in the lower level functions of their sklearn API before being sent to the fit/predict methods (see here for example: https://github.com/dmlc/xgboost/blob/4d7a187cb0ebc2194abc1efe0aaee6f06af937c5/python-package/xgboost/sklearn.py#L514)

I stumbled upon this because my shap values were not exactly equal to my predicted probabilities (a check I was performing) where I calculated the probabilities from the `predict_proba` method and I calculated the shap values using the TreeExplainer and then converted the shap values to shap probabilities (since the values are output as margin and not probabilities). The reason for the difference was that I use a custom value for representing missing data in my models (i.e. I set `missing = 1e-8` on my XGBClassifier instance) and when `predict_proba` gets called this missing value argument is passed lower down to the booster object via the DMatrix. SHAP however does not propagate these class level params further to the DMatrix as can be seen on these lines: https://github.com/shap/shap/blob/5121282dfd6fb66a21a4866ed3fa6ba6917a7efa/shap/explainers/_tree.py#L357-L358

For now the solution is to build the DMatrix up front before passing it to the explainer but I wanted to raise this as an issue because its a very silent bug and if users are not careful they will actually get the wrong shap values for their model.

These lines of the TreeEnsemble class would potentially need to be changed to account for reading in these parameters:
https://github.com/shap/shap/blob/5121282dfd6fb66a21a4866ed3fa6ba6917a7efa/shap/explainers/_tree.py#L971-L1015

```python

import pandas as pd
import xgboost
import shap

df = pd.DataFrame(your_data)

X, y, X_val, y_val = train_test_split(df[features], df[label], 0.3)

# Assuming this is a binary classification problem)
clf = xgboost.XGBClassifier(missing=1e-8)
clf.fit(X,y)
pred_probs = clf.predict_proba(X_val)[:,1]

explainer = shap.TreeExplainer(clf)
shap_values = explainer.shap_values(X_val)

pred_probs != sigmoid(explainer.expected_value + shap_values.sum(axis=1))
```
In above snippet if you build the DMatrix upfront and pass it to the shap_values call, the shap probabilities are equal to the predicted probabilities

### Expected Behavior
I would expect the sum of the calculated shap values to be equal to the predictions of the model (in both margin and probability terms). The `assert_additivity` method of shap only asserts to a 1e-2 order of magnitude that the margin predictions are correct but if additivity had been asserted on the probaibilities this would have been picked up earlier on.

### Bug report checklist

### Installed versions

Clone of the master branch of the repo

vancromy:
## Overview

Closes #3313 

- Changed the TreeEnsemble class to keep track of xgboost sklearn specific class parameters that may be useful for propogating to DMatrices later on
- Adapted the shap_values calculation method of the TreeExplainer to use DMatrix kwargs if they have been set on the explainer class

## Checklist

### DIFF
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -355,10 +355,8 @@ class TreeExplainer(Explainer):
             if self.model.model_type == "xgboost":
                 import xgboost
                 if not isinstance(X, xgboost.core.DMatrix):
-                    # Retrieve any DMatrix properties if they have been set on the class
-                    dmatrix_props = getattr(self, 'dmatrix_props', {})
-                    if "n_jobs" in dmatrix_props:
-                        dmatrix_props["nthread"] = dmatrix_props.pop("n_jobs")
+                    # Retrieve any DMatrix properties if they have been set on the TreeEnsemble Class
+                    dmatrix_props = getattr(self.model, "_xgb_dmatrix_props", {})
                     X = xgboost.DMatrix(X, **dmatrix_props)
                 if tree_limit == -1:
                     tree_limit = 0
@@ -996,8 +994,8 @@ class TreeEnsemble:
                 else:
                     self.model_output = "probability"
             # Some properties of the sklearn API are passed to a DMatrix object in xgboost
-            # We need to make sure we do the same here - gh3313
-            self.dmatrix_props = get_dmatrix_properties(model)
+            # We need to make sure we do the same here - GH #3313
+            self._xgb_dmatrix_props = get_xgboost_dmatrix_properties(model)
         elif safe_isinstance(model, "xgboost.sklearn.XGBRegressor"):
             self.original_model = model.get_booster()
             self.model_type = "xgboost"
@@ -1010,8 +1008,8 @@ class TreeEnsemble:
             if xgb_loader.num_class > 0:
                 self.num_stacked_models = xgb_loader.num_class
             # Some properties of the sklearn API are passed to a DMatrix object in xgboost
-            # We need to make sure we do the same here - gh3313
-            self.dmatrix_props = get_dmatrix_properties(model)
+            # We need to make sure we do the same here - GH #3313
+            self._xgb_dmatrix_props = get_xgboost_dmatrix_properties(model)
         elif safe_isinstance(model, "xgboost.sklearn.XGBRanker"):
             self.original_model = model.get_booster()
             self.model_type = "xgboost"
@@ -1024,8 +1022,8 @@ class TreeEnsemble:
             if xgb_loader.num_class > 0:
                 self.num_stacked_models = xgb_loader.num_class
             # Some properties of the sklearn API are passed to a DMatrix object in xgboost
-            # We need to make sure we do the same here - gh3313
-            self.dmatrix_props = get_dmatrix_properties(model)
+            # We need to make sure we do the same here - GH #3313
+            self._xgb_dmatrix_props = get_xgboost_dmatrix_properties(model)
         elif safe_isinstance(model, "lightgbm.basic.Booster"):
             assert_import("lightgbm")
             self.model_type = "lightgbm"
@@ -1661,7 +1659,7 @@ class IsoTree(SingleTree):
             self.features = np.where(self.features >= 0, tree_features[self.features], self.features)
 
 
-def get_dmatrix_properties(model):
+def get_xgboost_dmatrix_properties(model):
     """
     Retrieves properties from an xgboost.sklearn.XGBModel instance that should be passed to the xgboost.core.DMatrix object before calling predict on the model
     """
@@ -1671,6 +1669,10 @@ def get_dmatrix_properties(model):
         value = getattr(model, attribute)
         if value is not None:
             dmatrix_attributes[attribute] = value
+
+    # Convert sklearn n_jobs to xgboost nthread
+    if "n_jobs" in dmatrix_attributes:
+        dmatrix_attributes["nthread"] = dmatrix_attributes.pop("n_jobs")
     return dmatrix_attributes
 
 def get_xgboost_json(model):
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -1666,9 +1666,8 @@ def get_xgboost_dmatrix_properties(model):
     properties_to_pass = ["missing", "n_jobs", "enable_categorical", "feature_types"]
     dmatrix_attributes = {}
     for attribute in properties_to_pass:
-        value = getattr(model, attribute)
-        if value is not None:
-            dmatrix_attributes[attribute] = value
+        if hasattr(model, attribute):
+            dmatrix_attributes[attribute] = getattr(model, attribute)
 
     # Convert sklearn n_jobs to xgboost nthread
     if "n_jobs" in dmatrix_attributes:
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -991,6 +991,12 @@ class TreeEnsemble:
                     self.model_output = "probability_doubled" # with predict_proba we need to double the outputs to match
                 else:
                     self.model_output = "probability"
+            # Some properties of the sklearn API are passed to a DMatrix object in xgboost
+            # We need to make sure we do the same here - gh3313
+            properties_to_pass = ["missing", "n_jobs", "enable_categorical", "feature_types"]
+            for prop in properties_to_pass:
+                if hasattr(model, prop):
+                    setattr(self, f"dmatrix_{prop}", getattr(model, prop))
         elif safe_isinstance(model, "xgboost.sklearn.XGBRegressor"):
             self.original_model = model.get_booster()
             self.model_type = "xgboost"
@@ -1002,6 +1008,12 @@ class TreeEnsemble:
             self.tree_limit = getattr(model, "best_ntree_limit", None)
             if xgb_loader.num_class > 0:
                 self.num_stacked_models = xgb_loader.num_class
+            # Some properties of the sklearn API are passed to a DMatrix object in xgboost
+            # We need to make sure we do the same here - gh3313
+            properties_to_pass = ["missing", "n_jobs", "enable_categorical", "feature_types"]
+            for prop in properties_to_pass:
+                if hasattr(model, prop):
+                    setattr(self, f"dmatrix_{prop}", getattr(model, prop))
         elif safe_isinstance(model, "xgboost.sklearn.XGBRanker"):
             self.original_model = model.get_booster()
             self.model_type = "xgboost"
@@ -1013,6 +1025,12 @@ class TreeEnsemble:
             self.tree_limit = getattr(model, "best_ntree_limit", None)
             if xgb_loader.num_class > 0:
                 self.num_stacked_models = xgb_loader.num_class
+            # Some properties of the sklearn API are passed to a DMatrix object in xgboost
+            # We need to make sure we do the same here - gh3313
+            properties_to_pass = ["missing", "n_jobs", "enable_categorical", "feature_types"]
+            for prop in properties_to_pass:
+                if hasattr(model, prop):
+                    setattr(self, f"dmatrix_{prop}", getattr(model, prop))
         elif safe_isinstance(model, "lightgbm.basic.Booster"):
             assert_import("lightgbm")
             self.model_type = "lightgbm"
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -356,11 +356,10 @@ class TreeExplainer(Explainer):
                 import xgboost
                 if not isinstance(X, xgboost.core.DMatrix):
                     # Retrieve any DMatrix properties if they have been set on the class
-                    dmatrix_props = [p for p in dir(self) if p.startswith("dmatrix_")]
-                    dmatrix_kwargs = {p.split("dmatrix_")[-1]: getattr(self, p) for p in dmatrix_props}
-                    if "n_jobs" in dmatrix_kwargs:
-                        dmatrix_kwargs["nthread"] = dmatrix_kwargs.pop("n_jobs")
-                    X = xgboost.DMatrix(X, **dmatrix_kwargs)
+                    dmatrix_props = getattr(self, 'dmatrix_props', {})
+                    if "n_jobs" in dmatrix_props:
+                        dmatrix_props["nthread"] = dmatrix_props.pop("n_jobs")
+                    X = xgboost.DMatrix(X, **dmatrix_props)
                 if tree_limit == -1:
                     tree_limit = 0
                 try:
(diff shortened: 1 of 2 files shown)


vancromy:
@CloseChoice should we ping maintainers for a review/merge or is this something you can do?

I also noticed the build pipeline does not seem to run pytest. Are unit tests not run as part of a merge?

shap/explainers/_tree.py
```diff
@@ -991,6 +996,12 @@ def __init__(self, model, data=None, data_missing=None, model_output=None):
                     self.model_output = "probability_doubled" # with predict_proba we need to double the outputs to match
                 else:
                     self.model_output = "probability"
+            # Some properties of the sklearn API are passed to a DMatrix object in xgboost
+            # We need to make sure we do the same here - gh3313
+            properties_to_pass = ["missing", "n_jobs", "enable_categorical", "feature_types"]
```
CloseChoice: Can we have a function like this at someplace:
```
    def get_dmatrix_properties(model):
        properties_to_pass = ["missing", "n_jobs", "enable_categorical", "feature_types"]
        return {attribute: getattr(model, f'dmatrix_{attribute}') for attribute in properties_to_pass if getattr(model, f'dmatrix_{attribute}') is not None}
```

And then just call this function at all 3 places where you duplicated code and do
`self.dmatrix_props = get_dmatrix_properties(model)`

shap/explainers/_tree.py
```diff
@@ -355,7 +355,12 @@ def shap_values(self, X, y=None, tree_limit=None, approximate=False, check_addit
             if self.model.model_type == "xgboost":
                 import xgboost
                 if not isinstance(X, xgboost.core.DMatrix):
-                    X = xgboost.DMatrix(X)
+                    # Retrieve any DMatrix properties if they have been set on the class
+                    dmatrix_props = [p for p in dir(self) if p.startswith("dmatrix_")]
```
CloseChoice: With the suggested change we have
```python
dmatrix_props = getattr(self, 'dmatrix_props', {})
```
And can just pass this as kwargs to the DMatrix init

CloseChoice:
> @CloseChoice should we ping maintainers for a review/merge or is this something you can do?
> 
> I also noticed the build pipeline does not seem to run pytest. Are unit tests not run as part of a merge?

The maintainers will have a look at this once they find time. No need to ping them aswell. They also approve the remaining pipeline runs.

The PR already looks pretty good to me. I added some minor change proposals to not repeat code and use `setattr` less often but this seems to me pretty close to be ready.

shap/explainers/_tree.py
vancromy: Sure. I'll amend. I was thinking the same but since the TreeEnsemble class was meant to tackle all possible model types I didn't feel it was right creating a class method. But a general method is certainly a better idea :)

vancromy:
@CloseChoice apologies for a bit of spam. I introduced a typehint that seemed to break things so quickly reversed it. PR should be good to go for running unittests :)

shap/explainers/_tree.py
```diff
@@ -1671,6 +1661,15 @@ def _recalculate_value(tree, i , level):
             self.features = np.where(self.features >= 0, tree_features[self.features], self.features)
 
 
+def get_dmatrix_properties(model: xgboost.sklearn.XGBModel):
```
CloseChoice: shap does not use any type hints yet, so not sure what the maintainers will say about this. But if you use them could you also please add the correct return type `Dict[str, Union[int, float, str, ...]]` (not sure if I missed something hence `...`. And import Dict and Union from typing? Since we need to support python3.7 and onwards

CloseChoice (approved):
LGTM

shap/explainers/_tree.py
```diff
@@ -355,7 +355,11 @@ def shap_values(self, X, y=None, tree_limit=None, approximate=False, check_addit
             if self.model.model_type == "xgboost":
                 import xgboost
                 if not isinstance(X, xgboost.core.DMatrix):
-                    X = xgboost.DMatrix(X)
+                    # Retrieve any DMatrix properties if they have been set on the class
+                    dmatrix_props = getattr(self, 'dmatrix_props', {})
```
thatlittleboy: Shouldn't this be `getattr(self.model, "dmatrix_props", {})` instead? Since it's set on the TreeEnsemble object rather than the Explainer itself.

shap/explainers/_tree.py
```diff
@@ -355,7 +355,11 @@ def shap_values(self, X, y=None, tree_limit=None, approximate=False, check_addit
...
+                    if "n_jobs" in dmatrix_props:
+                        dmatrix_props["nthread"] = dmatrix_props.pop("n_jobs")
```
thatlittleboy: Why not do this check/adjustment in `get_dmatrix_properties` instead?

shap/explainers/_tree.py
```diff
@@ -1013,6 +1023,9 @@ def __init__(self, model, data=None, data_missing=None, model_output=None):
             self.tree_limit = getattr(model, "best_ntree_limit", None)
             if xgb_loader.num_class > 0:
                 self.num_stacked_models = xgb_loader.num_class
+            # Some properties of the sklearn API are passed to a DMatrix object in xgboost
+            # We need to make sure we do the same here - gh3313
```
thatlittleboy: Really small nitpick, but please change it to `GH #3313` so we stick to the same convention in the codebase

shap/explainers/_tree.py
```diff
@@ -1013,6 +1023,9 @@ def __init__(self, model, data=None, data_missing=None, model_output=None):
...
+            self.dmatrix_props = get_dmatrix_properties(model)
```
thatlittleboy: I'm gonna need @connortann to weigh in on this as well, but here's my proposal

```python
self._xgb_dmatrix_props = get_dmatrix_properties(model)
# instead of `self.dmatrix_props`
```

My logic is this: `TreeEnsemble` is meant to be a **uniform interface** for all tree-like models; it has a standard set of attributes and properties like `model_type`, `original_model` etc.  
I prefer not to add in attributes that are not generalizable to other tree-based models wherever possible, but seems its not possible in this case.  
(NB: There is also precedence of doing this if you look at how Catboost is wrapped in `TreeEnsemble`, there is a `self.cat_feature_indices` only for catboost to use. This was added before we took over maintainership.)

In the event we have to add in custom attributes (such as now), I prefer to 1) precede the attribute name with an underscore, to make it clear it is an internal implementation, and 2) prefix with some abbreviation of the model type, in this case `_xgb_xxxx`.

Thoughts?

shap/explainers/_tree.py
```diff
@@ -1648,6 +1661,18 @@ def _recalculate_value(tree, i , level):
             self.features = np.where(self.features >= 0, tree_features[self.features], self.features)
 
 
+def get_dmatrix_properties(model):
```
thatlittleboy: Suggest ot rename the function to `get_xgboost_dmatrix_properties()` for consistency with the other functions, and clarity

shap/explainers/_tree.py
```diff
@@ -1648,6 +1661,18 @@ def _recalculate_value(tree, i , level):
...
+    """
+    Retrieves properties from an xgboost.sklearn.XGBModel instance that should be passed to the xgboost.core.DMatrix object before calling predict on the model
+    """
+    properties_to_pass = ["missing", "n_jobs", "enable_categorical", "feature_types"]
+    dmatrix_attributes = {}
+    for attribute in properties_to_pass:
+        value = getattr(model, attribute)
+        if value is not None:
+            dmatrix_attributes[attribute] = value
```
thatlittleboy: May I understand, what the concern is here when you decided to only add to the dmatrix attribute dictionary if value is not None?

tests/explainers/test_tree.py
```diff
@@ -1197,6 +1197,30 @@ def test_xgboost_regression(self):
         expected_diff = np.abs(explanation.values.sum(1) + explanation.base_values - predicted).max()
         assert expected_diff < 1e-4, "SHAP values don't sum to model output!"
 
+    def test_xgboost_dmatrix_propagation(self):
+        xgboost = pytest.importorskip("xgboost")
+
+        def sigmoid(x):
+            return 1 / (1 + np.exp(-x))
```
thatlittleboy: You can just use `scipy.special.expit` instead of defining your own. scipy is a core dependency of shap, you can assume it always exists

EDIT: maybe not required, if you accept my suggestion from below to compare the output in margin space instead of probability space.

tests/explainers/test_tree.py
```diff
@@ -1197,6 +1197,30 @@ def test_xgboost_regression(self):
         expected_diff = np.abs(explanation.values.sum(1) + explanation.base_values - predicted).max()
         assert expected_diff < 1e-4, "SHAP values don't sum to model output!"
 
+    def test_xgboost_dmatrix_propagation(self):
```
thatlittleboy, line 1200: Please add a short test description and a link back to the GH issue (`GH #3313`)

tests/explainers/test_tree.py
```diff
@@ -1197,6 +1197,30 @@ def test_xgboost_regression(self):
...
+        xgboost = pytest.importorskip("xgboost")
+
+        def sigmoid(x):
+            return 1 / (1 + np.exp(-x))
+
+        X, y = shap.datasets.adult()
+
+        # Randomly set some features of X to NaN to test propagation of missing param
```
thatlittleboy: The comment is a little incoherent with the code below, you're setting some values of X to 1e-8, not NaN

tests/explainers/test_tree.py
```diff
@@ -1197,6 +1197,30 @@ def test_xgboost_regression(self):
...
+        X_nan = X.copy()
```
thatlittleboy, line 1210: Is the copy necessary here?

tests/explainers/test_tree.py
```diff
@@ -1197,6 +1197,30 @@ def test_xgboost_regression(self):
         expected_diff = np.abs(explanation.values.sum(1) + explanation.base_values - predicted).max()
         assert expected_diff < 1e-4, "SHAP values don't sum to model output!"
 
+    def test_xgboost_dmatrix_propagation(self):
+        xgboost = pytest.importorskip("xgboost")
+
+        def sigmoid(x):
+            return 1 / (1 + np.exp(-x))
+
+        X, y = shap.datasets.adult()
```
thatlittleboy: Actually the `adult` dataset is relatively huge (32k rows) for a dataset to use in a unit test. If it's not critical, maybe we could bump the number of samples down a little, e.g. `shap.datasets.adult(n_samples=1000)` as long as it still reproduces the issue

tests/explainers/test_tree.py
```diff
@@ -1197,6 +1197,30 @@ def test_xgboost_regression(self):
...
+        ] = 1e-8
+
+        clf = xgboost.XGBClassifier(missing=1e-8, random_state=42)
+        clf.fit(X_nan, y)
+        pred_probs = clf.predict_proba(X_nan)[:, 1]
+        explainer = shap.TreeExplainer(clf)
+        shap_values = explainer.shap_values(X_nan)
+        # check that SHAP values sum to model output
+        expected_diff = np.max((pred_probs - sigmoid(explainer.expected_value + shap_values.sum(axis=1))).abs())
+        assert expected_diff < 1e-5, "SHAP values don't sum to model output!"
```
thatlittleboy: I'm thinking if it's really necessary to do the comparison in probability space? Would it be easier to compare in margin space, then we don't even need the sigmoid? (You can refer to the other xgboost tests for an example)

thatlittleboy (changes requested):
Overall the PR makes sense to me, and thanks for the in-depth issue raised @vancromy , really appreciate it! (and thanks @CloseChoice for the initial pass at the code while we were away)

Just a couple of implementation details / conventions that need sorting out, then I think we're good to go.

shap/explainers/_tree.py
vancromy: Yes. Good spot! Will amend

shap/explainers/_tree.py
vancromy: Agreed. Will change this too.

shap/explainers/_tree.py
vancromy: I'm not an xgboost maintainer so can't say with certainty but I don't think there would be an issue if they were none. In fact, I originally included them all even if None but @CloseChoice asked me to add the if check. IMO there is no need to pollute the kwargs with Nones since those are the defaults in most cases in the xgboost library so could make sense to leave out. The only issue is if a user explicitly sets a setting to None and we omit it then there could be some unwanted behaviour so maybe best to keep?

tests/explainers/test_tree.py
vancromy: I don't think so but I think it may prevent setting with copy warnings from pandas.

tests/explainers/test_tree.py
vancromy: It makes it easier to do it in the margin space so I have amended the test to do so. The reason it was in the probaiblity space is because that is how I uncovered the bug when using `predict_proba` from the sklearn-like `xgboost.sklearn.XGBClassifier` implementation and comparing it to the lower level `predict` method on the `xgboost.core.Booster` which is what shap was using. The bug still exists either way since it was related to the DMatrix implementation in the `Tree` class of shap.

shap/explainers/_tree.py
vancromy: I agree with this and also noticed the same with the CatBoost specific attributes too. For me it makes sense to make it clear what the attribute is for so I have implemented your suggestion :)

vancromy:
@thatlittleboy I've hopefully addressed all your comments. Let me know if I missed something :) Should be ready for a final review.

shap/explainers/_tree.py
connortann: At present, `getattr(model, attribute)` will raise an `AttributeError` if the given attribute is not found, which seems undesirable. It would be better to pass on only those attributes which are present.

Also, as you say, we should probably allow users to define `None` as a possible value. So, we could avoid using `None` as a missing value sentinel.

How about something like:

```python
if hasattr(model, attribute):
    dmatrix_attributes[attribute] = getattr(model, attribute)
```

shap/explainers/_tree.py
connortann: I agree with you @thatlittleboy . In future I'm keen to refactor this rather huge method, but for now I think using that attribute naming convention is a great place to start.

shap/explainers/_tree.py
vancromy: Good spot. I agree its better practice to safeguard this. All `xgboost.sklearn.XGBModel` instances initialise the above attributes with sane defaults either way but in case their API changes in the future this is a more robust way of doing it. I have amended with your suggestion :)

vancromy:
It seems like all tests passed but one on macOS and Python 3.11. Its not a test I'm familier with and the error message suggests a connection failure. @thatlittleboy   or @connortann is this something that could be caused by my PR or is it a just a coincidence?

connortann:
> It seems like all tests passed but one on macOS and Python 3.11. Its not a test I'm familier with and the error message suggests a connection failure. @thatlittleboy or @connortann is this something that could be caused by my PR or is it a just a coincidence?

That's probably unrelated to this PR, as we do get intermitted connection errors with HuggingFace models. I've run the tests a second time; it will probably pass.

CloseChoice:
@thatlittleboy are you still missing something here?

shap/explainers/_tree.py
thatlittleboy: yep makes sense, thanks

thatlittleboy (approved):
Thanks @vancromy ! LGTM


simon-cowell-agt-thumbsup

Issue #2675, opened 2022-09-09
Why is the output format of shap_value of xgboost and RF different?


zevellong:
I use the exact same data, y has 2 classes, the output of **lightGBM and RF** are **lists**, and the list contains a two-dimensional matrix with the same dimension as the data, but the output of **xgboost** is **only a two-dimensional matrix**. 
, the xgboost plot and beeswam are the same

The `summary_plot` is also different, the xgboost plot is the same as `beeswam`.
``` python
# XGBoost: <class 'numpy.ndarray'> (32561, 12)
print("XGBoost:", type(shap_value), shap_value.shape)

# lightGBM: <class 'list'> (32561, 12) (32561, 12)
print("lightGBM:", type(shap_values1), shap_values1[0].shape, shap_values1[1].shape) 

# RF: <class 'list'> (32561, 12) (32561, 12)
print("RF:", type(shap_values2), shap_values2[0].shape, shap_values2[1].shape)
```

The test code is as follows:
``` python
from sklearn.model_selection import train_test_split
from sklearn.ensemble import RandomForestClassifier
import lightgbm as lgb
import shap
import numpy as np
import pandas as pd
import xgboost 

X,y = shap.datasets.adult()
X_display,y_display = shap.datasets.adult(display=True)
# create a train/test split
X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.2, random_state=7)

# XGBoost
d_train = xgboost.DMatrix(X_train, label=y_train)
d_test = xgboost.DMatrix(X_test, label=y_test)
params = {
    "eta": 0.01,
    "objective": "binary:logistic",
    "subsample": 0.5,
    "base_score": np.mean(y_train),
    "eval_metric": "logloss"
}
model = xgboost.train(params, d_train, 5000, evals = [(d_test, "test")], verbose_eval=100, early_stopping_rounds=20)
explainer = shap.TreeExplainer(model)
shap_values = explainer.shap_values(X)
shap.summary_plot(shap_values, X_display, plot_type="bar")
print("XGBoost:", type(shap_values), shap_values.shape)

# lightGBM
l_train = lgb.Dataset(X_train, label=y_train)
l_test = lgb.Dataset(X_test, label=y_test)
params = {
    "max_bin": 512,
    "learning_rate": 0.05,
    "boosting_type": "gbdt",
    "objective": "binary",
    "metric": "binary_logloss",
    "num_leaves": 10,
    "verbose": -1,
    "min_data": 100,
[...]

PR #3318, opened 2023-10-08
BREAKING: make shap explanations consistent for xgboost and lightgbm

### LINKED ISSUE #2887: TreeExplainer support for GradientBoostingClassifier in multiclass classification
TreeExplainer does not support GradientBoostingClassifier for multiclass classification. 
https://github.com/slundberg/shap/blob/45b85c1837283fdaeed7440ec6365a886af4a333/shap/explainers/_tree.py#L776

I would love to see this feature. Are there any plans to implement it or is anyone working on it?

Thanks!

### LINKED ISSUE #3187: BUG: Inconsistency and error in returned explainer objects with TreeExplainer
### Issue Description

I use shap to explain LightGBM models for different tasks. Depending on the explaination task (classification bin/multi or regression; with or without feature interactions), the TreeExplainer class returns ShapExplaination objects in various inconsistent shapes or in the case of task multiclass+interactions it crashes.

Here my testings:

Kind of task | Explainer object shape | My assessment
--
1. Regression+no_interactions | samples x features | OK
2. Regression+feature_interactions | samples x features x features | OK
3. Classification binary+no_interactions | samples x features x classes | inconsistent with 4.
4. Classification binary+feature_interactions | samples x features x features | inconsistent with 3. and 5.
5. Classification multiclass+no_interactions | samples x features x classes | inconsistent with 4.
6. Classification multiclass+feature_interactions | crashes with the traceback shown below | error

```python
from sklearn.datasets import load_diabetes
from sklearn.datasets import load_digits
from sklearn.datasets import load_breast_cancer
from lightgbm import LGBMRegressor
from lightgbm import LGBMClassifier
from shap.explainers import Tree as TreeExplainer

# data | regression | binary | multi class
data_reg = load_diabetes(as_frame=True)
data_bin = load_breast_cancer(as_frame=True)
data_mult = load_digits(as_frame=True)

# train models | regression | binary | multi class
model_reg = LGBMRegressor(**{'verbosity': -1,}).fit(data_reg.data, data_reg.target)
model_bin = LGBMClassifier(**{'verbosity': -1,}).fit(data_bin.data, data_bin.target)
model_mult = LGBMClassifier(**{'verbosity': -1,}).fit(data_mult.data, data_mult.target)

# Explainer
explainer_reg = TreeExplainer(model_reg)
explainer_bin = TreeExplainer(model_bin)
explainer_mult = TreeExplainer(model_mult)

# Explainations
explainations_reg = explainer_reg(data_reg.data, interactions=False)
explainations_reg_inter = explainer_reg(data_reg.data, interactions=True)
explainations_bin = explainer_bin(data_bin.data, interactions=False)
explainations_bin_inter = explainer_bin(data_bin.data, interactions=True)
explainations_mult = explainer_mult(data_mult.data, interactions=False)
explainations_mult_inter = explainer_mult(data_mult.data, interactions=True)
```

### Traceback

```shell
File ~\anaconda3\envs\mlenv\Lib\site-packages\shap\explainers\_tree.py:238 in __call__
    v = self.shap_interaction_values(X)
File ~\anaconda3\envs\mlenv\Lib\site-packages\shap\explainers\_tree.py:525 in shap_interaction_values
    X, y, X_missing, flat_output, tree_limit, _ = self._validate_inputs(X, y, tree_limit, False)
File ~\anaconda3\envs\mlenv\Lib\site-packages\shap\explainers\_tree.py:265 in _validate_inputs
    tree_limit = self.model.values.shape[0]

```

### Expected Behavior

1. No error in task case 6. (Classification multiclass+feature_interactions)
2. Consistent shape of the returned explainer objects

### Bug report checklist

### Installed Versions

conda-forge shap 0.42.1
conda-forge lightgbm 4.0.0
conda-forge python 3.11.4
Windows 11

CloseChoice:
## Overview

Closes #3187  
Closes #2887

- make the shap explanations outputs consistent for the binary feature interaction case for xgboost and lightgbm
- adapt test to the new output shape

NOTE: This is a breaking change.

## Checklist


CloseChoice:
Here are a couple of things that I think could need refactoring:
 - I think it would be good to have some kind `prediction_type` which can take the values: `binary_classification`, `multi_class_classifiction`, `multi_label_classification`, `regression`, `multi_target_regression`. With this, we could simplify some code
 - the whole structure of the code, [here](https://github.com/shap/shap/blob/master/shap/explainers/_tree.py#L234) and [here](https://github.com/shap/shap/blob/master/shap/explainers/_tree.py#L407-L418) looks to me like it is build bit by bit but could be refactored nicely and some complexity could be removed in the process aswell.

CloseChoice:
The tests will pass once PR #3325 gets merged

thatlittleboy:
I'm supportive of the proposed changes.

We'll just have to make sure the implementation is thorough and documentations are updated accordingly to educate the users on the API changes.

CloseChoice:
> I'm supportive of the proposed changes.
> 
> We'll just have to make sure the implementation is thorough and documentations are updated accordingly to educate the users on the API changes.

Anything else you want to be tested or added in this PR?

connortann:
May I suggest adding to the docstring:
1. A description of precisely what the shapes will be, depending on the prediction type
2. A change notice, e.g. "Changed in v0.44.0: the shape of returned shap values is changed from ... to ..."

We can also ensure we put this change notice in the release notes. The PR title and description is currently a little vague as to what exactly has changed in the API.

CloseChoice:
connected to #2675. Note to myself. Check if output of random forest is consistent with the other 3 outputs.

CloseChoice:
@connortann : I've added another comment to the discussion, see https://github.com/shap/shap/issues/3187#issuecomment-1890751284. This PR is almost done, just one existing test is failing but that would require quite a bit of work and I am not sure if we really want to go in this direction. Maybe we can sort the future design out first before I put in the work.

@connortann: Gentle push ;)

connortann:
Sorry for the delay in replying, thanks for the prompt.

Now that #3462 has been merged, there are a couple of merge conflicts to resolve. Other than that, I think it looks good to go - might as well try to bundle up the breaking changes together for the next release!

CloseChoice:
> Sorry for the delay in replying, thanks for the prompt.
> 
> Now that #3462 has been merged, there are a couple of merge conflicts to resolve. Other than that, I think it looks good to go - might as well try to bundle up the breaking changes together for the next release!

Nice, then I'll finish this up and ping you again. As mentioned in the comments, I'll align the shape to the model outputs!

CloseChoice:
So this is ready for another review round. Note that we are not consistent in the sense that random forest classifier for binary classifier has the shap values of shape: `observations x features x classes` compared to xgboost, lightgbm and catboost shapes result in:  `observations x features`. This has to do that random forest classifier works differently under the hood. The general rule for the output is that the shap values shape mirrors the shape of the prediction with scores (e.g. for random forest classifier the output of `predict_proba`).
All other cases are as expected and consistent among the classifiers

connortann:
Nice! I'll take a proper look tomorrow.

Just to reiterate my [comment above](https://github.com/shap/shap/pull/3318#issuecomment-1832315908), I think the new behaviour should be explicitly documented, together with a change note (which we can copy into the release notes).

shap/explainers/_tree.py
```diff
@@ -501,6 +495,8 @@ def shap_values(self, X, y=None, tree_limit=None, approximate=False, check_addit
         if check_additivity and self.model.model_output == "raw":
             self.assert_additivity(out, self.model.predict(X))
 
+        if isinstance(out, list):
```
connortann, line 510: Nit: suggest add a comment here explaining the interpretation of the various dimensions and why they are stacked.

tests/explainers/test_tree.py
```diff
@@ -1753,6 +1725,126 @@ def test_lightgbm_call_explanation(self):
...
+    assert (50, 8, 8) == e_lgbm.shape == e_xgb.shape == e_rfc.shape, \
+        f"Interactions LightGBM: {e_lgbm.shape}, XGBoost: {e_xgb.shape}, RandomForest: {e_rfc.shape}"
+    for outputs, pred in [(e_lgbm_bin, lgbm_pred), (e_xgb_bin, xgb_pred), (e_rfc_bin, rfc_pred), (e_cat_bin, cat_pred)]:
+        assert np.allclose(outputs.values.sum(1) + outputs.base_values, pred, atol=1e-4)
+    # todo: add catboost here once #1438 is fixed
+    for outputs, pred in [(e_lgbm, lgbm_pred), (e_xgb, xgb_pred), (e_rfc, rfc_pred)]:
+        assert np.allclose(outputs.values.sum((1, 2)) + outputs.base_values, pred, atol=1e-4)
+
+
+@pytest.mark.skip("Currently breaking on master, see GH: #3457 for discussion.")
```
connortann: Do we need to reopen #3457 or open an issue to capture this failure?

connortann (changes requested):
Looks great, very much looking forward to getting this merged. Just a few suggested changes, on top of the in-line suggestions above.

1. Please add some documentation of the new convention for what shape the shap_values will have under the various circumstances. The `Explanation` class docstring could be a place.
2. In the notebook, suggest restart & run all cells, so the outputs remain in linear execution order.
3. In the notebook, there are two cells with the warning below. I believe this can be fixed by updating to the new-style `shap_values = explainer(X)` rather than `shap_values = explainer.shap_values(X)`.

```
C:\Users\Tobias Pitters\programming\shap\shap\explainers\_tree.py:434: UserWarning: LightGBM binary classifier with TreeExplainer shap values output has changed to a list of ndarray
  warnings.warn('LightGBM binary classifier with TreeExplainer shap values output has changed to a list of ndarray')
```

CloseChoice:
> Looks great, very much looking forward to getting this merged. Just a few suggested changes, on top of the in-line suggestions above.
> 
>     1. Please add some documentation of the new convention for what shape the shap_values will have under the various circumstances. The `Explanation` class docstring could be a place.
> 
>     2. In the notebook, suggest restart & run all cells, so the outputs remain in linear execution order.
> 
>     3. In the notebook, there are two cells with the warning below. I believe this can be fixed by updating to the new-style `shap_values = explainer(X)` rather than `shap_values = explainer.shap_values(X)`.
> 
> 
> ```
> C:\Users\Tobias Pitters\programming\shap\shap\explainers\_tree.py:434: UserWarning: LightGBM binary classifier with TreeExplainer shap values output has changed to a list of ndarray
>   warnings.warn('LightGBM binary classifier with TreeExplainer shap values output has changed to a list of ndarray')
> ```

I implemented your points. 
A huge point to consider is that while the outputs of the tree explainer are consistent, the outputs between different explainers aren't with this change. Deep/Kernelexplainer (and the others) return a  List[(#observations x #features); #outputs] while the tree explainer returns a matrix (#observations x #features x #outputs). So this PR can only be the first of others which align the output of the others (should be an easy fix and test adjustments).

So not sure if we really want to merge this now or wait for the others to be ready. I would prefer the second option.

connortann:
> So not sure if we really want to merge this now or wait for the others to be ready. I would prefer the second option.

That's a good point. I agree it would make sense to update them all at the same time.

If we hold off merging, this could become a bit messy to maintain. To try to keep with the tidy "trunk-based development" we could:

1. Keeping this PR open and update the other explainers on this branch too;
2. Merge this into a feature branch, do separate PRs into that branch for the other explainers, then finally merge the feature branch into master when it's all ready; or
3. Use a feature flag for the new behaviour, so it can be implemented bit by bit then turned on all at once.

As you've taken the lead on implementing this change I'm happy to go with whatever works for you. Maybe (1) is simplest if it's not too much work to update the other explainers, or (2) or (3) is better if this is a longer-term thing (say longer than another week or so).

connortann (approved):

connortann:
PS - another option (4) could be just to merge this now as-is, and aim to get the other explainers updated in due course before we make a next release. That's probably simplest, although I'd only be comfortable if we could get it completed fairly quickly - say, within a week. Otherwise we'd be leaving `master` in an un-releasable state.

CloseChoice:
> > So not sure if we really want to merge this now or wait for the others to be ready. I would prefer the second option.
> 
> That's a good point. I agree it would make sense to update them all at the same time.
> 
> If we hold off merging, this could become a bit messy to maintain. To try to keep with the tidy "trunk-based development" we could:
> 
>     1. Keeping this PR open and update the other explainers on this branch too;
> 
>     2. Merge this into a feature branch, do separate PRs into that branch for the other explainers, then finally merge the feature branch into master when it's all ready; or
> 
>     3. Use a feature flag for the new behaviour, so it can be implemented bit by bit then turned on all at once.
> 
> 
> As you've taken the lead on implementing this change I'm happy to go with whatever works for you. Maybe (1) is simplest if it's not too much work to update the other explainers, or (2) or (3) is better if this is a longer-term thing (say longer than another week or so).

I would suggest we go with (1). I already fixed the deepexplainer on another (yet unpushed) branch and am almost through with the kernelexplainer. Need to check the other explainers after that but they should be even quicker than the 3 most complex ones.

CloseChoice:
@connortann I changed the code based on your remarks. Added the docstring to the shap_values methods for the changed explainers since the return varies (e.g. tree explainers only have one input while deepexplainers can have multiple). 

Another review of yours is greatly appreciated

shap/explainers/_deep/__init__.py
```diff
@@ -114,9 +114,11 @@ def shap_values(self, X, ranked_outputs=None, output_rank_order='max', check_add
...
         -------
         array or list
-            For a models with a single output this returns a tensor of SHAP values with the same shape
-            as X. For a model with multiple outputs this returns a list of SHAP value tensors, each of
-            which are the same shape as X. If ranked_outputs is None then this list of tensors matches
+            The return type and shape depend on the number of model inputs and outputs:
+              - one input one output: matrix of shape (#num_samples, *X.shape[1:]).
+              - one input multiple outputs: matrix of shape (#num_samples, *X.shape[1:], #num_outputs)
+              - multiple inputs one or more outputs: list of matrices, with shapes of one of the above.
+            If ranked_outputs is None then this list of tensors matches
```
connortann: Nit: I think the RST formatting is slightly wrong, I believe we need to:
- use `*` rather than `-` for the bullets
- Indent the bullets to the same level as the containing text
- and add a blank line before & after

We should also enclose the shapes in backticks as `*X.shape` bit seems to render as a kind of hyperlink.

```suggestion
            The return type and shape depend on the number of model inputs and outputs:

            * one input, one output: matrix of shape ``(#num_samples, *X.shape[1:])``.
            * one input, multiple outputs: matrix of shape ``(#num_samples, *X.shape[1:], #num_outputs)``
            * multiple inputs, one or more outputs: list of matrices, with shapes of one of the above.

            If ranked_outputs is None then this list of tensors matches
```

shap/explainers/_kernel.py
```diff
@@ -223,9 +226,10 @@ def shap_values(self, X, **kwargs):
             # vector-output
             s = explanation.shape
             if len(s) == 2:
-                outs = [np.zeros(s[0]) for j in range(s[1])]
+                outs = [np.zeros(s[0]) for _ in range(s[1])]
```
connortann: Can we get a test to cover this part of the diff, and the other lines which are flagged by CodeCov?

CloseChoice: That's a good point, will write tests for all those cases aswell

tests/explainers/test_gradient.py
```diff
@@ -97,12 +97,12 @@ def test_tf_keras_mnist_cnn(random_seed):
...
-    sess.run(model.layers[-1].input, feed_dict={model.layers[0].input: x_train[inds, :, :]}).mean(0)
+    outputs = sess.run(model.layers[-1].input, feed_dict={model.layers[0].input: x_test[:1]})
+    background = sess.run(model.layers[-1].input, feed_dict={model.layers[0].input: x_train[inds, :, :]})
+    expected_value = background.mean(0)
 
-    sums = np.array([shap_values[i].sum() for i in range(len(shap_values))])
-    d = np.abs(sums - diff).sum()
-    assert d / (np.abs(diff).sum() + 0.01) < 0.1, "Sum of SHAP values does not match difference! %f" % (d / np.abs(diff).sum())
+    sums = shap_values.sum((1, 2, 3))
+    assert np.allclose(sums + expected_value, outputs, atol=1e-4)
```
connortann: This new syntax for the assert is a lot clearer, nice. As a minor nitpick, `np.testing.assert_allclose` might be slightly better than `assert np.allclose`. It's the same function, but with slightly more appropriate defaults and more informative error tracebacks.

shap/explainers/_deep/__init__.py
connortann: Same comment for the other occurrences of this docstring :)

connortann:
As this is a fairly significant breaking change, I would suggest including the [versionchanged](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionchanged) directive in the relevant docstrings. This is probably something we should do for all breaking changes.

I believe it is used like this:

```
Returns
-------
array
    Returns a matrix. The shape depends on the number....

    .. versionchanged:: 0.45.0.
       Return type for models with multiple outputs changed from [etc...]
```

connortann:
This comment caught my eye : https://github.com/microsoft/LightGBM/pull/6348#issuecomment-1973471486

> I will take a guess that the reason why it's output like this is because SciPy's sparse matrices do not support n-dimensional arrays and the output here is 3-dimensional.

I wonder if SHAP was originally set up as lists of 2D arrays for the same reason. I don't think we have any test cases of this type of model. There are some cases where the inputs are sparse, but I don't know if we have any cases where the SHAP outputs are also sparse?

I saw this comment from Scott:

> We should really support returning a csr at some point, but what you did seems like a good stopgap.

_Originally posted by @slundberg in https://github.com/shap/shap/issues/138#issuecomment-407878718_

Related:
- #138
- #200

CloseChoice:
> This comment caught my eye : [microsoft/LightGBM#6348 (comment)](https://github.com/microsoft/LightGBM/pull/6348#issuecomment-1973471486)
> 
> > I will take a guess that the reason why it's output like this is because SciPy's sparse matrices do not support n-dimensional arrays and the output here is 3-dimensional.
> 
> I wonder if SHAP was originally set up as lists of 2D arrays for the same reason. I don't think we have any test cases of this type of model. There are some cases where the inputs are sparse, but I don't know if we have any cases where the SHAP outputs are also sparse?
> 
> I saw this comment from Scott:
> 
> > We should really support returning a csr at some point, but what you did seems like a good stopgap.
> 
> _Originally posted by @slundberg in [#138 (comment)](https://github.com/shap/shap/issues/138#issuecomment-407878718)_
> 
> Related:
> 
>     * [Support for csr sparse matrices. #138](https://github.com/shap/shap/issues/138)
> 
>     * [Adding CSR support for KernelExplainer #200](https://github.com/shap/shap/pull/200)

I looked at this and it does not seem like there is an easy-to-use solution for n-dimensional sparse arrays (with n > 2). Though the approach taken in this PR might limit us in this regard in the future to only 2d sparse matrix outputs I would still go for this.

If we really come to the point of implementing the sparse matrices everywhere we could still create our own class that stores the sparse 2d matrices in a list but allows slicing as usual. Or we could just be non-consistent in that case, that would be the price for more efficiency. Either way I believe for the overwhelming majority of the usecases the proposed shape makes sense.

CloseChoice:
@connortann : I added tests for all lines except the linear one. The reason for the missing linear line is that there seems to be a bug for multi-class linear models. I added a test and let this xfail. Would be really glad if you could take a look at this PR (hopefully) one final time

Note to myself: check https://shap--3318.org.readthedocs.build/en/3318/generated/shap.TreeExplainer.html#shap.TreeExplainer

shap/explainers/_deep/__init__.py
```diff
@@ -126,7 +126,7 @@ def shap_values(self, X, ranked_outputs=None, output_rank_order='max', check_add
             ranked_outputs, and indexes is a matrix that indicates for each sample which output indexes
             were chosen as "top".
 
-           .. versionchanged:: 0.45.0.
+           .. versionchanged:: 0.45.0
            Return type for models with multiple outputs and one input changed from list to np.ndarray.
```
connortann: I believe RST directives need to be indented like so:

```suggestion
           .. versionchanged:: 0.45.0
              Return type for models with multiple outputs and one input changed from list to np.ndarray.
```

Same comment for the other directives below.

shap/explainers/_gradient.py
```diff
@@ -142,7 +142,7 @@ def shap_values(self, X, nsamples=200, ranked_outputs=None, output_rank_order="m
             ranked_outputs, and indexes is a matrix that tells for each sample which output indexes
             were chosen as "top".
 
-           .. versionchanged:: 0.45.0.
+           .. versionchanged:: 0.45.0
            Return type for models with multiple outputs and one input changed from list to np.ndarray.
```
connortann, line 146:
```suggestion
           .. versionchanged:: 0.45.0
              Return type for models with multiple outputs and one input changed from list to np.ndarray.
```

shap/explainers/_tree.py
```diff
@@ -550,15 +550,20 @@ def shap_interaction_values(self, X, y=None, tree_limit=None):
...
         Returns
         -------
-        array or list
-            For models with a single output, this returns a tensor of SHAP values
-            (# samples x # features x # features). The matrix (# features x # features) for each sample sums
+        array
+            Returns a matrix. The shape depends on the number of model outputs:
+                - one output: matrix of shape (#num_samples, #features, #features).
+                - multiple outputs: matrix of shape (#num_samples, #features, #features, #num_outputs).
+            The matrix (#num_samples, # features, # features) for each sample sums
```
connortann: I believe to render bullet points in Restructured text there needs to be a newline before&after, and the indentation of the bullets should match the surrounding text:

```suggestion
            Returns a matrix. The shape depends on the number of model outputs:

            * one output: matrix of shape (#num_samples, #features, #features).
            * multiple outputs: matrix of shape (#num_samples, #features, #features, #num_outputs).

            The matrix (#num_samples, # features, # features) for each sample sums
```

connortann (changes requested):
Some slight formatting adjustments but otherwise LGTM!

CloseChoice:
> Some slight formatting adjustments but otherwise LGTM!

Merged your suggestions. Thanks once again for the review.

connortann (approved):
Thanks again for all your work on this. There are a few remaining docs formatting issues, but I can address them in a subsequent PR. Let's get it in!

connortann:
@CloseChoice  I noticed a test failure on `master`, which might be related. Would you kindly take a look - perhaps we need to loosen the tolerances of a few of the `np.asset_all_close` calls, and/or ensure the random seed is fixed?

Example failing run:
https://github.com/shap/shap/actions/runs/8186093687/job/22383771181

I'll re-run the failed job to see if it's reproducible...

CloseChoice:
@connortann yes, looks like we the allowed tolerance is too low. Will add a pull request for this.

From the failed run logs:
```
Mismatched elements: 1 / 96 (1.04%)
Max absolute difference: 1.53473644e-05
Max relative difference: 4.40842603e-05
```

imatiach-msft:
The title and description for this PR is completely wrong. It does not just "make the shap explanations outputs consistent for the binary feature interaction case for xgboost and lightgbm". It breaks the output format of every explainer in shap from ```(# classes x # examples x # features)``` to ```(# examples x # features x # classes)``` and changes the output from list of 2D np.ndarrays to 3D np.ndarray. This is a huge headache for downstream users and libraries that have used the old format for many years now - in just our OSS library this shap update has caused 60 test failures when we can usually update without any issues. Also, I don't understand why every explainer needs to be changed instead of xgboost. The old format seemed much better and was much easier to understand - it was a lot easier to tell if the model is a classifier or regressor. Indexing into the class to get the shap values for that class is much easier than trying to figure out how to get the third dimension and slice the array for users - often this is the most common operation and what users pass to visualizations. Many other text/vision explainers and other libraries follow this logic as well. The new format is harder to understand and use in my opinion and xgboost should have been changed instead of lightgbm and every shap explainer (kernel, deep tf/pytorch, gradient, linear, tree) in this library.

connortann:
@imatiach-msft I'm troubled to hear that this change is causing you a huge headache to accommodate. I'm keen to see if we can collectively figure out the best way forward.

The original inconsistency described in #3187 was quite a compelling argument that the shapes of the Explanation object were hard to understand and inconsistent.  There's still a huge amount of inconsistent & legacy code in shap, and the documentation is lacking, so I think it can be hard to understand and interpret the various attributes of Explainer and Explanation objects. The issue tracker was (and still is) swamped with bug reports that relate to this kind of inconsistency in shapes, which I suppose fed the motivation to try to standardise it.

As a long-time contributor to the ML interpretability ecosystem, do you have a view of what would be a better resolution to the discussion in #3187? It's theoretically possible to make another breaking change I suppose, if that is really merited.

It's worth emphasising as well that this package is direly short of maintainers, so whilst the intent is always to try to make the package more internally consistent and helpful for downstream users, changes like this do not get as much discussion and review as they perhaps deserve. #3559 has more context. Anything we can do to grow the pool of maintainers would be helpful, especially to include folk involved in downstream packages.

imatiach-msft:
@connortann  I think there are two parts to this:
1.) How can we reduce breaking changes in the future
2.) Is the new format ```(# examples x # features x # classes)``` superior to the old one ```(# classes x # examples x # features)```

In regards to the first point, actually shap did this very well several years ago. Scott Lundberg added a new API, ```__call__```, which returned an explanation object instead of changing the old shap_values API and format. It would be better if we could do this in the future if we want to change the output and we don't have a choice.
Another part of this is to define clearly what the explanation format from shap_values is and have tests to validate it. It's one thing to fix one explainer because it's outputting something off, it's quite another to modify every single explainer's output format. That's going to cause major issues for users, since as soon as they install a newer or older version of shap they will need to change a lot of their code, and downstream libraries that handle shap explanations now have to deal with multiple explanation formats based on the shap package version.

In regards to the second point, I still believe the old format was superior to the new one for shap_values and this format makes less sense. However, given that this breaking change was already made, and we want to reduce breaking changes, I'm not sure if doing another breaking change is a good idea. We should just make sure we don't do any more breaking changes like this. In addition, Scott added the new ```__call__``` API on explainers which returns an explanation object which is the new output format. Our library still uses the legacy shap_values API and it's interesting that shap users/contributors are making a lot of changes to it - which seems to indicate that it is still widely used despite being the old API.

imatiach-msft:
For more context on the ```__call__``` API change see this PR from 2020:
"Major refactor to support new API (backward compat retained)"
https://github.com/shap/shap/commit/ddcfa16b2811015bf3e3a264097c1418240d0f82
Note how there was an emphasis to keep backwards compatibility and add this new ```__call__``` API that returns an explanation object, whereas this change completely breaks the shap_values API format that has been there for more than 7 (?) years now

connortann:
It would be great to hear @CloseChoice & @thatlittleboy 's perspective, as I think they are more knowledgeable about the various tradeoffs in this PR. Unfortunately there haven't been any active maintainers recently so I really value getting more input.

We are overdue for a release, so I wonder if there is an opportunity to try to mitigate the headache for downstream packages somewhat. I'm struggling to come up with a workable proposal but this seems like a very important issue so I'd like to help out if I can.

>  I'm not sure if doing another breaking change is a good idea.

Yeah, I'm very hesitant about this idea too. On the flip side though, it could be worthwhile if a) it makes the package better in the long term, and b) we find a way to mitigate harm to downstream packages.

Maybe there could be a configurable setting, like `output_shape: Literal["classfirst", "classlast"]`, so that users could opt-out of the new behaviour? That would add a lot of complexity, and wouldn't work for 0.45 & 0.46 unless this was somehow back-ported as a patch release.

Or alternatively, hypothetically if we switched back to the old-style output format and made a new release, downstream packages could include a dependency pin to exclude the problematic versions. We could retrospectively edit the release notes to highlight the changed behaviour and encourage users to upgrade. Whilst it's not ideal to make two breaking changes in fairly quick succession, do you think it would be better to restore the prior behaviour?

imatiach-msft:
@connortann I like the idea of having the ```output_shape: Literal["classfirst", "classlast"]```. I think my main complaint is that:
1.) The PR title and description is wrong. The scope of changes implemented here were much bigger than in the title and description.
2.) shap should not make breaking changes so easily to every explainer in the library. There needs to be tests, verification and documentation about the expected format and before making breaking changes like this there needs to be a lot more discussion and agreement as this affects many notebooks, existing production code, etc.
3.) Ideally, instead of making breaking changes, a new API like the ```__call__``` API change should be implemented. Also, I'm not sure why we are making changes to shap_values method since it is the legacy way to call explainers and the call API was supposed to be the new preferred way. The fact that we are seeing changes makes me wonder if users actually still prefer to use the shap_values API over the call API. If we are to make breaking changes to the explanation object from call API, I think it would be useful to have some version on the explanation object so it will be easier for downstream libraries to maintain code and have different logic based on the explanation object version. Our library was written several years before the call API was added and it still uses the legacy shap_values method which returns the simple array/list of arrays format.

CloseChoice:
Totally agree, that the title is misleading and we should have better documentation on this. I saw a couple issues that relate to the breaking changes in here, so it's a couple users that are affected. I would still argue that the format is easier to grasp, but that might just be my implementation bias and doesn't help with the problem at hand.

For future changes I totally agree, we should introduce new APIs instead of breaking old ones, and if not possible otherwise, either back-ported new changes to older versions to give users time to adapt or throw deprecation warnings.

EDIT: in retrospect adding a `output_shape: Literal["classfirst", "classlast"]` would have still be a breaking change but probably kept most of the downstream code working. Would it be an option to add these parameters, but still keep "classlast" as default, to avoid another breaking change? That would make it the most convenient for users to upgrade? I am also willing to do this.

I saw that we'll changed the default value of KernelExplainer`s `shap_values` function in 0.47.0. And while I agree that `shap_values` is somehow the legacy interface, this is not clearly communicated in the docs, so IMO we'll need to guarantee consistency between `__call__` and `shap_values` or make it transparent that we'll freeze the `shap_values` API, so not adding any new keywords but also saveit from breaking. @connortann @imatiach-msft  what do you think about that?

connortann:
To address @imatiach-msft 's comments:

> The PR title and description is wrong. The scope of changes implemented here were much bigger than in the title and description.

Yes, this is totally fair. I should have picked up on that as a reviewer. Also, I think the release notes aren't clear enough about the nature of the change and they can be improved.

> shap should not make breaking changes so easily to every explainer in the library. There needs to be tests, verification and documentation about the expected format and before making breaking changes like this there needs to be a lot more discussion and agreement as this affects many notebooks, existing production code, etc.

Again, fair. I think the existing tests are actually pretty ok, but we're definitely lacking sufficient documentation about the expected outputs. I know @CloseChoice and others are trying to improve that, e.g. on #3939, which will be v. helpful. We should prioritise documentating the shapes of the attributes and how to slice them.

In retrospect one mistake was not going through a careful deprecation cycle as we usually do, as tracked on #3507. 

> Also, I'm not sure why we are making changes to shap_values method since it is the legacy way to call explainers and the call API was supposed to be the new preferred way

On this point, I think it's preferable to have the output shapes consistent between the two methods. It would be quite confusing if the `shap_values` array had a different shape depending on how it was created.

There is lots of legacy code in shap, which makes it a nightmare to maintain in many ways. For example, as of a year ago half the plots expected numpy arrays, and the other half expected Explanation objects; and the main "summary plot" is an alias for a legacy function that has a newer duplicate implementation, is poorly tested, and is completely missing from the documentation.

So, if we want a hope of being able to fix bugs and even add new features that the community is requesting, I think we absolutely need to be able to deprecate old functionality to make the library a little more consistent and less duplicative. But I agree with your point, we nonetheless should still be extremely cautious about breaking changes.

> in retrospect adding a `output_shape: Literal["classfirst", "classlast"]` would have still be a breaking change but probably kept most of the downstream code working. Would it be an option to add these parameters, but still keep "classlast" as default, to avoid another breaking change?

Yep, agreed about the default. However - thinking it through a bit would it actually help downstream projects ? Supposing hyothetically shap added & released this new functionality, there would still be several versions without this option, so potentially quite a large range of versions would need to be excluded. Or, downstream packages could add a version constraint to only the latest `shap>=1.48` (or whatever the future release is). What would be best for you @imatiach-msft : would you be interested in having this option in a future version, or at this point would it be simplest just to migrate to the new output format?

imatiach-msft:
```Would it be an option to add these parameters, but still keep "classlast" as default, to avoid another breaking change?```
Not sure about others who might experience these but I am just trying to upgrade to latest now. My hope is that we won't have such breaking changes in the future or they can be more limited. My fix is to reshape the output so that even though shap has breaking changes our library won't, since we have many internal and external customers depending on this middle layer. My hope is that I can try to support both shap before and after these changes but keep our API as is so all of the downstream code that consumes our library won't need to be changed. I would not recommend to add this classlast parameter at this point since it would just complicate the shap source code more and possibly add more issues.

```Or, downstream packages could add a version constraint to only the latest shap>=1.48 (or whatever the future release is).```
You always want to support as much as possible, so as many users as possible can use the package across their various different python versions and dependencies. However, at some point the spaghetti code becomes too much to handle and this is the only remaining choice left.
The other big problem is that we serialize and deserialize explanations (upload and download explanations and explainers) in various different scenarios, so on one environment you might have a newer version of shap while in another an older version. Just using the version of the shap package can cause issues in these situations, the code has to be written carefully with this in mind.

The main problem for our team is that shap is used everywhere by a ton of customers, 1P and 3P, and with this breaking change there are a ton of unknowns. In addition, I am supposed to be working on new genai projects now. Just trying to support the legacy software that is still used by many customers and then this sort of change just turns the legacy code maintenance into a big project.

Issue #2727, opened 2022-10-14
Support for Adaboost Regressor


M-Schreurs:
Similar to the issue solved in #335, I was wondering if there is some way to implement Adaboost Regressor in Shap values.
I tried the fix for the Classifier, but the force plot and summary don't make a lot of sense (all my SHAP values are tiny, < 10^-16).

M-Schreurs:
I managed to get it to work myself by looking at the other regressors and the solution in #335 ([posted here on StackExchange](https://stackoverflow.com/questions/60433389/how-to-calculate-shap-values-for-adaboost-model)).

`Option 1: scaling = 1.0 . len(model.estimators_)`

        elif safe_isinstance(model, ("sklearn.ensemble.AdaBoostClassifier", "sklearn.ensemble._weighted_boosting.AdaBoostClassifier", "imblearn.ensemble.RUSBoostClassifier", "imblearn.ensemble._weight_boosting.RUSBoostClassifier")):
            assert hasattr(model, "estimators_"), "Model has no `estimators_`! Have you called `model.fit`?"
            self.internal_dtype = model.estimators_[0].tree_.value.dtype.type
            self.input_dtype = np.float32
            scaling = 1.0 / len(model.estimators_) # output is average of trees
            self.trees = [SingleTree(e.tree_, normalize=True, scaling=scaling, data=data, data_missing=data_missing) for e in model.estimators_]
            self.objective = objective_name_map.get(model.base_estimator_.criterion, None) #This line is done to get the decision criteria, for example gini.
            self.tree_output = "probability"
        elif safe_isinstance(model, ("sklearn.ensemble._weighted_boosting.AdaBoostRegressor", "sklearn.ensemble.AdaBoostRegressor")):
            assert hasattr(model, "estimators_"), "Model has no `estimators_`! Have you called `model.fit`?"
            self.internal_dtype = model.estimators_[0].tree_.value.dtype.type
            self.input_dtype = np.float32
            scaling = 1.0 / len(model.estimators_) # output is average of trees
            self.trees = [SingleTree(e.tree_, scaling=scaling, data=data, data_missing=data_missing) for e in model.estimators_]
            self.objective = objective_name_map.get(model.base_estimator_.criterion, None) #This line is done to get the decision criteria, for example gini.
            self.tree_output = "raw_value"


` Option 2: scaling = model.estimator_weights_/sum(model.estimator_weights_) `


[...]

PR #3319, opened 2023-10-08
feat: add support for AdaBoostClassifier

Helias:
## Overview

related of https://github.com/shap/shap/issues/335
after this PR we could close https://github.com/shap/shap/pull/1219, https://github.com/shap/shap/pull/1546  

add support for AdaBoostClassifier using this code https://github.com/shap/shap/pull/1546#issuecomment-1209576650
It worked locally so I think that we can add it to the `shap` repository.

I tested it running `pip install git+https://github.com/helias/shap.git@support-adaboost` then I used shap with my AdaBoostClassfiier model.

Co-authored-by: tk27182
Co-authored-by: ArkanEmre
Co-authored-by: diarmaidfinnerty

@tk27182 @ArkanEmre @diarmaidfinnerty

## Checklist

I tried to run `pytest` but I get this error:

```
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: --mpl
  inifile: /Users/.../projects/PhD/shap/pyproject.toml
  rootdir: /Users/.../projects/PhD/shap
```

so I removed that parameter in the pyproject file to make it run.

P.S. I installed also matplotlib and nose manually, but I would like to suggest adding a `requirements_dev.txt` file as a reference for the tests dependencies.


thatlittleboy:
Hi @Helias , thanks for the PR! We'll get round to reviewing the PR, just a quick response to your question about the dependencies.

You may read our CONTRIBUTING guide here: https://github.com/shap/shap/blob/master/CONTRIBUTING.md#installing-from-source

`--mpl` comes from the `pytest-mpl` package, which should be installed if you install the `shap[test]` optional dependency.

Helias:
> Hi @Helias , thanks for the PR! We'll get round to reviewing the PR, just a quick response to your question about the dependencies.
> 
> You may read our CONTRIBUTING guide here: https://github.com/shap/shap/blob/master/CONTRIBUTING.md#installing-from-source
> 
> `--mpl` comes from the `pytest-mpl` package, which should be installed if you install the `shap[test]` optional dependency.

I read very quickly CONTRIBUTING.md, thanks for the tip.

Let me know if I have to update my PR, I hope to add the [AdaBoostRegressor](https://github.com/shap/shap/issues/2727)  when this PR will be merged

ArkanEmre:
Hi @Helias,

it looks like you used the test I provided, which stated that it passes for an ensemble with a single tree. That's why the test has an AdaBoost with one tree, which is basically not an ensemble but a single decision tree. Hence the test passes. Once the AdaBoost ensemble has more than one tree, the test will surely fail. Have you tried a test with multiple trees as well?

Helias:
Hi, unfortunately I did not try other tests than your unit test.
For what I use it, the code seems working, it could help more users so it should be included in the repo, if it does not work for multiple trees, we could add a warning or error for it.

ArkanEmre:
@Helias, if you're using an ensemble with multiple trees, the sum of the SHAP values + expected value will not equal the output of the ensemble. The results are most likely invalid. This is basically what the test is checking and it is an essential for the SHAP algorithm. Could you check this with your use case please?

Helias:
Sorry for the late reply, but from my assumptions the output of this feature seems quite correct.

I am using the classifier to distinguish audio synthetic between fake and real in three datasets SOS, Fake-or-Real and In-the-Wild. In all three datasets the `energy` feature seems a meaningful one for the classification task.

image

Indeed, training the classifier on multiple features extracted from SOS shows that the main feature used to classify is the `energy`, if I remove the feature energy it will use other features.

image

ArkanEmre:
@Helias, How many trees are you using?

Helias:
I am using the default parameters, so I am using DecisionTreeClassifier initialized with max_depth=1 as estimator.

PR #3322, opened 2023-10-09
MAINT: Simplify pre-commit hook specification

connortann:
## Overview

Simplifies the configuration of pre-commit hooks, now that all notebooks can be linted.

Previously we used regular expressions to only lint specific notebooks. Now though, we have all notebooks passing Ruff and Black! 🎉

So, we can remove the regular expressions and just run Ruff a single time on both `.py` and `.ipynb` files.


thatlittleboy (approved):
cat-wearing-shades-letsgo

PR #3325, opened 2023-10-09
Fix CatboostClassifier explanations with feature interactions on windows

### LINKED ISSUE #3324: BUG: CatboostClassifier explanations with feature interactions fail on Windows
### Issue Description

This a pretty strange bug that I encountered [here](https://github.com/shap/shap/actions/runs/6458284627/job/17531689364?pr=3318#step:6:116) that only appears under Windows.

```python
import catboost

X, y = shap.datasets.adult(n_points=50)
model = catboost.CatBoostClassifier(depth=1, iterations=10).fit(X, y)
predicted = model.predict(X, prediction_type="RawFormulaVal")

ex_cat = shap.TreeExplainer(model)

# catboost explanations
explanation = ex_cat(X, interactions=True)
```

### Traceback

```shell
def _validate_inputs(self, X, y, tree_limit, check_additivity):
        # see if we have a default tree_limit in place.
        if tree_limit is None:
            tree_limit = -1 if self.model.tree_limit is None else self.model.tree_limit

        if tree_limit < 0 or tree_limit > self.model.values.shape[0]:
>           tree_limit = self.model.values.shape[0]
E           AttributeError: 'TreeEnsemble' object has no attribute 'values'
```

### Expected Behavior

Should result in the same as under linux.

The problem here is with the tempfile creation

### Bug report checklist

### Installed Versions

shap master

CloseChoice:
## Overview

Closes #3324  

- handle delete manually so that catboost interactions can work for windows

## Checklist

### DIFF
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -6,6 +6,7 @@ import warnings
 import numpy as np
 import pandas as pd
 import scipy.special
+import os
 from packaging import version
 
 from .. import maskers
@@ -1837,10 +1838,11 @@ class CatBoostTreeModelLoader:
         # cb_model.save_model("cb_model.json", format="json")
         # self.loaded_cb_model = json.load(open("cb_model.json", "r"))
         import tempfile
-        tmp_file = tempfile.NamedTemporaryFile()
+        tmp_file = tempfile.NamedTemporaryFile(delete=False)
         cb_model.save_model(tmp_file.name, format="json")
         self.loaded_cb_model = json.load(open(tmp_file.name))
         tmp_file.close()
+        os.unlink(tmp_file.name)
 
         # load the CatBoost oblivious trees specific parameters
         self.num_trees = len(self.loaded_cb_model['oblivious_trees'])
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -1,4 +1,5 @@
 import json
+import os
 import struct
 import time
 import warnings
@@ -6,7 +7,6 @@ import warnings
 import numpy as np
 import pandas as pd
 import scipy.special
-import os
 from packaging import version
 
 from .. import maskers
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -1835,14 +1835,11 @@ class XGBTreeModelLoader:
 
 class CatBoostTreeModelLoader:
     def __init__(self, cb_model):
-        # cb_model.save_model("cb_model.json", format="json")
-        # self.loaded_cb_model = json.load(open("cb_model.json", "r"))
         import tempfile
-        tmp_file = tempfile.NamedTemporaryFile(delete=False)
-        cb_model.save_model(tmp_file.name, format="json")
-        self.loaded_cb_model = json.load(open(tmp_file.name))
-        tmp_file.close()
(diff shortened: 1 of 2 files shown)


CloseChoice:
Btw. it seems like we just want to get the model parameters as a dictionary here. Isn't there a better way then to writing and loading the whole model to do so? Maybe this could be resolved in a seperate issue

connortann:
Thanks for the bug report and PR. I might be missing something, but currently I don't see how this change relates to the bug report.

What leads you to believe the cause of the test failure you found is with tempfile deletion specifically? Moreover, do we know what specifically is going wrong with the current implementation?

The tempfile is a python builtin, so I'd expect it to be very robust and well tested. Do we have reason to believe that `NamedTemporaryFile` is not behaving as it should according to the docs, by deleting the file after the object is closed?

CloseChoice:
Sorry, the connection is not obvious at all. So the error happens [here](https://github.com/shap/shap/blob/master/shap/explainers/_tree.py#L1094), under Windows there happens a `Permission Denied` error. And this error has its cause in the tempfile implementation. This error leads to `self.trees` being `None`, which leads to `self.values` to not be initialized, s. [here](https://github.com/shap/shap/blob/master/shap/explainers/_tree.py#L1158-L1176). 

Unfortunately it is not platform independent, from the [documentation](https://docs.python.org/3.9/library/tempfile.html#tempfile.NamedTemporaryFile):

>  Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows). 

[Here](https://stackoverflow.com/questions/23212435/permission-denied-to-write-to-my-temporary-file/54768241#54768241) is the stackoverflow discussion around Tempfile but it seems like its implementation has a couple of implementation issues (e.g. [here](https://stackoverflow.com/questions/23212435/permission-denied-to-write-to-my-temporary-file/58955530#58955530)).

connortann:
Thanks for the explanation, that makes sense.

As per the docs as you say, there might still be an issue is with re-opening a temp file by name, as is done here: `json.load(open(tmp_file.name))`.

So instead, may I suggest using a temporary directory? I think that would solve any permission errors. We can also use it as a context manager to ensure it stays open as long as we need it, and that it's deleted cleanly afterwards. Something like:

```python
with tempfile.TemporaryDirectory() as tmp_dir:
    tmp_file = os.path.join(tmp_dir, "model")
    cb_model.save_model(tmp_file, format="json")
    self.loaded_cb_model = json.load(open(tmp_file))
```

CloseChoice:
> Thanks for the explanation, that makes sense.
> 
> As per the docs as you say, there might still be an issue is with re-opening a temp file by name, as is done here: `json.load(open(tmp_file.name))`.
> 
> So instead, may I suggest using a temporary directory? I think that would solve any permission errors. We can also use it as a context manager to ensure it stays open as long as we need it, and that it's deleted cleanly afterwards. Something like:
> 
> ```python
> with tempfile.TemporaryDirectory() as tmp_dir:
>     tmp_file = os.path.join(tmp_dir, "model")
>     cb_model.save_model(tmp_file, format="json")
>     self.loaded_cb_model = json.load(open(tmp_file))
> ```

This looks good, I've update the PR and the tests run locally

connortann (approved):
LGTM!

Issue #882, opened 2019-11-05
Possible problem with using conditional vs marginal expectation for dropped features in Tree SHAP?


ashermullokandov:
Hi Scott,

First, thanks for this great package!

I'd like to mention a recent paper which describes a conceptual critique of an aspect of the SHAP package, and to get your opinion on the matter. In [Feature relevance quantification in explainable AI: A causality problem](https://arxiv.org/pdf/1910.13413v1.pdf) it is claimed that in attributing the difference between f(x) and E[f(x)] to individual features, using the conditional expectation for the feature dropping is misleading. The critique seems to rest on a causal argument, described in section 3 of the paper. The paper further claims that using a 'marginal expectation', without conditioning, produces a more reliable attribution, demonstrating this with a very simple example and some numerics.

Thanks again!

slundberg:
Thanks for asking! That paper [Feature relevance quantification in explainable AI: A causality problem](https://arxiv.org/pdf/1910.13413v1.pdf) provides a very nice way of justifying why it is a good idea to perturb sets of features independently from the other features we are not perturbing (which is what the SHAP package already does when possible). The argument is that we want to know what would happen if we set a feature's value to something (written as E[f(X) | do(X = x_s)]), not just that we observe a feature's value to be something (written as E[f(X) | do(X = x_s)]).

To help clarify that SHAP aligns with casual interventional perturbations (as described in that paper) I have actually renamed the `feature_dependence="independent"` option to `feature_perturbation="interventional"`. Ironically I had used the "interventional" wording a while ago for the same reason this paper describes (#269), but then dropped it because I feared it might confuse people. Now that there is a good paper to point to for details I think it is worth using it again.

It is important to note that one context where using `feature_perturbation="interventional"` is not possible is when we are using `TreeExplainer(model)` without giving a background dataset. When you don't give a background dataset it is not possible to perturb features entirely causally (since you only know how the training samples traversed the tree, not the feature values of those samples). In this case SHAP defaults (now with a warning) to `feature_perturbation="tree_path_dependent"`. This option is not exactly a causal explanation (as defined in that paper), but it is the best you can do without a background dataset, and importantly it does not suffer from the problem of putting weight on feature not explicitly used by the model.

[...]

PR #3340, opened 2023-10-15
Remove deprecated option checks

thatlittleboy:
## Overview

<s>Closes #XXXX</s>

This PR is proposing to remove the warnings for using the old parameter names for TreeExplainer and LinearExplainer. The renaming of parameters was done 4 years ago now, see #882. This should have been sufficient time for everyone to shift over to the new option names.

 e.g.

```
feature_dependence -> feature_perturbation
independent -> interventional
correlation -> correlation_dependent
logloss -> log_loss
```

I propose we remove the old code for doing these checks to simplify our code base slightly and make maintenance easier on ourselves.

## Checklist


connortann:
This looks like a sensible initiative, nice one. I've added the "Breaking" label to ensure we highlight this in the release notes.

One thing caught my eye:

> feature_perturbation option is now deprecated in favor of using the appropriate masker

Should we update the docstring of LinearExplainer to remove the deprecated option from the arguments, or at least mention that it is deprecated?

thatlittleboy:
> Should we update the docstring of LinearExplainer to remove the deprecated option from the arguments, or at least mention that it is deprecated?

Sure, I'll take a look at this later tonight. Need some time to better understand the rationale for its deprecation

tests/explainers/test_linear.py
```diff
@@ -185,6 +186,17 @@ def test_sparse():
     assert np.max(np.abs(scipy.special.expit(explainer.expected_value + shap_values.sum(1)) - model.predict_proba(X)[:, 1])) < 1e-6
 
 
+def test_invalid_feature_perturbation_raises():
+    Ridge = pytest.importorskip("sklearn.linear_model").Ridge
```
connortann: `sklearn` is one of the package's core dependencies, so `Ridge` can just be imported in the usual way at the top of the module

thatlittleboy: done

shap/explainers/_gpu_tree.py
```diff
@@ -24,34 +24,37 @@ class GPUTreeExplainer(TreeExplainer):
...
+        If ``model_output`` is the name of a supported prediction method on the ``model`` object, then we
+        explain the output of that model method name. For example, ``model_output="predict_proba"``
+        explains the result of calling ``model.predict_proba``.
+        If "probability", then we explain the output of the model transformed into probability space
+        (note that this means the SHAP values now sum to the probability output of the model).
+        If "log_loss", then we explain the log base e of the model loss function, so that the SHAP values
+        sum up to the log loss of the model for each sample. This is helpful for breaking down model
+        performance by feature.
+        Currently the "probability" and "log_loss" options are only supported when
+        ``feature_perturbation="interventional"``.
```
connortann: Nit: it could be helpful to format these in bullet points or similar, so each option is on its own line in the sphinx docs. At the moment, it seems to render as a single paragraph:

image

connortann: Same comment for `Tree` as for `GPUTree`

connortann (approved):
LGTM with suggestions above

shap/explainers/_gpu_tree.py
connortann: I pushed a commit to fix this

connortann:
@CloseChoice I took up this PR as it would be great to get this merged. @thatlittleboy did most of the work, and I pushed some further commits to tidy up the docstrings.

I was quite surprised to see that some notebooks fail. Here's the traceback of one example:

<details>

```
FAILED: notebooks/tabular_examples/tree_based_models/Front page example (XGBoost).ipynb with error: An error occurred while executing the following cell:
------------------
import xgboost

import shap

# train an XGBoost model
X, y = shap.datasets.california()
model = xgboost.XGBRegressor().fit(X, y)

# explain the model's predictions using SHAP
# (same syntax works for LightGBM, CatBoost, scikit-learn, transformers, Spark, etc.)
explainer = shap.Explainer(model)
shap_values = explainer(X)

# visualize the first prediction's explanation
shap.plots.waterfall(shap_values[0])
------------------


TypeError                                 Traceback (most recent call last)
Cell In[1], line 11
      7 model = xgboost.XGBRegressor().fit(X, y)
      9 # explain the model's predictions using SHAP
     10 # (same syntax works for LightGBM, CatBoost, scikit-learn, transformers, Spark, etc.)
---> 11 explainer = shap.Explainer(model)
     12 shap_values = explainer(X)
     14 # visualize the first prediction's explanation

File /opt/hostedtoolcache/Python/3.11.8/x64/lib/python3.11/site-packages/shap/explainers/_explainer.py:188, in Explainer.__init__(self, model, masker, link, algorithm, output_names, feature_names, linearize_link, seed, **kwargs)
    186 elif algorithm == "tree":
    187     self.__class__ = explainers.TreeExplainer
--> 188     explainers.TreeExplainer.__init__(self,self.model,self.masker,link=self.link,feature_names=self.feature_names,linearize_link=linearize_link,**kwargs)
    189 elif algorithm == "additive":
    190     self.__class__ = explainers.AdditiveExplainer

TypeError: TreeExplainer.__init__() got an unexpected keyword argument 'link'
```

</details>

It looks like the `shap.Explainer(model)` interface failed when using a TreeExplainer, as the `link` argument is not used. What's surprising to me is that this wasn't caught in the unit tests, give that this is pretty core to the main API. I added a quick unit test for this particular case, and now the unit tests fail (as expected). This issue has made me realise that our coverage is still quite patchy!

connortann:
So, it looks like the current implementation of TreeExplainer entirely ignores the `link` and `linearize_link` arguments which are passed from `Explainer` here. Created a new issue: #3513

As a short-term suggest, I've allowed `**kwargs` in TreeExplainer which means the "link" argument remains allowed but ignored. This doesn't seem particularly elegant, so any thoughts are most welcome. @CloseChoice , what do you think?

shap/explainers/_tree.py
```diff
@@ -68,53 +73,68 @@ def __init__(
         feature_perturbation="interventional",
         feature_names=None,
         approximate=False,
-        **deprecated_options,
+        # FIXME: Do not allow unexpected arguments.
+        # `link` and `linearize_link` may be passed in by Explainer, but are unused.
+        **kwargs,
```
connortann: I added this `**kwargs` so that the Explainer interface will still pass. It seems like an ugly hack, but it should at least fix the failing test for now. I suppose it's a separate issue that the main thrust of this PR.

CloseChoice: Why don't we specify link and linearize_link explicitly (and both should default to `None` so that users cannot hand over (other) arguments without effect. But this is also fine for me for now

shap/explainers/_tree.py
```diff
@@ -68,53 +73,68 @@ def __init__(
...
     ):
         """ Build a new Tree explainer for the passed model.
 
         Parameters
         ----------
         model : model object
-            The tree based machine learning model that we want to explain. XGBoost, LightGBM, CatBoost, Pyspark
-            and most tree-based scikit-learn models are supported.
+            The tree based machine learning model that we want to explain.
```
CloseChoice, line 106: Did you somehow limit the comment width programmatically, e.g. through a formatting tool?

shap/explainers/_tree.py
```diff
@@ -132,28 +152,6 @@ def __init__(
         if getattr(self.masker, "clustering", None) is not None:
             raise ExplainerError("TreeExplainer does not support clustered data inputs! Please use shap.Explainer or pass an unclustered masker!")
 
-        # check for deprecated options
```
CloseChoice, line 156: don't we need this part of the code? At least the model_output parameter?

tests/explainers/test_linear.py
```diff
@@ -116,7 +112,7 @@ def test_perfect_colinear():
     X.iloc[:, 3] = 0 # test null features
     model = LinearRegression()
     model.fit(X, y)
-    explainer = shap.LinearExplainer(model, X, feature_dependence="correlation")
+    explainer = shap.LinearExplainer(model, X, feature_perturbation="correlation_dependent")
```
CloseChoice, line 115: so is this a breaking change? Seems like you just aligned the code to the docs but this might break people's code, doesn't it?

CloseChoice (commented):
Just a couple of questions that I have here.

tests/explainers/test_linear.py
connortann: Yep, this is a breaking change.

The code previously raised a warning when the old deprecated parameter names were used; as per the PR description this was over 4 years ago, so now seems like a fair time to go ahead and drop support for them.

CloseChoice: Sorry, did not realize the breaking label was there.

shap/explainers/_tree.py
connortann: Yes, I used the [rewrap](https://[marketplace.visualstudio.com/items?itemName=stkb.rewrap](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap)) VS Code extension to semi-automatically rewrap a few sections. I find wrapping can help long paragraphs be readable both as source code and in the rendered HTML.

shap/explainers/_tree.py
connortann: Yes, that seems like a better approach: I'll add those in. In any case, after this PR I think we should try to fix the root issue of #3513

shap/explainers/_tree.py
connortann: Which part is still needed? As I understand, the deleted lines just handled deprecated aliases for the function arguments. The `model_output` parameter is still defined as a keyword argument to the function, so that variable still exists

CloseChoice: You're right, looks fine to remove these lines.

CloseChoice (approved):
LGTM. Let's merge this one and wait for #3340 so that we can have one breaking changes release.

connortann:
> LGTM. Let's merge this one and wait for #3340 so that we can have one breaking changes release.

@CloseChoice I'm afraid I don't quite follow what you mean - #3340 is this PR, is there another PR that we should include in this release's milestone?

A bit off topic, but I was also wondering if we still wish to keep the "visualisation"-related items in the [milestone](https://github.com/shap/shap/milestone/7) or if it would be better to release sooner. I'm personally erring on the side of releasing little and often, so I'm tempted to bump the viz items to `0.46.0`

CloseChoice:
> > LGTM. Let's merge this one and wait for #3340 so that we can have one breaking changes release.
> 
> @CloseChoice I'm afraid I don't quite follow what you mean - #3340 is this PR, is there another PR that we should include in this release's milestone?
> 
> A bit off topic, but I was also wondering if we still wish to keep the "visualisation"-related items in the [milestone](https://github.com/shap/shap/milestone/7) or if it would be better to release sooner. I'm personally erring on the side of releasing little and often, so I'm tempted to bump the viz items to `0.46.0`

Yes, there is #3318 which I would also like to include in 0.45.0. Will spend some time on this PR this weekend.
Let's push the visualization further. I think we do not even have a concept how to tackle #3411.

PR #3341, opened 2023-10-15
Remove unnecessary `safe_isinstance` calls

thatlittleboy:
## Overview

<s>Closes #XXXX</s>  

Just a simple refactoring PR. There should not be any noticeable effect on the user-facing API.

* Prefer usage of the builtin `isinstance` instead of `safe_isinstance` for pandas and numpy, since they are core dependencies of shap (they are already imported at top-level).

## Checklist

### DIFF
--- a/shap/benchmark/_explanation_error.py
+++ b/shap/benchmark/_explanation_error.py
@@ -5,7 +5,7 @@ from tqdm.auto import tqdm
 
 from shap import Explanation, links
 from shap.maskers import FixedComposite, Image, Text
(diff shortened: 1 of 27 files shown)


connortann:
Looks good! I'll find some time to properly review this one later in the week, as the diff is fairly long. I'm surprised we have so many isinstance checks.

I am pondering why the pandas check was previously done with the strange `str(type(df)).endswith` method. It's so unusual, I wonder if there might have been a good reason for it? I can't think of a reason why though.

connortann (approved):
LGTM, this is a nice clean-up.

Issue #1864, opened 2021-03-08
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 337: invalid start byte


Allenhuang0708:
Hi, my envs as follow:
xgboost                    1.3.3
shap                       0.38.1
python  3.7.7
When I am running my program, there came a issue when decoding, I printed out my model.get_booster().save_raw(), it's like this:
`bytearray(b'binfn\xac\xedA\x08\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x03\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x10\x00\x00\x00\x00\x00\x00\x00reg:squarederror\x06\x00\x00\x00\x00\x00\x00\x00gbtree\x0f\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00...') (I deleted the last characters)`

I noticed that we have already lstrip 'binf' but i guess the problem came out of the 'n\xac\xedA' part, cause when I printed out the base score etc, the number is clearly wrong:
`base 5.835820683183942e-34 num features 0 num class 16777216 extra attrs 0 eval metrics 16777216`

The orginial base score by the model is: 29.709193548387095

I  am wondering why is this happend? The xgboost saved the model in an unregular way?

By the way, we are using xgboost regession model

Allenhuang0708:
A related decoding issue is on #1215
[...]

Issue #1302, opened 2020-07-08
Xgboost binary format compatibility


RAMitchell:
In #1215, changes to the xgboost binary format break integration with the shap library and lead users to downgrade xgboost or create other hack solutions.

Xgboost makes no promises about the consistency of its binary format, except that it may save and load its own models, so this is likely to occur again in future xgboost releases.

I propose that shap migrates to the newer json based serialisation format that has better compatibility guarantees: https://xgboost.readthedocs.io/en/latest/tutorials/saving_model.html (cc the author of this feature @trivialfis)

After migration I propose adding shap tests to xgboost CI in order to prevent this from our end.

YannDubs:
@RAMitchell  I think this would be great. Currently, there are some new features that are not usable due to using the old binary format, for example categorical features (see #2662 ).

YannDubs:
This seems even more important given that Xgboost plans to drop support from old binary format (see: [here](https://github.com/dmlc/xgboost/7547) and [here](https://github.com/dmlc/xgboost/blob/ee382c4153a675b0ecd21cc940e55ae8ae509a25/NEWS.md#new-serialization-format) ).

It seems that SHAP used JSON [before](https://github.com/slundberg/shap/commit/397ed9278fc81c858d3188171dda7591531ccc0d) but it was reverted by @slundberg  [here](https://github.com/slundberg/shap/commit/397ed9278fc81c858d3188171dda7591531ccc0d) due to conversion issues. Maybe the newer UBJSON format of Xgboost can help here?

CloseChoice:
Resolved via #3345

PR #3345, opened 2023-10-16
Update XGBoost parsing to use ubjson format, replacing deprecated binary format

### LINKED ISSUE #3321: BUG: XGboost binary format is being deprecated
### Issue Description

Based on the roadmap here https://github.com/dmlc/xgboost/issues/7547,
the binary format for storing xgboost models is going to be deprecated in xgboost v2.2, when saving to raw format is no longer supported.

We currently rely on this in our `XGBTreeModelLoader` to load in xgboost models to our TreeExplainer.

Warnings are already starting to show up in our test logs,

image

so I think we need to prioritize this over the next couple of months. It's not clear when v2.2 will be released (xgb is on v2.0 as of time of writing the issue), if it does release before this issue is resolved, then we'll need to set an upper bound on the xgboost version we support.

(If anyone is interested in tackling this, feel free to holler, so we can assign the issue appropriately.)
The change required doesn't look trivial, though.

```python
N.A.
```

### Traceback

```shell
N.A.
```

### Expected Behavior

N.A.

### Bug report checklist

### Installed Versions

master

CloseChoice:
## Overview

Closes #3321  

- add a parser for UBJSON that is capable to parse xgboost dumps. Note that this is not a generic parser for ubjson.
- add tests to check that ubjson is correctly parsed and xgboost still works correctly

Limitations of the parser:
 - overall container must be an object, not an array
 - unoptimized arrays and optimized objects with types & count is not supported, though this is not much work to implement
 - no support for high-precision numbers
 - does not support arrays in arrays (objects in arrays is implemented)

Todo:
 - [x] add docstring for ubjson
 - [x] decide whether the binary format should still be used for versions lower than `1.6.0`
 - [x] remove error throwing lines from `print_info`
 - [x] What do we do with `max_depth`? -> Just set to 0, which is the same behaviour as we had before (at least in the example I checked)

## Checklist


thatlittleboy:
Thanks @CloseChoice for the PR.

Just curious, is there anything specific about your implementation here of the ubjson decoder for xgboost models or is it essentially equivalent (minus the caveats you mentioned in your PR description) to this? https://github.com/Iotic-Labs/py-ubjson

`xgboost` itself I think is using this library for saving and loading, so I wonder if it's easier to use the same as xgboost than reimplementing part of the ubj spec ourselves.

CloseChoice:
I tried py-ubjson and it works ok but cannot encode all arrays correctly and leaves them in a binary format. Furthermore the library seems pretty much unmaintened (last commit was Feb 1 2023) to me and has just 39 stars. 

Here is an example of the encoding of a very simple model:
```json
{
    "learner": {
        "attributes": {},
        "feature_names": [],
        "feature_types": [],
        "gradient_booster": {
            "model": {
                "gbtree_model_param": {
                    "num_parallel_tree": "1",
                    "num_trees": "2"
                },
                "iteration_indptr": [
                    0,
                    1,
                    2
                ],
                "tree_info": [
                    0,
                    0
                ],
                "trees": [
                    {
                        "base_weights": [
                            -0.0012426718603819609
                        ],
                        "categories": [],
                        "categories_nodes": [],
                        "categories_segments": [],
                        "categories_sizes": [],
                        "default_left": b"\x00",
                        "id": 0,
                        "left_children": [
                            -1
                        ],
                        "loss_changes": [
                            0.0
                        ],
                        "parents": [
                            2147483647
                        ],
                        "right_children": [
                            -1
                        ],
                        "split_conditions": [
                            -0.0003728015872184187
                        ],
                        "split_indices": [
                            0
                        ],
                        "split_type": b"\x00",
                        "sum_hessian": [
                            1.714766025543213
                        ],
                        "tree_param": {
                            "num_deleted": "0",
                            "num_feature": "3",
                            "num_nodes": "1",
                            "size_leaf_vector": "1"
                        }
                    },
                    {
                        "base_weights": [
                            -0.0010072523728013039
                        ],
                        "categories": [],
                        "categories_nodes": [],
                        "categories_segments": [],
                        "categories_sizes": [],
                        "default_left": b"\x00",
                        "id": 1,
                        "left_children": [
                            -1
                        ],
                        "loss_changes": [
                            0.0
                        ],
                        "parents": [
                            2147483647
                        ],
                        "right_children": [
                            -1
                        ],
                        "split_conditions": [
                            -0.00030217572930268943
                        ],
                        "split_indices": [
                            0
                        ],
                        "split_type": b"\x00",
                        "sum_hessian": [
                            1.7146751880645752
                        ],
                        "tree_param": {
                            "num_deleted": "0",
                            "num_feature": "3",
                            "num_nodes": "1",
                            "size_leaf_vector": "1"
                        }
                    }
                ]
            },
            "name": "gbtree"
        },
        "learner_model_param": {
            "base_score": "4.2905337E-1",
            "boost_from_average": "1",
            "num_class": "0",
            "num_feature": "3",
            "num_target": "1"
        },
        "objective": {
            "name": "binary:logistic",
            "reg_loss_param": {
                "scale_pos_weight": "1"
            }
        }
    },
    "version": [
        2,
        0,
        0
    ]
}
```
As you can see `default_left` and `split_type` are not encoded correctly. We do not need these fields but it did not build trust in the library.

As far as I have seen xgboost is directly encoding the stuff in C++ and just uses py-ubjson for tests, s [here](https://github.com/search?q=repo%3Admlc%2Fxgboost%20ubjson&type=code).

I also thought about using the library but decided not to because of the reasons mentionef above. Could still go back to py-ubjson if you think that's better.

EDIT: We do actually need the field `default_left`. So if we used py-ubjson one has to encode this seperately (or fix the bug in py-ubjson)

thatlittleboy:
Got it, thanks for the clarification. I'll need to take a deeper look at the PR first, will update here later

shap/explainers/_tree.py
```diff
@@ -1697,43 +1699,34 @@ class XGBTreeModelLoader:
...
-        self.name_gbm_len = self.read('Q')
-        self.name_gbm = self.read_str(self.name_gbm_len)
+        import xgboost
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            if version.parse(xgboost.__version__) >= version.parse("1.6.0"):
+                tmp_file = os.path.join(tmp_dir, "model.ubj")
+                xgb_model.save_model(tmp_file)
+                xgb_params = decode_ubjson_buffer(open(tmp_file, 'rb'))
+            else:
+                tmp_file = os.path.join(tmp_dir, "model.json")
```
CloseChoice: If the xgboost version is lower than 1.6 (which is the version ubjson was introduced), should we keep the old binary parser?

CloseChoice:
@thatlittleboy Did you already have time to look at this?

thatlittleboy:
@CloseChoice Thanks for the reminder, but not yet, sorry. I'm currently on vacation, I should have some bandwidth to look at this when I'm back in a couple of days.

shap/explainers/_tree.py
connortann: We generally try to follow [SPEC 0](https://scientific-python.org/specs/spec-0000/), which means supporting XBoost versions that were released in the past two years.

From a scan of the XGB releases:
- XGB 1.5.0 was released Oct 2021
- XGB 1.6.0 was released Apr 2022

So, at time of writing I think it would be appropriate for us to drop support for XGB 1.6, if that means this code can be simplified significantly. We should raise an informative error message if the XGB version is not supported.

shap/explainers/_tree.py
```diff
@@ -1715,43 +1717,34 @@ class XGBTreeModelLoader:
...
-        self.name_obj_len = self.read('Q')
-        self.name_obj = self.read_str(self.name_obj_len)
-        self.name_gbm_len = self.read('Q')
-        self.name_gbm = self.read_str(self.name_gbm_len)
+        import xgboost
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            if version.parse(xgboost.__version__) >= version.parse("1.6.0"):
+                tmp_file = os.path.join(tmp_dir, "model.ubj")
+                xgb_model.save_model(tmp_file)
+                xgb_params = decode_ubjson_buffer(open(tmp_file, 'rb'))
```
connortann: Suggest refactoring this section out as a separate static method, e.g.

```python
@staticmethod
def read_xgb_params(xgb_model) -> dict
```

Maybe `xgb_params["learner"]` should be returned, as I don't think anything else in the dict is used.

The docstring of this function could be a good place to document the rationale behind why the model is saved and loaded, the limitations of JSON, and the different binary formats. I don't think that context would be obvious to the reader.

shap/explainers/_tree.py
connortann: As I understand, at present the code uses the JSON representation for models below 1.6. What are the consequences of that compared to the binary format? Is it worth raising a warning about any limitations?

CloseChoice: Json is less precise than the binary format. So raising a warning about this is appropriate IMO

connortann: If you'd kindly go ahead an implement this warning, then apart from the other minor comment about refactoring I think this PR will be pretty much ready to go 👍

connortann:
@CloseChoice happy new year! It would be great to finish off this PR: would you be happy to tie up the loose ends in the comments above?

CloseChoice:
> @CloseChoice happy new year! It would be great to finish off this PR: would you be happy to tie up the loose ends in the comments above?

Thanks and happy new year. Will do so this week/weekend. Just a few things on my plate right now.

shap/explainers/_tree.py
CloseChoice: Implemented

shap/explainers/_tree.py
CloseChoice: Done.

CloseChoice:
@connortann Sorry to keep you waiting for that long to reiterate over this PR. I implemented your suggestions. Would be great if you could have a final look (especially over the wording for the warning). Thanks.

shap/explainers/_tree.py
```diff
@@ -1786,39 +1772,43 @@ def __init__(self, xgb_model):
...
+            else:
+                warnings.warn("You are using an XGBoost version below 1.6.0 which is not fully supported by shap. "
+                              "Shap falls back to encoding the model as JSON which can lead to numerical precision issues. "
+                              "Please consider upgrading to XGBoost 1.6.0 or higher.")
+                tmp_file = os.path.join(tmp_dir, "model.json")
+                xgb_model.save_model(tmp_file)
+                xgb_params = json.load(open(tmp_file))
```
connortann: Probably nothing, but I think it's safer to use a `with` statement here to ensure the file handle is closed properly before the temporary directory is deleted. I've seen some unexpected errors in the past on different OSs relating to file handle behaviour, so probably worth it to avoid surprises.

```suggestion
                with open(tmp_file) as fh:
                    xgb_params = json.load(fh)
```

connortann: @CloseChoice As this is a v minor change I'll go ahead and make this tweak so we can complete the PR :)

connortann (approved):
LGTM!

connortann:
@CloseChoice think we can now close these related issues?
- #1864
- #1302

There is also another recent PR which is relevant but has more going on: #3462

trivialfis:
Wow, I didn't know about this PR, otherwise, I wouldn't write https://github.com/shap/shap/pull/3462 , will try to rebase and cherry-pick the remaining fixes.

CloseChoice:
@trivialfis thanks for your PR, I am pretty sure we can get a lot of helpful things out of your PR. 

@connortann : I would close #1864 for sure. There is no reproducible example and we completely rewrote the loading mechanism. I think I just forgot to link #1302 here, that is fixed now.

trivialfis:
Thank you for the work on the decoder. I have revised my PR to handle other data.

PR #3352, opened 2023-10-19
BUG: Explicitly specify constraint of xgboost>=1.4

### LINKED ISSUE #3326: BUG: TypeError for `shap_values()` and `shap_interaction_values()` for xgboost<1.4
### Issue Description

Following #2987 (released in shap v0.42), the `shap_values()`and `shap_interaction_values()` for xgboost models rely on the `iteration_range` parameter for xgboost model `predict()` method introduced xgboost v1.4 (with the `ntree_limit` parameter deprecated in that version, and then removed in xgboost v2.0). Thus, there should be a lower bound on the version of xgboost supported by shap. I can see that xgboost is NOT a shap dependency, but still assumptions are made about the version of that package. Perhaps a similar snippet to the one starting here https://github.com/shap/shap/blob/master/shap/explainers/_tree.py#L187 should be introduced, so that the user does not have to resolve this 'dependency' issue themselves.

```python
NA
```

### Traceback

```shell
...
phi = self.model.original_model.predict(X, iteration_range=(0, xgb_tree_limit),
pred_interactions=True, validate_features=False)
E           TypeError: predict() got an unexpected keyword argument 'iteration_range'

venv/lib/python3.8/site-packages/shap/explainers/_tree.py:519: TypeError
```

### Expected Behavior

_No response_

### Bug report checklist

### Installed Versions

shap==0.43.0
xgboost==1.3.3

mtlulka:
## Overview

Closes #3326 

- Restricts the test dependency `xgboost>=1.4`
- Modifies the minimal version `xgboost>=1.4` for `TreeExplainer`

## Checklist


shap/explainers/_tree.py
```diff
@@ -185,11 +185,12 @@ def __init__(
...
-        # A bug in XGBoost fixed in v0.81 makes XGBClassifier fail to give margin outputs
-        if safe_isinstance(model, "xgboost.sklearn.XGBClassifier") and self.model.model_output != "raw":
+        # A change in the signature of `xgboost.Booster.predict()` method has been introduced in XGBoost v1.4:
+        # The introduced `iteration_range` parameter is used when obtaining SHAP (incl. interaction) values from XGBoost models.
+        if self.model.model_type == 'xgboost':
             import xgboost
-            if version.parse(xgboost.__version__) < version.parse('0.81'):
-                raise RuntimeError("A bug in XGBoost fixed in v0.81 makes XGBClassifier fail to give margin outputs! Please upgrade to XGBoost >= v0.81!")
+            if version.parse(xgboost.__version__) < version.parse('1.4'):
+                raise RuntimeError("The parameters of a method used to obtain SHAP values have changed in XGBoost v1.4! Please upgrade to XGBoost >= v1.4!")
```
connortann: I think we can simplify the error message a little, as our users probably don't need to bother with the reasoning behind a dependency constraint. How about:

```suggestion
                raise RuntimeError(f"SHAP requires XGBoost >= v1.4 , but found version {xgboost.__version__}. Please upgrade XGBoost!")
```

pyproject.toml
```diff
@@ -58,7 +58,7 @@ test = [
   "pytest",
   "pytest-mpl",
   "pytest-cov",
-  "xgboost",
+  "xgboost>=1.4",
```
connortann: I suggest removing this line:
- I don't think this pin will have any effect, as the test suite on CI currently lets pip resolve the latest versions of all dependencies.
- Moreover, it may give other contributors the false impression that this config option is used to specify constraints on optional dependencies.

connortann:
I think this PR looks good. I agree with your proposal to enforce the constraint at runtime, give that XGBoost is not a dependency of the main package.

However, I'd like to get a second opinion from @thatlittleboy on XGBoost support:

1. On one hand, we could consider handling both the new `iteration_range` and the deprecated `ntree_limit` parameters in XGBoost, to maintain compatibility with older XGBoost versions at the expense of more complexity.
2. On the other hand, XGBoost `1.4` is now over two years old which is the time at which [SPEC 0](https://scientific-python.org/specs/spec-0000/) recommends dropping support for core package dependencies.

IMO, supporting only 1.4+ seems like the right way forward.

connortann (dismissed):
LGTM, with some suggestions above.

thatlittleboy:
> However, I'd like to get a second opinion from @thatlittleboy on XGBoost support:
> 
>     On one hand, we could consider handling both the new iteration_range and the deprecated ntree_limit parameters in XGBoost, to maintain compatibility with older XGBoost versions at the expense of more complexity.
>     On the other hand, XGBoost 1.4 is now over two years old which is the time at which [SPEC 0](https://scientific-python.org/specs/spec-0000/) recommends dropping support for core package dependencies.
> 
> IMO, supporting only 1.4+ seems like the right way forward.


@connortann I'm with you that we should support 1.4+ only. The added complexity doesn't make sense for me at this stage. If there is sufficient interest from the community (in the form of issues and upvotes) to do so, we could consider making a patch release.

pyproject.toml
mtlulka: Reverted the change. I thought it would be a good idea as anyone running the tests will require this constraint to be satisfied for the tests to pass. However, they will still be able to identify the issue from the error message should it arise.

mtlulka:
Thanks @connortann for your review, and @thatlittleboy for your input. I've applied the suggestions. This tiny PR is ready for your review once again.

connortann (approved):
LGTM!

PR #3355, opened 2023-10-20
ENH: Enable `ax` passing to `group_difference()`

### LINKED ISSUE #3354: ENH: Enable using `group_difference()` plots in subplots
### Problem Description

I would like to plot multiple `group_difference()` plots within as a single `matplotlib.figure.Figure` plot using `matplotlib.pyplot.subplots()` routine. This is currently not possible as each call to `group_difference()` creates a new `matplotlib.figure.Figure`:
https://github.com/shap/shap/blob/v0.43.0/shap/plots/_group_difference.py#L57

### Alternative Solutions

Without this change, I will be forced to concatenate these `matplotlib.figure.Figure` outside of `matplotlib` (using PIL?).

### Additional Context

This is part of a bigger issue of SHAP plots not handling a provision of `ax` objects. Described in #2822 and related comments.

### Feature request checklist

mtlulka:
## Overview

Closes #3354

- Modifies `group_difference` so that an `ax` (`matplotlib.axes.Axes`) object can be passed: backward compatible

## Checklist


mtlulka:
I've checked and there is currently no test for the module/function that I'm submitting changes to. At this point, I am not planning to add a test for this plotting functionality but I am open to suggestions on the scope of it should you want me to create one. I've run the 'Explaining quantitative measures of fairness.ipynb' notebook with no issues.

thatlittleboy:
Hi @mtlulka , thanks for the contribution. The changes look good to me in principle, I would request for a test here to ensure everything is running as expected.

Scope-wise, it should be sufficient to just create a test that plots the group difference plot with providing axes, in a `test_group_difference.py` file.
And use [`pytest-mpl`](https://github.com/matplotlib/pytest-mpl) to generate & save the output into `tests/plots/baseline/` directory.

We have several examples of how to do this in the `tests/plots/` directory, feel free to holler if you encounter any issues.

mtlulka:
@thatlittleboy Thanks for you guidance. I have added a test in line with your specification.

shap/plots/_group_difference.py
```diff
@@ -52,36 +52,34 @@ def group_difference(shap_values, group_mask, feature_names=None, xlabel=None, x
 
     if max_display is not None:
         inds = inds[:max_display]
-    # draw the figure
-    figsize = [6.4, 0.2 + 0.9 * len(inds)]
-    pl.figure(figsize=figsize)
+    if ax:
+        # Disable plotting out if an ax has been provided
+        show = False
```
thatlittleboy, line 57: I _think_ I see your reasoning for this, but for the sake of record, do you mind explaining why we choose to explicitly set `show` to False and override the user's argument?

mtlulka: In short, my thinking is that if `ax` is provided, then the corresponding `fig` has already been instantiated outside of this function and showing the plot should be handled there after all axes of the figure have been filled in with plots. Besides, multiple calls to `matplotlib.pyplot.show()` seem to work weird, and cannot be replaced by `ax.figure.show()` or `fig.show()` calls to backward compatibility.

thatlittleboy: I see where you're coming from. I'm just not sure if overriding the user's input is the best user experience (e.g. if the user explicitly states `show=True, ax=ax`, and the figure isn't shown, the user might be surprised as to why that is.

It's alright for now, we can standardize this behaviour across all plotting functions in a future PR.

thatlittleboy (approved):
Approving, thanks and sorry it took a while to get this merged in!

PR #3356, opened 2023-10-21
chore: Code cleanup and general maintenance

thatlittleboy:
## Overview

<s>Closes #XXXX</s>

Just some additional cleaning up of the code base, including:

* [Update error and warning messages formatting](https://github.com/shap/shap/commit/155f82def1ab9fe98499299aec0ba425bdbb6ed4)
* [Don't use assert, raise appropriate Exception](https://github.com/shap/shap/commit/e801ad3a198c77dab19c05f6ecc9e5acd8c28bbc)
* [Remove all pylint ignores](https://github.com/shap/shap/commit/ab1e89861591d8ec2823bad1bcbfe4d2d11a5972)
  * Almost all of the pylint ignores don't apply anymore
* [Use f-string debug now that we are on py3.8+](https://github.com/shap/shap/commit/208e9483436dd3cb7e4d6b9e7860fd60f8d9ee0c)
* [Clear dead code](https://github.com/shap/shap/commit/a6de5a34b27cf574f19c7bf6a1b66332f0470be7)
* [Standardize Github issue number formatting](https://github.com/shap/shap/pull/3356/commits/178bcefb92c0726b5977dcbc7bee7dd2df068299) to GH #xxxx

This PR is not meant to be comprehensive, just an effort towards better codebase maintainability.

## Checklist


connortann:
This looks generally good, nice work!

I do still feel a little uncomfortable with the exception formatting pattern, as we discussed previously.  I think the vanilla approach of:

```python
raise ValueError("The thing is bad")
```

is more commonly used than the more verbose alternative:

```python
emsg = "The thing is bad"
raise ValueError(emsg)
```

I had a look at the origin of the recommendation to avoid in-line exception messages, which seems to be the [TRY003](https://github.com/guilatrova/tryceratops/blob/main/docs/violations/TRY003.md) linting rule from tryceratops. There is also a discussion about this rule in the Ruff implementation [here](https://github.com/astral-sh/ruff/issues/2246).
The motivation was to reduce repetition of error messages by defining custom exception classes. This doesn't seem relevant in our case, as we're not re-using any custom classes.

Another reason for this pattern is that the exception traceback is slightly shorter. This doesn't seem to be particularly important given that the traceback is typically several dozens of lines long, and in any case python programmers are accustomed to the typical layout. In terms of readability for developers, any advantage in a slightly shorter traceback is arguably offset by devs being slightly confused about why the exception is raised in a roundabout way with an `emsg` variable.

I suppose it's also a matter of determining what is more "readable", which I think is largely informed by what is most common. Is there any consensus in support of this new pattern in any other popular libs? If not, should we avoid being esoteric in style?

thatlittleboy:
@connortann  Yea we've had this discussion before, it's not TRY003, it's [EM101](https://docs.astral.sh/ruff/rules/raw-string-in-exception/).
And the reasoning for adopting this style is explained in the `ruff` documentation linked above. I think their reasoning makes sense, but alas it's largely just a matter of style and taste; I'm not really interested to have an extensive discussion about this as it's not really productive. If you're opposed to it, I'll revert those changes.

I had the rule set up in my neovim editor so it was just a little annoying to see the warnings in the gutter.

But I'm abit preoccupied over the next few days, I'll have to come back to this PR later.

connortann:
I think we should arrive at a decision as to whether this is a recommendation for all other contributions that should be documented and enforced, or just a optional pattern in your PRs.

connortann:
I accidentally pressed "close with comment", my apologies. I'm working on mobile at the moment, not the first time I've hit that button by mistake.

Overall I'm not too fussed, but as mentioned I'd be interested in understanding if this convention is widespread and worth adopting / enforcing.

In any case, your PR overall looks great, though I haven't yet had time to go through line by line.

shap/maskers/_text.py
```diff
@@ -50,8 +50,8 @@ def __init__(self, tokenizer=None, mask_token=None, collapse_mask_token="auto",
             try:
                 self.tokenizer = SimpleTokenizer(tokenizer)
             except Exception:
-                raise Exception( # pylint: disable=raise-missing-from
-                    "The passed tokenizer cannot be wrapped as a masker because it does not have a __call__ " + \
+                raise Exception(
```
connortann, line 53: Note to self for future: perhaps a more specific Exception would be better

shap/explainers/tf_utils.py
```diff
@@ -63,10 +63,12 @@ def _get_model_inputs(model):
...
         isinstance(model, tf.keras.Model):
         return model.inputs
-    elif str(type(model)).endswith("tuple'>"):
+    if str(type(model)).endswith("tuple'>"):
         return model[0]
-    else:
-        assert False, str(type(model)) + " is not currently a supported model type!"
+
+    emsg = f"{type(model)} is not currently a supported model type!"
+    raise ValueError(emsg)
```
connortann, line 70: FWIW I slightly prefer the former style on this bit, as the `if / elif / else` style makes it clear that the options are mutually exclusive. They're both functionally equivalent though so happy either way.

connortann (approved):
Coming back to this PR - looks great! A few minor nitpicks above for future, but I'll go ahead and merge this now. Apologies for the far-too-long time to properly review.

PR #3360, opened 2023-10-23
Default to "identity" instead of a callable in KernelExplainer

thatlittleboy:
## Overview

A simple PR to not use the callable `IdentityLink()` as default argument. We have `convert_to_link()` in `__init__` which handles the conversion automatically from the string "identity" so this should not affect user experience at all.

Plus some minor docstring changes.

## Checklist


connortann (approved):
LGTM!

PR #3379, opened 2023-11-03
fix for shap.maskers.Impute class

### LINKED ISSUE #3378: BUG: shap.maskers.Impute() throws TypeError
### Issue Description

`shap.maskers.Impute()` in version 0.42.1 is throwing the following TypeError when used in an `Explainer` object. This is regardless of model or data type (minimal reproducible example below uses a toy example that was not the first time I encountered this problem). It seems like this problem would likely be encountered in other (albeit much older) issues like #1723. It's entirely possible that I am not using `Impute` or `Explainer` properly, in which case I would appreciate any corrections. :smile:

From what I can tell in the source code, `Impute` has not implemented a `__call__` function and is inheriting the blank `__call__` function from `Masker`, its parent. There's a comment that it should eventually be inheriting from `Tabular` once arbitrary masking is supported. Should this be changed, or can someone expand on this? Thanks for the help!

P.S. Given the latest source code and release notes, it is not clear to me that 0.43.0 resolves this issue, or that the issue is resolved on the master branch.

```python
# relevant packages
from sklearn.datasets import make_regression
from sklearn.neural_network import MLPRegressor
from sklearn.model_selection import train_test_split
import shap

# setup toy data
X, y = make_regression(n_samples=100)
X_train, X_test, y_train, y_test = train_test_split(X, y, train_size = 0.75)

# train toy model
model = MLPRegressor()
model.fit(X_train, y_train)
model.score(X_test, y_test)

background = shap.maskers.Impute(X_train)
# TypeError here:
explainer = shap.Explainer(model.predict, masker=background)

shap_values = explainer(X_test)
exp = shap.Explanation(shap_values.values, shap_values.base_values, shap_values.data)
```

### Traceback

```shell
TypeError                                 Traceback (most recent call last)
     17 # TypeError here:
     18 explainer = shap.Explainer(model.predict, masker=background)
---> 20 shap_values = explainer(X_test)
     21 exp = shap.Explanation(shap_values.values, shap_values.base_values, shap_values.data)

File ~/miniconda3/lib/python3.9/site-packages/shap/explainers/_permutation.py:76, in Permutation.__call__(self, max_evals, main_effects, error_bounds, batch_size, outputs, silent, *args)
     72 def __call__(self, *args, max_evals=500, main_effects=False, error_bounds=False, batch_size="auto",
     73              outputs=None, silent=False):
     74     """ Explain the output of the model on the given arguments.
     75     """
---> 76     return super().__call__(
     77         *args, max_evals=max_evals, main_effects=main_effects, error_bounds=error_bounds, batch_size=batch_size,
     78         outputs=outputs, silent=silent
     79     )

File ~/miniconda3/lib/python3.9/site-packages/shap/explainers/_explainer.py:264, in Explainer.__call__(self, max_evals, main_effects, error_bounds, batch_size, outputs, silent, *args, **kwargs)
    262     feature_names = [[] for _ in range(len(args))]
    263 for row_args in show_progress(zip(*args), num_rows, self.__class__.__name__+" explainer", silent):
--> 264     row_result = self.explain_row(
    265         *row_args, max_evals=max_evals, main_effects=main_effects, error_bounds=error_bounds,
    266         batch_size=batch_size, outputs=outputs, silent=silent, **kwargs
    267     )
    268     values.append(row_result.get("values", None))
    269     output_indices.append(row_result.get("output_indices", None))

File ~/miniconda3/lib/python3.9/site-packages/shap/explainers/_permutation.py:134, in Permutation.explain_row(self, max_evals, main_effects, error_bounds, batch_size, outputs, silent, *row_args)
    131     i += 1
    133 # evaluate the masked model
--> 134 outputs = fm(masks, zero_index=0, batch_size=batch_size)
    136 if row_values is None:
    137     row_values = np.zeros((len(fm),) + outputs.shape[1:])

File ~/miniconda3/lib/python3.9/site-packages/shap/utils/_masked_model.py:66, in MaskedModel.__call__(self, masks, zero_index, batch_size)
     64         full_masks = np.zeros((int(np.sum(masks >= 0)), self._masker_cols), dtype=bool)
     65         _convert_delta_mask_to_full(masks, full_masks)
---> 66         return self._full_masking_call(full_masks, zero_index=zero_index, batch_size=batch_size)
     68 else:
     69     return self._full_masking_call(masks, batch_size=batch_size)

File ~/miniconda3/lib/python3.9/site-packages/shap/utils/_masked_model.py:106, in MaskedModel._full_masking_call(self, masks, zero_index, batch_size)
    103     masked_inputs = (masked_inputs,)
    105 # masked_inputs = self.masker(mask, *self.args)
--> 106 num_mask_samples[i] = len(masked_inputs[0])
    108 # see which rows have been updated, so we can only evaluate the model on the rows we need to
    109 if i == 0 or self._variants is None:

TypeError: object of type 'NoneType' has no len()
```

### Expected Behavior

_No response_

### Bug report checklist

### Installed Versions

0.42.1

stompsjo:
In brief, `Impute` now uses `Tabular` as a parent and uses the sklearn.impute objects for imputation. See #3378 for more discussion (h/t @CloseChoice).

## Overview

Closes #3378

#### Description of the changes proposed in this pull request:

This addresses the `TypeError` generated when `shap.maskers.Impute()` is used in an `Explainer` object. It currently calls `Masker` as a parent instead of `Tabular` and therefore does not have a `__call__` method.

Instead, `Impute` now inherits `Tabular` as a parent (including its `__call__` method and functionality). Upon `Impute.__init__`, an `sklean.impute` object is used for imputing masked data in `Tabular.__call__`. I added four supported keywords to `method`: "mean", "median", "mode", and "knn." The first three refer to the different strategies used by [`sklearn.impute.SimpleImputer`](https://scikit-learn.org/stable/modules/generated/sklearn.impute.SimpleImputer.html#sklearn.impute.SimpleImputer.transform). The last one creates an [`sklearn.impute.KNNImputer`](https://scikit-learn.org/stable/modules/generated/sklearn.impute.KNNImputer.html#sklearn.impute.KNNImputer) object with default inputs instead. There are more Scikit-Learn imputers (e.g. [`sklearn.impute.IterativeImputer`](https://scikit-learn.org/stable/modules/generated/sklearn.impute.IterativeImputer.html#sklearn.impute.IterativeImputer), which is experimental) and many more settings, but if that amount of control is needed, I have added the functionality for `method` to be a pre-initialized `sklearn.impute` object supplied by the user.

## Checklist
I have added two additional tests similar to those for `Independent` and `Partition`. I recognize that there's probably a place for making _all_ of these tabular tests more meaningful for each type of masker.

### DIFF
--- a/shap/maskers/_tabular.py
+++ b/shap/maskers/_tabular.py
@@ -331,6 +331,8 @@ class LinearImpute:
             Array to impute missing values for, should be masked
             using missing_value.
         """
+        if len(data.shape) != 2:
+            raise NotImplementedError(f"Currently only 2 dimensional data can by processed with the LinearImpute class. You provided {len(data.shape)}. If this is crucial to you, feel free to open an issue: https://github.com/shap/shap/issues.")
         self.data = pd.DataFrame(data)
         self.data = self.data.replace(self.missing_value, np.NaN)
         interpolated = self.data.interpolate(
@@ -376,7 +378,7 @@ class Impute(Tabular):
         methods = ["linear", "mean", "median", "mode", "knn"]
         if isinstance(method, str):
             if method not in methods:
-                raise NotImplementedError("Given imputation method is not supported.")
+                raise NotImplementedError(f"Given imputation method is not supported. Please provide one of the following methods: {', '.join(methods)}")
             elif method == "knn":
                 impute = KNNImputer(missing_values=0)
             elif method == "linear":
--- a/shap/maskers/_tabular.py
+++ b/shap/maskers/_tabular.py
@@ -299,7 +299,7 @@ class Partition(Tabular):
         super().__init__(data, max_samples=max_samples, clustering=clustering)
 
 
-class Impute(Masker): # we should inherit from Tabular once we add support for arbitrary masking
+class Impute(Tabular):
     """ This imputes the values of missing features using the values of the observed features.
 
     Unlike Independent, Gaussian imputes missing values based on correlations with observed data points.
--- a/shap/maskers/_tabular.py
+++ b/shap/maskers/_tabular.py
@@ -308,6 +308,39 @@ class Partition(Tabular):
         super().__init__(data, max_samples=max_samples, clustering=clustering)
 
 
+class LinearImpute():
+    """ Simple class for imputing missing values using pandas.Series.interpolate.
+    """
+
+    def __init__(self, missing_value=0):
+        """ Build a linear imputer for Impute classes when method=linear
+
+        Parameters
+        ----------
+        missing_value : int, numpy.NaN
+            The missing value to impute.
+        """
+        self.missing_value = missing_value
+
+    def fit(self, data):
+        """ Linearly impute missing values in the data and return as array.
+        
+        Parameters
+        ----------
+        data : numpy.ndarray
+            Array to impute missing values for, should be masked
+            using missing_value.
+        """
+        self.data = pd.Series(data)
+        self.data = self.data.replace(self.missing_value, np.NaN)
+        interpolated = self.data.interpolate(
+            method="linear",
+            limit_direction="both"
+        )
+
+        return interpolated.values()
+
+
 class Impute(Tabular):
     """ This imputes the values of missing features using the values of the observed features.
 
@@ -340,12 +373,14 @@ class Impute(Tabular):
                 mode - SimpleImputer with elements replaced by most frequent of feature in data.
                 knn - KNNImputer with elements replaced by mean value within 5 NNs of feature in data.
         """
-        methods = ["mean", "median", "mode", "knn"]
+        methods = ["linear", "mean", "median", "mode", "knn"]
         if isinstance(method, str):
             if method not in methods:
                 raise NotImplementedError("Given imputation method is not supported.")
             elif method == "knn":
                 impute = KNNImputer(missing_values=0)
+            elif method == "linear":
+                impute = LinearImpute(missing_value=0)
             else:
                 impute = SimpleImputer(missing_values=0, strategy=method)
         else:
--- a/shap/maskers/_tabular.py
+++ b/shap/maskers/_tabular.py
@@ -1,4 +1,5 @@
 import logging
+from typing import Callable
 
 import numpy as np
 import pandas as pd
@@ -65,7 +66,12 @@ class Tabular(Masker):
         # prepare by fitting sklearn imputer
         self.impute = impute
         if self.impute is not None:
-            self.impute.fit(self.data)
+            if len(self.data.shape) == 1:
+                self.impute.fit(self.data.reshape(1, -1))
+            elif len(self.data.shape) == 2:
+                self.impute.fit(self.data)
+            elif len(self.data.shape) >= 2:
+                raise NotImplementedError(f"Currently only 1 and 2 dimensional data can by processed with the LinearImpute class. You provided {len(self.data.shape)}. If this is crucial to you, feel free to open an issue: https://github.com/shap/shap/issues.")
 
         # # warn users about large background data sets
         # if self.data.shape[0] > 100:
@@ -124,7 +130,12 @@ class Tabular(Masker):
             return (masked_inputs_out,), varying_rows_out
 
         if self.impute is not None:
-            self.data = self.impute.transform(x)
+            if len(x.shape) == 1:
+                self.data = self.impute.transform(x.reshape(1, -1))
+            elif len(x.shape) == 2:
+                self.data = self.impute.transform(x)
+            elif len(x.shape) >= 2:
+                raise NotImplementedError(f"Currently only 1 and 2 dimensional data can by processed with the LinearImpute class. You provided {len(x.shape)}. If this is crucial to you, feel free to open an issue: https://github.com/shap/shap/issues.")
 
         # otherwise we update the whole set of masked data for a single sample
         self._masked_data[:] = x * mask + self.data * np.invert(mask)
@@ -323,6 +334,9 @@ class LinearImpute:
         self.missing_value = missing_value
 
     def fit(self, data):
+        self.data = pd.DataFrame(data)
+
+    def transform(self, x):
         """ Linearly impute missing values in the data and return as array.
 
         Parameters
@@ -331,16 +345,23 @@ class LinearImpute:
             Array to impute missing values for, should be masked
             using missing_value.
         """
-        if len(data.shape) != 2:
-            raise NotImplementedError(f"Currently only 2 dimensional data can by processed with the LinearImpute class. You provided {len(data.shape)}. If this is crucial to you, feel free to open an issue: https://github.com/shap/shap/issues.")
-        self.data = pd.DataFrame(data)
-        self.data = self.data.replace(self.missing_value, np.NaN)
-        interpolated = self.data.interpolate(
+        self.x = x
+        if len(x.shape) == 1:
+            self.x = x.reshape(1, -1)
+        elif len(x.shape) > 2:
+            raise NotImplementedError(f"Currently only 1 and 2 dimensional data can by processed with the LinearImpute class. You provided {len(x.shape)}. If this is crucial to you, feel free to open an issue: https://github.com/shap/shap/issues.")
+        self.x = pd.DataFrame(self.x)
+        # Pandas interpolate uses NaN as missing value
+        self.x = self.x.replace(self.missing_value, np.NaN)
+        # number of imputed samples for indexing later
+        n_samples = self.x.shape[0]
+        # combine with background data for interpolation
+        interpolated = pd.concat([self.data, self.x], axis=0).interpolate(
             method="linear",
             limit_direction="both"
         )
 
-        return interpolated.values()
+        return interpolated.values[-n_samples]
 
 
 class Impute(Tabular):
@@ -375,7 +396,7 @@ class Impute(Tabular):
                 mode - SimpleImputer with elements replaced by most frequent of feature in data.
                 knn - KNNImputer with elements replaced by mean value within 5 NNs of feature in data.
         """
-        methods = ["linear", "mean", "median", "mode", "knn"]
+        methods = ["linear", "mean", "median", "most_frequent", "knn"]
         if isinstance(method, str):
             if method not in methods:
                 raise NotImplementedError(f"Given imputation method is not supported. Please provide one of the following methods: {', '.join(methods)}")
@@ -385,7 +406,9 @@ class Impute(Tabular):
                 impute = LinearImpute(missing_value=0)
             else:
                 impute = SimpleImputer(missing_values=0, strategy=method)
-        else:
+        elif isinstance(method, Callable):
             impute = method
+        else:
+            raise NotImplementedError(f"Given imputation method is not supported. Please provide one of the following methods: {', '.join(methods)}")
 
         super().__init__(data, max_samples=max_samples, impute=impute)
--- a/shap/maskers/_tabular.py
+++ b/shap/maskers/_tabular.py
@@ -308,7 +308,7 @@ class Partition(Tabular):
         super().__init__(data, max_samples=max_samples, clustering=clustering)
 
 
-class LinearImpute():
+class LinearImpute:
     """ Simple class for imputing missing values using pandas.Series.interpolate.
     """
 
@@ -324,7 +324,7 @@ class LinearImpute():
 
     def fit(self, data):
         """ Linearly impute missing values in the data and return as array.
-        
+
         Parameters
         ----------
         data : numpy.ndarray
--- a/shap/maskers/_tabular.py
+++ b/shap/maskers/_tabular.py
@@ -3,6 +3,7 @@ import logging
 import numpy as np
 import pandas as pd
 from numba import njit
+from sklearn.impute import KNNImputer, SimpleImputer
 
 from .. import utils
 from .._serializable import Deserializer, Serializer
@@ -17,7 +18,7 @@ class Tabular(Masker):
     """ A common base class for Independent and Partition.
     """
 
-    def __init__(self, data, max_samples=100, clustering=None):
+    def __init__(self, data, max_samples=100, clustering=None, impute=None):
         """ This masks out tabular features by integrating over the given background dataset.
 
         Parameters
@@ -60,6 +61,11 @@ class Tabular(Masker):
         self.data = data
         self.clustering = clustering
         self.max_samples = max_samples
+
+        # prepare by fitting sklearn imputer
+        self.impute = impute
+        if self.impute is not None:
+            self.impute.fit(self.data)
 
         # # warn users about large background data sets
         # if self.data.shape[0] > 100:
@@ -116,6 +122,9 @@ class Tabular(Masker):
                 return (pd.DataFrame(masked_inputs_out, columns=self.feature_names),), varying_rows_out
 
             return (masked_inputs_out,), varying_rows_out
+
+        if self.impute is not None:
+            self.data = self.impute.transform(x)
 
         # otherwise we update the whole set of masked data for a single sample
         self._masked_data[:] = x * mask + self.data * np.invert(mask)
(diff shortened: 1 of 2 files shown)


shap/maskers/_tabular.py
```diff
@@ -299,24 +308,47 @@ def __init__(self, data, max_samples=100, clustering="correlation"):
...
+        methods = ["mean", "median", "mode", "knn"]
+        if isinstance(method, str):
+            if method not in methods:
+                raise NotImplementedError("Given imputation method is not supported.")
+            elif method == "knn":
+                impute = KNNImputer(missing_values=0)
+            else:
+                impute = SimpleImputer(missing_values=0, strategy=method)
+        else:
+            impute = method
```
stompsjo: Seems to me that this should check whether `method` is an appropriate object (`sklearn.impute.SimpleImputer`, `sklearn.impute.KNNImputer`, etc.) rather than any other object that will result in an error when `self.impute.fit` is called later.

shap/maskers/_tabular.py
```diff
@@ -299,24 +308,47 @@ def __init__(self, data, max_samples=100, clustering="correlation"):
...
-            self.cov = data.get("cov", None)
-            data = np.expand_dims(data["mean"], 0)
+        methods = ["mean", "median", "mode", "knn"]
+        if isinstance(method, str):
+            if method not in methods:
+                raise NotImplementedError("Given imputation method is not supported.")
+            elif method == "knn":
+                impute = KNNImputer(missing_values=0)
+            else:
+                impute = SimpleImputer(missing_values=0, strategy=method)
```
stompsjo: The default behavior of `sklearn.impute` objects is to have a `missing_value` used to determine what elements to impute. AFAIK, `0` is valid given a standardized mask above, right?

stompsjo (commented):
A couple of questions/comments.

stompsjo:
Hm, I missed that (legacy) code/tests specified a "linear" impute method, which was missing from my implementation. (I'm kind of surprised unit tests are passing when `method='linear'`, although I guess it doesn't do anything as is?) I added a simple `LinearImpute` class that linearly interpolates missing values using the `pandas.Series.interpolate()` method. I am sure there are other ways to do this, happy to hear suggestions.

CloseChoice:
@connortann could this PR get your approval to run the workflows?

shap/maskers/_tabular.py
```diff
@@ -299,24 +308,82 @@ def __init__(self, data, max_samples=100, clustering="correlation"):
...
+    def fit(self, data):
+        """ Linearly impute missing values in the data and return as array.
+
+        Parameters
+        ----------
+        data : numpy.ndarray
+            Array to impute missing values for, should be masked
+            using missing_value.
+        """
+        self.data = pd.DataFrame(data)
```
CloseChoice:
```suggestion
        if len(data.shape) != 2:
            raise NotImplementedError(f"Currently only 2 dimensional data can by processed with the LinearImpute class. You provided {len(data.shape)}. If this is crucial to you, feel free to open an issue: https://github.com/shap/shap/issues.")
        self.data = pd.DataFrame(data)
```

I would like to check explicitly that the data can be converted into a pd Dataframe

shap/maskers/_tabular.py
```diff
@@ -299,24 +308,82 @@ def __init__(self, data, max_samples=100, clustering="correlation"):
...
+                knn - KNNImputer with elements replaced by mean value within 5 NNs of feature in data.
         """
-        if data is dict and "mean" in data:
-            self.mean = data.get("mean", None)
-            self.cov = data.get("cov", None)
-            data = np.expand_dims(data["mean"], 0)
+        methods = ["linear", "mean", "median", "mode", "knn"]
+        if isinstance(method, str):
+            if method not in methods:
+                raise NotImplementedError("Given imputation method is not supported.")
```
CloseChoice:
```suggestion
                raise NotImplementedError(f"Given imputation method is not supported. Please provide one of the following methods: {', '.join(methods)}")
```

shap/maskers/_tabular.py
CloseChoice: NIT: this could result in weird error messages in the unlikely case that someone provides an non-callable object. Maybe we can check for this explicitly by importing

```python
from typing import Callable
...
elif isinstance(method, Callable):
    impute = method
else:
    raise TypeError("some error message")
```

tests/maskers/test_tabular.py
```diff
@@ -54,7 +54,7 @@ def test_serialization_independent_masker_numpy():
         temp_serialization_file.seek(0)
 
         # deserialize masker
-        new_independent_masker = shap.maskers.Masker.load(temp_serialization_file)
+        new_independent_masker = shap.maskers.Independent.load(temp_serialization_file)
```
CloseChoice, line 60: why this change?

tests/maskers/test_tabular.py
```diff
@@ -107,7 +107,60 @@ def test_serialization_partion_masker_numpy():
...
+
+    with tempfile.TemporaryFile() as temp_serialization_file:
+
+        # serialize partition masker
+        original_partition_masker.save(temp_serialization_file)
+
+        temp_serialization_file.seek(0)
+
+        # deserialize masker
+        new_partition_masker = shap.maskers.Impute.load(temp_serialization_file)
```
CloseChoice, line 167: Could you please add some tests where you actually impute something, the housing dataset does not have missing data AFAIK. Parameterizing the tests for each method would be great aswell.

CloseChoice (changes requested):
Sorry for the late review, I added myself as reviewer so please ping me whenever you want another review. Thanks for the PR, this looks really promising. I added a couple of comments, feel free to ignore the NIT prefixed one.

CloseChoice:
Please also add the reproducible example from issue #3378 as a test

tests/maskers/test_tabular.py
stompsjo: Maybe I misinterpreted what this test is doing. I assumed that we want to use the load method specifically of the masker being tested, rather than the parent `Masker` class. Happy to revert it if needed. This and line 110 below are changed:
https://github.com/shap/shap/blob/43d47f63e9f747243f2f3f56fadda87265e75fe0/tests/maskers/test_tabular.py#L110

tests/maskers/test_tabular.py
stompsjo: Here's a start:
https://github.com/shap/shap/blob/75c88e8de2c1efaabce06ab6982f44db7720b6fe/tests/maskers/test_tabular.py#L175-L191

I tried to come up with something reproducible for all methods. Let me know if you have feedback.

stompsjo:
Alright, changes made. @CloseChoice, feel free to give it another review when you're ready.

connortann:
Thanks for the PR! Chipping in with my two cents.

It's not clear to me what the original implementation intended; it does indeed look half-finished. So, it would be great to get this sorted. Two thoughts:

## 1. Imputing strategy

The current docstring for the `Impute` states that Gaussian imputing is used:

https://github.com/shap/shap/blob/63223e11eb651dc22b802647526b6bb8d0a02293/shap/maskers/_tabular.py#L302-L305

This doesn't seem to be consistent with the new implementation; should we look to add it in?

##  2.  Use from Explainer

I noticed the Imputer is used in the LinearExplainer here. Will this this need to be rewritten?

https://github.com/shap/shap/blob/63223e11eb651dc22b802647526b6bb8d0a02293/shap/explainers/_linear.py#L76-L83

shap/maskers/_tabular.py
```diff
@@ -61,6 +63,16 @@ def __init__(self, data, max_samples=100, clustering=None):
         self.clustering = clustering
         self.max_samples = max_samples
 
+        # prepare by fitting sklearn imputer
```
CloseChoice, line 66: I would rather see this in the impute than in the generic tabular class

shap/maskers/_tabular.py
```diff
@@ -17,7 +19,7 @@ class Tabular(Masker):
     """ A common base class for Independent and Partition.
     """
 
-    def __init__(self, data, max_samples=100, clustering=None):
+    def __init__(self, data, max_samples=100, clustering=None, impute=None):
```
CloseChoice, line 22: I would add this keyword only for the impute class ant not for tabular. See my comment below.

shap/maskers/_tabular.py
```diff
@@ -117,6 +129,14 @@ def __call__(self, mask, x):
 
             return (masked_inputs_out,), varying_rows_out
 
+        if self.impute is not None:
```
CloseChoice, line 132: This should then also just be initialized in impute

tests/maskers/test_tabular.py
```diff
@@ -107,11 +110,103 @@ def test_serialization_partion_masker_numpy():
...
 
     mask = np.ones(X.shape[1]).astype(int)
     mask[0] = 0
     mask[4] = 0
 
     # comparing masked values
     assert np.array_equal(original_partition_masker(mask, X[0])[0], new_partition_masker(mask, X[0])[0])
+
+def test_imputation():
```
CloseChoice, line 175: I think we should get another test where we have values to impute (so explicitly set some values to `np.nan`).

shap/maskers/_tabular.py
stompsjo: I'm not sure I understand. Do you suggest overwriting the `__call__` method for `Impute`? I agree that this shaping check could be moved to the `LinearImpute` and `Impute` classes (sklearn imputers expect 2-dimensional arrays, so I kept that consistent in `LinearImpute`). However, some form of `self.impute.fit()` and `self.impute.transform()` still need to be called during the main algorithm as written.

shap/maskers/_tabular.py
```diff
@@ -299,24 +319,96 @@ def __init__(self, data, max_samples=100, clustering="correlation"):
...
-        if data is dict and "mean" in data:
-            self.mean = data.get("mean", None)
-            self.cov = data.get("cov", None)
-            data = np.expand_dims(data["mean"], 0)
+        methods = ["linear", "mean", "median", "most_frequent", "knn"]
+        if isinstance(method, str):
+            if method not in methods:
+                raise NotImplementedError(f"Given imputation method is not supported. Please provide one of the following methods: {', '.join(methods)}")
+            elif method == "knn":
+                impute = KNNImputer(missing_values=0)
```
CloseChoice, line 404: IMO the missing values should be `np.nan` otherwise replacing `0` can lead to unintended results

CloseChoice (changes requested):

stompsjo:
@connortann Thanks for the comments! Responding,
1. Yes, the docstring should be updated to better reflect that this class is imputing based on some "interpolation" method. I will make that update.
2. I don't believe any change should be necessary(?) since I have added a "linear" method with `LinearImpute`.

tests/maskers/test_tabular.py
CloseChoice: also can we do 

```python
@pytest.mark.parametrize("method", ["linear", "mean", "median", "most_frequent", "knn"])
def test_imputation(method):
    ...
```
and then get rid of the loop within the function? That helps us locate what method throws an error

shap/maskers/_tabular.py
stompsjo: I think line 142 may cause issues:
https://github.com/shap/shap/blob/75c88e8de2c1efaabce06ab6982f44db7720b6fe/shap/maskers/_tabular.py#L141
If the missing value is `np.nan`, and that is present in `x`, then `x*mask` will still leave the np.nan in place. I _think_ the current behavior is to have masked values set to `0` via the mask.

(That said, I recognize that as is, every time `0` is present in the data it will be imputed (or replaced by the other maskers like `Partition` or `Independent`), regardless of whether that `0` was masked or part of the inherent data.)

CloseChoice:
There seems to be something wrong with the implementation since it is not correctly imputing nans. I just changed the `test_imputation` to:
```python
def test_imputation():
    # toy data
    x = np.full((5, 5), np.arange(1,6)).T

    methods = ["mean", "median", "most_frequent", "knn"]
    # toy background data
    bckg = np.full((5, 5), np.arange(1,6)).T
    for method in methods:
        # toy sample to impute
        x = np.arange(1, 6)
        x = x.astype(np.float32)
        x[3] = np.nan
        masker = shap.maskers.Impute(np.full((1,5), 1), method=method)
        # only mask the second value
        mask = np.ones_like(bckg[0])
        mask[1] = 0
        # masker should impute the original value (toy data is predictable)
        imputed = masker(mask.astype(bool), x)
        # assert np.all(x == imputed)
```
and the imputed result was almost always (`(array([[          1,           2,           3, -2147483648,           5]]),)`). Also had to change the `missing_values=0` parameter to `missing_values=np.nan`

stompsjo:
Yeah, check [my comment above](https://github.com/shap/shap/pull/3379#discussion_r1423128467), it has to do with line 142 and the way that masks are applied.

connortann:
> @connortann Thanks for the comments! Responding,
> 
> 1. Yes, the docstring should be updated to better reflect that this class is imputing based on some "interpolation" method. I will make that update.
> 2. I don't believe any change should be necessary(?) since I have added a "linear" method with `LinearImpute`.

On 1., my thinking was actually more along the lines that perhaps the code should implement what was originally described in the docstring: using a Gaussian model with means and a covariance matrix to impute the missing values. The Gaussian imputing strategy seems particularly useful in the context of Shapley values as it accounts for the correlations between variables- perhaps it should be the default?

Regarding the other imputing strategies- I'm a little uncomfortable with the "interpolation" method. I think the rest of the code base generally assumes that samples are I.I.D. and _exchangable_, so that any two rows in the dataset could be swapped. The "interpolation" strategy breaks this assumption as it depends on the row order. @CloseChoice what do you think about this one, is the "interpolation" strategy problematic?

Thank you again for helping us get this into shape! Apologies that this PR has a fair bit of discussion; that's just a consequence of the original code being quite poorly defined and documented.

stompsjo:
> On 1., my thinking was actually more along the lines that perhaps the code should implement what was originally described in the docstring: using a Gaussian model with means and a covariance matrix to impute the missing values. The Gaussian imputing strategy seems particularly useful in the context of Shapley values as it accounts for the correlations between variables- perhaps it should be the default?

This is where I think it might be useful to review the implementation presented in #1489 that was merged into the `shap:benchmark_utility` branch. That seems to do as you describe, but I have not tested it (or know the history behind it/that branch). Maybe it's time to include that in `shap:master`? I like what you suggest: having that behavior as a default, with the other options (sans interpolation) as additional methods.

CloseChoice:
> > @connortann Thanks for the comments! Responding,
> > 
> > 1. Yes, the docstring should be updated to better reflect that this class is imputing based on some "interpolation" method. I will make that update.
> > 2. I don't believe any change should be necessary(?) since I have added a "linear" method with `LinearImpute`.
> 
> On 1., my thinking was actually more along the lines that perhaps the code should implement what was originally described in the docstring: using a Gaussian model with means and a covariance matrix to impute the missing values. The Gaussian imputing strategy seems particularly useful in the context of Shapley values as it accounts for the correlations between variables- perhaps it should be the default?
> 
> Regarding the other imputing strategies- I'm a little uncomfortable with the "interpolation" method. I think the rest of the code base generally assumes that samples are I.I.D. and _exchangable_, so that any two rows in the dataset could be swapped. The "interpolation" strategy breaks this assumption as it depends on the row order. @CloseChoice what do you think about this one, is the "interpolation" strategy problematic?
> 
> Thank you again for helping us get this into shape! Apologies that this PR has a fair bit of discussion; that's just a consequence of the original code being quite poorly defined and documented.

Sorry, I somehow missed this. So yes, I think it's not ideal to break iid + exchangability assumption. I would change this in such a fashion, that we create a linear model, that takes the well defined subset of X (where now np.nan values are present), and then we fit the model to each column with missing values and replace the missing ones with the model output. This might be tedious if there are a lot of columns with missing values but IMO the most straight forward approach.

I can even pick this one up and implement that approach if need be.

stompsjo:
@CloseChoice That sounds like a plan. If you can get to this in the near-term, feel free to work on an implementation. I can otherwise try to pick this back up soonish. Going with your approach, presumably the columns the model is fitting to would have to be complete (i.e. no missing values)? Otherwise how would the model estimate a value for that row? I concede I probably didn't fully comprehend your comment 😃

CloseChoice:
> @CloseChoice That sounds like a plan. If you can get to this in the near-term, feel free to work on an implementation. I can otherwise try to pick this back up soonish. Going with your approach, presumably the columns the model is fitting to would have to be complete (i.e. no missing values)? Otherwise how would the model estimate a value for that row? I concede I probably didn't fully comprehend your comment 😃

Yes, my approach would be to first filter for the subset of rows that have null values in only the target column or no null values at all and then fit the model on it.

CloseChoice:
Thanks for the PR, due to faded activity I'll close this here for now. Feel free to reopen if you want to come back and continue to work on this

PR #3386, opened 2023-11-16
ENH: Add support for Explanation.display_data in bar plot

### LINKED ISSUE #3387: ENH: add support for display_data in bar plot
### Problem Description

shap.plots.bar doesn't seem to use `.display_data` when plotting which is useful when categorical features have been encoded.

### Alternative Solutions

I've added a PR here: https://github.com/shap/shap/pull/3386 that uses the same logic as the waterfall plot. The feature values will be taken from `.display_data` if it exists and then tries `.data`

### Additional Context

_No response_

### Feature request checklist

fountaindive:
Closes: #3387

For `shap.plots.bar` if attribute `.display_data` given then use that otherwise use `.data`.

## Overview

I noticed that `shap.plots.bar` doesn't make use of `explanation.display_data` if it exists which I've found to be really useful in other shap plots e.g. the waterfall or scatter plots. For example when you have categorical features that you have encoded but would like the plot to show the "unencoded" feature value.

The change in the code is basically taken straight from the waterfall plot code.

## Checklist

### DIFF
--- a/shap/plots/_bar.py
+++ b/shap/plots/_bar.py
@@ -88,7 +88,7 @@ def bar(shap_values, max_display=10, order=Explanation.abs, clustering=None, clu
         # TODO: check other attributes for equality? like feature names perhaps? probably clustering as well.
 
     # unpack the Explanation object
-    features = cohort_exps[0].data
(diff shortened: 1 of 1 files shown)


connortann (approved):
Looks good to me, thanks for the PR!

fountaindive:
awesome thank you!

PR #3391, opened 2023-11-17
ENH: return axis in shap.plots.beeswarm when show=False

stompsjo:
## Overview

A _very_ simple change (could probably be included with other changes?) to return the figure object from `matplotlib` when creating a beeswarm plot. This aligns it with the return of `shap.plots.waterfall` as the figure object may be useful for a user to have.

Is an additional unit test needed?

## Checklist


connortann:
Thanks for the PR!

May I suggest returning `ax` rather than the figure?

There is a more general discussion over on #3411 on how we can improve the plotting API. An axis-level interface should be more flexible than a figure-level interface, as multiple plots could be composed together.

shap/plots/_beeswarm.py
```diff
@@ -421,6 +421,8 @@ def beeswarm(shap_values, max_display=10, order=Explanation.abs.mean(0),
     pl.xlabel(labels['VALUE'], fontsize=13)
     if show:
         pl.show()
+    else:
+        return pl.gcf()
```
connortann: Suggest returning the axes:

```suggestion
        return pl.gca()
```

It would also be helpful to update the docstring to document the new return value.

connortann (changes requested):

stompsjo:
I made that change, updated the docstring, and did the same for `_waterfall` since I was looking at it. I am happy to defer to the community regarding what should be returned. I concur that the axis-level would be preferable.

connortann (approved):
Thanks for making that update, LGTM

PR #3402, opened 2023-11-28
MAINT: Add version number from git tag using setuptools scm

### LINKED ISSUE #3059: MAINT: adopt dynamic versioning to facilitate TestPyPi releases
## Proposal

Adopt setuptools-scm for versioning.

### Commitless releases

Enables us to publish by just making a GitHub release, without having to make a new commit to modify the hard-coded version.

### Publishing to TestPyPI

Publishing to PyPi will fail if the given version already exists. This is not an issue for regular releases, as the tagged versions should always be unique.

However, we may wish to publish to TestPyPi more regularly, as recommended in the [python packaging guide](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/).

At present, the `build_wheels.yaml` release pipeline will only pass once for a given hardcoded version, then thereafter it will fail. The PyPi publish action mentions:

> [In order to publish to TestPyPi regularly] generate a unique version number for each upload so that they'd not create a conflict [... which is] possible if you use setuptools_scm package

Setuptools-scm will generate unique versions for non-tagged versions, which is useful in itself. As a side benefit, there will be a single source of truth of the version.

connortann:
## Overview

Closes #3059

- Single-source the version from git tag, to enable "commitless releases"
- Import `__version__` from the dynamically generated `shap/_version.py
- Add `/shap/_version.py` to `.gitignore`, and also sort the existing rules alphabetically
- Remove the TestPyPI step from the publishing workflow, to avoid any unexpected surprises

Link to docs: <https://setuptools-scm.readthedocs.io/en/latest/usage/>

Test run of `build wheels` workflow (succeeded): <https://github.com/shap/shap/actions/runs/7020852838>


.github/workflows/build_wheels.yml
```diff
@@ -101,11 +112,20 @@ jobs:
...
+
+      # The step below can be uncommented for testing / debugging purposes.
+
+      # We do not publish to TestPyPI here, as the job will fail if the version
+      # number already exists on TestPyPI.
+
+      # Note: if the non-release builds may fail to upload to TestPyPI if the version
+      # number is "non-local": https://github.com/pypa/setuptools_scm/issues/342
+
+      # - name: Publish package to TestPyPI
```
CloseChoice: think this is fine. I might need a mental note to check if we can parameterize the github actions run so that one can pass a "test" parameter to the run which will then either publish to TestPyPI or just skip the last step. 

This is what I found with a quick search:
 - https://stackoverflow.com/a/62736541
 - https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

CloseChoice:
Did you test this?

CloseChoice (approved):
If this is tested, then we're good to go.

.github/workflows/build_wheels.yml
connortann: I am unsure if we really need to enable this final publish step for anything other than real releases. I used it a while ago when we updated the GitHub action, but I don't think it's something we would run routinely.

If we're testing the release workflow, I think it will typically be on a non-tagged release. PyPI ( and also TestPyPI) will not allow an upload if the version number is not well formatted and unique, so we wouldn't expect the step to pass unless we have actually made a real tagged release.

CloseChoice: With testing I mean uncommenting the release step and just running the pipeline up to that point.

connortann:
> Did you test this?

I've tested it locally like this:

```bash
# Make a tag (but do not push!)
git tag -a v0.0.999 -m "my test tag"

# Build the project
pip install build
python -m build

# See the built package with the correct version
ls dist
> shap-0.0.999-cp311-cp311-linux_x86_64.whl  shap-0.0.999.tar.gz
```

I also manually triggered the build_wheels workflow, as per the PR description.

That said, I think we will still want to keep a close eye on the workflow when making the next release `0.44.0` just in case anything fails.

Two releases ago, the release pipeline happened to fail because tensorflow made a breaking release just as I triggered a shap release, which caused the unit tests to fail. I deleted the tag, fixed the tests, and re-tagged; and it was then published to PyPI successfully.

.github/workflows/build_wheels.yml
connortann: On anything other than a tagged version, the version number will look something like this: `0.44.0.dev4+gdf99fe2`

If we were to try publishing such a version to TestPyPI, it would be rejected as the second bit (the "local" bit) isn't accepted.

connortann: From a different thread I see:

> usage of local versions is intended to prevent accident release uploads

> Setuptools-scm is looking out for folks to prevent blunders from happening.

which makes sense. From https://github.com/pypa/setuptools_scm/issues/342

connortann:
Thanks for the review @CloseChoice ! I'll merge it so we're ready to make a release.

connortann:
Actually, on further reflection I think we should be able to get the TestPyPI working for when the workflow is manually triggered. I'll give that a shot...

connortann:
OK @CloseChoice I've edited the build_wheels workflow so that the package will be published to TestPyPI whenever we trigger the workflow manually, even if it's not a tagged release.

To get that working, I added `local_scheme = "no-local-version"` to our setuptools-scm config in pyproject.toml. This means that for dev releases the version number will be a PyPI-compliant number like e.g. `0.43.1.dev46` , meaning "46 new commits since last tag 0.43.1". 

I re-requested a review if you'd kindly take a look?

connortann:
Huh. I did a test run of build_wheels : https://github.com/shap/shap/actions/runs/7116720022 .

The run mostly worked as expected, and there is a new release on TestPyPI:
https://test.pypi.org/project/shap/0.43.1.dev48/

However, the upload of the source distribution failed, as `shap-0.43.1.dev48.tar.gz` file size of 159.0MB is over the 100MB limit. Looking at the [logs](https://github.com/shap/shap/actions/runs/6454916907/job/17522789269) of our last release upload, the file `shap-0.43.0.tar.gz` was only `420kB`. 

So, next step for me is to figure out why the source tarball is so big.

connortann:
OK, I think I determined the root cause. setuptools-scm changes what files are included by default in the source distribution, so the directories like `docs`, `data` and `notebooks` were being included, which pushed us over the PyPI file size limit. This is fixed by excluding these directories with a `MANIFEST.in` file.

Second manual run of build_wheels to ensure the PyPI publish step works: 
https://github.com/shap/shap/actions/runs/7118860427

Well, I suppose this just goes to show that @CloseChoice 's suggestion to get TestPyPI working was a good idea. I've re-requested a review for the new changes.

.github/workflows/build_wheels.yml
```diff
@@ -85,11 +96,10 @@ jobs:
           name: sdist_files
           path: dist/*.tar.gz
 
-  publish_wheels:
-    name: Publish wheels on pypi
+  publish_test_pypi:
+    name: Publish wheels to TestPyPI
```
CloseChoice, line 100: just a question, let's say we are building a release version from master and publish_test_pypi works but for some reason publish_pypi does not (image their api is too busy and timing out or whatever) then we would need to retrigger the job but publish_test_pypi will fail because the version we want to build is already published, right? So do we have a possibility to delete versions from TestPYPI? I assume that would be a manual process, but that is fine given the chances of this scenario.

pyproject.toml
```diff
@@ -96,13 +96,11 @@ packages = [
...
 ]
-zip-safe = false
 
-[tool.setuptools.dynamic]
-version = {attr = "shap.__version__"}
-
-[tool.setuptools.package-data]
-shap = ["plots/resources/*", "cext/*"]
+[tool.setuptools_scm]
+version_file = "shap/_version.py"
```
CloseChoice, line 101: does this file already exist? I would have thought it is added with this PR?

CloseChoice: Ah I understood. setuptools-scm extracts the git tag and writes this into the file, we'll use it to allow `shap.__version__` with the code you added in the `__init__.py`

CloseChoice (approved):
Looks good, nice work!

.github/workflows/build_wheels.yml
connortann: I was thinking about that scenario also; however, I think in the proposed setup we shouldn't encounter this problem. We have separated the two uploads as two separate jobs in the CI workflow, so if TestPyPI fails (e.g. because the version already exists), the main PyPI job can still proceed. So, hopefully this allays the concern that an issue with TestPyPI might interfere with a real release.

Regarding deletions: I believe it's possible to "yank" a release, but in general it seems to be better to simply increment the patch version and push a new release.

pyproject.toml
connortann: Yep, exactly.

PR #3408, opened 2023-11-29
MAINT: fix UP032, use f-string instead of format()

connortann:
## Overview

Fixes ruff linting rule UP032. Implemented by ruff's `--fix` flag.


CloseChoice (approved):
LGTM

PR #3409, opened 2023-11-29
MAINT: Fix ruff E402 (imports at top of file)

connortann:
Fix ruff rule E402 (imports not at top of file).

Configures ruff to ignore this rule in Jupyter notebooks, as imports are sometimes helpful in context.


tests/plots/test_decision.py
```diff
@@ -2,9 +2,9 @@
 import matplotlib.pyplot as pl
 import numpy as np
 
-matplotlib.use('Agg')
-import shap  # pylint: disable=wrong-import-position
```
connortann: It looks to me like previously these two lines were deliberately in an unusual order, as the pylint rule was suppressed. I can't think of any reason for this, and the tests pass fine after this change. But, thought I'd flag it in case I'm missing something.

pyproject.toml
```diff
@@ -153,7 +151,7 @@ exclude = ["**.py"]
 "notebooks/tabular_examples/tree_based_models/tree_shap_paper/*"  = ["ALL"]
 
 # Disable some unwanted rules on Jupyter notebooks
-"*.ipynb" = ["E703"]  # Trailing semicolons, used to supress outputs
+"*.ipynb" = ["E703", "E402"]  # Allow trailing semicolons, allow imports not at top
```
CloseChoice, line 153: why do we allow imports to not be at the top for notebooks? I do not really have any objections but am just curious.

connortann: Good question. Some of the notebooks have some imports in context under the relevant subheading, rather than at the top. For example [here](https://shap.readthedocs.io/en/latest/example_notebooks/tabular_examples/model_agnostic/Iris%20classification%20with%20scikit-learn.html). there are a few subheadings with examples of explaining different types of models, and each model is imported in the relevant section.

I think this is fine in documentation, as things can be a bit clearer when  everything is together in context.

CloseChoice (approved):

PR #3410, opened 2023-11-29
Fix ambiguous variable names

connortann:
## Overview

Replaces ambiguous variable names with more readable alternatives.  In some fonts, `l` and `1` are indistinguishable.

Fixes ruff rule E741.


shap/explainers/_gradient.py
```diff
@@ -291,27 +291,27 @@ def shap_values(self, X, nsamples=200, ranked_outputs=None, output_rank_order="m
                 for k in range(nsamples):
                     rind = np.random.choice(self.data[0].shape[0])
                     t = np.random.uniform()
-                    for l in range(len(X)):
+                    for u in range(len(X)):
```
CloseChoice, line 295: very minor NIT: why use `u` here instead of `i`. I would consider `i` more intuitive for row indices, and `j` for columns.

CloseChoice (approved):
The only comment I have is just very minor nit pick, maybe not even worth an answer. I also do not see that we should try to enforce consistency in there. This thought just came to my mind when I went over your changes.

shap/explainers/_gradient.py
connortann: I think `i` is already in use here, from line 281. Running out of letters actually!

CloseChoice (approved):
LGTM

PR #3423, opened 2023-12-06
MAINT: Enable "no response" bot

connortann:
## Overview

Closes #3422

Add the "no response" bot, to close issues with the "awaiting feedback" label where the original author has not replied. For more info about the bot see: https://github.com/lee-dohm/no-response

This is probably less impactful that the "stale bot" over on #3424 , so we should probably focus on that first.

## Checklist

- [X] All [pre-commit checks](https://pre-commit.com/#install) pass.
- [X] Unit tests added (if fixing a bug or adding a new feature)

PR #3424, opened 2023-12-06
MAINT: Enable stale bot in debug mode

connortann:
## Overview

- Implements stale bot as discussed in https://github.com/shap/shap/discussions/3051
- Updates maintainer docs with overview of issue triage and use of labels

I have suggested starting with "debug mode" so that no action is taken, but the output is written to the logs. I propose we start with a dry run, which requires merging into the master branch. Then if the logs look suitable then we can remove the debug option.

For more info about the action see https://github.com/actions/stale

## Discussion

This is related to #3423 , which addresses a slightly different situation. I think the two approaches could be complementary. I think this one is a good place to start, as the volume of "stale" issues is much higher than the "awaiting feedback" issues.

### DIFF
--- a/.github/workflows/stale_issues.yml
+++ b/.github/workflows/stale_issues.yml
@@ -1,0 +1,54 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    # Run at 2:30 AM every day
+    - cron: '30 2 * * *'
+  # Enable manual run
+  workflow_dispatch:
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v8
+        with:
+          # Set for dry-run only. TODO: remove debug-only
+          debug-only: true
+          # See discussion on generous time limits:
+          # https://github.com/shap/shap/discussions/3051
+          days-before-stale: 730  # 2 years
+          days-before-close: 90   # 3 months
+          stale-issue-message: "This issue has been inactive for two years, so it's
+            been automatically marked as 'stale'.
+            \n\n
+            We value your input! If you think this issue is still relevant, please
+            leave a comment to keep it open and remove the 'stale' label.
+            \n\n
+            If there's no activity in the next 30 days the issue will be closed."
+          stale-pr-message: "We appreciate your contribution! This Pull Request has
+            been inactive for two years, so it's been automatically marked as 'stale'.
+            \n\n
+            Please leave a comment if you would still like the PR to remain open, and
+            the 'stale' label will be removed.
+            \n\n
+            Otherwise, if there is no activity in the next 30 days the PR will be
+            closed."
+          close-issue-message: "This issue has been automatically closed due to lack of
+            recent activity.
+            \n\n
+            Your input is important to us! Please feel free to open a new issue if the
+            problem persists or becomes relevant again."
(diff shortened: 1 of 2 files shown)


.github/workflows/stale_issues.yml
```diff
@@ -0,0 +1,54 @@
...
+          # https://github.com/shap/shap/discussions/3051
+          days-before-stale: 730  # 2 years
+          days-before-close: 90   # 3 months
+          stale-issue-message: "This issue has been inactive for two years, so it's
+            been automatically marked as 'stale'.
+            \n\n
+            We value your input! If you think this issue is still relevant, please
+            leave a comment to keep it open and remove the 'stale' label.
+            \n\n
+            If there's no activity in the next 30 days the issue will be closed."
```
CloseChoice:
```suggestion
            If there's no activity in the next 90 days the issue will be closed."
```

Or did I miss something? I thought line 24 configures this time span?!

.github/workflows/stale_issues.yml
```diff
@@ -0,0 +1,54 @@
...
+          stale-pr-message: "We appreciate your contribution! This Pull Request has
+            been inactive for two years, so it's been automatically marked as 'stale'.
+            \n\n
+            Please leave a comment if you would still like the PR to remain open, and
+            the 'stale' label will be removed.
+            \n\n
+            Otherwise, if there is no activity in the next 30 days the PR will be
```
CloseChoice:
```suggestion
            Otherwise, if there is no activity in the next 90 days the PR will be
```

.github/workflows/stale_issues.yml
```diff
@@ -0,0 +1,54 @@
...
+            closed."
+          close-issue-message: "This issue has been automatically closed due to lack of
+            recent activity.
+            \n\n
+            Your input is important to us! Please feel free to open a new issue if the
```
CloseChoice, line 46: why not just reopen the ticket?

CloseChoice:
Hmm, I am a bit torn on this one. On the one hand I am worried that we might lose valualbe information if we'll close tickets just because the person who opened it does not respond in time. But I would see this as totally fine procedure regarding PRs. On the other hand my former scrum master always said: "What's important comes back in no time" and I guess he's right about that. And I think you wrote good, clear and appreciative messages to the authors and are pretty defensive about the time frames. 

So I am fine with this change.

CloseChoice (approved):
Looks good to me, just some minor things to sort out

.github/workflows/stale_issues.yml
connortann: Ah good spot, thanks

.github/workflows/stale_issues.yml
connortann: I believe than only maintainers can re-open issues, so most of the community will instead need to open a new one:

https://github.com/orgs/community/discussions/41926

connortann:
Example run in debug mode (which is capped to one batch at a time):
https://github.com/shap/shap/actions/runs/7126747237/job/19405262741

```
Processed items: 388
├── Processed issues: 359
└── Processed PRs   : 29
New stale issues: 3
No longer stale items: 3
├── No longer stale issues: 1
└── No longer stale PRs   : 2
Deleted items labels: 3
├── Deleted issues labels: 1
└── Deleted PRs labels   : 2
Added issues labels: 3
Added issues comments: 3
Fetched items: 400
Fetched items events: 10
Fetched items comments: 10
Operations performed: 33
```

PR #3425, opened 2023-12-07
FIX: Workaround for broken "fetch-tags" option

connortann:
Hot patch for failed release workflow due to https://github.com/actions/checkout/issues/1467


connortann:
I'll merge this now to get the release fixed; thought would appreciate a post-hoc code review.

PR #3427, opened 2023-12-07
Stale bot: process oldest issues first

connortann:
A suggested tweak to the stale bot, to process the oldest issues first. The "ascending" key controls whether the oldest or newest issues are sorted first:

https://github.com/actions/stale#ascending

Each run is limited to maximum 30 GH API operations, which so far ends up tagging ~5 issues per run. The workflow is currently scheduled to run once per day, and we can also manually trigger it. 

https://github.com/actions/stale#operations-per-run

I think this run limit could be a useful thing to avoid creating too many notifications at once, so we could leave it ticking over in the background rather than processing everything at once.

### DIFF
--- a/.github/workflows/stale_issues.yml
+++ b/.github/workflows/stale_issues.yml
@@ -17,9 +17,7 @@ jobs:
   stale:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


CloseChoice (approved):
LGTM

PR #3460, opened 2024-01-15
add script to run notebooks in CI

### LINKED ISSUE #3430: ENH: Run Notebooks in CI pipeline
### Problem Description

We are currently making a huge effort to get the notebooks up and running in #3036. But each and every breaking change we do might break the notebooks once again. So having a CI workflow where we check that the notebooks run through would be great. So my idea would be to have a CI workflow and a make command that runs the notebooks (maybe we need to skip or refactor some due to execution time limits).

E.g. #3318 will break a lot of notebooks.

### Alternative Solutions

_No response_

### Additional Context

We could use [papermill](https://papermill.readthedocs.io/en/latest/usage-execute.html#execute-a-notebook-with-parameters) or [nbconvert](https://nbconvert.readthedocs.io/en/latest/execute_api.html) for this

### Feature request checklist

CloseChoice:
## Overview

Closes #3430  

- add script to run notebooks in CI. Note that the notebooks time out after 30 seconds, so a lot of the notebooks are not run but this is a good first step.

## Checklist

~- [ ] Unit tests added (if fixing a bug or adding a new feature)~


CloseChoice:
@connortann I would put the currently failing notebooks on an allowlist and then we can cross them off one by one while fixing them via #3036. Your thoughts on this are very welcome

pyproject.toml
```diff
@@ -75,6 +75,15 @@ test = [
   "opencv-python",
 ]
 
+notebooks = [
```
connortann: Would `test_notebooks` be a better name?

scripts/run_notebooks_timeouts.py
```diff
@@ -0,0 +1,72 @@
...
+    # Add Matplotlib configuration to use a non-interactive backend
+    exporter = ScriptExporter()
+    content, _ = exporter.from_filename(notebook_path)
+    content = 'import matplotlib\nmatplotlib.use("Agg")\n' + content
+
+    with open(output_path, 'w', encoding='utf-8') as f:
+        f.write(content)
+
+
+def execute_python_script(script_path, timeout_seconds=30):
```
connortann: Rather than convert to python files, would it be simpler to use the "execute" API of `nbconvert` directly?
https://nbconvert.readthedocs.io/en/latest/execute_api.html

At the CLI this is something like: 

```bash
jupyter nbconvert --to notebook --execute mynotebook.ipynb
```

The python API would probably better, to avoid having to call a subprocess. Eg:

```python
ep = ExecutePreprocessor(timeout=TIMEOUT)

for filename in Path('notebooks').rglob('*.ipynb'):
    with open(filename) as f:
        nb = nbformat.read(f, as_version=4)
    ep.preprocess(nb)
```

connortann:
> @connortann I would put the currently failing notebooks on an allowlist and then we can cross them off one by one while fixing them via #3036. Your thoughts on this are very welcome

This seems like a very sensible idea! Pragmatically, not all notebooks will make the allowlist, but having partial coverage of the docs would still be a neat improvement.

I added some more general thoughts to the issue above.

scripts/run_notebooks_timeouts.py
```diff
@@ -0,0 +1,72 @@
+import os
```
connortann: Nitpick: I'd advocate for `pathlib.Path` as a modern high-level replacement for `os.path`

pyproject.toml
CloseChoice: I am indifferent to a change here. Technically these are also the dependencies needed to execute notebooks. Your call how we name this.

scripts/run_notebooks_timeouts.py
CloseChoice: done

scripts/run_notebooks_timeouts.py
CloseChoice: Thanks, think this is a great solution

pyproject.toml
CloseChoice: I renamed to `test_notebooks`

scripts/run_notebooks_timeouts.py
```diff
@@ -0,0 +1,103 @@
+import time
+from pathlib import Path
+
+import nbformat
+from nbconvert.preprocessors import CellExecutionError, ExecutePreprocessor
+
+TIMEOUT = 20  # seconds
+
+allow_to_fail = [
```
connortann, line 9: Nitpick: it might be a little simpler to make these paths relative to the `notebooks` directory.

connortann:
This looks great - nice work. Ready for me to merge?

connortann (approved):
LGTM! One minor suggestion above but otherwise looks great

connortann:
@CloseChoice  the new job appears to fail on master. Would you kindly take a look?

Recent failing job on master:
https://github.com/shap/shap/actions/runs/7640505479

EDIT: Interestingly a more recent job on master passes:
https://github.com/shap/shap/actions/runs/7642306906/job/20821620592

Issue #3251, opened 2023-09-08
BUG: No GPU Support for Modern CUDA


alexk101:
### Issue Description

This is somewhat related to a number of GPU issues people have had, and probably more directly to #3150, but as is, the GPU kernels for this project do not compile with modern versions of CUDA. From my understanding this should be all versions >11.7, as that was when the `sm_37` compute capability was [deprecated](https://docs.nvidia.com/cuda/archive/11.7.1/cuda-toolkit-release-notes/index.html#title-new-cuda-libraries) from CUDA, which is required for building the kernels, as referenced here https://github.com/shap/shap/blob/2262893cf441478418abac5fd8cdd38e436a867b/setup.py#L77-L78

I think that there are a couple of ways that this could be approached. The first, I would say is a short term solution, and that is capping the maximum allowed CUDA version to 11.7. I don't have a machine to test this, but from my understanding of this error, that should fix it.

The second approach, which is more sustainable in the long term, but will require more work, is removing support for `sm_37` (which is the kepler architecture btw). Removing the previously referenced lines would be the first step, but as I have found will still require some changes to the kernels themselves for CUDA 12 and Thrust 2.0.0, as I get compile errors related to `__device__`-only lambda’s return type being queried from host code here https://github.com/shap/shap/blob/2262893cf441478418abac5fd8cdd38e436a867b/shap/cext/gpu_treeshap.h#L292

I don't know which approach the maintainers would like to take, but it would be nice to have shap on modern versions of CUDA. I am willing to help with that, but I must confess that my experience with c++ is fairly limited, so I would need some help from someone that is more informed. Familiarity with those kernels would also probably be helpful.

Additionally, I think that CUDA/GPU support should be a required parameter provided by the user when building, since as it is now, the setup simply has `with_cuda` hardcoded to True. This way, if the cuda build fails, the entire build can fail, though this would have to be checked for in this function by monitoring stderr https://github.com/shap/shap/blob/2262893cf441478418abac5fd8cdd38e436a867b/setup.py#L67

As it stands, even if the CUDA build fails, the setup will still continue. I didn't notice the CUDA compile errors until later because they are never propagated in the logs and just sit at the top of the install log.

Thanks, Alex

### Minimal Reproducible Example

```python
git clone https://github.com/shap/shap.git
cd shap
python setup.py install
```

### Traceback

```shell
Modern CUDA compile error snippet:

NVCC ==>  /opt/cuda/bin/nvcc
[...]

Issue #2662, opened 2022-08-26
SHAP TreeExplainer is not compatible with XGBoost models that use Pandas categorical features


rwjmiller:
As of XGBoost 1.6.0, it is possible to use Pandas categorical dtypes as inputs to models, avoiding the need for encoding:

https://xgboost.readthedocs.io/en/latest/tutorials/categorical.html

However, if you do this, XGBoost only allows you to serialise the model using JSON / UBJSON. SHAP currently loads the trees in a model by reading a raw binary dump generated via the save_raw() method, so it can't load models that make use of this functionality. Attempting to do so results in the following XGBoost error:

```
  explainer = TreeExplainer(model)
  File "/usr/local/lib/python3.9/site-packages/shap/explainers/_tree.py", line 149, in __init__
    self.model = TreeEnsemble(model, self.data, self.data_missing, model_output)
  File "/usr/local/lib/python3.9/site-packages/shap/explainers/_tree.py", line 824, in __init__
    xgb_loader = XGBTreeModelLoader(self.original_model)
  File "/usr/local/lib/python3.9/site-packages/shap/explainers/_tree.py", line 1431, in __init__
    self.buf = xgb_model.save_raw()
  File "/usr/local/lib/python3.9/site-packages/xgboost/core.py", line 2216, in save_raw
    _check_call(
  File "/usr/local/lib/python3.9/site-packages/xgboost/core.py", line 203, in _check_call
    raise XGBoostError(py_str(_LIB.XGBGetLastError()))
xgboost.core.XGBoostError: [16:41:34] ../src/tree/tree_model.cc:871: Check failed: !HasCategoricalSplit(): Please use JSON/UBJSON for saving models with categorical splits.

```

Are there any plans to add support for this type of model, e.g. by using the JSON dump instead to load tree data if enable_categorical is true for the model?

Apologies if this is a known issue or there is a simple workaround that allows users to continue using this XGBoost functionality with SHAP - I searched briefly on the issues page but I couldn't find anything that seemed to match this. I got the error above when using the 0.41.0 release.

dataversenomad:
Hi @rwjmiller . I was struggling with this issue a lot of time. The solution is to use xgb.train (the traditional method), apparently there is still no support for one of scikit-learn version (XGBClassifier). 

One other solution might be NOT TO USE three_method = 'gpu_hist' in your model (at least on the Explainer object). I was facing the same issue for several time and removed it from my model object. After that, everything works fine.

[...]

PR #3311, opened 2023-10-05
WIP: ENH: support cuda 12

CloseChoice:
## Overview

Closes #3251  

Description of the changes proposed in this pull request:
- these are just the changes @alexk101 proposed in #3251. I ran some code on a cuda capable GPU with CUDA 12.2 installed
- WIP: reverted to legacy build system so that `pip install -e .` installs cuda if it is available (note that this does not work right now, since setup.py does not have a fallback for non-cuda GPUs, need to fix this)

## Checklist

- [x] All [pre-commit checks](https://pre-commit.com/#install) pass.
- [ ] ~Unit tests added (if fixing a bug or adding a new feature)~

PR #3462, opened 2024-01-18
Fixes for XGBoost model load and GPU build.

### LINKED ISSUE #1302: Xgboost binary format compatibility

### LINKED ISSUE #1864: UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 337: invalid start byte

trivialfis:
## Overview

Closes #1302 
Closes #1864

- Require XGBoost >= 1.6
- ~Use JSON or UBJSON for loading XGBoost model.~ (handled in https://github.com/shap/shap/pull/3345)
- Handle link functions for various objectives.
- Correctly handle `best_iteration`.
- Fix GPU build.
- Code for loading split types. See below.

I have added some parsing code for loading categorical splits, but I am not sure how to support it in the SHAP package https://github.com/shap/shap/issues/2662. For future reference, the SHAP implementation in XGBoost can output SHAP values with categorical splits. In order to support it here in the SHAP package, we need to modify the plotting logic and the implementation of SHAP algorithms (C and Python).

Alternatively, we can have more SHAP algorithms inside XGBoost, and the shap package simply reroutes the input data to the `booster.predict` method for all algorithms. Currently, both the original SHAP contribution and interaction are handled this way.

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -76,11 +76,11 @@ def compile_cuda_module(host_args):
 
     print("NVCC ==> ", nvcc)
     arch_flags = (
-        "-arch=sm_37 "
-        "-gencode=arch=compute_37,code=sm_37 "
+        "-gencode=arch=compute_60,code=sm_60 "
         "-gencode=arch=compute_70,code=sm_70 "
         "-gencode=arch=compute_75,code=sm_75 "
-        "-gencode=arch=compute_75,code=compute_75"
+        "-gencode=arch=compute_75,code=compute_75 "
+        "-gencode=arch=compute_80,code=sm_80"
     )
     nvcc_command = (
         f"-allow-unsupported-compiler shap/cext/_cext_gpu.cu -lib -o {lib_out} "
--- a/shap/cext/gpu_treeshap.h
+++ b/shap/cext/gpu_treeshap.h
@@ -17,10 +17,11 @@
 #pragma once
 #include <thrust/device_allocator.h>
 #include <thrust/device_vector.h>
+#include <thrust/host_vector.h>
 #include <thrust/iterator/discard_iterator.h>
 #include <thrust/logical.h>
 #include <thrust/reduce.h>
-#include <thrust/host_vector.h>
+#include <thrust/sort.h>
 #if (CUDART_VERSION >= 11000)
 #include <cub/cub.cuh>
 #else
@@ -849,7 +850,7 @@ void GetBinSegments(const PathVectorT& paths, const SizeVectorT& bin_map,
 
 struct DeduplicateKeyTransformOp {
   template <typename SplitConditionT>
-  __device__ thrust::pair<size_t, int64_t> operator()(
+  __host__ __device__ thrust::pair<size_t, int64_t> operator()(
       const PathElement<SplitConditionT>& e) {
     return {e.path_idx, e.feature_idx};
   }
@@ -875,13 +876,17 @@ void DeduplicatePaths(PathVectorT* device_paths,
   // Sort by feature
   thrust::sort(thrust::cuda::par(alloc), device_paths->begin(),
                device_paths->end(),
-               [=] __device__(const PathElement<SplitConditionT>& a,
-                              const PathElement<SplitConditionT>& b) {
-                 if (a.path_idx < b.path_idx) return true;
-                 if (b.path_idx < a.path_idx) return false;
+               [=] __host__ __device__(const PathElement<SplitConditionT> &a,
+                                       const PathElement<SplitConditionT> &b) {
+                 if (a.path_idx < b.path_idx)
+                   return true;
+                 if (b.path_idx < a.path_idx)
+                   return false;
 
-                 if (a.feature_idx < b.feature_idx) return true;
-                 if (b.feature_idx < a.feature_idx) return false;
+                 if (a.feature_idx < b.feature_idx)
+                   return true;
+                 if (b.feature_idx < a.feature_idx)
+                   return false;
                  return false;
                });
 
@@ -895,13 +900,13 @@ void DeduplicatePaths(PathVectorT* device_paths,
   size_t* h_num_runs_out;
   CheckCuda(cudaMallocHost(&h_num_runs_out, sizeof(size_t)));
 
-  auto combine = [] __device__(PathElement<SplitConditionT> a,
-                               PathElement<SplitConditionT> b) {
+  auto combine = [] __host__ __device__(PathElement<SplitConditionT> a,
+                                        PathElement<SplitConditionT> b) {
     // Combine duplicate features
     a.split_condition.Merge(b.split_condition);
     a.zero_fraction *= b.zero_fraction;
     return a;
-  };  // NOLINT
+  }; // NOLINT
   size_t temp_size = 0;
   CheckCuda(cub::DeviceReduce::ReduceByKey(
       nullptr, temp_size, key_transform, DiscardOverload<Pair>(),
@@ -926,18 +931,24 @@ void SortPaths(PathVectorT* paths, const SizeVectorT& bin_map) {
   auto d_bin_map = bin_map.data();
   DeviceAllocatorT alloc;
   thrust::sort(thrust::cuda::par(alloc), paths->begin(), paths->end(),
-               [=] __device__(const PathElement<SplitConditionT>& a,
-                              const PathElement<SplitConditionT>& b) {
+               [=] __host__ __device__(const PathElement<SplitConditionT> &a,
+                                       const PathElement<SplitConditionT> &b) {
                  size_t a_bin = d_bin_map[a.path_idx];
                  size_t b_bin = d_bin_map[b.path_idx];
-                 if (a_bin < b_bin) return true;
-                 if (b_bin < a_bin) return false;
+                 if (a_bin < b_bin)
+                   return true;
+                 if (b_bin < a_bin)
+                   return false;
 
-                 if (a.path_idx < b.path_idx) return true;
-                 if (b.path_idx < a.path_idx) return false;
+                 if (a.path_idx < b.path_idx)
+                   return true;
+                 if (b.path_idx < a.path_idx)
+                   return false;
 
-                 if (a.feature_idx < b.feature_idx) return true;
-                 if (b.feature_idx < a.feature_idx) return false;
+                 if (a.feature_idx < b.feature_idx)
+                   return true;
+                 if (b.feature_idx < a.feature_idx)
+                   return false;
                  return false;
                });
 }
--- a/shap/explainers/_gpu_tree.py
+++ b/shap/explainers/_gpu_tree.py
@@ -97,10 +97,16 @@ class GPUTreeExplainer(TreeExplainer):
         assert not approximate, "approximate not supported"
 
         X, y, X_missing, flat_output, tree_limit, check_additivity = \
-            self._validate_inputs(X, y,
-                                  tree_limit,
-                                  check_additivity)
-        transform = self.model.get_transform()
+            self._validate_inputs(X, y, tree_limit, check_additivity)
+
+        model = self.model
+        if model.model_type == "xgboost" and model.cat_feature_indices is not None:
+            raise ValueError(
+                "Categorical split is not yet supported. You can still use GPU"
+                " with TreeExplainer and `feature_perturbation=tree_path_dependent`"
+                " when the input XGBoost model is set to use GPU."
+            )
+        transform = model.get_transform()
 
         # run the core algorithm using the C extension
         assert_import("cext_gpu")
--- a/shap/explainers/_gpu_tree.py
+++ b/shap/explainers/_gpu_tree.py
@@ -2,7 +2,12 @@
 import numpy as np
 
 from ..utils import assert_import, record_import_error
-from ._tree import TreeExplainer, feature_perturbation_codes, output_transform_codes
+from ._tree import (
+    TreeExplainer,
+    _xgboost_cat_unsupported,
+    feature_perturbation_codes,
+    output_transform_codes,
+)
 
 try:
     from .. import _cext_gpu
@@ -100,12 +105,7 @@ class GPUTreeExplainer(TreeExplainer):
             self._validate_inputs(X, y, tree_limit, check_additivity)
 
         model = self.model
-        if model.model_type == "xgboost" and model.cat_feature_indices is not None:
-            raise NotImplementedError(
-                "Categorical split is not yet supported. You can still use GPU"
-                " with TreeExplainer and `feature_perturbation=tree_path_dependent`"
-                " when the input XGBoost model is set to use GPU."
-            )
+        _xgboost_cat_unsupported(model)
         transform = model.get_transform()
 
         # run the core algorithm using the C extension
--- a/shap/explainers/_gpu_tree.py
+++ b/shap/explainers/_gpu_tree.py
@@ -101,7 +101,7 @@ class GPUTreeExplainer(TreeExplainer):
 
         model = self.model
         if model.model_type == "xgboost" and model.cat_feature_indices is not None:
-            raise ValueError(
+            raise NotImplementedError(
                 "Categorical split is not yet supported. You can still use GPU"
                 " with TreeExplainer and `feature_perturbation=tree_path_dependent`"
                 " when the input XGBoost model is set to use GPU."
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -59,6 +59,14 @@ def _check_xgboost_version(v: str):
         )
 
 
+def _xgboost_n_iterations(tree_limit: int, num_stacked_models: int) -> int:
+    """Convert number of trees to number of iterations for XGBoost models."""
+    if tree_limit == -1:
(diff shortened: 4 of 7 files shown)


tests/explainers/test_tree.py
```diff
@@ -55,6 +57,72 @@ def test_front_page_xgboost():
...
+
+    explainer = shap.TreeExplainer(clf)
+    assert (
+        explainer.model.tree_limit == (clf.best_iteration + 1) * len(np.unique(y)) * 3
+    )
+
+    # Test ranker
+    ltr = xgboost.XGBRanker(n_estimators=5, num_parallel_tree=3)
+    qid = np.zeros(X_train.shape[0])
+    qid[qid.shape[0] :] = 1
```
trivialfis: Need a fix.

shap/explainers/_tree.py
```diff
@@ -1710,188 +1718,222 @@ def get_xgboost_dmatrix_properties(model):
...
+        learner_model_param = learner["learner_model_param"]
+        objective = learner["objective"]
+
+        booster = learner["gradient_booster"]
+        # check the input model doesn't have vector-leaf
+        iteration_indptr = np.asarray(
+            booster["model"]["iteration_indptr"], dtype=np.int32
+        )
+        diff = np.diff(iteration_indptr)
+        if diff.mean() != iteration_indptr[1] - iteration_indptr[0]:
```
trivialfis: Use all instead, check the trees as well.

shap/explainers/_tree.py
```diff
@@ -1710,188 +1718,222 @@ def get_xgboost_dmatrix_properties(model):
...
-        self.pad_32bit = self.read('i')
-        self.num_pbuffer_deprecated = self.read('Q')
-        self.num_output_group = self.read('i')
-        self.size_leaf_vector = self.read('i')
-        self.read_arr('i', 32) # reserved
+        import xgboost as xgb
+
+        model: xgb.Booster = xgb_model
+        if is_binary:
+            buf = model.save_raw(raw_format="ubj")
```
trivialfis: Check the XGB version.

CloseChoice:
@connortann Could you please allow the workflows? I do not have the permission :(

trivialfis:
Hi @connortann could you please help trigger it again? apologies for the inconvenience.

connortann:
Thanks for the PR @trivialfis , fantastic to have you on board as a contributor. Thank you for updating this branch on top of the changes in #3345.

The changes all look good to me from a first read. There is quite a lot going on here, so I will add a few inline comments in places not least to build my own understanding of a few points.

There is one failing test, but once that is passing this looks good to me.

setup.py
```diff
@@ -76,11 +76,11 @@ def compile_cuda_module(host_args):
...
     print("NVCC ==> ", nvcc)
     arch_flags = (
-        "-arch=sm_37 "
-        "-gencode=arch=compute_37,code=sm_37 "
+        "-gencode=arch=compute_60,code=sm_60 "
         "-gencode=arch=compute_70,code=sm_70 "
         "-gencode=arch=compute_75,code=sm_75 "
-        "-gencode=arch=compute_75,code=compute_75"
+        "-gencode=arch=compute_75,code=compute_75 "
+        "-gencode=arch=compute_80,code=sm_80"
```
connortann, line 83: These flags make sense although I haven't tested them.

I note the repo doesn't have any automated tests of the GPU build, which is concerning - it's hard to verify whether a bug has crept in. I'd love for shap to improve in this area. Out of interest, how does XGBoost test the GPU build on CI - do you have self-hosted runners or something like that?

shap/explainers/_tree.py
```diff
@@ -1023,8 +1030,12 @@ def __init__(self, model, data=None, data_missing=None, model_output=None):
...
             self.objective = objective_name_map.get(xgb_loader.name_obj, None)
             self.tree_output = tree_output_name_map.get(xgb_loader.name_obj, None)
-            if xgb_loader.num_class > 0:
-                self.num_stacked_models = xgb_loader.num_class
+            self.num_stacked_models = xgb_loader.n_trees_per_iter
+            self.cat_feature_indices = xgb_loader.cat_feature_indices
+            best_iteration = getattr(
+                model, "best_iteration", model.num_boosted_rounds() - 1
+            )
+            self.tree_limit = (best_iteration + 1) * self.num_stacked_models
```
connortann: As an aside, this method init method seems much too long. It could be good to refactor this as perhaps as series of class methods: `cls.from_xbg_booster()`, `cls.from_xgb_classifier()` and so on. But let's save that for another future PR though.

shap/cext/gpu_treeshap.h
```diff
@@ -849,7 +850,7 @@ void GetBinSegments(const PathVectorT& paths, const SizeVectorT& bin_map,
 
 struct DeduplicateKeyTransformOp {
   template <typename SplitConditionT>
-  __device__ thrust::pair<size_t, int64_t> operator()(
+  __host__ __device__ thrust::pair<size_t, int64_t> operator()(
```
connortann, line 853: For my own understanding, what is the motivation behind adding the `__host__` decorator here, and also the new `#include <thrust/sort.h>`?

shap/explainers/_tree.py
```diff
@@ -1732,157 +1751,214 @@ def get_xgboost_dmatrix_properties(model):
...
+        import xgboost as xgb
 
+        _check_xgboost_version(xgb.__version__)
+        model: xgb.Booster = xgb_model
 
-class XGBTreeModelLoader:
-    """ This loads an XGBoost model directly from a raw memory dump.
+        raw = xgb_model.save_raw(raw_format="ubj")
+        with io.BytesIO(raw) as fd:
+            jmodel = decode_ubjson_buffer(fd)
```
connortann, line 1809: Much nicer than the previous approach of saving to disk!

CloseChoice:
from my understanding this will also close #3251

shap/explainers/_tree.py
```diff
@@ -1023,8 +1030,12 @@ def __init__(self, model, data=None, data_missing=None, model_output=None):
             self.base_offset = xgb_loader.base_score
             self.objective = objective_name_map.get(xgb_loader.name_obj, None)
             self.tree_output = tree_output_name_map.get(xgb_loader.name_obj, None)
-            if xgb_loader.num_class > 0:
-                self.num_stacked_models = xgb_loader.num_class
+            self.num_stacked_models = xgb_loader.n_trees_per_iter
+            self.cat_feature_indices = xgb_loader.cat_feature_indices
```
CloseChoice: This is still causing problems for me in combination with [this line](https://github.com/shap/shap/pull/3462/files#diff-63ac901e45ff042f4dc0d4a7346c898c16a7254bc940bc19e014cc99052854b1R103) since `None != np.array([])` (which is what we set here in the example in the GPUTree notebook.

shap/explainers/_gpu_tree.py
```diff
@@ -97,10 +97,16 @@ def shap_values(self, X, y=None, tree_limit=None, approximate=False, check_addit
...
 
         X, y, X_missing, flat_output, tree_limit, check_additivity = \
-            self._validate_inputs(X, y,
-                                  tree_limit,
-                                  check_additivity)
-        transform = self.model.get_transform()
+            self._validate_inputs(X, y, tree_limit, check_additivity)
+
+        model = self.model
+        if model.model_type == "xgboost" and model.cat_feature_indices is not None:
```
CloseChoice:
```suggestion
        if model.model_type == "xgboost" and (model.cat_feature_indices is not None and len(model.cat_feature_indices) != 0):
```
See my comment above referencing this line

CloseChoice:
This looks absolutely amazing. Thanks a lot for the PR. I was testing and while installing via
`python setup.py install` I was always getting the error:
`ImportError: cannot import name '_cext_gpu' from partially initialized module 'shap' (most likely due to a circular import) (/home/tobias/programming/github/shap/shap/__init__.py)`. But this also happens on master.

By any chance: do you have a test snippet for me to work with?

This is my test:
```python
import xgboost
import numpy as np

# import shap
from shap import explainers
from shap import datasets

# get a dataset on income prediction
X, y = datasets.adult()

# train an XGBoost model (but any other model type would also work)
model = xgboost.XGBClassifier()
model.fit(X, y)
predicted = model.predict(X, output_margin=True)

# build a Permutation explainer and explain the model predictions on the given dataset
explainer = explainers.GPUTree(model, X)
shap_values = explainer(X)

# get just the explanations for the positive class
shap_values = shap_values
assert np.allclose(shap_values.values.sum(1) + shap_values.base_values, predicted, 1e-2)
```

Edit: Okay so this works and the assert passes on CUDA 12.2 (so this definitely closes #3251), if I add the changes from #3311 in the pyproject.toml and the changes from my comment and install with `pip install -e .`. But would be great if you could elaborate a bit on your testing/installation process.

connortann:
I merged master into this branch to sort a small merge conflict. (Since dropping python 3.8, we now have `list[foo]` rather than `List[foo]`.)

tests/explainers/test_tree.py
```diff
@@ -60,18 +62,95 @@ def test_xgboost_predictions():
...
+        data_missing=None,
+        model_output="raw",
+    )
     y_pred = model.predict(xgboost.DMatrix(X))
     y_pred_tree_ensemble = tree_ensemble.predict(X)
     # this is pretty close but not exactly the same
     assert np.allclose(y_pred, y_pred_tree_ensemble, atol=1e-7)
 
 
+def test_tree_limit() -> None:
```
connortann: Suggest moving this test to be part of the xgboost test class down below:

https://github.com/shap/shap/blob/bf2e44920b85d4f1173d42deff01ec7e6429bd47/tests/explainers/test_tree.py#L1210-L1211

It might be clearer distribute some of these various checks to within the existing test functions in that class.

Somewhat related: #3493

trivialfis:
Thank you for the reviews! I'm currently on holiday and don't have the device for work, will pick up the PR once the holiday is over.

setup.py
trivialfis: Yes, XGBoost has GPU CI with workers on AWS managed by Buildkite. @hcho3 is more knowledgeable about the CI in XGBoost.

> These flags make sense although I haven't tested them.

Actually, the fix here is half-baked. I see that shap is moving toward `pyproject.toml`, but C extensions are only available through `setup.py`. I use the `setup.py` to build GPU extensions, not sure what it takes to get it working with the toml file.

shap/explainers/_tree.py
trivialfis: I agree. We can split up some files and methods for improved modularity. The model loader is another one in mind.

shap/cext/gpu_treeshap.h
trivialfis: The new include because there's a use of the `thrust::sort` function. The newer CUDA version has improved header dependency. Hence, one needs to include a specific header for each feature in thrust.

> what is the motivation behind adding the __host__ decorator here

I think the newer CUDA version requires this to deduce return type for device lambda function.

trivialfis:
@CloseChoice 

>  python setup.py install I was always getting the error:

I'm using the old `setup.py` for building the C extension. The transition to `pyproject.toml` might need some more work from people who are familiar with the SHAP package building process.

shap/explainers/_gpu_tree.py
trivialfis: Hmm, would be great to have more standardized Python attribute like:

```
@property
def cat_features() -> Optional[List[str]]
   return None if ...
```

For now, I will check length.

tests/explainers/test_tree.py
trivialfis: Will move it into the test class.

shap/explainers/_gpu_tree.py
trivialfis: Never mind, will fix it inside the loader.

trivialfis:
> I merged master into this branch to sort a small merge conflict. (Since dropping python 3.8, we now have list[foo]

I have changed all use of `List` into `list`.

shap/explainers/_tree.py
trivialfis: Fixed in the tree loader, this is caused by non-empty feature types, but no categorical feature is used; apologies for the oversight.

trivialfis:
let me add some fixes for RF.

trivialfis:
I have disabled the support for XGBoost random forest as I think it's not quite ready yet, and enabling it might generate invalid results silently or with difficult to verify results. When the random forest is used, the `num_stacked_models` is not the same as `num_outputs`. Please note that XGBoost can boost random forest, which means in some cases, both attributes can be greater than one, which violates some assumptions made in the SHAP library (correct me if I'm wrong).

To enable it, I think some refactorings to reduce the complexity of the codebase might be beneficial. Please let me know if there are plans to do so.

setup.py
CloseChoice: Yes, you are right but currently we only support GPUs if built through `setup.py`. This is still an open issue we need to tackle at some point, so this looks all good. Will check if I can run this today evening

connortann: I think `setup.py` remains the appropriate place for this. Most of the "static" metadata has been moved from `setup.py` to `pyproject.toml`; however, as I understand the configuration of extension modules [needs to remain](https://stackoverflow.com/a/66479252) in setup.py.

connortann:
> I have disabled the support for XGBoost random forest as I think it's not quite ready yet, and enabling it might generate invalid results silently or with difficult to verify results.

This makes sense, I agree it's preferable to explicitly declare which models are supported and which are not. I'd love to improve this over time, so we handle any XGBoost model including boosted random forests.

FYI, recently we added some simple tests with XGBoost RF models here: #3493

> To enable it, I think some refactorings to reduce the complexity of the codebase might be beneficial. Please let me know if there are plans to do so.

Absolutely. I'm keen to help out improving shap in this area as much as I can, and any & all contributions are hugely appreciated. There's quite a bit to do. It's very helpful to have you on board as someone experienced with the internals of the XGBoost tree representation.

connortann (approved):
This looks great, approved from my side. Hoping we can continue with further PRs to refactor and add support for random forests.

Over to you @CloseChoice if you're happy to approve?

trivialfis:
> FYI, recently we added some simple tests with XGBoost RF models here

Thank you for the reference. I looked into the PR, it's great that RF support is of interest. I can look into it to gain more understanding of the SHAP package, and see if we can at least get some cases working.

>  I'm keen to help out improving shap in this area as much as I can,

Thank you! That's a great news!

CloseChoice (approved):
LGTM

CloseChoice:
@trivialfis Thanks a lot. This is great work. Am really happy that you helped out with this one. Would also be happy to see you again if your time allows ;)

PR #3485, opened 2024-02-05
MAINT: Increase operations limit of stale action

connortann:
## Summary

Increases the operations-per-run from 30 to 500.

## Discussion

The GitHub API is limited to 1000 operations per hour, so a limit of 500 means we never come near the limit. See: https://github.com/actions/stale/tree/main?tab=readme-ov-file#operations-per-run

At present the stale action only processes about a dozen items per run. This was useful at first so that we could watch how the action worked whilst avoiding changing too much at once. Now we can see it working, it would be useful to update the issue backlog (e.g. removing `stale` labels upon activity) in a more timely manner.


CloseChoice (approved):
LGTM

PR #3488, opened 2024-02-06
MAINT: xfail on transformers connection errors

connortann:
## Overview

Label a test as XFAIL if a connection error to `transformers` occurs.

## Discussion

A recent CI job failed on `master` due to a connection error when downloading a model from `transformers`:
https://github.com/shap/shap/actions/runs/7798964523/attempts/1

I have seen this happen a few times previously. I don't think this should cause the whole CI workflow to fail, as it seems to be indicative of a flaky network connection rather than a genuine issue with shap. So, suggest marking such failures as "XFAIL" (and thus allowing the overall suite to pass), to avoid this phenomenon making our test suite "flaky".

### DIFF
--- a/tests/models/test_teacher_forcing_logits.py
+++ b/tests/models/test_teacher_forcing_logits.py
@@ -3,6 +3,7 @@
 
 import numpy as np
 import pytest
+from requests.exceptions import RequestException
 
 import shap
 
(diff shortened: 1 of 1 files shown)


tests/models/test_teacher_forcing_logits.py
```diff
@@ -14,8 +14,11 @@ def test_method_get_teacher_forced_logits_for_encoder_decoder_model():
     transformers = pytest.importorskip("transformers")
 
     name = "hf-internal-testing/tiny-random-BartModel"
-    tokenizer = transformers.AutoTokenizer.from_pretrained(name)
-    model = transformers.AutoModelForSeq2SeqLM.from_pretrained(name)
+    try:
+        tokenizer = transformers.AutoTokenizer.from_pretrained(name)
+        model = transformers.AutoModelForSeq2SeqLM.from_pretrained(name)
+    except OSError:  # OSError includes requests.exceptions.RequestException
```
CloseChoice: Why don't we except only `RequestException`?

CloseChoice (approved):

tests/models/test_teacher_forcing_logits.py
connortann: OK, updated accordingly.

I initially kept it as an `OSError` as I think I once encountered an exception that did not inherit from `RequestException` when testing locally, but I can't reproduce it. I think you're right, might as well be specific.

Issue #3411, opened 2023-11-29
[Meta-issue] Improve plotting API for consistency


connortann:
This is a meta-issue to co-ordinate our efforts to on improving shap's plotting API. 

## Summary

The SHAP plotting API itself is currently inconsistent in a few ways, for example in return values, and in styles.

In a similar vein, the shap code itself is inconsistent in the use of matplotlib. We should standardise on using the explicit matplotlib interface (see below).

## Related

I added the `visualisation` label to help track relates issues.

Consistency of the object-oriented interface:

- #3070
- #3206
- #3351
- #3391
- #3819

Relating to style and customisation:

- #1430
- #1672
- #2335
- #2807
- #3213
- #3266 

## TODO

I would suggest taking PR #3523 as an example. We'd like to the same for all matplotlib plotting functions.

Checklist for each function:

- **Unit tests**
  - Coverage at least ~80%
  - Use pytest-mpl, to verify that any refactors do not cause unintended changes
- Uses explicit matplotlib interface 

  - Accepts an optional `ax` parameter, and return `ax`
  - Prefer existing `plt.gcf()` rather than creating new figure
  - Any legacy versions are suitably deprecated
- **Docs**: 
  - Notebook in the "API Examples" section runs and is up-to-date
  - Docstring is complete and up-to-date
- **Typed**: type hints for inputs
- Uses **Explanation** interface, not old-style shap_values

We can track the status of this overhaul in https://github.com/orgs/shap/projects/1 . 

Functions to overhaul:

- [x] #3523
- [x] #3809
- [x] #3793
- [x] #3794
- [ ] #3795
- [ ] #3796
- [ ] #3797
- [x] #3798
- [ ] #3799
- [x] #3800
- [ ] #3801
- [ ] #3802
- [ ] #3803
- [ ] #3805
- [x] #3804
- [ ] #3806
- [ ] #3807
- [ ] #3808

## Context: Implicit vs Explicit matplotlib interface

Adapted from the matplotlib [docs](https://matplotlib.org/stable/users/explain/figure/api_interfaces.html#summary):

Matplotlib has two major interfaces, or styles of using the library:

1. An **explicit** "Axes"  or "object-oriented" interface that uses methods on a Figure or Axes object to create other Artists, and build a visualization step by step.
```python
f, ax = plt.subplots()
ax.plot(x, y)
```

2. An **implicit** "pyplot" interface that keeps track of the last Figure and Axes created, and adds Artists to the object it thinks the user wants.

```python
[...]

PR #3494, opened 2024-02-08
Support returning the actual matplotlib Figure

richarddli:
## Overview

I'm using SHAP in software (as opposed to a Jupyter notebook). One use case I would like to do is display the SHAP bar chart from an Explanation object. With Force plots, there is an HTML method that returns raw HTML. I've implemented a simple alternative to adding the HTML method everywhere.

In essence, if show=False, return the actual current matplotlib Figure. This means that the calling function can then manipulate the actual figure. This approach also means that a lot of the styling / color options that are part of the library can be made directly on the Figure itself, instead of passing through to the actual plot.

If this approach is acceptable, I'd suggest modifying all the other plot functions as well to behave similarly.

Please let me know your thoughts! cc @CloseChoice

### DIFF
--- a/shap/plots/_bar.py
+++ b/shap/plots/_bar.py
@@ -342,6 +342,8 @@ def bar(shap_values, max_display=10, order=Explanation.abs, clustering=None, clu
 
     if show:
         pl.show()
+    else:
+        return pl.gcf()
 
 
 def bar_legacy(shap_values, features=None, feature_names=None, max_display=None, show=True):


JSchoeck:
I wanted to make quite the same PR, as for example shap.plots.scatter does currently not return anything, even if show=False.

Definitely should be fixed and harmonized across the different plot types.
Some return nothing, others a plt.Figure, others a plt.Axes.

CloseChoice:
Shouldn't calling `pl.cfg()` outside of calling `bar` return the same figure? E.g.

```python
import xgboost
import shap

# train XGBoost model
X, y = shap.datasets.adult(n_points=1000)
model = xgboost.XGBClassifier().fit(X, y)

# compute SHAP values
shap_values = explainer(X)

shap.plots.bar(shap_values, show=False)
fig = plt.gcf()
fig.savefig('shap_bar_plot.png')
```
This still gives me the same figure as expected.
Hmm, at beeswarm we return the axis, often we do not return anything. Actually I do not really see an upgrade in this, especially with `show=False` might not be really intuitive.

Edit: @JSchoeck I totally agree that this should be harmonized and the solution provided by @richarddli might be a generally applicable but that needs to be discussed if it is viable for all plots. @connortann what do you think about this?

richarddli:
The Jupyter notebook implementation that I'm using in VSCode automatically invokes `pl.show()` which clears the figure before I can save it. You can argue whether or not that the VSCode behavior is correct, but I see no downside to being explicit about the semantics.

I would also suggest that this approach seems like a much cleaner separation of concerns than something like https://github.com/shap/shap/blob/5768d114eb3ae34081ff61b67f352067b59b6aba/shap/plots/_force.py#L508. `_force.py` in particular takes a bunch of args that it passes to `matplotlib` such as `figsize`. I think having each plot return a matplotlib figure (or equiv), and then a separate set of functions (if desired) to manipulate these figures, is a much cleaner implementation.

CloseChoice (approved):
Also do not see a downside to this and if this helps users, then it's fine for me.

connortann:
Thanks for the PR! This is a good step towards #3411

connortann (approved):

PR #3506, opened 2024-02-14
ENH: Add support for special characters in catboost models

### LINKED ISSUE #3475: BUG: AttributeError when using CatBoost with columns containing special characters
### Issue Description

I encountered a bug when using SHAP (version 0.44.0) with CatBoost (version 1.2.2) when some columns contain special characters. Specifically, when column names have accents, it leads to the following error:

```python
import numpy as np
import pandas as pd
from catboost import CatBoostClassifier
import shap

# Seed
np.random.seed(42)

# Simulate a dataset
data = {
        'x1': np.random.rand(1000),
        'x2': np.random.choice([0, 1], size=1000),
        'x3': np.random.choice([0, 1], size=1000),
        'x4': np.random.choice([0, 1], size=1000),
        'x5=ROMÁNIA': np.random.choice([0, 1], size=1000),
        'label': np.random.choice([0, 1], size=1000)
    }

df = pd.DataFrame(data)

# Separate features and labels
x_train = df.drop('label', axis=1)
y_train = df['label']
# Fit a CatBoostClassifier
cb_best = CatBoostClassifier(random_state=42, allow_writing_files=False)
cb_best.fit(x_train, y_train)

# Create a SHAP TreeExplainer
        cb_best,
        data=x_train,
        model_output='probability',
        feature_perturbation='interventional'
    )
```

### Traceback

```shell
File "***", line 58, in start_catboost
  File "****", line 208, in __init__
    self.expected_value = self.model.predict(self.data).mean(0)
  File "***", line 1328, in predict
    print(self.values)
```

### Expected Behavior

I expect the SHAP explainer to work correctly even when columns have special characters.

If I replace the special characters in the column names with non-accented characters (e.g., 'x5=ROMANIA' instead of 'ROMÁNIA'), the code works as expected without raising the AttributeError.

### Bug report checklist

### Installed Versions

SHAP version: 0.44.0
CatBoost version: 1.2.2
numpy version: 1.25.2
pandas: 1.5.3

CloseChoice:
## Overview

Closes #3475  

Use utf-8 encoding when loading catboost json file

## Checklist

### DIFF
--- a/tests/explainers/test_tree.py
+++ b/tests/explainers/test_tree.py
@@ -1718,25 +1718,16 @@ def test_lightgbm_interactions():
 
 def test_catboost_column_names_with_special_characters():
     # GH #3475
-    # Seed
     catboost = pytest.importorskip("catboost")
+    # Seed
     np.random.seed(42)
 
     # Simulate a dataset
-    data = {
-            'x1': np.random.rand(1000),
-            'x2': np.random.choice([0, 1], size=1000),
-            'x3': np.random.choice([0, 1], size=1000),
-            'x4': np.random.choice([0, 1], size=1000),
-            'x5=ROMÁNIA': np.random.choice([0, 1], size=1000),
-            'label': np.random.choice([0, 1], size=1000)
-        }
+    x_train = pd.DataFrame({
+            'x5=ROMÁNIA': np.random.choice([0, 1], size=10),
+        })
 
-    df = pd.DataFrame(data)
-
-    # Separate features and labels
-    x_train = df.drop('label', axis=1)
-    y_train = df['label']
+    y_train =np.random.choice([0, 1], size=10)
     # Fit a CatBoostClassifier
     cb_best = catboost.CatBoostClassifier(random_state=42, allow_writing_files=False, iterations=3, depth=1)
     cb_best.fit(x_train, y_train)
--- a/tests/explainers/test_tree.py
+++ b/tests/explainers/test_tree.py
@@ -1714,3 +1714,39 @@ def test_lightgbm_interactions():
 
     explanation = explainer(X, interactions=True)
     assert np.allclose(np.stack(explanation.values, axis=-1).sum(axis=(1, 2)) + explanation.base_values, predicted)
+
+
+def test_catboost_column_names_with_special_characters():
+    # GH #3475
+    # Seed
+    catboost = pytest.importorskip("catboost")
+    np.random.seed(42)
+
+    # Simulate a dataset
+    data = {
+            'x1': np.random.rand(1000),
+            'x2': np.random.choice([0, 1], size=1000),
+            'x3': np.random.choice([0, 1], size=1000),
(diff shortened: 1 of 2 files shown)


CloseChoice:
@connortann Maybe we can adopt the approach that is taken in #3462 for catboost aswell without saving the file. But I leave that for another PR.

tests/explainers/test_tree.py
```diff
@@ -1714,3 +1714,39 @@ def test_lightgbm_interactions():
...
+
+    # Simulate a dataset
+    data = {
+            'x1': np.random.rand(1000),
+            'x2': np.random.choice([0, 1], size=1000),
+            'x3': np.random.choice([0, 1], size=1000),
+            'x4': np.random.choice([0, 1], size=1000),
+            'x5=ROMÁNIA': np.random.choice([0, 1], size=1000),
+            'label': np.random.choice([0, 1], size=1000)
+        }
```
connortann: Nitpick: I think the unit test can be simplified somewhat to be a minimal reproducer of the error. For example, I don't think the test needs x2, x3 & x3; and perhaps it needs only say 10 rows rather than 1000?

connortann (approved):
LGTM, with an optional suggestion

tests/explainers/test_tree.py
CloseChoice: fixed. Would be great if you could merge the changes, can't do that myself due to permissions ;)

CloseChoice:
@connortann once again off-topic but I noticed that the run_notebooks job failed again for the last PR but this time due to unexpected versions. Will take a look (or just try to use the cache for dependencies) when I have the time.

connortann:
Sounds good. On the topic of caching, I can share my recollections from when we added the cache step to the tests to speed up the CI jobs, if that helps you with any future Ci stuff.

- The GitHub Cache limit wasn't big enough to allow us to cache the full set of dependencies for every version of python that we tested against. So, we restricted the scope cache to just a few of the larger packages. I think the GH cache size limit increases now and then, so over time we might be able to cache more.
- The speedup we got from caching different packages was hard to predict: sometimes it was useful, sometimes it was actually slower. It was worthwhile doing experiments to measure the speedup, averaging over multiple runs to account for noise.
- We had to be quite careful with the cache key, to ensure the cache was only restored in the exact same environment as the one in which it was created. Some tricky issues boiled down to subtle differences between those envs, such as different minor versions of the python interpreter.

CloseChoice:
Thanks for the information. I thought that was easier/more useful. Will keep that in mind when I implement the caching

connortann:
> the run_notebooks job failed again for the last PR but this time due to unexpected versions. Will take a look (or just try to use the cache for dependencies) 

Another thought: I don't think caching will help as a _solution_ to any dependency issues, per-se. Conceptually the cache allows us to take short-cuts in what should be already a complete deterministic end-to-end workflow; so the cache job shouldn't change the behaviour of the workflow other than being faster.

PR #3514, opened 2024-02-19
Do not overwrite default warning filter or formatter

### LINKED ISSUE #3287: ENH: Do not overwrite global warning formatter
The global warnings formatter is overwritten here:

https://github.com/shap/shap/blob/e51a87db4c9355968524dd6f788f9ec781e81924/shap/explainers/_tree.py#L23

This pattern of overwriting the builtin implementation is actually recommended in the [docs](https://docs.python.org/3/library/warnings.html) of the builtin `warnings` module. However, this kind of import-time side-effect seems inappropriate for a package, as it will interfere with other packages' use of warnings.

A similar case caused issues in the pypdf2 package: https://github.com/py-pdf/pypdf/issues/67

connortann:
## Overview

Closes #3287

- Removes modifications of the default warning formatter
- Removes the global supression of numba deprecation warnings

## Discussion

As per the issue above, the unexpected import-time side-effects can be surprising and cause nefarious bugs. As a general rule, importable libraries should avoid modifying or monkeypatching other libraries.

One consequence of this change is that the warnings produced by SHAP will be the default verbose formatting, rather than the prettier "message-only" format:

image

However, whilst I admittedly do prefer the concise formatting, I don't think this preference for formatting has anything to do with shap specifically, so it should be out-of-scope for the shap package to fiddle with generic python behaviour like this. Users are still able to modify the handling of warnings as they see fit.

### DIFF
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -26,8 +26,6 @@ from ..utils._legacy import DenseData
 from ._explainer import Explainer
 from .other._ubjson import decode_ubjson_buffer
 
-warnings.formatwarning = lambda msg, *args, **kwargs: str(msg) + '\n' # ignore everything except the message
-
 try:
     from .. import _cext
 except ImportError as e:
--- a/shap/maskers/_image.py
+++ b/shap/maskers/_image.py
@@ -1,11 +1,9 @@
 # TODO: heapq in numba does not yet support Typed Lists so we can move to them yet...
 import heapq
-import warnings
 
 import numba.typed
 import numpy as np
 from numba import njit
-from numba.core.errors import NumbaPendingDeprecationWarning
 
 from .._serializable import Deserializer, Serializer
 from ..utils import assert_import, record_import_error, safe_isinstance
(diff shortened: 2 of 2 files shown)


connortann:
As an alternative, we could preserve the current warning formatting with a function like this that tracks and restores the original global state. This would involve replacing all instances of `warnings.warn`. 

```python
def pretty_warn(msg: str) -> None:
    """Raise a UserWarning with a custom formatter that just shows the message"""
    previous_formatter = warnings.formatwarning
    warnings.formatwarning = lambda msg, *args, **kwargs: str(msg) + '\n'
    warnings.warn(msg)
    # Restore previous global state
    warnings.formatwarning = previous_formatter
```

I'm actually inclined to just leave the default warning formatter in place for simplicity. Open to suggestions. Whatever we do, it seems like we should not permanently override the global state.

CloseChoice (approved):
very nice! LGTM

connortann:
Just to clarify, @CloseChoice do you have any preferences regarding removing the formatting tweak entirely vs implementing the "pretty_warn" func?

CloseChoice:
I think removing it entirely as you did is totally fine. I wouldn't spend time to find a solution that keeps the old formatting AND doesn't overwrite the global warnings. If you have a quick solution please go ahead, but I would not spend even 15 minutes on that, tbh. If the formatting disturbs us in any way we can still open another issue and find a solution for that.

PR #3517, opened 2024-02-23
MAINT: update deprecation warnings for l1_reg="auto" and auto_size_plot=False

### LINKED ISSUE #3507: [Meta-issue] deprecation tracker for upcoming releases
This is a meta-issue to track any deprecation cycles within the shap package.

Maintainers, feel free to edit this tracker as necessary,

## Recommendations for which warning to use:

For deprecated parameters in e.g. plotting functions, we probably want `FutureWarning`.

Excerpt from [PEP0565](https://peps.python.org/pep-0565/):

> - `PendingDeprecationWarning`: hidden by default for all code. The intended audience is Python developers that take an active interest in ensuring the future compatibility of their software (e.g. professional Python application developers with specific support obligations).
> - `DeprecationWarning`: reported by default for code that runs directly in the `__main__` module (as such code is considered relatively unlikely to have a dedicated test suite), but hidden by default for code in other modules. The intended audience is Python developers that are at risk of upgrades to their dependencies (including upgrades to Python itself) breaking their software (e.g. developers using Python to script environments where someone else is in control of the timing of dependency upgrades).
> - `FutureWarning`: reported by default for all code. The intended audience is users of applications written in Python, rather than other Python developers (e.g. warning about use of a deprecated setting in a configuration file format).

## Scheduled for removal

Items that raise a DeprecationWarning due to be removed in an specific release:

### 0.45.0

- [x] #3511
- [x] #3340

Parameters which were renamed:
```
feature_dependence -> feature_perturbation
independent -> interventional
correlation -> correlation_dependent
logloss -> log_loss
margin -> raw
```

Begin raising warnings for:

- [x] Deprecate KernelExplainer.shap_values: `l1_reg="auto"` 
- [x] Deprecate Beeswarm / Violin plot `auto_size_plot`
- Covered by #3517

### 0.46.0

To be removed:

- [x] Remove Beeswarm / Violin plot `auto_size_plot`
- [x] Deprecate unused `title` parameter of `shap.plots._violin.py`
- Covered by #3717

### 0.47.0

- [x] KernelExplainer.shap_values: change default behaviour of `l1_reg` away from `"auto"` 
  - #3858
- [x] Raise warning for `shap.bar_plot`
  - #3739
- [x] Change default feature_peturbation to "auto"
  - #2624
- [x] deprecate ` approximate` in `TreeExplainer` initialisation
  - #3834

Remove unused / dead code:

- [x] `shap/utils/_masked_models.py:_build_delta_masked_inputs()`
- [x] `shap/explainers/_explainer.py:Explainer._compute_main_effects()`
  - #3856
 
### 0.48.0

To be removed:

- [ ] `shap.plots._violin.py`: remove unused `title` parameter

### 0.49.0

- [ ] Update `shap.bar_plot` to use new plotting API, delete `shap.plots._bar.bar_legacy`
- [ ] Raise error if `feature_peturbation="interventional"` and `data=None`, as per #2624
- [ ] remove ` approximate` in `TreeExplainer` initialisation, as per #3834
- [ ] KernelExplainer.shap_values: removed deprecated argument `l1_reg="auto"` 

### At some point, at our leisure

Remove unused / dead code:
- [ ] `shap/plots/_text.py:text_old()
   - We should remove this once we have overhauled the rest of `_text.py`, got the test coverage up and looked over other functions in that module e.g. `_text.saliency_plot()` and `_text.heatmap()`.

CloseChoice:
## Overview

Supports #3507  

 - deprecate `l1_reg` parameter for kernelexplainer
 - update deprecation warning for auto_size_plot (though this might even be removed directly if we have a breaking release 0.45.0)

## Checklist


CloseChoice:
@connortann: I saw that you triggered the pipelines again. It is pretty weird to have the segmentation fault appear twice at the same test for macos. Could you please trigger master? Are there some version updates in the background that might cause this? 

Edit: don't think we need a master run because there is already one showing the exact same error: https://github.com/shap/shap/actions/runs/8017606141/job/21902418720#step:7:116

shap/explainers/_kernel.py
```diff
@@ -599,11 +598,12 @@ def solve(self, fraction_evaluated, dim):
...
-        #     warnings.warn(
-        #         "l1_reg=\"auto\" is deprecated and in the next version (v0.29) the behavior will change from a " \
-        #         "conditional use of AIC to simply \"num_features(10)\"!"
-        #     )
+        if self.l1_reg == "auto":
+            warnings.warn(
+                "l1_reg=\"auto\" is deprecated and in a future version the behavior will change from a " \
+                "conditional use of AIC to simply \"num_features(10)\"!",
+                DeprecationWarning
+            )
```
connortann, line 606: If the explainer is called with the `__call__` interface, I don't think it's possible for users to control the value of `l1_reg`, so the warning will be impossible to fix. That's not necessarily an issue, but it may be a little annoying for users.

Would it be better to simply change the behaviour of "auto" to the desired new behaviour, as a breaking change? The fact that the regularisation behaviour will change in a future version has been declared in the docstring for several years now, so I think enough time has passed.

I'd be happy either way, just a thought.

connortann:
Suggest changing in the description to "supports #3507" rather than "Closes", as we still want to use that meta-issue to track the lifecycle of these warnings

shap/explainers/_kernel.py
CloseChoice: I think in that case I would just set l1_reg directly to "num_features(10)" when going through the `__call__` interface. I would prefer this since it is more explicit.

connortann: I think the `__call__` and `shap_values` interfaces should have the same default behaviour, so if we update one we might as well change the other. May I suggest changing the default behaviour as part of this PR?

CloseChoice: I would rather change the behaviour after some time so that users can adapt because changing defaults is a breaking change

connortann: What is the alternative?
1. If the default of `__call__` is changed but, `shap_values` is not changed, they will have different behaviour which is probably rather unexpected and potentially confusing
2. If the defaults of both are unchanged, then users will get a DeprecationWarning with the "call" interface with no way to fix it

Maybe adding a parameter `l1_reg` to "call" so that users can opt-in to the new behaviour to fix the warning?

CloseChoice: I guess that is a good approach and raise a warning that says something like "The default value for `l1_reg=auto` is deprecated and will change in the future to `num_features(10)`. Please provide on of the following values ... to silence this warning."?

connortann: Yes, let's do that.

connortann: I'll merge this in as-is, and add the parameter in a follow-up PR...

connortann (approved):
LGTM!

PR #3518, opened 2024-02-23
fix macos segmentation fault

CloseChoice:
## Overview

Supports #3524

 - pin pytorch version on macos to avoid segmentation fault

## Checklist


connortann:
Thanks for looking it to this. Have you been able to make any headway into determining what caused has caused the MacOS tests to suddenly start failing?

It looks like the failure on the latest commit has progressed to something related to lightgbm, which seems like progress. Might I suggest restricting the macos tests to `extras: "test-core"` , to see if at least the core shap tests pass?

CloseChoice:
> Thanks for looking it to this. Have you been able to make any headway into determining what caused has caused the MacOS tests to suddenly start failing?
> 
> It looks like the failure on the latest commit has progressed to something related to lightgbm, which seems like progress. Might I suggest restricting the macos tests to `extras: "test-core"` , to see if at least the core shap tests pass?

My commits don't feel like progress but more like guessing. I first thought that it is a new python version (since the last successful run was executed with python 3.11.7) that caused the problem on macos and there was a historical parallel in python2.x. Am now to reverting to macos11 but this shouldn't be permanent solution. Good idea with the adjustment of extras, will try it

connortann:
Here's an issue for us to discuss / debug: #3524

I've proposed that temporary hotfix over on #3525 to fix the CI so that other PRs are unblocked. However, it would be good to get to the root cause of the issue. Thanks again for taking this one on @CloseChoice !

connortann:
Could also try the MacOS tests on `3.12` ?

CloseChoice:
Just put this here since it's a good reference:

Here are a couple notes:
The last successful run of the macos pipeline on master was this: https://github.com/shap/shap/actions/runs/7972563240/job/21765144274.
I debugged the macos pipeline using https://github.com/mxschmitt/action-tmate and found that the segmentation faults happen in the pytorch tests, in the lines w[here](https://github.com/shap/shap/blob/master/tests/explainers/test_deep.py#L384) one calls the model on data, e.g. here. From the successful run I found that we used torch version 2.2.0 there instead of 2.2.1. Will check if it works if I pin the version.

CloseChoice:
Should probably file an issue in the torch repo, but will leave that for tomorrow

CloseChoice:
Opened an issue on pytorch for this: https://github.com/pytorch/pytorch/issues/121101

.github/workflows/run_tests.yml
```diff
@@ -35,7 +35,7 @@ jobs:
             extras: "test"
           - os: macos-latest
             python-version: "3.11"
-            extras: "test-core"  # FIXME: the full set of tests should pass
+            extras: "test"  # FIXME: the full set of tests should pass
```
connortann:
```suggestion
            extras: "test"
```

connortann:
Great job for figuring this out! This looks good to me. There is "WIP" in the title - is this ready to merge?

After this is merged, I think we should probably keep #3524 open until things are working again with the latest release of pytorch.

connortann (approved):

CloseChoice:
> Great job for figuring this out! This looks good to me. There is "WIP" in the title - is this ready to merge?
> 
> After this is merged, I think we should probably keep #3524 open until things are working again with the latest release of pytorch.

this is ready to merge

PR #3523, opened 2024-02-26
ENH: Prototype new plotting API for bar plot, accept and return axes

connortann:
## Overview

Supports #3411
Supports #3036

A prototype new standard API interface for the plots, using `plots.bar` as an example:
- Add an `ax` parameter to accept an existing matplotlib axes
- Return the axes (if show=False)
- Internal refactor: use the explicit matplotlib interface rather than the imperative

## Discussion

I do wonder if the `show` argument is really needed. In jupyter notebooks, I believe `plt.show()` is called anyway after each cell, so this seems a little redundant. However, it does have the nice effect of returning `None` which avoids cluttering the output with the `__repr__` of the axes object.

I think changing the return type to `ax` from `fig` is technically a breaking change; but this was only a fairly recent addition, so it's _probably_ not too significant.

## Checklist

From #3411

- [x] Uses explicit object-oriented matplotlib interface
- [x] Accepts an optional `ax` parameter
- [x] Suitable tests using pytest-mpl
- [x] Docstring is complete and up-to-date
- [x] Notebook in the "API Examples" section runs and is up-to-date


connortann:
@DanGolding you may wish to take a skim through this PR if you get a minute

shap/plots/_bar.py
```diff
@@ -273,33 +278,33 @@ def bar(shap_values, max_display=10, order=Explanation.abs, clustering=None, clu
...
+    ax.tick_params('x', labelsize=11)
 
-    xmin,xmax = pl.gca().get_xlim()
-    ymin,ymax = pl.gca().get_ylim()
+    xmin,xmax = ax.get_xlim()
+    ymin,ymax = ax.get_ylim()
 
     if negative_values_present:
-        pl.gca().set_xlim(xmin - (xmax-xmin)*0.05, xmax + (xmax-xmin)*0.05)
+        ax.set_xlim(xmin - (xmax-xmin)*0.05, xmax + (xmax-xmin)*0.05)
```
DanGolding: nit: could be marginally clearer to refactor `(xmax-xmin)*0.05` out as a variable called padding or buffer maybe?

connortann: Good shout. Incidentally there is probably a lot more we could refactor in this function; but I'll aim to keep the scope of this PR fairly small - demonstrating the main change we want to make to the plotting API.

DanGolding: yes, the function is super long, would be good to break it up in the future maybe.

Also, more controversial, but why import matplotlib as pl instead of plt (which I think is pretty standard)?

DanGolding:
> I do wonder if the `show` argument is really needed.

It seems unnecessary to me, unless this has become an established pattern in other libraries?

> ... avoids cluttering the output with the __repr__ of the axes object

users can always end the line with a `;` if they want to suppress the output being printed

connortann:
I added a proposed "checklist" for things that we wish to overhaul for each plot to #3411. I'll work through that for this PR as an example - in particular, we need more tests and to update the API example notebook.

shap/plots/_bar.py
```diff
@@ -97,16 +110,21 @@ def bar(shap_values, max_display=10, order=Explanation.abs, clustering=None, clu
...
     if len(values[0]) == 0:
-        raise Exception("The passed Explanation is empty! (so there is nothing to plot)")
+        raise ValueError("The passed Explanation is empty! (so there is nothing to plot)")
 
-    # we show the data on auto only when there are no transforms
+    # we show the data on auto only when there are no transforms (excluding getitem calls)
     if show_data == "auto":
-        show_data = len(op_history) == 0
+        transforms = [t for t in op_history if t.get("name") != "__getitem__"]
+        show_data = len(transforms) == 0
```
connortann, line 132: Here is one functional change that I needed to make in order to reproduce more closely the plots in the API example notebook.

In the API example notebook, many plots have show_data=True. This adds the small grey numbers on the left axis. It looks like this behaviour was then changed at some point to only show data values when there are no transforms applied.

However, I think `show_data` still makes sense as `True` for a single explanation when the only operation applied has been `__getitem__`.

### Before (show_data = False)
<details>

image

</details>

### After (show_data = True)
<details>

image

</details>

connortann:
Whilst we're at it, I updated the API examples notebook. Hopefully it should be executable on CI now. Reducing the sample size of the toy dataset from 32k to 2k rows sped up the runtime significantly, without changing the overall look or complexity of the example plots.

shap/plots/_bar.py
```diff
@@ -20,31 +20,49 @@
 # TODO: improve the bar chart to look better like the waterfall plot with numbers inside the bars when they fit
 # TODO: Have the Explanation object track enough data so that we can tell (and so show) how many instances are in each cohort
 def bar(shap_values, max_display=10, order=Explanation.abs, clustering=None, clustering_cutoff=0.5,
-        merge_cohorts=False, show_data="auto", show=True):
```
connortann, line 23: This `merge_cohorts` parameter was unused and undocumented, so I've removed it.

CloseChoice: technically this makes it a breaking change. I added the BREAKING label

shap/plots/_bar.py
```diff
@@ -20,31 +20,49 @@
...
 
+        Passing a dictionary of Explanation objects will create a multiple-bar
+        plot with one bar type for each of the cohorts represented by the
+        explanation objects.
     max_display : int
         How many top features to include in the bar plot (default is 10).
-
+    order : OpChain or numpy.ndarray
+        A function that returns a sort ordering given a matrix of SHAP values
+        and an axis, or a direct sample ordering given as an ``numpy.ndarray``.
```
CloseChoice:
```suggestion
        and an axis, or a direct sample ordering given as a ``numpy.ndarray``.
```

shap/plots/_bar.py
```diff
@@ -20,31 +20,49 @@
...
+
+        By default, take the absolute value.
+    clustering: Optional np.ndarray
```
CloseChoice:
```suggestion
    clustering: np.ndarray or None
```

CloseChoice (approved):
LGTM. Just a couple doc nitpicks from me

connortann:
Fab - I applied those changes, and shall merge shortly.

Regarding the `show` param: I agree we could theoretically remove it. I'm inclined to leave that for later, to avoid making too many breaking changes.

PR #3527, opened 2024-02-27
ENH: Specify default clustering as "cosine" if y is None

connortann:
## Overview

Slight tweak to the behaviour of h_clust, so that a valid metric is chosen even if y=None.

This came up in writing a unit test for the bar plots. Previously if y=None, hclust would throw an exception via scipy: `unrecognised metric "auto"`

### DIFF
--- a/shap/utils/_clustering.py
+++ b/shap/utils/_clustering.py
@@ -146,6 +146,8 @@ def hclust(X, y=None, linkage="single", metric="auto", random_state=0):
     if metric == "auto":
         if y is not None:
             metric = "xgboost_distances_r2"
+        else:
+            metric = "cosine"
 
     # build the distance matrix
     if metric == "xgboost_distances_r2":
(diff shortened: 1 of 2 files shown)


shap/utils/_clustering.py
```diff
@@ -146,6 +146,8 @@ def hclust(X, y=None, linkage="single", metric="auto", random_state=0):
     if metric == "auto":
         if y is not None:
             metric = "xgboost_distances_r2"
+        else:
+            metric = "cosine"
```
CloseChoice, line 180: can we document this behaviour somewhere? Maybe in the (yet non-existent) docstring of the function

CloseChoice (changes requested):

shap/utils/_clustering.py
connortann: Good idea, done.

CloseChoice (approved):
Perfect, nice work. Let's merge this

PR #3533, opened 2024-03-01
ENH: 2x import time speedup via lazy importing of pytorch

### LINKED ISSUE #3528: ENH: Faster import performance
### Problem Description

The package import time is rather slow- over 2s on my machine. We could profile the import (e.g. with tuna) to identify any bottlenecks, and potentially improve performance.

### Alternative Solutions

We could copy scipy's approach and import subpackages lazily, with some `getattr` magic. This would probably be a bit more complex and nonstandard.

https://peps.python.org/pep-0562/

### Additional Context

_No response_

### Feature request checklist

connortann:
## Overview

Closes #3528

Implements lazy importing for pytorch

## Timing

Created a profile in a local environment with the dev dependencies installed:

```bash
python -X importtime -c "import shap" > profile.log
```

### Before: 2.293s

Visualising with [tuna](https://github.com/nschloe/tuna):

image

### After: 1.122s

image

## Discussion

At present pytorch is imported lazily in _some_ of the modules, but not all, so there is no net speedup. This PR implements lazy loading of torch everywhere it is used.

Presently some lazy importing is done via global variables, such as:

```python
foo = None

def bar():
    if foo is None:
        import foo
```

This seems needlessly complex, as `import foo` will be a simple lookup in sys.modules if foo has already been imported. So, I propose just importing the relevant modules at the top of each function in which it is used:

```python
def bar():
   import foo
```

### LazyLoader

The standard library has [LazyLoader](https://docs.python.org/3/library/importlib.html#importlib.util.LazyLoader), which can be used to postpone the execution of the loader of a module until the module has an attribute accessed.

However, this has some downsides and is discouraged unless really necessary:

>  For projects where startup time is not essential then use of this class is heavily discouraged due to error messages created during loading being postponed and thus occurring out of context.

There are a significant number of open issues relating to DeepExplainer, so I think it is wise to keep complexity to an absolute minimum and avoid complex approaches like this.

### Lazy subpackages

We could copy scipy's approach and import entire subpackages of shap lazily, with some `getattr` magic. This would probably be a bit more complex to implement. However, it could offer significant speedups. I don't think it's a priority for now.

https://peps.python.org/pep-0562/

### DIFF
--- a/shap/explainers/_deep/deep_pytorch.py
+++ b/shap/explainers/_deep/deep_pytorch.py
@@ -6,18 +6,13 @@ from packaging import version
 from .._explainer import Explainer
 from .deep_utils import _check_additivity
 
-torch = None
-
 
 class PyTorchDeep(Explainer):
 
     def __init__(self, model, data):
-        # try and import pytorch
-        global torch
-        if torch is None:
-            import torch
-            if version.parse(torch.__version__) < version.parse("0.4"):
-                warnings.warn("Your PyTorch version is older than 0.4 and not supported.")
+        import torch
+        if version.parse(torch.__version__) < version.parse("0.4"):
+            warnings.warn("Your PyTorch version is older than 0.4 and not supported.")
 
         # check if we have multiple inputs
         self.multi_input = False
@@ -102,6 +97,7 @@ class PyTorchDeep(Explainer):
                     pass
 
     def gradient(self, idx, inputs):
+        import torch
         self.model.zero_grad()
         X = [x.requires_grad_() for x in inputs]
         outputs = self.model(*X)
@@ -133,6 +129,7 @@ class PyTorchDeep(Explainer):
             return grads
 
     def shap_values(self, X, ranked_outputs=None, output_rank_order="max", check_additivity=True):
+        import torch
         # X ~ self.model_input
         # X_data ~ self.data
 
@@ -247,6 +244,7 @@ def add_interim_values(module, input, output):
     """The forward hook used to save interim tensors, detached
     from the graph. Used to calculate the multipliers
     """
+    import torch
     try:
         del module.x
     except AttributeError:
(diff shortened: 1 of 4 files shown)


connortann:
@DanGolding FYI if you'd like to review?

shap/explainers/_deep/deep_pytorch.py
```diff
@@ -102,6 +97,7 @@ def remove_attributes(self, model):
                     pass
 
     def gradient(self, idx, inputs):
+        import torch
```
CloseChoice, line 100: We need the imports here because a DeepExplainer can be unpickled and the a call to gradient can be made, right?

CloseChoice (approved):
LGTM

shap/explainers/_deep/deep_pytorch.py
connortann: I just added the import here as `torch` functions are used further down in this function. I'm not aware of any pickle-related reasons for the torch import, but please do correct me if I'm wrong.

CloseChoice: No, the thing is, that I suspected that doing an import in the __init__ would be enough (ok, linters will complain). But then we might run into the pickle errors

connortann: If we removed the import from the gradient method, wouldn't this line fail as `torch` is not defined within the local scope?

https://github.com/shap/shap/blob/0b73b8a41d005f120d909220d865a490cb002311/shap/explainers/_deep/deep_pytorch.py#L109

CloseChoice: Well, my thinking was that people first creating the model, thus importing pytorch by calling the `__init__` method. Then calling `gradient` wouldn't be an issue at all. But as I mentioned there might be cases where we do not call init but gradient, e.g. the unpickle and retrain case. Nevermind I was just thinking loud. 

This looks totally reasonable.

CloseChoice:
I think this is ready to be merged.

shap/explainers/_deep/deep_pytorch.py
connortann: I suppose we're on a bit of a tangent, but for what it's worth I think even if `__init__` is called first, the import would still needed again in `gradient` for the function to work as the module needs to be bound to a variable called `torch` in the local scope.

In any case I'll merge and we can discuss later, as I may well be missing something.

PR #3536, opened 2024-03-01
Fix text masking with certain tokenizers

costrau:
## Overview

Fixed a bug in text masker for a tokenizer where keep_prefix is >0 but keep_sufix is 0. In this case, the full mask is overwritten to True. This PR contains the (small) fix and some tests.

## Checklist

### DIFF
--- a/tests/maskers/test_text.py
+++ b/tests/maskers/test_text.py
@@ -95,7 +95,6 @@ def test_sentencepiece_tokenizer_output():
     # since we expect output wrapped in a tuple hence the indexing [0][0] to extract the string
     assert sentencepiece_tokenizer_output_processed[0][0] == expected_sentencepiece_tokenizer_output_processed
 
-@pytest.mark.skip(reason="fails on travis and I don't know why yet...Ryan might need to take a look since this API will change soon anyway")
 def test_keep_prefix_suffix_tokenizer_parsing():
     """ Checks parsed keep prefix and keep suffix for different tokenizers.
     """
--- a/tests/maskers/test_text.py
+++ b/tests/maskers/test_text.py
@@ -105,18 +105,22 @@ def test_keep_prefix_suffix_tokenizer_parsing():
     tokenizer_mt = AutoTokenizer.from_pretrained("Helsinki-NLP/opus-mt-en-es")
     tokenizer_gpt = AutoTokenizer.from_pretrained("gpt2")
     tokenizer_bart = AutoTokenizer.from_pretrained("sshleifer/distilbart-xsum-12-6")
+    tokenizer_mistral = AutoTokenizer.from_pretrained("mistralai/Mistral-7B-v0.1")
 
     masker_mt = shap.maskers.Text(tokenizer_mt)
     masker_gpt = shap.maskers.Text(tokenizer_gpt)
     masker_bart = shap.maskers.Text(tokenizer_bart)
+    masker_mistral = shap.maskers.Text(tokenizer_mistral)
 
     masker_mt_expected_keep_prefix, masker_mt_expected_keep_suffix = 0, 1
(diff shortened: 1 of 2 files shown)


tests/maskers/test_text.py
```diff
@@ -140,19 +147,25 @@ def test_text_infill_with_collapse_mask_token():
     mask_ex4 = np.array([False, False, False, False, False, False, False, False, False])
 
     text_infilled_ex1 = masker(mask_ex1, s)[0][0]
+    text_infilled_ex1_mist = masker_mistral(np.append(True, mask_ex1), s)[0][0]
```
CloseChoice, line 149: Just for me to clarify: we need to append True from the left because mistralai has `self.keep_prefix == 1` correct?

tests/maskers/test_text.py
```diff
@@ -105,18 +105,22 @@ def test_keep_prefix_suffix_tokenizer_parsing():
     tokenizer_mt = AutoTokenizer.from_pretrained("Helsinki-NLP/opus-mt-en-es")
     tokenizer_gpt = AutoTokenizer.from_pretrained("gpt2")
     tokenizer_bart = AutoTokenizer.from_pretrained("sshleifer/distilbart-xsum-12-6")
+    tokenizer_mistral = AutoTokenizer.from_pretrained("mistralai/Mistral-7B-v0.1")
```
CloseChoice, line 107: Thanks for adding this to the test. This test  is skipped though, could we try to let it run in the CI and check whether it works? If not I am totally willing to revert to the current state, just would be great to increase the test coverage if possible.

CloseChoice (changes requested):
Thanks a lot for the PR. This looks great. Thanks for catching this untested case.

I would kindly ask you to try to run the skipped test, if that works perfect, if not then let's just skip it. Totally happy with this PR, let's get this merged quickly

CloseChoice:
@connortann could you please let the workflows run?

tests/maskers/test_text.py
```diff
@@ -128,6 +132,9 @@ def test_text_infill_with_collapse_mask_token():
     tokenizer = AutoTokenizer.from_pretrained("gpt2")
     masker = shap.maskers.Text(tokenizer, mask_token='...', collapse_mask_token=True)
 
+    tokenizer_mistral = AutoTokenizer.from_pretrained("mistralai/Mistral-7B-v0.1")
```
connortann, line 134: This is a very large model. We should keep an eye on CI times to ensure the unit tests are not overly compute or network intensive.

Are there other smaller models that we could use instead for the purposes of unit tests?

connortann: Ah - the tokenizer actually seems to be reasonably small and loads within ~2s, so I think this is fine.

tests/maskers/test_text.py
costrau: Yes, exactly. In the text masker special tokens are replaced by "", which is why it does not appear in the output.

tests/maskers/test_text.py
costrau: Locally the test works as expected.

tests/maskers/test_text.py
costrau: Yeah, that's why I didn't investigate any further. If you are fine with it the way it is, I'd leave it with mistral.

CloseChoice: Am totally fine with it like that.

tests/maskers/test_text.py
CloseChoice: Yes, I also found that so I hope that it also runs in the CI.

CloseChoice:
From my side the code looks good. Just waiting for the CI checks to pass. @connortann gentle push ;)

connortann (approved):

connortann:
Fantastic, thanks for the PR @costrau !

PR #3538, opened 2024-03-03
Dummy to reproduce seg fault

CloseChoice:
## Overview

This PR is only there to find a reproducible example for the segmentation fault issue on macos, so that we can open an issue in the torch repository

## Checklist


CloseChoice:
no longer needed, since a reproducibe example was found and an issue was opened. See here: https://github.com/pytorch/pytorch/issues/121101

PR #3541, opened 2024-03-05
Remove deprecation warning for permutation explainer

connortann:
Supports #3507

Removes the deprecation warning for the permutation explainer


CloseChoice (approved):
LGTM

PR #3542, opened 2024-03-05
Fix issue with KernelExplainer when explaining tensorflow models

### LINKED ISSUE #3492: BUG: Problem of dimension during KernelExplainer
### Issue Description

tensorflow.python.framework.errors_impl.InvalidArgumentError: In[0] and In[1] has different ndims: [144,1000] vs. [1000] [Op:MatMul]
Problem of dimension in the line w = np.linalg.solve(X.T @ WX, WX.T @ y) of the solve function in shap>explainers>kernel.
I have no issue with the previous solve function in shap version : '0.41.0'.

```python
import shap

explainer = shap.KernelExplainer(model, data, nsamples, silent=True)

explainer.shap_values(data)
```

### Traceback

```shell
shap/explainers/_kernel.py", line 747, in solve
    w = np.linalg.solve(X.T @ WX, WX.T @ y)
tensorflow.python.framework.errors_impl.InvalidArgumentError: In[0] and In[1] has different ndims: [144,1000] vs. [1000] [Op:MatMul]
```

### Expected Behavior

_No response_

### Bug report checklist

### Installed Versions

0.44.1

connortann:
## Overview

Closes #3492

Casts `y` to a numpy array, to fix a bug where matrix multiplication operator `@` does not function as expected when a tensorflow tensor is passed.


connortann:
@CloseChoice  FYI, I'd appreciate your thoughts on this one. I added a test for using a tensorflow model with KernelExplainer, to fix an issue with matrix multiplication as described in https://github.com/shap/shap/issues/3492#issuecomment-1978011002

The fix which passes with python 3.12 but not earlier versions, as we get a different error relating to trying to transpose a tensor:

```
/opt/hostedtoolcache/Python/3.11.8/x64/lib/python3.11/site-packages/shap/explainers/_kernel.py:124: in __init__
    self.fnull = np.sum((model_null.T * self.data.weights).T, 0)
E       AttributeError: SymbolicTensor object has no attribute 'T'. 
```

https://github.com/shap/shap/blob/dbb275978abe27def1afe73fdd097d071c7fbe38/shap/explainers/_kernel.py#L122-L124

I don't use tensorflow that often - it looks like we get different behaviour of different versions. It looks like we sometimes get `EagerTensor` and sometimes get `SymbolicTensor`. Do you know a robust way to handle these?

connortann:
Possibly relevant: https://discuss.tensorflow.org/t/how-to-convert-a-tensor-to-a-numpy-array-without-enabling-the-run-eagerly-flag-in-keras/4180

Tagging @lorentzenchr in case you have suggestions, given your experience with this area

CloseChoice:
> @CloseChoice FYI, I'd appreciate your thoughts on this one. I added a test for using a tensorflow model with KernelExplainer, to fix an issue with matrix multiplication as described in [#3492 (comment)](https://github.com/shap/shap/issues/3492#issuecomment-1978011002)
> 
> The fix which passes with python 3.12 but not earlier versions, as we get a different error relating to trying to transpose a tensor:
> 
> ```
> /opt/hostedtoolcache/Python/3.11.8/x64/lib/python3.11/site-packages/shap/explainers/_kernel.py:124: in __init__
>     self.fnull = np.sum((model_null.T * self.data.weights).T, 0)
> E       AttributeError: SymbolicTensor object has no attribute 'T'. 
> ```
> 
> https://github.com/shap/shap/blob/dbb275978abe27def1afe73fdd097d071c7fbe38/shap/explainers/_kernel.py#L122-L124
> 
> I don't use tensorflow that often - it looks like we get different behaviour of different versions. It looks like we sometimes get `EagerTensor` and sometimes get `SymbolicTensor`. Do you know a robust way to handle these?

The difference here (I guess) is not the versions but the order in which tests are executed. If you just run the test only by itself it passes at my computer on python3.10 and python3.11 but adding `tf.compat.v1.disable_eager_execution()` to the test reproduces the issue. So I suspect that on the other runs this line was executed in some previous test. To be honest, it worries me quite a bit that tests seem not really reproducible across runs with this compatibility mode.

Fixing this is a mess and this is one of the caveats of tensorflow. I got a solution and file a PR into your branch, you can find it [here](https://github.com/shap/shap/pull/3545)

Whether you want to have the compatibility in the test is totally up to you. One the one hand I think testing the latest behaviour of tensorflow makes the most sense on the other hand the tests are better reproducible if we enable compatibility mode.

connortann:
Thanks for the help! I'll merge your PR into this branch.

> Whether you want to have the compatibility in the test is totally up to you. One the one hand I think testing the latest behaviour of tensorflow makes the most sense on the other hand the tests are better reproducible if we enable compatibility mode.

I'm not sure which is better. I wonder if we can parameterise the tests to run with either mode, given that we don't know what users will have enabled?

CloseChoice:
> ot sure

This should be possible with something like

```python
@pytest.mark.parametrize("enable_disable", [tf.compat.v1.disable_eager_execution, tf.compat.v1.disable_enable_execution])
def test_stuff(enable_disable):
    enable_disable()
    ...
```

Haven't tested this but I see no reason why it shouldn't work

lorentzenchr:
@connortann It seems like a general design decision of shap: what do you support. Is a model allowed to return tensorflow objects, pytorch tensors, numpy arrays, dataframes.
One option is to try to conform to the python array api, but solve is not (yet) part of it, so it doesn’t help.
I would avoid too much model library specific code and clearly state the allowed API of a model passed to the explainers.

CloseChoice (approved):
LGTM, nice work

CloseChoice:
> @connortann It seems like a general design decision of shap: what do you support. Is a model allowed to return tensorflow objects, pytorch tensors, numpy arrays, dataframes. One option is to try to conform to the python array api, but solve is not (yet) part of it, so it doesn’t help. I would avoid too much model library specific code and clearly state the allowed API of a model passed to the explainers.

We support specific libraries (sklearn, tensorflow, pytorch, ngboost, etc.) and whatever the supported models of the library returns should be handled by the corresponding shap explainer (e.g. DMatrix is supported by TreeExplainer). Internally we convert these to numpy arrays, so I would only support the library specific objects if an easy conversion to numpy is available. 

I guess the reason why you bring this up in here is that we convert symbolic tensors to numpy arrays. We won't get around that for the tensorflow library because the shap values are returned as `tf.tensor` and due to change from `tf1` to `tf2` from `SymbolicTensor` to `EagerTensor` we need to handle both (especially since a lot of people are still using the compatibility mode of tensorflow which causes the model to return `SymbolicTensor`s). We could think of dropping support for `tf1` but since shap has A LOT of problems with tensorflow models (see the open issues with the deep explainer label) I think that might be a bit premature. I personally would like to get shap into a state where we can explain every model that we already could at some point back in time (e.g. LSTM, models with embedding, etc.).

 @lorentzenchr Did I understand the context of your question correctly or did I miss something? Feel free to elaborate on this in here. Really appreciate your thoughts on that.

PR #3543, opened 2024-03-05
Enhance dataset functions by adding docstrings and removing unused `d…

LakshmanKishore:
…isplay` parameters

## Overview

This pull request enhances the dataset functions in the `shap/datasets.py` file by adding informative docstrings and removing unused `display` parameters where applicable. The goal is to improve the clarity and documentation of these functions.

1. Added detailed docstrings to the dataset functions to provide users with clear information about each dataset and its structure.
2. Removed unused `display` parameters in functions where they were not being utilized.

## Checklist

~- [ ] Unit tests added (if fixing a bug or adding a new feature)~


shap/datasets.py
```diff
@@ -10,9 +10,24 @@
...
 
 
-def imagenet50(display=False, resolution=224, n_points=None):
-    """ This is a set of 50 images representative of ImageNet images.
+def imagenet50(resolution=224, n_points=None):
+    """
+    This is a set of 50 images representative of ImageNet images.
+
+    Parameters:
+    - resolution (int, optional): The resolution of the images. Default is 224.
```
connortann: Although this parameter has a default and therefore is an optional parameter, the _type_ of the variable is not Optional.  In the context of typing, [optional](https://docs.python.org/3/library/typing.html#typing.Optional) means that `None` is an accepted value. Here we have a regular `int`, not `int | None`.

It looks like `224` is the only resolution supported at present, so I suggest mentioning that in the parameter description.

connortann (changes requested):
Thanks for the PR! There are a few formatting issues that need to be addressed before this is ready.

The docs are built with Sphinx, which uses RST formatting rather than markdown formatting. We use the "napoleon" extension which parses **numpy-style docstrings**.

There is a guide with examples here: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

Please take a look at the examples, and update the docstrings to have the same formatting. For example, taken from the link above:

```
def my_function(param1, param2=None):
    """This is an example of a module level function.

    Function parameters should be documented in the ``Parameters`` section.
    The name of each parameter is required. The type and description of each
    parameter is optional, but should be included if not obvious.

    Example
    -------
    Examples are formatted with a literal block::

        $ python example_numpy.py

    Parameters
    ----------
    param1 : int
        The first parameter.
    param2 str, optional
        The second parameter.
    
    Returns
    -------
    bool
        True if successful, False otherwise.

    """
```

connortann:
The docs are built on CI on each PR, so you can preview the built documentation to check the various sections render correctly. The link is next to the "docs/readthedocs.org" action, which for this PR points to here:

https://shap--3543.org.readthedocs.build/en/3543/generated/shap.datasets.california.html#shap.datasets.california

shap/datasets.py
LakshmanKishore: Sure, will correct this!

LakshmanKishore:
Thank you for the feedback! 
I acknowledge the formatting issues in the docstrings and will update them accordingly.

connortann:
FYI we recently expanded the linting rules to include pydocstyle over on #3550 , to ensure the docs render properly in HTML. So, you may wish to merge the master branch into your branch and re-run the precommit hooks (e.g. with `pre-commit run --all-files`).

Thanks again for the PR, let me know when you'd like another review!

connortann:
It looks like we have some unrelated CI failures, possibly due to an upstream issue from huggingface and transformers. Apologies, we'll try to get that fixed - it doesn't seem to be related to this PR.

LakshmanKishore:
I'm highly interested in doing code contributions in this repo.
Do you have any suggestions for me to start with?

connortann:
Glad to hear it, we'd love to have you on board as a regular contributor. To get some ideas of issues to tackle,  I'd recommend looking at issues with the `goodfirstissue` label.

connortann:
The should pass on this branch once #3563 is merged 👍

connortann (approved):
Looks good, thanks for the PR!

PR #3547, opened 2024-03-06
Fix force_plot contribution threshold for negative contributions

### LINKED ISSUE #2889: shap.force_plot() > contribution_threshold not properly displaying negative contributing features
### This is a fantastic package that is very useful for model explainability - thank you.

A parameter in shap.force_plot() seems to behaving inconsistently with expectations - **any clarifiers/pointers if this is user error is greatly appreciated**.
- I saw this was added in [PR 1584 ](https://github.com/slundberg/shap/commit/f9bcf8a9610dcd0716dde8270f1266c2a24bd394) by @shiva-z a while ago.
- I didn't immediately see any other issues mentioning this.

To be clear, talking specifically when I'm generating a force plot using the following:
`shap.force_plot(base_value, shap_values, contribution_threshold, matplotlib=True, show=False)`

### **Expectation**:
- `contribution_threshold` is some float between 0 & 1, where setting to 0 would show a label for **every** feature (including negative contributing features).
- Setting above 0 will show **every** contribution in which absolute value / sum(absolute values) is greater than the threshold.

### **Reality**:
- Setting `contribution_threshold` arbitrarily low will show all matching **positive** contributors, but cut off at a random number of **negative** ones.
- Anecdotally, the more features there are, the fewer negative ones show.
### Python 3.9.5, Windows machine:
```
shap==0.41.0
matplotlib==3.6.3
numpy==1.23.5
```

### Reproducible example:
```
import shap
import numpy as np

base_value= -5.96385670e+00
contributions=np.array(
    [
        2.62079178e-04,  1.03836309e-03, -4.80703916e-03,  3.47535330e-04,
       -7.19331438e-03,  2.00609677e-03,  7.54139200e-02, -1.20536340e-02,
       -3.57733807e-03,  2.56002939e-04,  5.51301939e-03, -4.02830541e-03,
        1.91734266e-02,  1.76921394e-03,  3.41030653e-03,  3.97255382e-04,
        8.85525224e-05, -2.79973671e-02,  3.44877085e-03, -1.18466816e-03,
       -3.29642966e-02,  3.60896252e-06,  1.12464791e-03, -3.37032927e-03
    ]
)

for i in [0.05, 0.025, 0.01, 0]:

    shap.force_plot(base_value=base_value,
        shap_values=contributions, 
        contribution_threshold = i,
        matplotlib = True, # use matplotlib
        show = True
    )
```

### Visually, note how in the `contribution_threshold=0` chart not all negative contributions are labeled.
- `contribution_threshold=0`
- Expected to see all labels.
- Practically, I have a larger model/feature set where I can't even get two negative contributing features to show.
image

connortann:
## Overview

Closes #2889

Removes redundant logic that controls the display of labels, so that the `contribution_threshold` parameter works as expected for both positive and negative values

## Screenshots

For the example in #2889, if `contribution_threshold=0` we expect to see all labels present (even though this is rather visually cluttered, which is why by default a higher threshold is applied).

Before (note labels missing on right hand side):

image

After:

image

### DIFF
--- a/shap/plots/_force_matplotlib.py
+++ b/shap/plots/_force_matplotlib.py
@@ -134,12 +134,6 @@ def draw_labels(fig, ax, out_value, features, feature_type, offset_text, total_e
         else:
             box_end_ = box_size.get_points()[1][0]
 
-        # If the feature goes over the side of the plot, we remove that label
-        # and stop drawing labels
-        if box_end_ > ax.get_xlim()[1]:
-            text_out_val.remove()
-            break
-
         # Create end line
         if (sign * box_end_) > (sign * val):
             x, y = np.array([[val, val], [0, -0.18]])


CloseChoice (approved):
Nice catch. LGTM.

Great to see that we have the resources to fix older bugs ;). Amazing

CloseChoice:
An improvement to the plotting would be to provide a keyword that limits the number of features shown. All others will be aggregated (so one aggregated arrow for positive and one for negative contributions). That would help the visualization a lot IMO. Please don't get me wrong, this is nothing for this PR just something we could look into in the future and I am curious about your thoughts on this.

connortann:
I agree, that sounds like it could be useful

PR #3549, opened 2024-03-07
DOCS: Formatting and phrasing updates for SHAP return values

connortann:
## Overview

A few fixups following #3318

- Update RST formatting for `versionchanged` directive and bullet points
- Rephrase docstrings slightly for clarity (feedback/suggestions welcome)

PR #3550, opened 2024-03-07
DOCS: Enable subset of pydocstyle rules, bulk update ruff fixes

connortann:
## Overview

Fixes a large number of [pydocstyle](https://docs.astral.sh/ruff/rules/#pydocstyle-d) linting rules.

This should help us avoid the Sphinx formatting issues that we've encountered quite a few times, with automated fixes on pre-commit.

## Changes

- Adds a number of pydocstyle to our ruff config
- Defines the convention as "numpy"-style, to match the existing style
- Fixes about 14 rules, mostly the ones which have automatic fixes available

## Reviewing

This PR is both simple and complex. It's simple as it comprises only automated fixes to docstrings, but it's complex as the diff is very large.

Whilst this is initially rather tiresome to fix & review, I think that have automated `ruff` rules with autofixes will help the code maintenance be simpler & easier in future.

For ease of review, I have fixed one rule per commit. I would suggest reviewing the [commits](https://github.com/shap/shap/pull/3550/commits) individually: the majority are very simple whitespace changes that can be skim-read.

Most commits were made automatically with a single command like `ruff check . select=D411 --fix`.

Note to self: this branch builds upon #3549, so that should be reviewed/merged first. Once this PR is squashed & merged, we could include the commit hash in `.git-blame-ignore-revs`.


CloseChoice (approved):
This looks good to me. I just had one question that arose from going through the commits in the wrong order but figured that out quickly. 

Thanks for the great work and an extra thanks for thinking about the reviewer when doing that work. That really helped A LOT.

connortann:
As a follow-up thought, please do feel free to tweak the rules and per-file-ignores in pyproject.toml, if you feel anything should be changed. For example, I realised we can just ignore the pydocstyle rules in `./tests`, as it has no real benefit.

PR #3552, opened 2024-03-07
MAINT: Bump ruff version, fix updated rule UP032 (f-strings)

connortann:
## Overview

- Increments the ruff version from `0.2.1` to `0.3.1`
- Fixes the now-failing rule UP032 (use f-strings rather than `.format()`)


CloseChoice (approved):
LGTM

PR #3554, opened 2024-03-07
MAINT: Add l1_reg argument to KernelExplainer call()

connortann:
## Overview

Follow-up to #3517 based on discussion. Adds a parameter `l1_reg` to the `__call__` method, so a user can silence the deprecation warning.

### DIFF
--- a/shap/explainers/_kernel.py
+++ b/shap/explainers/_kernel.py
@@ -150,7 +150,7 @@ class KernelExplainer(Explainer):
                 tensor_as_np_array = sess.run(symbolic_tensor)
         return tensor_as_np_array
 
-    def __call__(self, X):
+    def __call__(self, X, l1_reg="auto"):
 
         start_time = time.time()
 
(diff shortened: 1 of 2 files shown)


shap/explainers/_kernel.py
```diff
@@ -150,7 +150,7 @@ def _convert_symbolic_tensor(symbolic_tensor) -> np.ndarray:
                 tensor_as_np_array = sess.run(symbolic_tensor)
         return tensor_as_np_array
 
-    def __call__(self, X):
+    def __call__(self, X, l1_reg="auto"):
```
CloseChoice: Would it be possible to adjust a test using the call interface to actually use this argument? That would be great IMO

CloseChoice (approved):
LGTM. Just an optional suggestion

shap/explainers/_kernel.py
connortann: Good idea, done. Would you kindly re-review?

CloseChoice (approved):
Perfect, thanks for adding the test

PR #3560, opened 2024-03-09
Update slicer pin to 0.0.8 and add test

### LINKED ISSUE #3522: BUG: Shap value cannot be calculated according to NLP tutorial on the original website
### Issue Description

I try to implement however shap_value.shape is (10, None, 4). Perhabs, because of this, I have encountered error : "setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (10,) + inhomogeneous part."

```python
import datasets
import numpy as np
import transformers

import shap
dataset = datasets.load_dataset("imdb", split="test")

# shorten the strings to fit into the pipeline model
short_data = [v[:500] for v in dataset["text"][:20]]
classifier = transformers.pipeline("sentiment-analysis", return_all_scores=True)
classifier(short_data[:2])
pmodel = shap.models.TransformersPipeline(classifier, rescale_to_logits=True)
explainer3 = shap.Explainer(pmodel, classifier.tokenizer)
shap_values3 = explainer3(short_data[:10])
shap.plots.text(shap_values3[:, :, 1])#===>error 1
shap.plots.bar(shap_values3[:, :, 1].mean(0))#===>error 2


```

### Traceback

```shell
#===>error 1
<ipython-input-145-b849a902a16a> in <cell line: 1>()
----> 1 shap.plots.text(shap_values3[:, :, 1])

4 frames
/usr/local/lib/python3.10/dist-packages/slicer/slicer_internal.py in tail_slice(cls, o, tail_index, max_dim, flatten)
    441                 import numpy
    442 
--> 443                 return numpy.array(inner)
    444             elif _safe_isinstance(o, "torch", "Tensor"):
    445                 import torch

ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (2,) + inhomogeneous part." 

#===>error 2
<ipython-input-146-2f2b208f9c91> in <cell line: 1>()
----> 1 shap.plots.bar(shap_values3[:, :, 1].mean(0))

4 frames
/usr/local/lib/python3.10/dist-packages/slicer/slicer_internal.py in tail_slice(cls, o, tail_index, max_dim, flatten)
    441                 import numpy
    442 
--> 443                 return numpy.array(inner)
    444             elif _safe_isinstance(o, "torch", "Tensor"):
    445                 import torch

ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (2,) + inhomogeneous part.
```

### Expected Behavior

_No response_

### Bug report checklist

### Installed Versions

version: (0.44.1)

CloseChoice:
## Overview

Closes #3522  

 - install latest changes from slicer
 - add failing example from issue #3522 as test

@connortann : I leave it to you whether we really merge this or just wait for the next release.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,6 @@ docs = [
 ]
 test-core = ["pytest", "pytest-mpl", "pytest-cov"]
 test = [
-  "datasets",
   "pytest",
   "pytest-mpl",
   "pytest-cov",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [
   'pandas',
   'tqdm>=4.27.0',
   'packaging>20.9',
-  'slicer==0.0.7',
+  'slicer @ git+https://github.com/interpretml/slicer.git@12964ec458c5169efe19df3c296fc5e3e3f6a84f',
   'numba',
   'cloudpickle'
 ]
@@ -56,6 +56,7 @@ docs = [
 ]
 test-core = ["pytest", "pytest-mpl", "pytest-cov"]
 test = [
+  "datasets",
   "pytest",
   "pytest-mpl",
   "pytest-cov",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [
   'pandas',
   'tqdm>=4.27.0',
   'packaging>20.9',
-  'slicer==0.08',
+  'slicer==0.0.8',
   'numba',
   'cloudpickle'
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [
   'pandas',
   'tqdm>=4.27.0',
   'packaging>20.9',
-  'slicer @ git+https://github.com/interpretml/slicer.git@12964ec458c5169efe19df3c296fc5e3e3f6a84f',  # waiting for next release
+  'slicer==0.08',
   'numba',
   'cloudpickle'
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [
   'pandas',
   'tqdm>=4.27.0',
   'packaging>20.9',
-  'slicer @ git+https://github.com/interpretml/slicer.git@12964ec458c5169efe19df3c296fc5e3e3f6a84f',
+  'slicer @ git+https://github.com/interpretml/slicer.git@12964ec458c5169efe19df3c296fc5e3e3f6a84f',  # waiting for next release
   'numba',
   'cloudpickle'
 ]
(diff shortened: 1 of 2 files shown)


connortann:
I believe `slicer 0.0.8` is now on PyPI so we could use that?

connortann:
Looks like there is an incompatibility between `transformers` and `keras v3`:

https://github.com/huggingface/transformers/pull/28588

https://github.com/huggingface/transformers/blob/0290ec19c901adc0f1230ebdccad11c40af026f5/src/transformers/activations_tf.py#L21-L31

Maybe we need to pin the keras version to less than 2, at least for now?

pyproject.toml
```diff
@@ -18,7 +18,7 @@ dependencies = [
   'pandas',
   'tqdm>=4.27.0',
   'packaging>20.9',
-  'slicer==0.0.7',
+  'slicer @ git+https://github.com/interpretml/slicer.git@12964ec458c5169efe19df3c296fc5e3e3f6a84f',  # waiting for next release
```
connortann:
```suggestion
  'slicer==0.08',
```

connortann (changes requested):
Slicer has a new release on PyPi and conda forge, so we can use it directly. Great that we have a test: I definitely think for every bug fix we should include a test that fails on master but passes on the new branch.

One minor suggestion, would it be possible to simplify the test a little, whilst still reproducing the bug on #3522? From reading the test, I don't find it clear exactly what it is testing

CloseChoice:
> Slicer has a new release on PyPi and conda forge, so we can use it directly. Great that we have a test: I definitely think for every bug fix we should include a test that fails on master but passes on the new branch.
> 
> One minor suggestion, would it be possible to simplify the test a little, whilst still reproducing the bug on #3522? From reading the test, I don't find it clear exactly what it is testing

Hmm, under the hood this test now passes because `np.array([np.array([...]), np.array([...]))` can be sliced correctly. But testing this explicitly would actually mean that we test slicer's behaviour without the context in which it appears in shap. IMO we should test that behaviour that we fixed once is usable from now on and in this case the behaviour is just what we have in the test. Could do some minor tweaks but what we test here should be in my opinion just be that the transformer model works correctly. What are your thoughts on this? Should we explicitly test the nested np.array slicing behaviour? Or should we just create the explanation object that is the output of the model + explainer and test that it is slicable + plots work?

pyproject.toml
connortann: I'll go ahead and add this change, now that slicer 0.08 is out to check the tests pass, if that's ok with you @CloseChoice

connortann (approved):
> we test here should be in my opinion just be that the transformer model works correctly.

Makes sense, I agree.

This LGTM - I updated the slicer pin and tests pass. Nice one

PR #3561, opened 2024-03-09
ENH: New plotting API for beeswarm so can accept and return Axes

### LINKED ISSUE #2822: Displaying two beeswarm plots side by side in the same figure using matplotlib
Hello,

I wonder how to place two independent beeswarm plots into the same figure but in different axes. I am using matplotlib version `3.4.2`, shap version `0.39.0`, and Python `3.8`. I am trying using a single figure and two axes but realize that the beeswarm method does not handle the axis as parameter (error: `TypeError: beeswarm() got an unexpected keyword argument 'ax'`). Any recommendation is welcome. Thanks in advance.

```
from matplotlib import pyplot as plt
import shap

fig, axes = plt.subplots(1, 2, figsize=(35, 7), gridspec_kw = {"wspace":1.0}) 

shap.plots.beeswarm(explainer_a(X_test_a), max_display=10, ax = axes[0])
shap.plots.beeswarm(explainer_b(X_test_b), max_display=10, ax = axes[1])

plt.show()
```

### LINKED ISSUE #3351: ENH: beeswarm plot doesn't return figure
### Problem Description

Took me some time to discover this interface:
```
shap.plots.beeswarm(...)
fig = plt.gcf()
fig.savefig(...)
```
Instead, `beeswarm` (and other plots) should return the figure directly.

### Alternative Solutions

_No response_

### Additional Context

_No response_

### Feature request checklist

chriscave:
## Overview

Supports https://github.com/shap/shap/issues/3411

Change beeswarm to adhere to new standard API interface for the plots:
* Add an ax parameter to accept an existing matplotlib Axes
* Return the axes (if show=False)

## Discussion
The parameter `plot_size` is only really suitable for when there is one axes in the figure. If there are multiple then the `default` value affects the figure size and not just the axes which the plot is being drawn to.

We could use `plot_size` to just affect the size of the axes but this I think this is discouraged from mpl and can be quite tricky to do. 

So for now I thought it would be easier to just stop the behaviour of `plot_size` when `ax` is not None. So if a user passes in `ax` into the plot then it is up to them to determine the size of the plot using the figure the axes has come from.

We can definitely change this if it is not in line with the current thoughts on the new API. 

## Checklist

(Edit by maintainers):
- Closes #2822
- Closes #3351


chriscave:
Tagging @connortann into this draft PR

connortann:
Thanks for the PR! The tests are currently failing due to an unrelated issue, which should be fixed shortly by#3563.

shap/plots/_beeswarm.py
```diff
@@ -415,21 +431,22 @@ def beeswarm(shap_values, max_display=10, order=Explanation.abs.mean(0),
...
+    ax.set_yticks(range(len(feature_inds)), reversed(yticklabels), fontsize=13)
+    ax.tick_params('y', length=20, width=0.5, which='major')
+    ax.tick_params('x', labelsize=11)
+    ax.set_ylim(-1, len(feature_inds))
+    ax.set_xlabel(labels['VALUE'], fontsize=13)
     if show:
         pl.show()
     else:
-        return pl.gca()
+        pl.close(fig)
```
connortann: May I ask about the motivation behind `pl.close`? I think it may be a breaking change; we may as well decide a suitable pattern and plan to do that same for the other plots.

connortann:
This PR looks good to me! One suggestion - I think it would be preferable to raise a ValueError if plot_size and ax are both passed, rather than just ignoring the size.

shap/plots/_beeswarm.py
chriscave: Of course. Jupyter will still render the plot even if `show = false`. Closing the figure will prevent that. But maybe I've misunderstood what show is meant to be doing.

Ah if its a breaking change then perhaps more thought should be put into it. How do you think it's a breaking change?

chriscave:
> This PR looks good to me! One suggestion - I think it would be preferable to raise a ValueError if plot_size and ax are both passed, rather than just ignoring the size.

Good suggestion, I'll change it to that.

connortann:
Sorry for the delay on merging this one. We've been rather short of maintainers recently and this one fell through the cracks.

I'm trying to overhaul a few of the plotting functions at the moment, so I'll take another look and sort out the merge conflicts. Thanks again for the PR!

shap/plots/_beeswarm.py
connortann: I think this is "breaking" just in the sense that this is different behaviour than before. For now, let's avoid calling `plt.close` just to keep this plot consistent with the other plots in the library.

I think the "show" argument is really not needed and I'd like to deprecate it entirely, leaving it to the user to call `plt.show()` if desired. But, I think it would be better to make this change in one fell swoop, updating all the plots to a uniform interface.

connortann: Note to self for future reference: It may be quite hard to "re-open" a plot that was closed with `plt.close()` , as described here:
https://stackoverflow.com/a/31731945

shap/plots/_beeswarm.py
```diff
@@ -157,8 +174,12 @@ def beeswarm(
     if feature_names is None:
         feature_names = np.array([labels["FEATURE"] % str(i) for i in range(num_features)])
 
+    fig = pl.gcf()
+    if ax is None:
+        ax = pl.gca()
```
connortann: We should use the figure that belongs to `ax`, which in general may be different to the current figure.

```suggestion
    if ax is None:
        ax = pl.gca()
    fig = ax.get_figure()
    assert fig is not None  # type narrowing for mypy
```

The `assert` statement will be helpful for mypy, when this function is type checked. The return type of `artist.get_figure()` is `Figure | None`, but in this case we never expect it to be `None` .

shap/plots/_beeswarm.py
connortann: I've created an issue to discuss this issue further: #3819

For now though, let's remove this line (which should fix the tests):

```suggestion
```

connortann (changes requested):
@chriscave I've fixed up the merge conflicts, and edited the description to link a couple of issues that this will close.

Just two suggested changes as commented above, and then the tests should pass and we should be good to merge! Thanks again. [EDIT: done]

connortann (approved):

Issue #3562, opened 2024-03-09
CI failing on tensorflow 2.16+ due to incompatibility between transformers & keras V3


connortann:
TODO:

- [x] Pin CI to use keras to < v3 to fix CI on master
- [ ] Await transformers compatibility with keras 3: https://github.com/huggingface/transformers/pull/26224
- [ ] Get tests passing with latest tensorflow 2.16 +
- [ ] Include tensorflow in py12 tests

rathorexd:
is the issue is still open to contribute?

PR #3563, opened 2024-03-09
MAINT: Pin tensorflow to <2.16

connortann:
## Overview

Supports #3562 , pins tensorflow to the most recent passing version for now.

Hopefully fixes CI as short-term patch


CloseChoice (approved):
nice, LGTM

PR #3564, opened 2024-03-09
Include catboost tests for py312

connortann:
## Overview

Enable catboost tests for python 3.12, following support in recent release:

https://github.com/catboost/catboost/releases/tag/v1.2.3


CloseChoice (approved):
very nice. Thanks for having an eye on that. LGTM

PR #3566, opened 2024-03-10
Minor datasets docs fixups

connortann:
Some minor fixups to RST formatting for `datasets`. Still probably not perfect but should improve the rendering a bit.


CloseChoice (approved):
I see this as a good improvement to the docs. LGTM

PR #3567, opened 2024-03-10
MAINT: Minor simplifications and removal of duplication

connortann:
A few relatively minor refactors:

- Removing some redundant calls, e.g. replacing `range(0, x)` with `range(x)`
- Removing some duplicated blocks of code in if/else blocks

I would appreciate a review to ensure the refactored code is identical in behaviour.


CloseChoice (approved):
LGTM

PR #3570, opened 2024-03-13
Enable ROCM support

hliuca:
## Overview

This PR adds a few files under cext/ folder and enables AMD GPUs using ROCm.

The setup.py has been modified to detect ROCm and to compile HIP codes.


CloseChoice:
Thanks for the PR. Would be great to have support for this. Could you elaborate a bit how we can install and test this? No need for CI tests at the moment but as part of the review I would like to install it and run a couple of tests with it

hliuca:
> Thanks for the PR. Would be great to have support for this. Could you elaborate a bit how we can install and test this? No need for CI tests at the moment but as part of the review I would like to install it and run a couple of tests with it

Thanks for the response.
1) The setup.py script will take care of the installation, such as "python3 setup.py install". Its compilation and installation are the same as CUDA. The setup.py looks for CUDA and HIP and compiles automatically.

2) The HIP version has the same APIs as CUDA version, so the same scripts/tests for CUDA work for HIP. The users don't need to change any code and the code works.

3) tests/explainers/test_gpu_tree.py is a script to test GPU version, "explainer = shap.GPUTreeExplainer(model)".

You will need AMD Instinct GPUs to run the HIP/ROCm version.

CloseChoice:
> > Thanks for the PR. Would be great to have support for this. Could you elaborate a bit how we can install and test this? No need for CI tests at the moment but as part of the review I would like to install it and run a couple of tests with it
> 
> Thanks for the response.
> 
>     1. The setup.py script will take care of the installation, such as "python3 setup.py install". Its compilation and installation are the same as CUDA. The setup.py looks for CUDA and HIP and compiles automatically.
> 
>     2. The HIP version has the same APIs as CUDA version, so the same scripts/tests for CUDA work for HIP. The users don't need to change any code and the code works.
> 
>     3. tests/explainers/test_gpu_tree.py is a script to test GPU version, "explainer = shap.GPUTreeExplainer(model)".
> 
> 
> You will need AMD Instinct GPUs to run the HIP/ROCm version.

Thanks for all the work. Is this ready for a first round of review?

hliuca:
> 
> Thanks for all the work. Is this ready for a first round of review?

Yes, it is ready. :-)

PR #3571, opened 2024-03-14
ENH: Add ax option to heatmap plotting function

sroener:
## Overview

Addresses #3544 and contributes to #3411 

- changes `mean()` to `sum()` in the f(x) calculation performed in the heatmap plotting function. This is a better representation of how SHAP workes, but does not change the general functionality
- Adds `ax` argument to heatmap mapping function, including documentation and return statement

## Checklist

### DIFF
--- a/shap/plots/_heatmap.py
+++ b/shap/plots/_heatmap.py
@@ -136,7 +136,7 @@ def heatmap(shap_values, instance_order=Explanation.hclust(), feature_values=Exp
 
     # plot the f(x) line chart above the heat map
     ax.axhline(-1.5, color="#aaaaaa", linestyle="--", linewidth=0.5)
-    fx = values.T.mean(0)
+    fx = values.T.sum(0)
     ax.plot(
         -fx / np.abs(fx).max() - 1.5,
         color="#000000",
--- a/shap/plots/_heatmap.py
+++ b/shap/plots/_heatmap.py
@@ -10,7 +10,7 @@ from ._utils import convert_ordering
 
 def heatmap(shap_values, instance_order=Explanation.hclust(), feature_values=Explanation.abs.mean(0),
             feature_order=None, max_display=10, cmap=colors.red_white_blue, show=True,
-            plot_width=8):
+            plot_width=8, ax=None):
     """Create a heatmap plot of a set of SHAP values.
 
     This plot is designed to show the population substructure of a dataset using supervised
@@ -46,8 +46,11 @@ def heatmap(shap_values, instance_order=Explanation.hclust(), feature_values=Exp
         Setting this to ``False`` allows the plot
         to be customized further after it has been created.
 
-    plot_width: int, default 8
+    plot_width : int, default 8
         The width of the heatmap plot.
+
+    ax : matplotlib Axes
+        Axes object to draw the plot onto, otherwise uses the current Axes.
 
     Examples
     --------
@@ -98,8 +101,9 @@ def heatmap(shap_values, instance_order=Explanation.hclust(), feature_values=Exp
 
     # define the plot size based on how many features we are plotting
     row_height = 0.5
-    pl.gcf().set_size_inches(plot_width, values.shape[1] * row_height + 2.5)
-    ax = pl.gca()
+    if ax is None:
+        pl.gcf().set_size_inches(plot_width, values.shape[1] * row_height + 2.5)
+        ax = pl.gca()
 
     # plot the matrix of SHAP values as a heat map
     vmin, vmax = np.nanpercentile(values.flatten(), [1, 99])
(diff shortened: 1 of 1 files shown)


sroener:
@connortann please have a look at the Pull Request and the workflows. 

I tried to keep the API changes as minimal as possible. I would like to change how the f(x) graph is plotted, but I will open a new issue for that and would like to discuss if my approach to that would be fitting the general design decisions of the project.

shap/plots/_heatmap.py
```diff
@@ -46,9 +46,12 @@ def heatmap(shap_values, instance_order=Explanation.hclust(), feature_values=Exp
         Setting this to ``False`` allows the plot
         to be customized further after it has been created.
 
-    plot_width: int, default 8
+    plot_width : int, default 8
         The width of the heatmap plot.
 
+    ax : matplotlib Axes
+        Axes object to draw the plot onto, otherwise uses the current Axes.
+
```
connortann, line 54: Suggest adding the new return value to the doscstring:

```
    Returns
    -------
    ax: matplotlib Axes
        Returns the Axes object with the plot drawn onto it.
```

shap/plots/_heatmap.py
```diff
@@ -178,3 +182,5 @@ def heatmap(shap_values, instance_order=Explanation.hclust(), feature_values=Exp
 
     if show:
         pl.show()
+
+    return ax
```
connortann, line 191: The bar plot currently has this behaviour, only returning `ax` if `show=False`:

```python
    if show:
        pl.show()
    else:
        return ax
```

Any thoughts @CloseChoice on which convention is better? I'm happy with either, but we should probably be consistent. I think we should also be mindful to minimise breaking changes wherever possible.

shap/plots/_heatmap.py
sroener: Good suggestion, already pushed.

shap/plots/_heatmap.py
sroener: Maybe I can add my thoughts. When I use shap in jupyter notebooks/lab, the default is to automatically show plots created by Matplotlib/Seaborn. For me,  the pl.show() approach without returning an ax was irritating because it always took a bit of extra code to get the figure for saving. 

In contrast, when the ax is returned, I can get both the ax/fig for saving and I can see the plot.

CloseChoice: Isn't the figure shown twice in notebooks if we do 
```python
if show:
    pl.show()
return ax
```
? If so, then I'd think an additional `else` would be good.
If not, then let's just avoid the additional `else`

sroener: @CloseChoice Good question. I tried it with the following code and it does not show the figure twice:

```python
from matplotlib import pyplot as pl
import numpy as np

# define test function
def simple_plot(ax=None, show=True):
    
    if ax is None:
        pl.gcf().set_size_inches(6,8)
        ax = pl.gca()
    data = np.arange(0,100)
    ax.plot(data)
    
    if show:
        pl.show()
    
    return ax

# test without passing an ax

testfig = simple_plot(show=True)

# test with passing an ax

fig,ax = pl.subplots()
test_fig = simple_plot(ax=ax)
```

Note that the function returns an ax and not a figure. If the figure should be accessed/saved it needs to be retrieved using the `.get_figure()` method.

CloseChoice: Nice, thanks for checking. Then I would suggest to go with the 
```python
if show:
    pl.show()
return ax
```
solution

shap/plots/_heatmap.py
```diff
@@ -136,7 +145,7 @@ def heatmap(shap_values, instance_order=Explanation.hclust(), feature_values=Exp
 
     # plot the f(x) line chart above the heat map
     ax.axhline(-1.5, color="#aaaaaa", linestyle="--", linewidth=0.5)
-    fx = values.T.mean(0)
+    fx = values.T.sum(0)
```
CloseChoice, line 148: This would change existing plots. Would you be so kind to explain why you'd like to change this?

sroener: I thought so too, but it seems not to change the plot due to the abs_max normalization before plotting. I shortly discussed this in #3544.

This works because of two operations. First, the `mean()` and `sum()` are computed over the same number of samples, and second, the values are scaled by a [max_abs scaling](https://github.com/shap/shap/blob/dbb275978abe27def1afe73fdd097d071c7fbe38/shap/plots/_heatmap.py#L141C1-L146C6) afterwards, bringing the resulting values to a common scale.

Because SHAP should be additive, I thought the `sum()` operation would be more intuitive.

sroener:
@connortann @CloseChoice 

Is there anyhting that is missing for my PR to be merged?

CloseChoice:
from my point of view this is fine if the tests pass.

cc @connortann

CloseChoice (approved):
LGTM, let's check if the tests pass

connortann (approved):
Thanks again for the PR!

PR #3573, opened 2024-03-15
add return figure for summary_plot

AlKul:
## Overview

Closes #XXXX  

## Checklist

### DIFF
--- a/shap/plots/_beeswarm.py
+++ b/shap/plots/_beeswarm.py
@@ -489,6 +489,8 @@ def summary_legacy(shap_values, features=None, feature_names=None, max_display=N
         by default.
 
     """
+
+    fig = pl.figure()
     # support passing an explanation object
     if str(type(shap_values)).endswith("Explanation'>"):
         shap_exp = shap_values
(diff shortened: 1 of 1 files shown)


CloseChoice:
Thanks for the PR. Could you please elaborate a bit what is your intention behind this? So stating your problem with the current implementation would be great for us, so that we have something to put in the release notes later on. 
Also at some point we would need to fix the tests.

Another thing that makes me a bit curious is that we plan to deprecate this function. Do you still actively use it?

AlKul:
> Thanks for the PR. Could you please elaborate a bit what is your intention behind this? So stating your problem with the current implementation would be great for us, so that we have something to put in the release notes later on. Also at some point we would need to fix the tests.
> 
> Another thing that makes me a bit curious is that we plan to deprecate this function. Do you still actively use it?

Hey!
Thanks for quick reply, actually, that was supposed to be pushed to my fork - my bad

And yeah, I’m trying to use this function right now, for analyzing the importance of features for my multi class classification model

I want the plot function return figure - not just plot it (for saving results for each experiment)

And thanks for the lib - just starting using it, looks useful and interesting

CloseChoice:
Thanks for your work and the reply. I think this is a totally understandable undertaking. Are you still working on this? If so, would be great if you could fix the tests, then I think we are pretty close to merging.

connortann:
Similar functionality implemented by #3561

Issue #3574, opened 2024-03-15
BUG: LightGBM with multiclass interaction TreeShap produces explainer error


NegatedObjectIdentity:
### Issue Description

This is an follow up to #3187. In #3187 I reported on inconsistencies in the returned SHAP explainer object depending on the explanation task (e.g. classification bin/multi or regression; with or without feature interactions). This has bin solved (see assessment below). Thank you very much for your awesome work! But there is still an error in the case of task multiclass+interactions.

My assessment:

Kind of task | Explainer object shape | My assessment
1. Regression+no_interactions | samples x features | OK
2. Regression+feature_interactions | samples x features x features | OK
3. Classification binary+no_interactions | samples x features  | OK.
4. Classification binary+feature_interactions | samples x features x features | OK.
5. Classification multiclass+no_interactions | samples x features x classes | OK.
6. Classification multiclass+feature_interactions | error with the traceback shown below | error

### Minimal Reproducible Example

```python
from sklearn.datasets import load_diabetes
from sklearn.datasets import load_digits
from sklearn.datasets import load_breast_cancer
from lightgbm import LGBMRegressor
from lightgbm import LGBMClassifier
from shap.explainers import Tree as TreeExplainer

# data | regression | binary | multi class
data_reg = load_diabetes(as_frame=True)
data_bin = load_breast_cancer(as_frame=True)
data_mult = load_digits(as_frame=True)

# train models | regression | binary | multi class
model_reg = LGBMRegressor(**{'verbosity': -1,}).fit(data_reg.data, data_reg.target)
model_bin = LGBMClassifier(**{'verbosity': -1,}).fit(data_bin.data, data_bin.target)
model_mult = LGBMClassifier(**{'verbosity': -1,}).fit(data_mult.data, data_mult.target)

# Explainer
explainer_reg = TreeExplainer(
    model_reg,
    data=None,
    feature_perturbation='tree_path_dependent')
explainer_bin = TreeExplainer(
    model_bin,
    data=None,
    feature_perturbation='tree_path_dependent')
explainer_mult = TreeExplainer(
    model_mult,
    data=None,
    feature_perturbation='tree_path_dependent')

# Explainations
explainations_reg = explainer_reg(data_reg.data, interactions=False)
explainations_reg_inter = explainer_reg(data_reg.data, interactions=True)
explainations_bin = explainer_bin(data_bin.data, interactions=False)
explainations_bin_inter = explainer_bin(data_bin.data, interactions=True)
[...]

PR #3576, opened 2024-03-17
throw warning and update error message

CloseChoice:
## Overview

Supports #3574   

 - throw warning again. I checked the commit of Scott, where he uncommented the warning (see [here](https://github.com/shap/shap/commit/be1fa942604f30afe05deda83740620a8ea95607)) but could not find any reason why he did, so no linked PR or something else.
 - update error message in case of uncovered leaves

## Checklist

### DIFF
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -185,7 +185,7 @@ class TreeExplainer(Explainer):
             self.data = data
         if self.data is None:
             feature_perturbation = "tree_path_dependent"
-            logger.warning("Setting feature_perturbation = \"tree_path_dependent\" because no background data was given.")
+            logger.info('Setting feature_perturbation = "tree_path_dependent" because no background data was given.')
         elif feature_perturbation == "interventional" and self.data.shape[0] > 1_000:
             wmsg = (
                 f"Passing {self.data.shape[0]} background samples may lead to slow runtimes. Consider "
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -74,6 +74,8 @@ def _xgboost_cat_unsupported(model):
         )
 
 
+logger = logging.getLogger(__name__)
+
 class TreeExplainer(Explainer):
     """Uses Tree SHAP algorithms to explain the output of ensemble tree models.
 
@@ -183,7 +185,7 @@ class TreeExplainer(Explainer):
             self.data = data
         if self.data is None:
             feature_perturbation = "tree_path_dependent"
-            logging.warning("Setting feature_perturbation = \"tree_path_dependent\" because no background data was given.")
+            logger.warning("Setting feature_perturbation = \"tree_path_dependent\" because no background data was given.")
         elif feature_perturbation == "interventional" and self.data.shape[0] > 1_000:
             wmsg = (
                 f"Passing {self.data.shape[0]} background samples may lead to slow runtimes. Consider "
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -1295,7 +1295,6 @@ class TreeEnsemble:
 
                 # ensure that the passed background dataset lands in every leaf
                 if np.min(self.trees[i].node_sample_weight) <= 0:
-                    # import pdb; pdb.set_trace()
                     self.fully_defined_weighting = False
 
             self.num_nodes = np.array([len(t.values) for t in self.trees], dtype=np.int32)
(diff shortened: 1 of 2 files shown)


CloseChoice:
@DanGolding: happy if you also take a look

shap/explainers/_tree.py
```diff
@@ -182,7 +182,7 @@ def __init__(
             self.data = data
         if self.data is None:
             feature_perturbation = "tree_path_dependent"
-            #warnings.warn("Setting feature_perturbation = \"tree_path_dependent\" because no background data was given.")
+            warnings.warn("Setting feature_perturbation = \"tree_path_dependent\" because no background data was given.")
```
connortann: This change may impact a large fraction of users, so I think we should carefully consider what level of notification is appropriate. Would `logging.info` be more appropriate here?

In the logging [HOWTO guide](https://docs.python.org/3/howto/logging.html), they recommend:

- To report events that occur "during normal operation of a program", use logging `info()`
- To issue a warning regarding a particular runtime event, if the issue is avoidable and the client application should be modified to eliminate the warning, use `warnings.warn`

To me, this seems like normal operation rather than an "issue", as this behaviour is described in the docs of tree explainer:

```
feature_perturbation : "interventional" (default) or "tree_path_dependent" (default when data=None)
```

My guess is that Scott commented this warning as it was becoming a bit annoying.

connortann (changes requested):
One suggestion as above

CloseChoice (commented):
Your suggestion is implemented. Sorry that this took so long

shap/explainers/_tree.py
```diff
@@ -182,7 +183,7 @@ def __init__(
             self.data = data
         if self.data is None:
             feature_perturbation = "tree_path_dependent"
-            #warnings.warn("Setting feature_perturbation = \"tree_path_dependent\" because no background data was given.")
+            logging.warn("Setting feature_perturbation = \"tree_path_dependent\" because no background data was given.")
```
connortann:
```suggestion
            logging.warning("Setting feature_perturbation = \"tree_path_dependent\" because no background data was given.")
```

connortann (changes requested):
I believe `logging.warn` is deprecated in favour of `logging.warning`.

Also, we should use a module-level logger rather than the root logger: `log = logging.getLogger(__name__)`.

From the [python howto docs](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library):

>  It is strongly advised that you do not log to the root logger in your library. Instead, use a logger with a unique and easily identifiable name, such as the __name__ for your library’s top-level package or module. Logging to the root logger will make it difficult or impossible for the application developer to configure the logging verbosity or handlers of your library as they wish.

CloseChoice:
> I believe `logging.warn` is deprecated in favour of `logging.warning`.
> 
> Also, we should use a module-level logger rather than the root logger: `log = logging.getLogger(__name__)`.
> 
> From the [python howto docs](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library):
> 
> > It is strongly advised that you do not log to the root logger in your library. Instead, use a logger with a unique and easily identifiable name, such as the **name** for your library’s top-level package or module. Logging to the root logger will make it difficult or impossible for the application developer to configure the logging verbosity or handlers of your library as they wish.

Thanks for the review. I implemented your suggestion to have a custom logger for the `_tree.py` file

connortann (approved):
LGTM!

connortann:
Ah - probably the same discussion applies here about logging.INFO vs logging.WARNING. I'll hold off merging, pending the dicussion in our other thread...

https://github.com/shap/shap/pull/3650#discussion_r1600365266

shap/explainers/_tree.py
```diff
@@ -182,7 +185,7 @@ def __init__(
             self.data = data
         if self.data is None:
             feature_perturbation = "tree_path_dependent"
-            #warnings.warn("Setting feature_perturbation = \"tree_path_dependent\" because no background data was given.")
+            logger.warning("Setting feature_perturbation = \"tree_path_dependent\" because no background data was given.")
```
connortann: Coming back to this PR from a few months ago, I think let's go with `info` rather than `warning` to avoid generating warnings in typical uses of the package.

```suggestion
            logger.info('Setting feature_perturbation = "tree_path_dependent" because no background data was given.')
```

CloseChoice:
closing this for now!

PR #3650, opened 2024-05-11
improve transparency concerning background samples

### LINKED ISSUE #3461: ENH: increase transparency of background dataset sub-sampling
### Issue Description

Given $x$ which is the sample that we wish to explain, we can compute the Shapley values of that sample using a background sample $x^b$. By providing the Explainer class with background data, it should compute the Shapley values for each sample in the background the background data and then take the average, which will be an approximation to the interventional SHAP. 

The averaging procedure means that if I for example split my background data in half, A and B, then I should be able to call explainer on both A and B to obtain the averaged SHAP, a and b, for each half. If I now take (a + b)/2, then this should equal calling SHAP on the entire dataset to begin with. 

From my experimentation, it seems that if the background dataset is over 100 samples, then it becomes inconsistent i.e. (a+b)/2 is not equal to the interventional approximation on the entire background dataset. However, the formula holds for datasets under 100 samples. 

```python
import numpy as np
import pandas as pd
import xgboost

import shap

N = 1000
M = 2

X = rng.standard_normal(size=(N, 2))
X[:, 0] = 0.2*X[:, 1] + X[:, 0]
y = -2*X[:, 0] + X[:, 1] + 0.5*X[:, 0]*X[:, 1]

X = pd.DataFrame(X, columns=["X1", "X2"])


model = xgboost.XGBRegressor(n_estimators=100, learning_rate=0.1, random_state=3)
model.fit(X, y)


def get_shap_values(model, X, sample):
        model,
        X,
    )
    explanation = explainer(sample)

    expected_value = explanation.base_values[0]
    shap_values = explanation.values[0]
    return shap_values, expected_value

# Consistent for when the background data has 100 or less samples


for i in range(50, 53): # i is number of samples in each half 
    midpoint = i
    double_mid = midpoint * 2
    # shap on two halves
    shap_values1, expected_value1 = get_shap_values(model, X.loc[1:midpoint, :], X.loc[[0], :])
    shap_values2, expected_value2 = get_shap_values(model, X.loc[(midpoint+1):double_mid, :], X.loc[[0], :])
    # Shap on full background data
    shap_values, expected_value = get_shap_values(model, X.loc[1:double_mid, :], X.loc[[0], :])

    print(len(X.loc[1:midpoint, :]), len(X.loc[(midpoint+1):double_mid, :]), len(X.loc[1:double_mid, :]))
    print(shap_values, (shap_values1 + shap_values2) / 2) # inconsistent here when i > 50
```

### Traceback

_No response_

### Expected Behavior

In the for loop the values should equal each other

### Bug report checklist

### Installed Versions

0.44.0

CloseChoice:
## Overview

Closes #3461 

## Checklist

### DIFF
--- a/shap/explainers/_explainer.py
+++ b/shap/explainers/_explainer.py
@@ -1,4 +1,5 @@
 import copy
+import logging
 import time
 import warnings
 
@@ -78,6 +79,11 @@ class Explainer(Serializable):
         self.output_names = output_names
         self.feature_names = feature_names
 
+        masker_length = None
+        if hasattr(masker, '__len__'):
+            masker_length = len(masker)
+
+
         # wrap the incoming masker object as a shap.Masker object
         if (
             isinstance(masker, pd.DataFrame)
@@ -117,6 +123,10 @@ class Explainer(Serializable):
                     models.TransformersPipeline(self.model), self.masker,
                     link=link, algorithm=algorithm, output_names=output_names, feature_names=feature_names, linearize_link=linearize_link, **kwargs
                 )
+        if masker_length is not None and hasattr(self.masker, 'max_samples') and masker_length > self.masker.max_samples:
+            logging.warn(f"Your background dataset is larger than the max_samples {self.masker.max_samples}. You can hand over max_samples by explicitly "
+                         "defining the masker to use: "
+                         "`shap.Explainer(model, masker=maskers.Independent(X, max_samples=1000))`")
 
         # wrap self.masker and self.model for output text explanation algorithm
         if is_transformers_lm(self.model):


shap/explainers/_explainer.py
```diff
@@ -117,6 +123,10 @@ def __init__(self, model, masker=None, link=links.identity, algorithm="auto", ou
                     models.TransformersPipeline(self.model), self.masker,
                     link=link, algorithm=algorithm, output_names=output_names, feature_names=feature_names, linearize_link=linearize_link, **kwargs
                 )
+        if masker_length is not None and hasattr(self.masker, 'max_samples') and masker_length > self.masker.max_samples:
+            logging.warn(f"Your background dataset is larger than the max_samples {self.masker.max_samples}. You can hand over max_samples by explicitly "
+                         "defining the masker to use: "
+                         "`shap.Explainer(model, masker=maskers.Independent(X, max_samples=1000))`")
```
connortann, line 129: A few suggestions:

1. This seems possibly an odd place for the warning to be raised. Would it be more appropriate to raise it in the masker itself, where the subsampling actually occurs, for example in `Tabluar.__init__()`? I think this would be a bit tidier.

https://github.com/CloseChoice/shap/blob/b23058409eb05f5cb9e7e5cc7c59ea22b38a5695/shap/maskers/_tabular.py#L57-L58

2. It could help to adopt the common pattern of using a module-specific logger, e.g. `log = logging.getLogger(__name__)`

3. I believe `.warn` is deprecated in favour of `.warning`. 

4. I think the warning message could perhaps explain the consequences of having more samples (i.e. that the dataset will just be subsampled). Otherwise, this warning might be taken as an error that something is wrong and needs to be fixed by the user. E.g.:
```python
log.warning(
    "The background dataset exceeds the masker's `max_samples`, "
    "so will be subsampled from {n} rows to {max_samples} rows."
)
```

CloseChoice: I agree that it's an odd place but with this approach we do not need to repeat code in the maskers itself but have one place where this check happens. Would you rather duplicate the code in the maskers (I believe we currently have 2 maskers with the `max_samples` keyword)?

As for the other points I totally agree and will implement these when we've settled the point above.

connortann: Good point. On balance I'd probably prefer to log a message in both maskers. That way, the same warning is raised whether the old shap_values or the new Explainer interface is used.

One other thing actually, going back to the discussion on #3461 : I think there is a risk that excessive logging could be annoying for most users. We had originally proposed logging.INFO, but here is logging.WARNING. I think the latter will display in an IPython or Jupyter output by default.

Is this desirable? IMO the subsampling is pretty usual behaviour, so "INFO" feels more appropriate. Subsampling will be invoked probably the majority of the time, so it seems excessive to have a warning message visible on most runs of the package! I would think a visible warning would only show for things that are unusual / unintended / require user attention.

CloseChoice: to be honest, might be that our behaviour is standard but I still think it's bad practice to change user behaviour without at least an info (I would even prefer a warning and think it would be better if the user manually configures the correct `max_samples` if the provided data exceeds the default.)

connortann: >  it's bad practice to change user behaviour without ...

Please would you elaborate on this point: does the subsampling code "change user behaviour"?

CloseChoice: I mean the user selects 1000 rows, potentially a well balanced dataset and we sample this down to 100 destroying the balance. Even if something like this is standard behaviour I wouldn't consider this good practice. At least we should inform the user how to use the balanced dataset WITHOUT sampling it down.

rozeni89: How does one use background dataset without sampling it down? It seems no matter what I do the background dataset gets subsampled. 

When looking at the code (_tree.py) , almost the very first lines in the code (starting line 166) are

        masker = data
        super().__init__(model, masker, feature_names=feature_names)

        if type(self.masker) is maskers.Independent:
            data = self.masker.data
        elif masker is not None:
            raise InvalidMaskerError(f"Unsupported masker type: {str(type(self.masker))}!")

The second line subsamples the 'masker' to 100 which then gets assigned to the data. This later on gets assigned to self.data (line 182) which then used to generate baseline on line 229

         self.expected_value = self.model.predict(self.data).mean(0)

Is this the intended behavior or a bug?

CloseChoice: A workaround is given in the warning on the line you are commenting right here. Isn't that working for you?

rozeni89: No, it isn't. I am using the command "shap.TreeExplainer" instead of "shap.Explainer". Does it make a difference? I've tried many different thing and nothing seems to work. The line with "super()." seems to downsample background data no matter what you do. The default value of 100 for "max_sample" ends up being used every time. 

I think the reason is the statement on line 56 of "_tabular.py"

        if hasattr(data, "shape") and data.shape[0] > max_samples:
            data = utils.sample(data, max_samples)

Unless "max_sample" can somehow be specified by user, which I don't see how, this statement will always execute with default which is 100.

CloseChoice: It should work the same for the tree explainer as given in the code snippet. You will get the warning

                f"Passing {self.data.shape[0]} background samples may lead to slow runtimes. Consider "
                "using shap.sample(data, 100) to create a smaller background data set."
                
but nothing should be subsampled. If not, could you please provide a minimal reproducible example and create a seperate ticket for that?

rozeni89: Please see https://github.com/shap/shap/issues/3680

CloseChoice:
this just got resolved in https://github.com/shap/shap/commit/0725e7a979eb30bc28aee70bb09d59d76efba482

PR #3658, opened 2024-05-14
MAINT: Update ruff and precommit, update f-strings

connortann:
Updates ruff linter version, and fixes some f-strings to get rule UP031 passing:

https://docs.astral.sh/ruff/rules/printf-string-formatting/


CloseChoice (approved):
LGTM

PR #3686, opened 2024-06-10
MAINT: Better bug report template

connortann:
Supports #3661.


CloseChoice (approved):
Great, thanks for this! LGTM

PR #3696, opened 2024-06-17
Create c-cpp.yml

abhirajputnikku7:
Makeup

## Overview

Closes #XXXX  

## Checklist

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -1,0 +1,23 @@
+name: C/C++ CI
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: configure
+      run: ./configure
+    - name: make
+      run: make
+    - name: make check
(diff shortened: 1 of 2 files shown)


.github/workflows/c-cpp.yml
```diff
@@ -0,0 +1,23 @@
+name: C/C++ CI
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
```
CloseChoice, line 7:
```suggestion
    branches: [ "master" ]
    #Enable manual run
    workflow_dispatch:
```
Would be great if we could trigger this manually

.github/workflows/c-cpp.yml
```diff
@@ -0,0 +1,23 @@
+name: C/C++ CI
```
CloseChoice, line 1: Thanks for your PR, very happy to see some progress in this area. Since I am not the most familiar with the C extensions, would you mind adding a comment to specify what this improves on?

connortann:
This PR seems to call commands that do not exist in the repository. There is no `./configure` script, nor is there a Makefile. I'll close this for now.

Please let us know if you'd like us to take another look. If so, kindly update the description to explain the purpose of the PR.

PR #3704, opened 2024-06-20
resolve deprecated obj2sctype in numpy 2.0

paulbkoch:
## Overview

Resolves the error:

AttributeError: `np.obj2sctype` was removed in the NumPy 2.0 release. Use `np.dtype(obj).type` instead.

### DIFF
--- a/shap/plots/colors/_colorconv.py
+++ b/shap/plots/colors/_colorconv.py
@@ -816,7 +816,7 @@ def convert(image, dtype, force_copy=False, uniform=False):
     #   is a subclass of that type (e.g. `np.floating` will allow
     #   `float32` and `float64` arrays through)
 
-    if np.issubdtype(dtype_in, np.obj2sctype(dtype)):
+    if np.issubdtype(dtype_in, np.dtype(dtype).type):
         if force_copy:
             image = image.copy()
         return image


CloseChoice:
Thanks for the PR and great that you found this. Will merge if our pipelines succeed

connortann:
Thanks for the PR! I have merged the master branch into your feature branch; the CI should hopefully pass now.

I was initially surprised that this issue was not picked up by the `NPY201` linting rule that picks up deprecated functions; but, I realise we have explicitly excluded this `_colorconv.py` from linting as it is vendored code. So, thanks for spotting this.

connortann (approved):

jonathanng:
Anyone else seeing this error now?

```
/shap/plots/colors/_colorconv.py:819: DeprecationWarning: Converting `np.inexact` or `np.floating` to a dtype is deprecated. The current result is `float64` which is not strictly correct.
  if np.issubdtype(dtype_in, np.dtype(dtype).type):
```

jonathanng:
I'm using numpy==1.26.4.

connortann:
@jonathanng thanks for the bug report! That looks like a separate issue, so let's track that over at #3782

Issue #3707, opened 2024-06-21
Numpy2 Support


CloseChoice:
### Issue Description

The last merge to master failed due to numpy2 being installed and we have a lot of dependencies that cannot deal with it.

These are the packages that pinned the numpy version below 2 (reproducible by installing numpy2 on a venv where one installed shap before):
```bash
tensorflow 2.16.1 requires numpy<2.0.0,>=1.23.5; python_version <= "3.11", but you have numpy 2.0.0 which is incompatible.
scipy 1.12.0 requires numpy<1.29.0,>=1.22.4, but you have numpy 2.0.0 which is incompatible.
scikit-learn 1.4.1.post1 requires numpy<2.0,>=1.19.5, but you have numpy 2.0.0 which is incompatible.
pandas 2.2.1 requires numpy<2,>=1.22.4; python_version < "3.11", but you have numpy 2.0.0 which is incompatible.
numba 0.59.0 requires numpy<1.27,>=1.22, but you have numpy 2.0.0 which is incompatible.
```

So this is a tracker of all packages we'll wait for in order to support numpy2:
 - [x] tensorflow
 - [x] scipy
 - [x] sklearn
 - [x] pandas
 - [x] numba

see this pipeline: https://github.com/shap/shap/actions/runs/9619331932/job/26535218641

EDIT: Our problems might be bigger than I first thought. It's not just the dependencies but two more things:
 - [x] merge PR #3704 (check that the newly used function exists on previous numpy version before)
 - [ ] It seems like the `<numpy/arrayobject.h>` is no longer there, see: https://github.com/shap/shap/actions/runs/9602964617/job/26556244608?pr=3704#step:8:381. We potentially need to rewrite the c code

EDIT2: as a first step I would restrict shap to numpy < 2, so:
 - [x] merge #3708

connortann:
Thanks for looking at this - agreed, it looks like a show-stopper at the moment.

I saw numpy created a migration guide which could be useful, including things like a new Ruff linter rule to help compatibility.

https://numpy.org/doc/stable/numpy_2_0_migration_guide.html#numpy-2-migration-guide

The C ABI compatibility break looks like a bigger deal though, so yes let's stabilise the current CI by pinning numpy. It might help to make a release that contains the new version pin to numpy 1.X, to avoid broken downstream builds?

codesorcery:
There's an overview page of which libraries already support NumPy 2.0: https://github.com/numpy/numpy/issues/26191

[...]

PR #3711, opened 2024-06-24
MAINT: Add numpy compatibility linter rule

connortann:
Supports #3707

Adds numpy linter rule [NPY201](https://docs.astral.sh/ruff/rules/numpy2-deprecation/) , as recommended by numpy migration guide. It currently passes with no issues.


CloseChoice (approved):
LGTM

PR #3712, opened 2024-06-24
ENH: Add support for numpy 2, and test against numpy 1.X and 2.X

### LINKED ISSUE #3707: Numpy2 Support

connortann:
## Overview

Adds support for numpy 2.0. Closes #3707

## Changes

- Changes the build-time dependency to be numpy 2.0, rather than oldest-supported-numpy
- Adds a CI job to test against the oldest supported numpy (`1.24.0`)

## References

- https://numpy.org/doc/stable/dev/depending_on_numpy.html#numpy-2-0-specific-advice
- https://numpy.org/doc/stable/dev/depending_on_numpy.html#build-time-dependency
- https://scientific-python.org/specs/spec-0000/

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -103,3 +103,22 @@ jobs:
           if-no-files-found: ignore
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
+
+
+  oldest_supported_numpy:
+    # As recommended by https://numpy.org/doc/stable/dev/depending_on_numpy.html#numpy-2-0-specific-advice
+    # Package is built with numpy 2.X, and tested against numpy 1.X
+    name: Test oldest supported numpy
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.9
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install oldest-supported-numpy '.[test-core,plots]'
+      - name: Test with pytest
+        run: pytest --durations=20 --import-mode=append
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -105,10 +105,12 @@ jobs:
         uses: codecov/codecov-action@v3
 
 
-  oldest_supported_numpy:
-    # As recommended by https://numpy.org/doc/stable/dev/depending_on_numpy.html#numpy-2-0-specific-advice
-    # Package is built with numpy 2.X, and tested against numpy 1.X
-    name: Test oldest supported numpy
+  test_oldest_supported_numpy:
+    # Package is built with numpy 2.X, and tested against numpy 1.X.
+    # This job is recommended by https://numpy.org/doc/stable/dev/depending_on_numpy.html#numpy-2-0-specific-advice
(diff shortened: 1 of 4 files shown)


connortann:
@CloseChoice I was getting stuck debugging the C++ build until I realised I had confused the runtime numpy with the build numpy. It is the latter that we need to bump to ensure compatibility with numpy 2.

This page was very helpful:
https://numpy.org/doc/stable/dev/depending_on_numpy.html#numpy-2-0-specific-advice

It looks like eventually we should aspire to _build_ against numpy 2.0, as that will be compatible with both numpy 1.X and 2.X. I've updated our build metadata in this PR accordingly.

However, I think we need to be careful about breaking compatibility with older versions. Previously it was necessary to compile against `oldest-supported-numpy`:
https://numpy.org/doc/stable/dev/depending_on_numpy.html#build-time-dependency

connortann:
Note to self: according to SPEC 0, the oldest numpy that we should support is `1.24.0`:

https://scientific-python.org/specs/spec-0000/

We could experiment adding a test to CI to ensure that the tests pass on this older numpy version.

CloseChoice:
Hmm, to me this sounds like maybe we are lucky and just changing the version in 
[NPY_1_7_API_VERSION](https://github.com/shap/shap/blob/master/shap/cext/_cext.cc#L1) might do the trick

connortann:
> Hmm, to me this sounds like maybe we are lucky and just changing the version in [NPY_1_7_API_VERSION](https://github.com/shap/shap/blob/master/shap/cext/_cext.cc#L1) might do the trick

Good shout, done. I set this to `NPY_2_0_API_VERSION`. Note to self, this is mentioned in the numpy docs [here](https://numpy.org/doc/stable/reference/c-api/deprecations.html#deprecation-mechanism-npy-no-deprecated-api).

CloseChoice (approved):
looks great, let's get this one in.

PR #3713, opened 2024-06-24
MAINT: Bump test-core to use python 312

connortann:
A small change to switch the "test-core" CI job to use python 3.12 rather than 3.11. 

The intent of this job is to run the test suite with a minimal set of test dependencies, so that we can test against the very latest compatible version of numpy and pandas without being restricted by e.g. tensorflow compatibility.


CloseChoice (approved):
Looks good, thanks a lot.

PR #3714, opened 2024-06-24
DOCS: Add mention of SPEC 0 to maintainer docs

connortann:
In light of recent activity about numpy, I thought it would be helpful to add this little section to our maintainer docs. This PR is an opportunity to check we're agreed on this strategy.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -341,6 +341,17 @@ a given release, such as those relating to deprecation cycles.
 We use `setuptools-scm` to source the version number from the git history
 automatically. At build time, the version number is determined from the git tag.
 
+### Minimum supported dependencies
+
+We aim to follow the [SPEC 0](https://scientific-python.org/specs/spec-0000/) convention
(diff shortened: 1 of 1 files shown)


CloseChoice (approved):
LGTM, let's get this in!

CloseChoice:
This also means that we could drop support for Python 3.9 and add type hints (using pipes ;)) aggressively wherever possible. Though I believe that python3.9 is still widely used and we might mess with the code of our users doing so. 

But maybe we could add a deprecation warning for python3.9 users and drop it around october (it'll be 4 years since its release then)

connortann:
We've previously dropped support for python versions whenever a new one becomes available, such that we retain a constant number of CI jobs. We could consider doing that again: dropping 3.9 when we adopt 3.13.

It feels a little unusual to add depreciation warnings for dropped python versions, as I haven't seen that elsewhere in the python ecosystem. I don't think it's something that really requires action from users to avoid breakages, as package managers ensure that only compatible libs are installed.

PR #3715, opened 2024-06-25
Add basic type hints

### LINKED ISSUE #3373: ENH: Introduce Type Annotations incrementally
### Problem Description

Shap does not support type hints yet, but this would make the code more approachable by newcomers and easier to debug for contributors and maintainers. 
Since shap only supports Python 3.8 and onwards it would be easy to begin introducing type hints for new PRs that update functionality where type hints can be used in a straight forward fashion. In order to introduce type hints for complex classes and functions (e.g. the explainer constructor) more work needs to put into that.

Note: I do not suggest to use mypy as of now. This could be something we talk about in the future.

### Alternative Solutions

_No response_

### Additional Context

_No response_

### Feature request checklist

- ~[ ] I'd be interested in making a PR to implement this feature~

CloseChoice:
## Overview

Closes #3373 

 - add type hint checks for the test-core pipeline
 - adjust existing type hints to pass mypy call (fixing things and also using a couple ignore statements)

## Checklist


.github/workflows/run_tests.yml
```diff
@@ -85,6 +85,9 @@ jobs:
           # Use "eager" update strategy in case cached dependencies are outdated
           # Using regular install NOT editable install: see GH #3020
           pip install --upgrade --upgrade-strategy eager '.[${{ matrix.extras }},plots]'
+      - name: Run mypy
+        if: matrix.extras == 'test-core'
+        run: mypy shap
```
connortann: Would this make more sense as a separate `job` that runs in parallel?

shap/explainers/_deep/__init__.py
```diff
@@ -87,14 +85,16 @@ def __init__(self, model, data, session=None, learning_phase_flags=None):
         super().__init__(model, masker)
 
         if framework == 'tensorflow':
+            from .deep_tf import TFDeep
             self.explainer = TFDeep(model, data, session, learning_phase_flags)
         elif framework == 'pytorch':
+            from .deep_pytorch import PyTorchDeep
```
connortann, line 92: Out of interest, what is the motivation behind moving these imports?

.github/workflows/run_tests.yml
CloseChoice: absolutely, will do. Would split the `test-core` job and add mypy to that. 

Note: Will check test coverage for the plotting functions I changed, to make sure that I didn't break anything. Afterwards I'll either add more tests or would consider this as ready. Just forgot to create this as a draft PR (and don't know if I can promote it back to "draft PR")

connortann:
I support this endeavour! I saw this in the [docs](https://mypy.readthedocs.io/en/stable/existing_code.html#start-small):

> ## Start small
> If your codebase is large, pick a subset of your codebase (say, 5,000 to 50,000 lines) and get mypy to run successfully only on this subset at first, before adding annotations. This should be doable in a day or two. The sooner you get some form of mypy passing on your codebase, the sooner you benefit.

May I suggest as a first step just getting mypy running on CI, and perhaps included in the dev dependencies?

shap/explainers/_deep/__init__.py
CloseChoice: mypy complaints since previously we created empty global variables in the tensorflow repository (tf, tf_backprop, etc.). When I removed those and imported tf at the top of the deepTf file I had to move these imports down here, so that the pytorch DeepExplainer didn't need a tensorflow dependency

CloseChoice:
> I support this endeavour! I saw this in the [docs](https://mypy.readthedocs.io/en/stable/existing_code.html#start-small):
> 
> > ## Start small
> > If your codebase is large, pick a subset of your codebase (say, 5,000 to 50,000 lines) and get mypy to run successfully only on this subset at first, before adding annotations. This should be doable in a day or two. The sooner you get some form of mypy passing on your codebase, the sooner you benefit.
> 
> May I suggest as a first step just getting mypy running on CI, and perhaps included in the dev dependencies?

That's what I am trying here. Since we had a couple of type hints already, mypy checks those and I just fixed the error. So starting really small would actually mean removing the existing ones and then start implementing those from scratch. But I thought fixing the existing ones won't be that big of a PR. I underestimated this a little but still would do this in one PR, except you suggest differently.

connortann:
Makes perfect sense, fully aligned and happy with your approach in this PR.

shap/explainers/_deep/__init__.py
connortann: Ok, that makes sense. In fact I think this approach of importing the optional dependencies within the relevant function like this is simpler and consistent with #3533

.github/workflows/run_tests.yml
connortann: I think conceptually the "test-core" job belongs with the unit tests, as it requires the other common steps like uploading code coverage and running pytest.

May I suggest leaving that unit test job with the others in that matrix, and making the new job just do type checking?

connortann: I think the new job could be pretty much as simple as `pip install mypy` , `mypy shap`. It doesn't need caching or unit tests

CloseChoice: done, would you be so kind and take a final look?

.github/workflows/run_tests.yml
```diff
@@ -124,3 +124,22 @@ jobs:
           pip install numpy==1.24.0 '.[test-core,plots]'
       - name: Test with pytest
         run: pytest --durations=20 --import-mode=append
+
+  run_ypy:
```
connortann:
```suggestion
  run_mypy:
```

.github/workflows/run_tests.yml
```diff
@@ -124,3 +124,22 @@ jobs:
...
+      - name: Set up Python 3.12
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.12
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          # Use "eager" update strategy in case cached dependencies are outdated
+          # Using regular install NOT editable install: see GH #3020
+          pip install --upgrade --upgrade-strategy eager '.[test-core,plots]'
```
connortann: This can be simplified a little as we're not caching this job:

```suggestion
          pip install '.[test-core]'
```

connortann (approved):
Two minor suggestions but otherwise LGTM!

PR #3717, opened 2024-06-26
Remove deprecated parameter auto_size_plot

connortann:
## Overview

Supports #3507


CloseChoice (approved):
Looks good, we're getting ready for the release, great!

Seems like we have a flaky test.

Issue #3690, opened 2024-06-12
[Meta-issue] Increase test coverage


CloseChoice:
### Problem Description

We currently only have a pretty low coverage which makes it hard to refactor certain parts of the code base. We would like to increase the coverage to make it safe for everyone to create PRs and increase the level of safety reviewers feel when accepting PRs.

Our latest code coverage can be seen on [CodeCov.io](https://app.codecov.io/gh/shap/shap/tree/master/shap).

| File                    | statements | missing | excluded | coverage |
|-------------------------|------------|---------|----------|----------|
| shap/explainers/pytree.py|	186	 | 186	| 0	| 0% |
| shap/utils/_keras.py	| 41	| 41	| 0	| 0% |
| shap/utils/image.py	| 81	| 81	| 0	| 0% |
| shap/plots/_waterfall.py	| 316	| 306	| 0	| 3% |
| shap/plots/_benchmark.py	| 125	| 119	| 0	| 5% |
| shap/plots/_partial_dependence.py	| 151	| 143	| 0	| 5% |
| shap/plots/_decision.py	| 243	| 228	| 0	| 6% |
| shap/plots/_violin.py	| 197	| 186	| 0	| 6% |
| shap/plots/_force_matplotlib.py	| 210	| 195	| 0	| 7% |
| shap/explainers/_kernel.py	| 426	| 390	| 0	| 8% |
| shap/plots/_group_difference.py	| 46	| 42	| 0	| 9% |
| shap/maskers/_composite.py	| 99	| 89	| 0	| 10% |
| shap/explainers/_sampling.py	| 115	| 102	| 0	| 11% |
| shap/plots/_heatmap.py	| 70	| 62	| 0	| 11% |
| shap/explainers/_linear.py	| 209	| 182	| 3	| 13% |
| shap/explainers/other/_maple.py	| 166	| 145	| 0	| 13% |
| shap/plots/_utils.py	| 150	| 130	| 0	| 13% |
| shap/maskers/_image.py	| 139	| 119	| 0	| 14% |
| shap/plots/_image.py	| 115	| 99	| 0	| 14% |
| shap/actions/_optimizer.py	| 55	| 47	| 0	| 15% |
| shap/plots/_monitoring.py	| 44	| 36	| 0	| 18% |
| shap/plots/_embedding.py	| 31	| 25	| 0	| 19% |
| shap/explainers/_permutation.py	| 91	| 73	| 0	| 20% |
[...]

PR #3720, opened 2024-06-27
test decision plot

CloseChoice:
## Overview

Works towards #3690 and predecessor to PR #3715  

 - add tests to decision plot
 - increase test coverage of `_decision.py` from 6% to 79%

## Checklist


shap/plots/_decision.py
```diff
@@ -582,6 +583,7 @@ def multioutput_decision(base_values, shap_values, row_index, **kwargs) -> Union
         Returns a DecisionPlotResult object if `return_objects=True`. Returns `None` otherwise (the default).
 
     """
+    # todo: adjust to breaking changes made in #3318
```
connortann, line 599: Do we need an issue to track this to-do? Is any functionality broken here at the moment?

tests/plots/test_decision.py
```diff
@@ -15,7 +17,108 @@ def test_random_decision(random_seed):
         rs.standard_normal(size=(20, 5)),
         show=False
     )
-    pl.close()
+    plt.close()
```
connortann: Nitpick but I don't we need to bother closing the plots, as that's done by a fixture in `tests/plots/contest.py`

tests/plots/test_decision.py
```diff
@@ -15,7 +17,108 @@ def test_random_decision(random_seed):
...
+    shap.decision_plot(result_values.base_values[0, 1],
+                       result_values.values[:, :, :, 1],
+                       features=X,
+                       show=False,
+                       )
+    plt.tight_layout()
+    return fig
+
+@pytest.fixture
+def values_features():
```
connortann: Can this fixture also be used in the tests above?

tests/plots/test_decision.py
```diff
@@ -15,7 +17,108 @@ def test_random_decision(random_seed):
...
+                                       )
+    with pytest.raises(ValueError, match="The shap_values arg should be a list of two or three dimensional SHAP arrays."):
+        adult_rfc_shap_values_list = [adult_rfc_shap_values.values[:, 0, i] for i in range(adult_rfc_shap_values.shape[2])]
+        base_values_list = list(adult_rfc_shap_values.base_values[0, :])
+        shap.multioutput_decision_plot(base_values_list,
+                                       adult_rfc_shap_values_list,
+                                       row_index=0,
+                                       )
+
+# (base_values, shap_values, row_index, **kwargs)
```
connortann, line 120: There seems to be a lot of commented-out code below. Shall we get this running in the test suite?

It would be interesting to know where it came from and why it was commented out; I'll look at the commit history..

notebooks/api_examples/plots/decision_plot.ipynb
connortann, line 1: As a nitpick, may I suggest running all cells from top to bottom with a fresh kernel?

I was thinking of adding a pre commit hook to test for this, as per #3721

connortann (changes requested):
Looks great, a few suggestions to keep things tidy

shap/plots/_decision.py
CloseChoice: The functionality is not broken but pretty uncomfortable for the user but only in the case where we previously returned a list
So previously
```python
shap_values = explainer(X)
decision_plot(shap_values.base_values[0], shap_values)
```
worked but now one needs to construct these lists from the array (see the code I needed to test this: https://github.com/shap/shap/pull/3720/files#diff-a1e81e6553432a82ff71dd49ecdfeaf7a11d0af3a500605c303f15ab81fcee3bR95)

So this could definitely use a clean up

tests/plots/test_decision.py
CloseChoice: done

tests/plots/test_decision.py
CloseChoice: done

tests/plots/test_decision.py
CloseChoice: I removed the ones I wrote. I think we could try and this will probably show us what's not working here (anymore), guess there is a reason why it's commented out. I wouldn't do this in this PR though

notebooks/api_examples/plots/decision_plot.ipynb
CloseChoice: I found that there are a couple more issues with the notebook further down which I did not really look into, so I removed it from this PR.

connortann (approved):
LGTM!

Unfortunately this may cause some merge conflicts with the impending #3722 . I think it would be easiest to review/merge that one first, and then I'd be happy to resolve any conflicts on this branch and do the merge. I think that would be easier than resolving the conflicts on the larger messier PR

CloseChoice:
alright, will merge this then

PR #3721, opened 2024-06-28
Add pre commit hook to check notebook order

connortann:
Add pre commit hook, to help ensure notebook cells are executed cleanly in order.

This should help ensure we remember to "restart and run all cells", which helps us ensure the notebooks are reproducible and don't contain hidden state.


CloseChoice (approved):
LGTM

PR #3722, opened 2024-06-28
Maint: apply ruff formatter to all python files

### LINKED ISSUE #683: Blacken source code?
Black is emerging as the standard code formatter for Python (https://github.com/python/black, https://www.youtube.com/watch?v=esZLCuWs_2Y)

Are you open to reformatting the code using this? If so, I can submit a PR for it.

connortann:
## Overview

Applies the "python black" style to all python files, via the ruff formatter.

Closes #683.

## Discussion

Now is as good a time as any to adopt an autoformatter, to help ease the dev process of future PRs. This autoformatter is already applied to jupyter notebooks; this removes the exclusion of `.py` files.

Line length chosen as per discussion on https://github.com/shap/shap/issues/683#issuecomment-2197120343

A follow-up PR will be needed to add this commit has to `.git-blame-ignore-revs`, to ensure the `git blame` command is not adversely affected.


connortann:
@CloseChoice it would be useful if we could review & close this PR reasonably swiftly to minimise the chance of merge conflicts, as this PR necessarily has a very large number of changed files.

CloseChoice:
Okay, had a look over this and it looks pretty standard. Some asserts are broken a bit ugly IMO, but nothing too bad, so I would just go for the merge. 

This is just applying the formatter, right? So I believe reviewing this PR is more like getting a feel for how it looks rather than going over all changes explicitly

connortann:
> Some asserts are broken a bit ugly IMO, but nothing too bad

Yeah I agree. Pragmatically, I think we should do the merge now and then at our leisure we can fixup any strange splits over time.

> This is just applying the formatter, right?

yep.

PR #3724, opened 2024-06-30
MAINT: Update git-blame-ignore-revs

connortann:
Adding some recent reformatting commits to `.git-blame-ignore-revs`, to help preserve the utility of the "blame" view in GitHub and the `git blame` tool.


CloseChoice (approved):
as dicussed, totally support this change

PR #3727, opened 2024-07-01
MAINT: Bump ruff, fix E721 linting errors

connortann:
Update the ruff linter, and fix a few linting issues due to comparison of types. Also add a minor rule from flake8-simplify.

### DIFF
--- a/shap/explainers/_deep/deep_pytorch.py
+++ b/shap/explainers/_deep/deep_pytorch.py
@@ -26,7 +26,7 @@ class PyTorchDeep(Explainer):
         self.interim = False
         self.interim_inputs_shape = None
         self.expected_value = None  # to keep the DeepExplainer base happy
-        if type(model) == tuple:
+        if isinstance(model, tuple):
             self.interim = True
             model, layer = model
             model = model.eval()
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ exclude: '.*tree_shap_paper.*|.*user_studies.*'
 
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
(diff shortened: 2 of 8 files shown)


shap/explainers/_deep/deep_pytorch.py
```diff
@@ -26,7 +26,7 @@ def __init__(self, model, data):
         self.interim = False
         self.interim_inputs_shape = None
         self.expected_value = None  # to keep the DeepExplainer base happy
-        if type(model) == tuple:
+        if isinstance(model, tuple):
```
CloseChoice, line 29: this is an E721 fix? If so I am missing this error code in E721 to be ignored previously or why is this fix in here?

connortann: I believe the behaviour of this rule has changed slightly in ruff 5.0:

> [type-comparison](https://docs.astral.sh/ruff/rules/type-comparison/) (E721) now allows idioms such as type(x) is int

connortann:
On the topic of ruff rules I was looking into flake8-simplify, which has a few potentially useful rules. I added one rule to simplify `isinstance` calls. I was initially considering getting further rules passing, but they seemed less helpful / more pedantic than I originally hoped so I'll just leave it at the one.

CloseChoice (approved):
LGTM

CloseChoice:
> On the topic of ruff rules I was looking into flake8-simplify, which has a few potentially useful rules. I added one rule to simplify `isinstance` calls. I was initially considering getting further rules passing, but they seemed less helpful / more pedantic than I originally hoped so I'll just leave it at the one.

Sounds reasonable. I like our current approach: taking the useful side of things but not being pedantic about these things.

PR #3728, opened 2024-07-01
MAINT: Incrementally update mypy config

connortann:
A slight marginal update to the mypy config:
- `ignore missing imports` only for explicit 3rd party libraries, but not globally. That way, incorrect shap imports should be flagged.
- Enable `check-uptyped-defs` globally and then opt-out for specific modules that current fail (which is most!), to set us up to progressively get it passing by module

### DIFF
--- a/shap/plots/_image.py
+++ b/shap/plots/_image.py
@@ -1,7 +1,7 @@
 import json
 import random
 import string
-from typing import Optional, Union, cast
+from typing import Literal, Optional, Union, cast
 
 import matplotlib.pyplot as pl
 import numpy as np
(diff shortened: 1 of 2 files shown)


shap/plots/_image.py
```diff
@@ -27,7 +27,7 @@ def image(
     true_labels: Optional[list] = None,
     width: Optional[int] = 20,
     aspect: Optional[float] = 0.2,
-    hspace: Optional[float] = 0.2,
+    hspace: Union[Optional[float], Literal["auto"]] = 0.2,
```
CloseChoice, line 30: ah, didn't now about literal. That will come in handy in a lot of places

CloseChoice (approved):
LGTM

PR #3555, opened 2024-03-08
Refactor deep explainer tests and relax allowed tolerance

CloseChoice:
## Overview

Description of the changes proposed in this pull request:
As mentioned in the title: Refactor deep explainer tests and relax allowed tolerance

## Checklist

- [ ] All [pre-commit checks](https://pre-commit.com/#install) pass.
- [ ] Unit tests added (if fixing a bug or adding a new feature)

PR #3729, opened 2024-07-01
MAINT: enable mypy on tests

connortann:
Run mypy on `./tests` as well as `./shap`.

- Excludes the `tests/benchmark` dir, as it is not currently working :(
- Fixes a numpy bug in `test_tf_keras_imdb_lstm` , a test which is currently skipped. (The first genuine bug mypy has found!)
- Adds a few type hints to help mypy pass
- Adds the numpy mypy plugin, which should help when working with arrays.

### DIFF
--- a/tests/explainers/test_deep.py
+++ b/tests/explainers/test_deep.py
@@ -253,7 +253,7 @@ def test_tf_keras_imdb_lstm(random_seed):
     diff = sess.run(mod.layers[-1].output, feed_dict={mod.layers[0].input: testx})[0, :] - sess.run(
         mod.layers[-1].output, feed_dict={mod.layers[0].input: background}
     ).mean(0)
-    np.testing.testing_allclose(sums, diff, atol=1e-02), "Sum of SHAP values does not match difference!"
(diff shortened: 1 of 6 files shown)


tests/explainers/test_deep.py
```diff
@@ -253,7 +253,7 @@ def test_tf_keras_imdb_lstm(random_seed):
     diff = sess.run(mod.layers[-1].output, feed_dict={mod.layers[0].input: testx})[0, :] - sess.run(
         mod.layers[-1].output, feed_dict={mod.layers[0].input: background}
     ).mean(0)
-    np.testing.testing_allclose(sums, diff, atol=1e-02), "Sum of SHAP values does not match difference!"
+    np.testing.assert_allclose(sums, diff, atol=1e-02), "Sum of SHAP values does not match difference!"
```
connortann, line 256: I think this was a typo that crept through in #3555 . The test is skipped for recent tensorflow versions so it slipped through the net.

In any case, nice to see mypy picking up a genuine issue rather than just false positives!

CloseChoice (approved):
Looks good.

Issue #3730, opened 2024-07-01
MAINT: Enable type checking on all modules with mypy's `check_untyped_defs` setting enabled


connortann:
We would like the mypy type checker to run successfully on the shap codebase with the setting `check_untyped_defs = true`.

However, at present this does not pass on many submodules, so they opt out from this configuration setting:

https://github.com/shap/shap/blob/86b6ee0236219723fc73cffaac48619f43beaba1/pyproject.toml#L104-L108

When the typing issues in a given module are fixed, the relevant exclusion in `pyproject.toml` should be removed.

Checklist of modules to fix:

- [x] `__init__.py`
- [x] `_explanation.py`
- [x] `_serializable.py`
- [x] `_version.py`
- [x] `actions/__init__.py`
- [x] `actions/_action.py`
- [x] `actions/_optimizer.py`
- [x] `datasets.py`
- [x] `explainers/__init__.py`
- [ ] `explainers/_additive.py`
- [x] `explainers/_deep/__init__.py`
- [ ] `explainers/_deep/deep_pytorch.py`
- [ ] `explainers/_deep/deep_tf.py`
- [x] `explainers/_deep/deep_utils.py`
- [ ] `explainers/_exact.py`
- [ ] `explainers/_explainer.py`
- [x] `explainers/_gpu_tree.py`
- [ ] `explainers/_gradient.py`
- [ ] `explainers/_kernel.py`
- [ ] `explainers/_linear.py`
- [ ] `explainers/_partition.py`
- [ ] `explainers/_permutation.py`
- [ ] `explainers/_sampling.py`
- [ ] `explainers/_tree.py`
- [x] `explainers/other/__init__.py`
- [x] `explainers/other/_coefficient.py`
- [x] `explainers/other/_lime.py`
- [x] `explainers/other/_maple.py`
- [ ] `explainers/other/_random.py`
- [x] `explainers/other/_treegain.py`
- [x] `explainers/other/_ubjson.py`
- [ ] `explainers/pytree.py`
- [ ] `explainers/tf_utils.py`
- [x] `links.py`
- [x] `maskers/__init__.py`
- [ ] `maskers/_composite.py`
- [x] `maskers/_fixed_composite.py`
- [ ] `maskers/_fixed.py`
- [ ] `maskers/_image.py`
- [ ] `maskers/_masker.py`
- [x] `maskers/_output_composite.py`
- [ ] `maskers/_tabular.py`
- [ ] `maskers/_text.py`
[...]

PR #3732, opened 2024-07-02
Enable type checking for _explaination.py

connortann:
Supports #3730

Gets mypy passing on `_explanation.py` with `check-untyped-defs=True`, and adds a couple of type hints.

### DIFF
--- a/shap/_explanation.py
+++ b/shap/_explanation.py
@@ -1,6 +1,6 @@
 import copy
 import operator
-from typing import Any
+from typing import Any, Union
 
 import numpy as np
 import pandas as pd
@@ -104,7 +104,9 @@ class Explanation(metaclass=MetaExplanation):
         values_shape = _compute_shape(values)
 
         if output_names is None and len(self.output_dims) == 1:
-            output_names = [f"Output {i}" for i in range(values_shape[self.output_dims[0]])]
+            num_names = values_shape[self.output_dims[0]]
+            assert num_names is not None, "Unexpected shape of values"
+            output_names = [f"Output {i}" for i in range(num_names)]
 
         if (
             len(_compute_shape(feature_names)) == 1
@@ -159,9 +161,10 @@ class Explanation(metaclass=MetaExplanation):
         )
 
     @property
-    def shape(self):
+    def shape(self) -> tuple[Union[int, None], ...]:
         """Compute the shape over potentially complex data nesting."""
-        return _compute_shape(self._s.values)
+        shape = _compute_shape(self._s.values)
+        return shape
 
     @property
     def values(self):
@@ -297,7 +300,7 @@ class Explanation(metaclass=MetaExplanation):
             out += "\n\n.data =\n" + self.data.__repr__()
         return out
 
-    def __getitem__(self, item):
+    def __getitem__(self, item) -> "Explanation":
         """This adds support for OpChain indexing."""
         new_self = None
         if not isinstance(item, tuple):
@@ -399,7 +402,7 @@ class Explanation(metaclass=MetaExplanation):
         # call slicer for the real work
         item = tuple(v for v in item)  # SML I cut out: `if not isinstance(v, str)`
         if len(item) == 0:
-            return new_self
+            return new_self  # type: ignore
         if new_self is None:
             new_self = copy.copy(self)
         new_self._s = new_self._s.__getitem__(item)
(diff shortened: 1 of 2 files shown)


shap/_explanation.py
```diff
@@ -158,9 +161,10 @@ def __init__(
         )
 
     @property
-    def shape(self):
+    def shape(self) -> tuple[Union[int, None], ...]:
```
connortann, line 164: I am a bit unsure about this one: should it be a tuple of `int`, or a tuple or `int|None` ?

The function `_compute_shape` seems to return a tuple that may contain `None`; however, the attribute `.shape` seems to be used in contexts that always expect an integer.

shap/_explanation.py
```diff
@@ -398,7 +402,7 @@ def __getitem__(self, item):
         # call slicer for the real work
         item = tuple(v for v in item)  # SML I cut out: `if not isinstance(v, str)`
         if len(item) == 0:
-            return new_self
+            return new_self  # type: ignore
```
connortann, line 405: Mypy complains here that `new_self` may be `None`. For now I have supressed the error, but it would be good to get a second opinion if the return type of this func should be `Explanation | None`. This would cause some further typing issues though.

shap/_explanation.py
CloseChoice: Two things here: we cannot use the pipe like `int | None` yet, since we still support Python 3.9 (this syntax is only available in Python 3.10).

Yes, shape should return an integer but if it doesn't return this as of now, then I would argue that we just add a todo in this PR and do not confuse PRs that solely add type annotations with fixing issues.

shap/_explanation.py
CloseChoice: I think it should only return an `Explanation` here but for this PR I think it's fine. 

I would just try to aggressively improve the coverage of mypy of the repo even if this means that we'll ignore some lines explicitely. In a second run I would go over the ignore statements and remove those one after the other.

Just my two cents on this.

shap/_explanation.py
```diff
@@ -776,7 +782,7 @@ def _first_item(x):
     return None
 
 
-def _compute_shape(x):
+def _compute_shape(x) -> tuple[Union[int, None], ...]:
```
CloseChoice, line 785: A general question here: I found that you only added type hints for the return types not the inputs parameters, right? Is there a particular reason for this? I could imagine that it's easier iterate this way, but happy to understand this deeper.

Don't get me wrong, just want to understand this. No need to add type hints for input parameters in here.

CloseChoice (commented):
This is a good first step. I begin to realize that adding type hints will be with us for quite a while.

shap/_explanation.py
connortann: Good question. There's no particular reason, I'm just starting with a small manageable chunk.

In a draft PR #3734 I was thinking about how to create a type hint for variables like `X` : e.g. "ArrayLike", "sliceable". I think that's a bit of a larger issue, so I thought I'd leave that for later for now and come back to it.

shap/_explanation.py
connortann: > we cannot use the pipe like `int | None`

Yep, that was just my shorthand in my comment above, but in the PR I've used the full `Union` syntax.

> shape should return an integer

Presume you mean a tuple of integers. Are you sure this should always be the case? I noticed on lines 788 & 816 that entries may be `None` if the array is sparse.

shap/_explanation.py
connortann: Sounds good!

shap/_explanation.py
```diff
@@ -866,12 +872,12 @@ def _auto_cohorts(shap_values, max_cohorts):
                         name += " >= "
                     name += str(threshold) + " & "
         path_names.append(name[:-3])  # the -3 strips off the last unneeded ' & '
-    path_names = np.array(path_names)
+    path_names_arr = np.array(path_names)
```
connortann, line 875: There have been a few cases like this where mypy complains if a variable starts life as e.g. a list and then is converted to an array. A simple fix is just to give one of the variables a different name.

shap/_explanation.py
CloseChoice: sounds good, I am all for the approach to cut things down into manageable chunks.

shap/_explanation.py
CloseChoice: yes, I think this is actually a useful feature of mypy since it is the cause of a lot of bugs I've seen/produced

shap/_explanation.py
CloseChoice: I am not sure tbh. So I wouldn't change this but just stick to what is implemented as of now.

Then in a later iteration we can check suspicious type hints and go over those pieces of the code.

connortann: OK - I'll add a TODO comment to remind myself to come back to this one. For now, I've used an `assert` on lines 108 and 639 to validate that the shape is not None

CloseChoice (approved):
LGTM

PR #3733, opened 2024-07-02
Enable type checking for shap/actions

connortann:
## Overview

Supports https://github.com/shap/shap/issues/3730

Gets mypy passing on shap/actions/ with check-untyped-defs=True, and adds a couple of type hints.

## Discussion

This was the first time I've read this module. I'm not clear what the original context or purpose was, and it might even be "dead code". If we can confirm with Scott, perhaps we might want to remove this from a future release. But for now, I've added a few hints which hopefully improve the clarity and get it passing on mypy.

### DIFF
--- a/shap/actions/_optimizer.py
+++ b/shap/actions/_optimizer.py
@@ -1,6 +1,7 @@
 import copy
 import queue
 import warnings
+from typing import Any
 
 from ..utils._exceptions import ConvergenceError, InvalidAction
 from ._action import Action
@@ -9,9 +10,9 @@ from ._action import Action
 class ActionOptimizer:
     def __init__(self, model, actions):
         self.model = model
-        warnings.warn("Note that ActionOptimizer is still in an alpha state and is subjust to API changes.")
+        warnings.warn("Note that ActionOptimizer is still in an alpha state and is subject to API changes.")
         # actions go into mutually exclusive groups
-        self.action_groups = []
+        self.action_groups: list[Any] = []
         for group in actions:
             if issubclass(type(group), Action):
                 group._group_index = len(self.action_groups)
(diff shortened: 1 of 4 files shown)


shap/actions/_optimizer.py
```diff
@@ -1,23 +1,24 @@
...
+    def __init__(self, model, actions: list[Union[Action, list[Action]]]):
         self.model = model
-        warnings.warn("Note that ActionOptimizer is still in an alpha state and is subjust to API changes.")
+        warnings.warn("Note that ActionOptimizer is still in an alpha state and is subject to API changes.")
         # actions go into mutually exclusive groups
-        self.action_groups = []
+        self.action_groups: list[list[Action]] = []
         for group in actions:
-            if issubclass(type(group), Action):
+            if isinstance(group, Action):
```
connortann, line 17: I believe `isinstance(foo, Bar)` is identically equivalent to `issubclass(type(foo), Bar)`, but it would be useful to have a second opinion on this. It seems a bit bizarre that it was written this way originally, so I wonder if there might have been a specific reason or edge case.

CloseChoice: Hmm, the only thing that I could think of is to not match instances of `Action` but just subclasses of it but `subclass` does not behave like this

```python
>>> class A:
...     pass
... 
>>> class B(A):
...     pass
... 
>>> a = A()
>>> b = B()
>>> issubclass(type(b), A)
True
>>> issubclass(type(a), A) 
True  # I expected this to be False
```

So I guess the only reason here was that someone didn't know better, especially when I consider the usage in here.

CloseChoice (approved):
Nice work, I think we can merge this

connortann:
Ok will do. Out of interest, have you seen this module used anywhere in shap - do you have any insight as to what the purpose is?

PR #3734, opened 2024-07-02
Enable type checking for shap/utils, and improve OpChain docs

connortann:
Supports https://github.com/shap/shap/issues/3730

Gets mypy passing on shap/utils with check-untyped-defs=True, and adds a couple of type hints.

### DIFF
--- a/shap/utils/_clustering.py
+++ b/shap/utils/_clustering.py
@@ -1,5 +1,5 @@
 import warnings
-from typing import Literal, Optional
+from typing import Literal, Optional, Union
 
 import numpy as np
 import pandas as pd
@@ -101,7 +101,13 @@ def hclust_ordering(X, metric="sqeuclidean", anchor_first=False):
 
 
 def xgboost_distances_r2(
-    X, y, learning_rate=0.6, early_stopping_rounds=2, subsample=1, max_estimators=10000, random_state=0
+    X,
+    y,
+    learning_rate: float = 0.6,
+    early_stopping_rounds: Optional[int] = 2,
+    subsample: Optional[float] = 1.0,
+    max_estimators: Optional[int] = 10000,
+    random_state: Union[int, np.random.RandomState] = 0,
 ):
     """Compute reducancy distances scaled from 0-1 among all the feature in X relative to the label y.
 
@@ -118,8 +124,8 @@ def xgboost_distances_r2(
     X_train, X_test, y_train, y_test = sklearn.model_selection.train_test_split(X, y, random_state=random_state)
 
     # fit an XGBoost model on each of the features
-    test_preds_list = []
     train_preds_list = []
+    test_preds_list = []
     for i in range(X.shape[1]):
         model = xgboost.XGBRegressor(
             subsample=subsample,
@@ -176,7 +182,7 @@ def hclust(
     y: Optional[_ArrayLike] = None,
     linkage: Literal["single", "complete", "average"] = "single",
     metric: str = "auto",
-    random_state: int = 0,
+    random_state: Union[int, np.random.RandomState] = 0,
 ) -> np.ndarray:
     """Fit a hierarcical clustering model for features X relative to target variable y.
 
@@ -262,4 +268,4 @@ def hclust(
     elif linkage == "average":
         return scipy.cluster.hierarchy.average(dist)
     else:
-        raise Exception("Unknown linkage: " + str(linkage))
+        raise ValueError("Unknown linkage: " + str(linkage))
--- a/shap/utils/_clustering.py
+++ b/shap/utils/_clustering.py
@@ -1,4 +1,5 @@
 import warnings
+from typing import Literal, Optional
 
 import numpy as np
 import pandas as pd
@@ -8,6 +9,7 @@ import sklearn
 from numba import njit
 
 from ._show_progress import show_progress
+from ._types import _ArrayLike
 
 
 def partition_tree(X, metric="correlation"):
@@ -116,8 +118,8 @@ def xgboost_distances_r2(
     X_train, X_test, y_train, y_test = sklearn.model_selection.train_test_split(X, y, random_state=random_state)
 
     # fit an XGBoost model on each of the features
-    test_preds = []
-    train_preds = []
+    test_preds_list = []
+    train_preds_list = []
     for i in range(X.shape[1]):
         model = xgboost.XGBRegressor(
             subsample=subsample,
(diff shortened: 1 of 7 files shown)


shap/utils/_types.py
```diff
@@ -0,0 +1,8 @@
+from typing import TypeAlias, TypeVar, Union
+
+import numpy as np
+import pandas as pd
+import scipy.sparse
+
+_ArrayLike: TypeAlias = Union[np.ndarray, pd.DataFrame, pd.Series, list, scipy.spare.csr_matrix]
+_ArrayT = TypeVar("_ArrayT", np.ndarray, pd.DataFrame, pd.Series, scipy.sparse.csr_matrix, list)
```
connortann: I thought it would be useful to have a new submodule for common type aliases, so here's a start.

I'm aware numpy has `numpy.typing.ArrayLike` which represents (anything that can be coerced into a numpy array), but I think that's too broad for what we need to represent datasets as it includes things like `float` and `int`.

Just to draw attention to something I found a bit subtle: the `Union` and `TypeVar` are very similar, but the latter was useful to annotate that a function will return exactly the same type as the input type.

More info: https://stackoverflow.com/a/58903907

Example use in `_general.py : sample()`

shap/utils/_legacy.py
```diff
@@ -93,7 +93,6 @@ def match_instance_to_data(instance, data):
                 instance.x[0, group[0]] if len(group) == 1 else "" for group in data.groups
             ]
         assert len(instance.group_display_values) == len(data.groups)
-        instance.groups = data.groups
```
connortann, line 96: Nb. the `instance` class has no attribute `groups` and this does not seem to be used anywhere, so I think this line is not needed.

connortann:
This has some new mypy failures now that `_explanation` and `actions` have more types, so i'll take another look. In particular, I want to get more familiar with `OpChain` and the behaviour of the custom metaclass.

Edit: done

shap/utils/_general.py
```diff
@@ -302,10 +305,10 @@ def __getitem__(self, item):
         new_self._ops.append(["__getitem__", [item], {}])
         return new_self
 
-    def __getattr__(self, name):
+    def __getattr__(self, name: str) -> "OpChain":
         # Don't chain special attributes
         if name.startswith("__") and name.endswith("__"):
-            return None
+            return None  # type: ignore
```
connortann, line 311: This `return None` appears to be a special case that was put in to avoid issues with ReadTheDocs when `inspect` is used to lookup special attributes of objects. It doesn't seem to be invoked outside of building docs, so I think it is ok to ignore the type error here.

connortann:
I added some comments to the `Explanation` docstring, as personally I found the behaviour of the class methods and instance methods rather confusing at first.

shap/utils/_masked_model.py
```diff
@@ -323,7 +324,7 @@ def _build_delta_masked_inputs(
         "If you rely on this function, please open an issue: https://github.com/shap/shap/issues.",
         DeprecationWarning,
     )
-    all_masked_inputs = [[] for a in args]
+    all_masked_inputs: list[list[Any]] = [[] for a in args]
```
thatlittleboy:
```suggestion
    all_masked_inputs: list[list[Any]] = [[] for _ in args]
```

shap/utils/_types.py
```diff
@@ -0,0 +1,9 @@
+from typing import TypeVar, Union
+
+import numpy as np
+import pandas as pd
+import scipy.sparse
+
+# TODO: use TypeAlias (when we drop python 3.9)
+_ArrayLike = Union[np.ndarray, pd.DataFrame, pd.Series, list, scipy.sparse.spmatrix]
+_ArrayT = TypeVar("_ArrayT", np.ndarray, pd.DataFrame, pd.Series, scipy.sparse.spmatrix, list)
```
thatlittleboy, line 9: Any reason why we're going for `_ArrayT` instead of `ArrayT`? (If it's for keeping it "private", I'm wondering if that's necessary?)

EDIT: Is it because we are following what `numpy` is doing?

thatlittleboy (commented):
I think in general it looks great (great that we're finally adding types to shap! :))

Just one question / discussion point to be had.

shap/utils/_clustering.py
```diff
@@ -116,8 +118,8 @@ def xgboost_distances_r2(
     X_train, X_test, y_train, y_test = sklearn.model_selection.train_test_split(X, y, random_state=random_state)
 
     # fit an XGBoost model on each of the features
-    test_preds = []
-    train_preds = []
+    test_preds_list = []
+    train_preds_list = []
```
thatlittleboy: Very very minor nitpick, but could we shift the initialization of `train_preds_list` on top of `test_preds_list`?

shap/utils/_clustering.py
```diff
@@ -169,7 +171,13 @@ def xgboost_distances_r2(
     return dist
 
 
-def hclust(X, y=None, linkage="single", metric="auto", random_state=0):
+def hclust(
+    X: _ArrayLike,
+    y: Optional[_ArrayLike] = None,
+    linkage: Literal["single", "complete", "average"] = "single",
+    metric: str = "auto",
+    random_state: int = 0,
```
thatlittleboy: not a dealbreaker, but shall we also include the `RandomState` and None as part of the allowed types for `random_state` (it's not just int).

(ps: Might even be worth doing a TypeAlias, since this pops up quite often in multiple places)

shap/utils/_types.py
connortann: By convention, a leading underscore designates a private object (strictly speaking "non-public"), that is not part of the public API for users. These objects are typically hidden from IDE tool tips and IPython auto complete.

Whilst public interfaces must be tested and protected from sudden breaking changes, private objects are for internal use only and can be refactored more freely.

It's helpful for us as maintainers to minimise the public API and keep any helper functions / variables as private, as it allows us to refactor the internal code more freely.

thatlittleboy: Sure, I appreciate the difference between private and public API, let's leave it as such since numpy is doing the same.

shap/utils/_clustering.py
connortann: Good shout, done. I decided against a type alias as it's a pretty simple type, and I think writing it in-line has better [Locality of Behaviour](https://htmx.org/essays/locality-of-behaviour/)

connortann:
Thanks for the review @thatlittleboy . I've implemented your feedback, would you kindly take another look?

thatlittleboy (approved):
LGTM

PR #3738, opened 2024-07-04
Fix TypeError in summary_plot 

### LINKED ISSUE #2893: TypeError: only integer scalar arrays can be converted to a scalar index
If anyone has the same problem in `_beeswarm.py`, you can try this code:
`before`
```
summary_legacy(
            proj_shap_values, features[:, sort_inds] if features is not None else None,
            feature_names=np.array(feature_names)[sort_inds].tolist(),
            sort=False, show=False, color_bar=False,
            plot_size=None,
            max_display=max_display
        )
```
`after`
```
summary_legacy(
            proj_shap_values, features[:, sort_inds] if features is not None else None,
            feature_names=np.array(feature_names)[sort_inds].tolist(), #new
            sort=False, show=False, color_bar=False,
            plot_size=None,
            max_display=max_display
        )
```

bedapisl:
## Overview

Closes #2893

**Description of the changes proposed in this pull request:**

Small fix in `summary_plot` that removes a `TypeError` caused by a specific input.

### DIFF
--- a/tests/plots/test_summary.py
+++ b/tests/plots/test_summary.py
@@ -1,3 +1,4 @@
+import lightgbm
 import matplotlib.pyplot as plt
 import numpy as np
 import pytest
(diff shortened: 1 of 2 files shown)


tests/plots/test_summary.py
```diff
@@ -121,3 +122,16 @@ def test_random_summary_with_log_scale():
     shap.summary_plot(np.random.randn(20, 5), use_log_scale=True, show=False)
     fig.set_layout_engine("tight")
     return fig
+
+
+def test_no_type_error_in_summary_plot():
```
connortann: Suggest adding a quick comment to link this test to the original issue:

```suggestion
def test_summary_plot_with_multiclass_model():
    # See GH #2893
```

connortann (approved):
Looks good to me, thanks for the PR! I pushed a small edit to fixup the lightgbm import, so the test is skipped if the package is not installed.

connortann:
The type check is currently failing due to an unrelated issue due to a recent release of matplotlib. That will be fixed by #3740 ; once that's merged, I will go ahead and merge this PR 👍

PR #3739, opened 2024-07-04
Deprecate the legacy bar plot, add migration guide to new Explainer API

connortann:
Supports #3156.

- Adds a deprecation warning to the legacy bar plot.
- Adds a notebook to explain the difference between the old- and new-style API

### DIFF
--- a/notebooks/api_examples/migrating-to-new-api.ipynb
+++ b/notebooks/api_examples/migrating-to-new-api.ipynb
@@ -1,0 +1,142 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Migrating to the new \"Explanation\" API\n",
+    "\n",
+    "This notebook explains some of the differences between the old- and new-style shap API.\n",
+    "\n",
+    "In short:\n",
+    "\n",
+    "```python\n",
+    "shap_values = explainer.shap_values(X)  # Old style\n",
+    "shap_values = explainer(X)              # New style\n",
+    "```"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Create an example dataset and model\n",
+    "import xgboost\n",
+    "\n",
+    "import shap\n",
+    "\n",
+    "X, y = shap.datasets.adult(n_points=100)\n",
+    "model = xgboost.XGBClassifier().fit(X, y)\n",
+    "explainer = shap.TreeExplainer(model, X)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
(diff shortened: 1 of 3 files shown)


connortann:
@CloseChoice  FYI I noticed the new "run mypy" job failed on this PR, which is strange as the failure does not seem to relate to any files modified in this branch.

I cannot reproduce the mypy failures locally, which is a bit concerning. Perhaps the type checking is flaky, or perhaps it's due to an updated version of a dependency?

EDIT: it seems to be an error that mypy now picks up thanks to some new type hints in matplotlib 3.9.1. PR to fix:
#3740

shap/plots/_bar.py
```diff
@@ -392,6 +392,13 @@ def bar(
 
 
 def bar_legacy(shap_values, features=None, feature_names=None, max_display=None, show=True):
+    warnings.warn(
+        "The behaviour of this function will change in a future version to the new plotting API."
```
CloseChoice, line 396: will this change or is this a candidate for removal?

notebooks/api_examples/migrating-to-new-api.ipynb
```diff
@@ -0,0 +1,159 @@
...
+   "source": [
+    "explanation = explainer(X)\n",
+    "explanation[:2]  # a shap.Explanation object"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The `Explanation` object is a much richer representation that includes the shap values\n",
```
CloseChoice: I would add a cell where it is explained that using `explanation.values[:2]` yields the same result as the calculation using the old style.

CloseChoice (approved):
1 question, 1 nitpick. LGTM

shap/plots/_bar.py
connortann: That's a good question, I was thinking the same. I propose switching `shap.bar_plot` to point to `shap.plots.bar`. Does that sound good to you?

notebooks/api_examples/migrating-to-new-api.ipynb
connortann: What is the purpose of the `[:2]` indexing, is the shape different?

CloseChoice: nope, it is just done for the .shap_values call as well. One could then make an equal vomparison or just state that the vlaues are equal

shap/plots/_bar.py
CloseChoice: Havent thought this through, but could be an idea to pursue

notebooks/api_examples/migrating-to-new-api.ipynb
connortann: Done 👍

PR #3740, opened 2024-07-04
Fix mypy failure: set plt.subplots squeeze=False in _image.py

connortann:
Fixes a build failure introduced by the new types hints in matplotlib `3.9.1`.

`_image.py` uses `f, axs = plt.subplots(nrows=nrows, ncols=cols)`. Mypy determines with the new matplotlib type hints that `axs` will be a ndarray (expected in our case) or an Axes object (if `nrows==ncols==1`), and complains that our code will fail in the latter case.

Passing `squeeze=False` to `plt.subplots` ensures that `axs` is always a numpy array.

### DIFF
--- a/shap/plots/_image.py
+++ b/shap/plots/_image.py
@@ -121,9 +121,7 @@ def image(
     fig_size = np.array([3 * (len(shap_values) + 1), 2.5 * (x.shape[0] + 1)])
     if fig_size[0] > width:
         fig_size *= width / fig_size[0]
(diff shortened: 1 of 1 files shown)


connortann:
@CloseChoice FYI, it would be good to get this one reviewed/merged, as the build is currently broken on all PRs

CloseChoice (approved):
very nice, looks like another true positive mypy found. LGTM

connortann:
Note to self- it would be great to have some more comprehensive tests for the image plot.

PR #3741, opened 2024-07-05
MAINT: Include ipywidgets in dev dependencies

connortann:
Small change to fix a warning that can appear in some notebooks about missing dependencies when tqdm is used.


thatlittleboy (approved):
joey-friends-nice

PR #3746, opened 2024-07-11
MAINT: Remove redundant test fixture

connortann:
Removes a mark applied to a test fixture, which has no effect as consequently is quite misleading. From the docs:

https://docs.pytest.org/en/stable/deprecations.html#applying-a-mark-to-a-fixture-function

> Applying a mark to a fixture function never had any effect, but it is a common user error.

### DIFF
--- a/tests/explainers/conftest.py
+++ b/tests/explainers/conftest.py
@@ -1,9 +1,6 @@
-import sys
-
 import pytest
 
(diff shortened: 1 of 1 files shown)


thatlittleboy (approved):
walkby-cat-high-five-shipit

PR #3752, opened 2024-07-16
DOCS: reformat the scatter notebook in API example

Xovee:
## Overview

Thoroughly reformat the scatter.ipynb in API example

- change the model to xgboost, consistent with other notebooks
- remove the annoying printing issues of "divided by zero"
- adapt to the new explanation api
- add a few more visualization examples

## Checklist

- (n/a) Unit tests added (if fixing a bug or adding a new feature)


connortann:
This looks great! Thanks for the PR. I have two small suggestions.

We recently added a pre-commit hook to ensure the notebook cells are executed in order. This notebook was previously "dirty" and on an exclude list, but now that the cells have been re-run in order, this notebook can be removed from the exclude list. So, I suggest try deleting this line:

https://github.com/shap/shap/blob/3c84f5f74ae8cd6b232f6cc68d7d7d3dde3de366/.pre-commit-config.yaml#L37

We also have a CI job that tests the notebooks can be executed without errors, and this notebook was also on an exclude list for that. Now the model has been changed, we might also be able to get this notebook running. To enable this, try deleting this line too:

https://github.com/shap/shap/blob/3c84f5f74ae8cd6b232f6cc68d7d7d3dde3de366/scripts/run_notebooks_timeouts.py#L40

Xovee:
Of course, I have removed the two lines in the corresponding files. Thanks for your suggestions and valuable works.

connortann (approved):
Looks great! Thanks again for the PR

PR #3753, opened 2024-07-16
MAINT: Enable type checking for shap.models

connortann:
## Overview

Supports #3730

Gets mypy passing on shap/models with check-untyped-defs=True, and adds a couple of type hints.


thatlittleboy (approved):
sorry this took so long,

corgi-excited-tippytaps

PR #3754, opened 2024-07-16
MAINT: Enable type checking for subset of shap.plots, update to PEP604

connortann:
## Overview

Supports #3730

Gets mypy passing on about half the modules in shap.plots with `check-untyped-defs=True`.

## Discussion

Also updates types hints to use the nice PEP604-style annotations:
- `X | Y` rather than `Union[X, Y]`
- `X | None` rather than `Optional[X]`.
 
This is made possible even with python 3.9 support with `__future__` imports.

As a supporting change, enables two typing-related linting rules.

### DIFF
--- a/shap/_explanation.py
+++ b/shap/_explanation.py
@@ -1,6 +1,8 @@
+from __future__ import annotations
+
 import copy
 import operator
-from typing import Any, Union
+from typing import Any
 
 import numpy as np
 import pandas as pd
@@ -170,7 +172,7 @@ class Explanation(metaclass=MetaExplanation):
         )
 
     @property
-    def shape(self) -> tuple[Union[int, None], ...]:
+    def shape(self) -> tuple[int | None, ...]:
         """Compute the shape over potentially complex data nesting."""
         # TODO: check if the return type should actually be tuple[int, ...]
         return _compute_shape(self._s.values)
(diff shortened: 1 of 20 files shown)


shap/_explanation.py
```diff
@@ -170,7 +172,7 @@ def __init__(
         )
 
     @property
-    def shape(self) -> tuple[Union[int, None], ...]:
+    def shape(self) -> tuple[int | None, ...]:
```
CloseChoice, line 175: this won't work well on python3.9. I think we could drop support for python3.9 altogether though.

CloseChoice (changes requested):
looks great just one comment we should sort out

shap/_explanation.py
connortann: My understanding is that this annotation works perfect fine on python 3.9 due to the [future](https://docs.python.org/3/library/__future__.html) import, specifically:

```python
```

I first noticed that this was possible on the streamlit project (e.g. [here](https://github.com/streamlit/streamlit/blob/develop/lib/streamlit/runtime/fragment.py) ), which supports python 3.8+ and yet uses the `X | Y` syntax for annotations.

In this PR the tests still pass on python 3.9, so there don't appear to be any runtime issues.

> this won't work well on python3.9.

Are there any other aspects or downsides that we should consider?

connortann: On a related note, I found two neat ruff linting rules which we can enable for consistent use of type hints:
- [FA](https://docs.astral.sh/ruff/rules/#flake8-future-annotations-fa): future annotations
- [TCH](https://docs.astral.sh/ruff/rules/#flake8-type-checking-tch): type checking

I pushed them to this branch, which required a few small modifications to get them passing.

CloseChoice (approved):
looks good

PR #3760, opened 2024-07-20
chore: Remove unnecessary double quotes after formatting

thatlittleboy:
## Overview

* remove spurious double quotes after formatting with black/ruff
* Had to include a small fix in `plots/_decision.py` because mypy was complaining.

## Checklist

### DIFF
--- a/shap/plots/colors/_colorconv.py
+++ b/shap/plots/colors/_colorconv.py
@@ -725,7 +725,7 @@ def _scale(a, n, m, copy=True):
             dtype = "uint{}".format(mnew)
         n = int(np.ceil(n / 2) * 2)
         warn(
-            "Downcasting {} to {} without scaling because max " "value {} fits in {}".format(
+            "Downcasting {} to {} without scaling because max value {} fits in {}".format(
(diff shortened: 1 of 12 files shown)


shap/plots/colors/_colorconv.py
```diff
@@ -725,7 +725,7 @@ def _scale(a, n, m, copy=True):
             dtype = "uint{}".format(mnew)
         n = int(np.ceil(n / 2) * 2)
         warn(
-            "Downcasting {} to {} without scaling because max " "value {} fits in {}".format(
+            "Downcasting {} to {} without scaling because max value {} fits in {}".format(
```
CloseChoice, line 728: NIT: I would prefer an f-string here but we can merge it in as is too.

CloseChoice (approved):
LGTM, feel free to merge

shap/plots/colors/_colorconv.py
thatlittleboy: Since colorconv is vendored code (it's not written by us), I'll prefer to leave it in as is for now. Thanks

PR #3761, opened 2024-07-21
refactor: Slight logic and docs cleanup of the clustering functions

thatlittleboy:
## Overview

A general code cleanup and logic improvement for the clustering utils, namely `xgboost_distances_r2` and `hclust` functions.

* Reduce unnecessary indentation in nested for loops by using itertools `product` and `combinations_with_replacement` where appropriate
* Early exit if linkage is not valid (`xgboost_distances_r2` is potentially time-consuming)
* Add some more tests for hclust
* Some other docstring fixes and improvements (e.g. wrong dtype, add references to scipy docs, clarify return types/shapes of numpy arrays)

I'll highlight some other changes directly as code review snippets.

## Checklist

### DIFF
--- a/shap/utils/_clustering.py
+++ b/shap/utils/_clustering.py
@@ -106,10 +106,10 @@ def xgboost_distances_r2(
     learning_rate: float = 0.6,
     early_stopping_rounds: Optional[int] = 2,
     subsample: Optional[float] = 1.0,
-    max_estimators: Optional[int] = 10000,
+    max_estimators: Optional[int] = 10_000,
     random_state: Union[int, np.random.RandomState] = 0,
-):
-    """Compute reducancy distances scaled from 0-1 among all the feature in X relative to the label y.
+) -> np.ndarray:
+    """Compute redundancy distances scaled from 0-1 among all the features in X relative to the label y.
 
     Distances are measured by training univariate XGBoost models of y for all the features, and then
     predicting the output of these models using univariate XGBoost models of other features. If one
@@ -184,10 +184,13 @@ def hclust(
     metric: str = "auto",
     random_state: Union[int, np.random.RandomState] = 0,
 ) -> np.ndarray:
-    """Fit a hierarcical clustering model for features X relative to target variable y.
+    """Fit a hierarchical clustering model for features X relative to target variable y.
 
-    For more information on clutering methods see:
+    For more information on clustering methods, see:
     https://docs.scipy.org/doc/scipy/reference/generated/scipy.cluster.hierarchy.linkage.html
+
+    For more information on scipy distance metrics, see:
+    https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.pdist.html
 
     Parameters
     ----------
--- a/shap/utils/_clustering.py
+++ b/shap/utils/_clustering.py
@@ -124,9 +124,10 @@ def xgboost_distances_r2(
     X_train, X_test, y_train, y_test = sklearn.model_selection.train_test_split(X, y, random_state=random_state)
 
     # fit an XGBoost model on each of the features
+    num_features = X.shape[1]
     train_preds_list = []
     test_preds_list = []
-    for i in range(X.shape[1]):
+    for i in range(num_features):
         model = xgboost.XGBRegressor(
             subsample=subsample,
             n_estimators=max_estimators,
@@ -141,9 +142,9 @@ def xgboost_distances_r2(
     test_preds = np.vstack(test_preds_list).T
 
     # fit XGBoost models to predict the outputs of other XGBoost models to see how redundant features are
-    dist = np.zeros((X.shape[1], X.shape[1]))
-    for i in show_progress(range(X.shape[1]), total=X.shape[1]):
-        for j in range(X.shape[1]):
+    dist = np.zeros((num_features, num_features))
+    for i in show_progress(range(num_features), total=num_features):
+        for j in range(num_features):
             if i == j:
                 dist[i, j] = 0
                 continue
--- a/shap/utils/_clustering.py
+++ b/shap/utils/_clustering.py
@@ -119,6 +119,14 @@ def xgboost_distances_r2(
     then the second feature is redundant with the first with respect to y. A distance of 1 corresponds
     to no redundancy while a distance of 0 corresponds to perfect redundancy (measured using the
     proportion of variance explained). Note these distances are not symmetric.
+
+    Returns
+    -------
+    np.ndarray
+        A square matrix of shape (n_features, n_features) containing the pairwise
+        redundancy distances between features. Each element [i, j] represents the
+        redundancy distance from feature i to feature j with respect to y.
+
     """
     import xgboost
 
--- a/shap/utils/_clustering.py
+++ b/shap/utils/_clustering.py
@@ -149,10 +149,12 @@ def xgboost_distances_r2(
                 continue
 
             # skip features that have not variance in their predictions (likely because the feature is a constant)
-            preds_var = np.var(test_preds[:, i])
+            preds_var: float = np.var(test_preds[:, i])
             if preds_var < 1e-4:
                 warnings.warn(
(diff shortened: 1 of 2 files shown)


thatlittleboy:
Note: `mypy` is complaining now because of

```
shap/plots/_decision.py:526: error: Need type annotation for "xmin"  [var-annotated]
shap/plots/_decision.py:526: error: Argument 1 has incompatible type "tuple[Any, float | ndarray[Any, Any]]"; expected "_SupportsArray[dtype[Never]] | _NestedSequence[_SupportsArray[dtype[Never]]]"  [arg-type]
shap/plots/_decision.py:527: error: Need type annotation for "xmax"  [var-annotated]
shap/plots/_decision.py:527: error: Argument 1 has incompatible type "tuple[Any, float | ndarray[Any, Any]]"; expected "_SupportsArray[dtype[Never]] | _NestedSequence[_SupportsArray[dtype[Never]]]"  [arg-type]
```
This is fixed in my other PR #3760 , so I won't duplicate the fix here. I'll wait for #3760 to be merged in first.

shap/utils/_clustering.py
```diff
@@ -267,5 +286,3 @@ def hclust(
         return scipy.cluster.hierarchy.complete(dist)
     elif linkage == "average":
         return scipy.cluster.hierarchy.average(dist)
-    else:
-        raise ValueError("Unknown linkage: " + str(linkage))
```
thatlittleboy, line 271: This check is irrelevant now since we already ensure the `linkage` parameter passed in is one of the 3 acceptable values above.

shap/utils/_clustering.py
```diff
@@ -228,37 +253,31 @@ def hclust(
...
-                    elif linkage == "complete":
-                        dist_list.append(max(dist_full[i, j], dist_full[j, i]))
-                    elif linkage == "average":
-                        dist_list.append((dist_full[i, j] + dist_full[j, i]) / 2)
-                    else:
-                        raise ValueError("Unsupported linkage type!")
+        dist_list: list[float] = []
+        for i, j in it.combinations_with_replacement(range(len(dist_full)), 2):
+            if i == j:
+                continue
```
thatlittleboy, line 262: This `it.combinations_with_replacement` is identical (element-wise) with the nested for loop, it just saves us one level of indentation.

Also note here that `dist_full` is necessarily a square matrix. So `shape[1] == shape[0]`.

image

shap/utils/_clustering.py
```diff
@@ -228,37 +253,31 @@ def hclust(
...
+            if linkage == "single":
+                dist_list.append(min(dist_full[i, j], dist_full[j, i]))
+            elif linkage == "complete":
+                dist_list.append(max(dist_full[i, j], dist_full[j, i]))
+            elif linkage == "average":
+                dist_list.append((dist_full[i, j] + dist_full[j, i]) / 2)
```
thatlittleboy, line 268: Note to myself (and other maintainers):

For curiosity's sake: I did try to vectorize this operation instead of using a nested for-loop, but at least on my machine, it seems to be slower (even when `num_features = 100`) so I dropped that implementation in the end.  
Also, the bulk of the time spent is on calculating the distances (`dist_full`) anyway, so it's not really worth the effort to optimize this too much.

For reference, here's the test with the attempted "vectorized" form and the original nested for loops:
```python
def test_hclust_building_triu():
    """Tests the refactoring of building a condensed upper triangular matrix within hclust,
    when using the 'xgboost_distances_r2' metric.
    """
    pytest.importorskip("xgboost")
    from shap.utils._clustering import xgboost_distances_r2
    import time

    linkage = "single"
    X = np.random.randint(1,30, size=(500, 50))
    y = np.where(X[:, 0] > 10, 1, 0)

    dist_full = xgboost_distances_r2(X, y, random_state=64)

    t0 = time.monotonic()
    # New logic (proposed)
    if linkage == "single":
        dist_new = np.minimum(dist_full, dist_full.T)
    elif linkage == "complete":
        dist_new = np.maximum(dist_full, dist_full.T)
    elif linkage == "average":
        dist_new = (dist_full + dist_full.T) / 2
    else:
        raise ValueError("Unsupported linkage type!")
    dist_new = dist_new[np.triu_indices_from(dist_new, 1)]
    t1 = time.monotonic()
    print(f">>> New {t1 - t0}")

    # Old logic (manual)
    dist = []
    for i in range(dist_full.shape[0]):
        for j in range(i + 1, dist_full.shape[1]):
            if i != j:
                if linkage == "single":
                    dist.append(min(dist_full[i, j], dist_full[j, i]))
                elif linkage == "complete":
                    dist.append(max(dist_full[i, j], dist_full[j, i]))
                elif linkage == "average":
                    dist.append((dist_full[i, j] + dist_full[j, i]) / 2)
                else:
                    raise ValueError("Unsupported linkage type!")
    dist_old = np.array(dist)
    t2 = time.monotonic()
    print(f">>> Old {t2 - t1}")

    assert (dist_new == dist_old).all()
```

And the results when running via pytest:

```
$ pytest -r P tests/utils/test_clustering.py::test_hclust_building_triu
...
_________________________________ test_hclust_building_triu _________________________________
--------------------------------------- Captured stdout call ----------------------------------------
>>> New 0.0005963749717921019
>>> Old 0.00045070797204971313
```

shap/utils/_clustering.py
```diff
@@ -228,37 +253,31 @@ def hclust(
...
-                if i != j:
-                    if linkage == "single":
-                        dist_list.append(min(dist_full[i, j], dist_full[j, i]))
-                    elif linkage == "complete":
-                        dist_list.append(max(dist_full[i, j], dist_full[j, i]))
-                    elif linkage == "average":
-                        dist_list.append((dist_full[i, j] + dist_full[j, i]) / 2)
-                    else:
-                        raise ValueError("Unsupported linkage type!")
+        dist_list: list[float] = []
```
CloseChoice, line 259: NIT: I would prefer to create the matrix here right away and then fill it line by line later. 

Pro: shape of the matrix is clear as one reads the code, memory is already allocated (though this is neglectable).
Con: we would need to enumerate and another index in the loop.

shap/utils/_clustering.py
CloseChoice: Interesting, did you test where the break even happens (I strongly assume that numpy gets way faster for larger matrices)?

CloseChoice (approved):
Looks good to me, thanks for cleaning this up

shap/utils/_clustering.py
thatlittleboy: @CloseChoice Not really, at X.shape=(500,100) (i.e., 500 samples, 100 feature dimensions), the xgboost `dist_full` was already taking 40s to run on my machine (IIRC, it's a non-trivial amount of time, let's just put it that way).  
Whereas building the linkages was performing at most 100**2 operations and finishing in <0.01s.  
The numpy vectorized way is still slower in this case.

I have the same expectations as you (that the numpy version should be faster, but I think it's unnecessary to test further :x).
In most realistic cases, we have around this order of magnitude of features anyway.

shap/utils/_clustering.py
thatlittleboy: @CloseChoice I'm assuming you mean the creation of the `dist_list`'s array here,

Agree on the memory reallocation point, and yea it probably doesn't matter here. As for the shape:  
I think the number of dimensions is clear enough from the code (i.e., ndim=1); and the length of the dist_list (k**2-k) I would argue isn't that important here since the `dist` array is input into the scipy hclust functions right after.

Hope it's alright with you that I close this

shap/utils/_clustering.py
connortann: @thatlittleboy if I understand correctly, is this logic:

```python
for i, j in it.combinations_with_replacement(range(len(dist_full)), 2):
    if i == j:
        continue
```

equivalent to just:

```python
for i, j in it.combinations(range(len(dist_full)), 2):
```

?

thatlittleboy: Thanks @connortann , nice catch. raised a separate PR to fix this. #3776

PR #3768, opened 2024-07-26
CI: Switch from pip to uv (~5min speedup), fix tensorflow/numpy constraint 

connortann:
## Overview

Switches CI jobs to use `uv` rather than `pip` to install dependencies. This speeds up all CI jobs by an average of 1min 40sec, and up to **5 minutes faster** on the slowest job (windows)!

## Discussion

[uv](https://github.com/astral-sh/uv) is *"An extremely fast Python package installer and resolver, written in Rust"*. It is becoming increasingly popular and dominant, following the success of the `ruff` linter by the same team.

Implementation notes:
1. I have disabled caching, as it does not lead to any speedup. 
   - I experimented with adding caching over at #3769 , and it does not lead to any improvement. This means we can just keep things simple.
   - Related: https://github.com/actions/setup-python/issues/822
1. We use the `--system` flag to install directly into system python, as recommended in CI environments. See https://github.com/astral-sh/uv/issues/1386
1. As part of this PR I encountered an unrelated issue of incompatible tensorflow and numpy versions, explained in https://github.com/shap/shap/issues/3707#issuecomment-2252511284 . To protect against this, I've added a new conditional numpy constraint into the test dependencies.

## Timings comparison

Comparison of timings of the "install dependencies" stage, versus the latest run on `master`:

| job  name     | Old (pip) | New (uv)   |
| -------- | ------- |------- |
| run_tests (ubuntu-latest, 3.9, test)           |  2m 14s  |  59s  |
| run_tests (ubuntu-latest, 3.10, test)         | 2m 53s   |  57s  |
| run_tests (ubuntu-latest, 3.11, test)         | 2m 44s   | 50s   |
| run_tests (ubuntu-latest, 3.12, test)         | 3m 53s   |  54s  |
| run_tests (windows-latest, 3.11, test)       |  10m 12s  | 4m 38s  |
| run_tests (macos-latest, 3.11, test)           |  1m48s  | 32s   |
| run_tests (ubuntu-latest, 3.12, test-core) | 46s        | 14s   |
| run tests (oldest supported numpy)        | 43s         | 20s  |
| run mypy                                                  | 42s        | 15s  |

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -69,6 +69,7 @@ jobs:
         if: matrix.os == 'macos-latest'
         run: brew install libomp
       - name: Install dependencies
+        # Do regular install NOT editable install: see GH #3020
         run: |
           python -m pip install --upgrade uv
           uv pip install --system '.[${{ matrix.extras }},plots]'
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -65,26 +65,24 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Cache python libs
-        uses: actions/cache@v4
-        if: matrix.extras == 'test'
-        with:
-          path: |
-            # Only cache a subset of libraries, ensuring cache size remains under 10GB. See GH dsgibbons#42
-            ${{ env.pythonLocation }}/**/site-packages/pyspark*
-            ${{ env.pythonLocation }}/**/site-packages/nvidia*
-            ${{ env.pythonLocation }}/**/site-packages/torch*
-            ${{ env.pythonLocation }}/**/site-packages/functorch*
-          key: ${{ runner.os }}-${{ matrix.python-version }}-${{ env.pythonLocation }}-${{ hashFiles('setup.py', 'pyproject.toml') }}-0
+      # - name: Cache python libs
+      #   uses: actions/cache@v4
+      #   if: matrix.extras == 'test'
+      #   with:
+      #     path: |
+      #       # Only cache a subset of libraries, ensuring cache size remains under 10GB. See GH dsgibbons#42
+      #       ${{ env.pythonLocation }}/**/site-packages/pyspark*
+      #       ${{ env.pythonLocation }}/**/site-packages/nvidia*
+      #       ${{ env.pythonLocation }}/**/site-packages/torch*
+      #       ${{ env.pythonLocation }}/**/site-packages/functorch*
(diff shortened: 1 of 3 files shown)


connortann:
The tests can fail due to an dependency specification bug in tensorflow: https://github.com/shap/shap/issues/3707#issuecomment-2252511284

### Root cause

Older versions of tensorflow (2.14) incorrectly miss it's upper bound dependency on `numpy<2`, so we can resolve a broken env with `tensorflow==2.14.0, numpy==2.0.1`. This leads to an error:

```python
>   _np_bfloat16 = pywrap_ml_dtypes.bfloat16()
E   TypeError: Unable to convert function return value to a Python type! The signature was
E   	() -> handle
```

Tensorflow will not fix this issue upstream, so we must ensure our constraints in the CI tests do not allow incompatible versions.

### Proposed fix

Now, we don't really want to pin `numpy<2` as we'd like to test against numpy 2, specifically when only the `tests-core` extras are installed. For later versions of tensorflow we don't need to worry, as they include the correct numpy bounds.

Ideally we would have a conditional constraint like `numpy<2.0; tensorflow< '2.15'` , but unfortunately this syntax is not supported. The conditional statement can only be an [environment marker](https://peps.python.org/pep-0345/#environment-markers).

My proposed solution is to add a constraint in our `tests` extras of: `numpy<2.0; python_version < '3.12'`. This should fix the tests, whilst allowing the latest tensorflow & numpy to be tested together on 3.12.

.github/workflows/run_tests.yml
```diff
@@ -65,26 +65,13 @@ jobs:
...
         if: matrix.os == 'macos-latest'
         run: brew install libomp
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip
-          # Use "eager" update strategy in case cached dependencies are outdated
-          # Using regular install NOT editable install: see GH #3020
-          pip install --upgrade --upgrade-strategy eager '.[${{ matrix.extras }},plots]'
+          python -m pip install --upgrade uv
+          uv pip install --system '.[${{ matrix.extras }},plots]'
```
thatlittleboy, line 75: @connortann shall we retain the 

```
# Using regular install NOT editable install: see GH #3020
```
comment? I think it's still relevant, right?

thatlittleboy (approved):
Just a small comment, otherwise good to go from me. Love to see the CI runs getting faster!

seal-of-approval-seal

.github/workflows/run_tests.yml
connortann: Ah yes, absolutely. I'll add that back in.

PR #3770, opened 2024-07-26
CI: use uv package's lowest-direct resolution strategy

connortann:
Draft PR, exploring some ideas.

Related: https://github.com/astral-sh/uv/issues/5492


connortann:
I shall abandon this for now, as this resolution strategy would require us to define lower bounds on all of our dev-dependencies, which doesn't seem ideal.

I think the present method of explicitly adding a lower bound numpy pin is good enough, and it's very simple.

In future though, it might be helpful to come back to this and explore automated ways to test against lower versions of other important dependencies.

PR #3331, opened 2023-10-13
Cleanup disk before running CI tests on ubuntu

CloseChoice:
## Overview

Closes #3330  

Description of the changes proposed in this pull request:

## Checklist

- [x] All [pre-commit checks](https://pre-commit.com/#install) pass.
- ~[ ] Unit tests added (if fixing a bug or adding a new feature)~

Issue #3330, opened 2023-10-13
CI: GH Actions encounter "No space left on device" unless unused CI tools are removed


connortann:
Our GitHub Actions workflow to run the unit tests has started failing. When trying to install pip-dependencies we now get an error `OSError: [Errno 28] No space left on device`:

```
Installing collected packages: wcwidth, sentencepiece, pytz, py4j, pure-eval, ptyprocess, pickleshare, mpmath, libclang, flatbuffers, executing, backcall, wrapt, wheel, urllib3, tzdata, typing-extensions, traitlets, tqdm, threadpoolctl, termcolor, tensorflow-io-gcs-filesystem, tensorflow-estimator, tensorboard-data-server, tenacity, sympy, slicer, six, setuptools, safetensors, regex, pyyaml, pyspark, pyparsing, pygments, pyasn1, protobuf, prompt-toolkit, pluggy, pillow, pexpect, parso, packaging, oauthlib, nvidia-nvtx-cu12, nvidia-nvjitlink-cu12, nvidia-nccl-cu12, nvidia-curand-cu12, nvidia-cufft-cu12, nvidia-cuda-runtime-cu12, nvidia-cuda-nvrtc-cu12, nvidia-cuda-cupti-cu12, nvidia-cublas-cu12, numpy, networkx, MarkupSafe, markdown, llvmlite, kiwisolver, keras, joblib, iniconfig, idna, grpcio, graphviz, gast, fsspec, fonttools, filelock, decorator, cycler, coverage, cloudpickle, charset-normalizer, certifi, cachetools, absl-py, werkzeug, triton, scipy, rsa, requests, python-dateutil, pytest, pyasn1-modules, plotly, opt-einsum, opencv-python, nvidia-cusparse-cu12, nvidia-cudnn-cu12, numba, ml-dtypes, matplotlib-inline, Jinja2, jedi, h5py, google-pasta, contourpy, astunparse, asttokens, xgboost, stack-data, scikit-learn, requests-oauthlib, pytest-cov, pandas, nvidia-cusolver-cu12, matplotlib, lightgbm, huggingface-hub, google-auth, torch, tokenizers, shap, pytest-mpl, pyod, ipython, gpboost, google-auth-oauthlib, catboost, transformers, torchvision, tensorboard, tensorflow
  Attempting uninstall: setuptools
    Found existing installation: setuptools 65.5.0
    Uninstalling setuptools-65.5.0:
      Successfully uninstalled setuptools-65.5.0
[...]

PR #3771, opened 2024-07-26
CI: remove the disk-space job

connortann:
## Overview

After having migrated to use `uv`, I think it's likely that we don't need the clean-disk-space job and can save several more minutes of CI time. This PR reverts #3331 , dropping that job.

Most test jobs are now much faster: in the region of 5-9 minutes, which is almost halved compared to the previous pip workflow. The tests on windows remain the clear bottleneck.

## Discussion

We never got to the bottom of what caused the disk space issue, but I think it's likely due to the face that pip will sometimes download many different wheels for the same package to determine which are compatible. uv does not have this flaw, so I think it's likely we will not encounter the disk space issue.

That said, after merging this PR we should keep an eye out for CI failures due to disk space, and if issue #3330 occurs again we can always re-instate the fix.

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -44,22 +44,6 @@ jobs:
       fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
-      - name: Free Disk Space (Ubuntu)
-        if: runner.os == 'Linux'
-        uses: jlumbroso/free-disk-space@main
-        with:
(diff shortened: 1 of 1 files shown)


thatlittleboy (approved):
seems reasonable, lgtm thanks

domino-effect-self-high-five

connortann:
> seems reasonable, lgtm thanks

@thatlittleboy a most excellent choice of LGTM gif today!

PR #3775, opened 2024-08-02
MAINT: fixup string literal concatenation

connortann:
## Overview

Gets ruff rule ["ISC"](https://docs.astral.sh/ruff/rules/single-line-implicit-string-concatenation/) passing, which checks for implicit string concatentation.

## Discussion

This involved fixing a few remnant string literal formatting issues from `ruff format`. The linting rule also highlighted a few places in `_text.py` where f-strings could be used more widely for a tidier implementation.

Example of changes:

```python
old = "foo" " bar"
new = "foo bar"
```

### DIFF
--- a/shap/plots/_image.py
+++ b/shap/plots/_image.py
@@ -214,19 +214,15 @@ def image_to_text(shap_values):
     for i in range(model_output.shape[0]):
         output_text_html += (
             "<div style='display:inline; text-align:center;'>"
-            + f"<div id='{uuid}_output_flat_value_label_"
-            + str(i)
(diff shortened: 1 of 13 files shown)


shap/plots/_image.py
connortann, line 1: Note for reviewer(s): whilst most changes are pretty trivial, I would suggest giving particular attention to the changes to `_image.py` and `_text.py`, to double-check for any syntax issues.

thatlittleboy (approved):
toddler-skiing-high-five

PR #3776, opened 2024-08-03
refactor: use it.combinations instead of it.combinations_with_replacement

thatlittleboy:
## Overview

A small follow up from my previous PR #3761, comment by Connor: [LINK](https://github.com/shap/shap/pull/3761#discussion_r1690223340)

## Checklist

### DIFF
--- a/shap/utils/_clustering.py
+++ b/shap/utils/_clustering.py
@@ -257,9 +257,7 @@ def hclust(
 
         # build a condensed upper triangular version by taking the max distance from either direction
         dist_list: list[float] = []
-        for i, j in it.combinations_with_replacement(range(len(dist_full)), 2):
-            if i == j:
-                continue
+        for i, j in it.combinations(range(len(dist_full)), 2):
             if linkage == "single":
                 dist_list.append(min(dist_full[i, j], dist_full[j, i]))
             elif linkage == "complete":


connortann (approved):

thatlittleboy:
@connortann Any idea why this job is now failing on Windows?

https://github.com/shap/shap/actions/runs/10232378534/job/28309628156

I'm not sure why this problem didn't surface in your previous 2 PRs (just merged in) and commits though.

connortann:
Hmm, strange. It looks like it might possibly be related to this issue:
https://github.com/astral-sh/uv/issues/1491

thatlittleboy:
> @connortann Any idea why this job is now failing on Windows?
> 
> https://github.com/shap/shap/actions/runs/10232378534/job/28309628156
> 
> I'm not sure why this problem didn't surface in your previous 2 PRs (just merged in) and commits though.

So it seems that Matplotlib recently deleted their 3.9.1 release wheel for Windows ([source](https://github.com/matplotlib/matplotlib/issues/28551#issuecomment-2267225291)) due to a critical bug, so that explains why this problem didn't surface in earlier runs.

And this in turn resulted in matplotlib being built from source via `uv`, and some sort of build error caused it to fail (still unsure what). Strangely though, I tested [here with a debug branch](https://github.com/shap/shap/actions/runs/10236308054/job/28318207052) that using `pip` instead of `uv` to install `matplotlib==3.9.1` (built from source) on windows **doesn't** cause an error.

So the actual fault seems to be a combination of `windows + uv + matplotlib being built from source`..

Regardless, I'm expecting that matplotlib 3.9.2 is going to be released soon (also reiterated by mpl's maintainers), where the wheels will be reintroduced back for windows. I suggest we just wait till then to rerun the CI.

connortann:
Thanks for the analysis!

As a short-term workaround, we could also add a constraint in the test dependencies of `matplotlib!=3.9.1`.

If you have time, it might also be helpful to raise a report with uv to investigate why that install fails on windows with uv. Perhaps the issue could be reproduced with the uv equivalent of:

```bash
pip wheel --no-cache-dir --use-pep517 "matplotlib (==3.9.1)"
```

tacaswell:
The 3.9.1 wheels on windows would, depending on import order, cause segfaults, see https://github.com/matplotlib/matplotlib/issues/28551#issuecomment-2266699619.

We deleted the bad wheels to avoid segfaults (which seemed like a good idea) but failed to anticipate that this would cause downstream projects to start trying to build mpl from source instead of pulling the 3.9.0 wheel.  We have now yanked the whole 3.9.1 release to fix these problems and will get 3.9.2 out as quick as we can.

To prevent this sort of thing in the future (trying to build something from source when you are not prepared to do so) I suggest using `matplotlib --only-binary "matplotlib" ` in your requirements file (which may be a good practice for all wheels that have extentions).

Sorry for the trouble.


----

As a side note, the install failing with `uv` but "working" with pip may in fact be a bug with pip as some of the other failure reports we got indicated building "worked" and Matplotlib was installed, but the c-extension modules were missing and we failed immediately on import.

PR #3777, opened 2024-08-04
feat: change OpHistoryItem data structure to dataclass

thatlittleboy:
## Overview

1. I'm proposing to use a dataclass structure for an OpHistory item instead of a dictionary. For several reasons: 1) it removes the need for "attribute"/"item" access via a string, a la `op.get('name')`, so we use the safer `op.name` instead. 2) it makes clear what attributes of an OpHistory item are mandatory, and what are optional. E.g. "name" should always be present based on the code, but "collapsed_instances" is not (defaults to False if not provided).
1. Side note: Chiming in on the TODO left from the ([review comment](https://github.com/shap/shap/pull/3732/files#r1662223509))  from PR #3732, I think specifically for `Explanation.shape`, it should return a `tuple[int,...]`, not a `tuple[int|None,...]`. Because the `Explanation.shape` calculates the shape of the underlying `Explanation.values` array, which should never contain a string, nor is it ever sparse. The reason why I'm addressing this TODO here is because `OpHistoryItem.shape` requires a type hint for `prev_shape` attribute.

### Discussion

I'm assuming the OpHistory is an internal implementation detail (because `Explanation.op_history` is not even a documented attribute), so this change shouldn't break anyone's code, in theory. I.e., this isn't breaking change territory.

## Checklist

### DIFF
--- a/shap/_explanation.py
+++ b/shap/_explanation.py
@@ -21,6 +21,8 @@ op_chain_root = OpChain("shap.Explanation")
 
 @dataclass
 class OpHistoryItem:
+    """An operation that has been applied to an Explanation object."""
+
     name: str
     prev_shape: tuple[int, ...]
     args: tuple[Any, ...] = ()
--- a/shap/_explanation.py
+++ b/shap/_explanation.py
@@ -2,7 +2,8 @@ from __future__ import annotations
 
 import copy
 import operator
-from typing import Any
+from dataclasses import dataclass, field
+from typing import Any, cast
 
 import numpy as np
 import pandas as pd
@@ -16,6 +17,15 @@ from .utils._exceptions import DimensionError
 from .utils._general import OpChain
 
 op_chain_root = OpChain("shap.Explanation")
+
+
+@dataclass
+class OpHistoryItem:
+    name: str
+    prev_shape: tuple[int, ...]
+    args: tuple[Any, ...] = ()
+    kwargs: dict[str, Any] = field(default_factory=dict)
+    collapsed_instances: bool = False
 
 
 class MetaExplanation(type):
@@ -100,7 +110,7 @@ class Explanation(metaclass=MetaExplanation):
         clustering=None,
         compute_time=None,
     ):
-        self.op_history = []
+        self.op_history: list[OpHistoryItem] = []
 
         self.compute_time = compute_time
 
@@ -172,10 +182,12 @@ class Explanation(metaclass=MetaExplanation):
         )
 
     @property
-    def shape(self) -> tuple[int | None, ...]:
+    def shape(self) -> tuple[int, ...]:
         """Compute the shape over potentially complex data nesting."""
-        # TODO: check if the return type should actually be tuple[int, ...]
-        return _compute_shape(self._s.values)
+        shap_values_shape = _compute_shape(self._s.values)
+        # impl: `Explanation.values` always corresponds to the shap values, which is a numpy array, so the
+        # shape will always be of tuple[int, ...] type, not tuple[int|None, ...].
+        return cast(tuple[int, ...], shap_values_shape)
 
     @property
     def values(self):
@@ -417,7 +429,7 @@ class Explanation(metaclass=MetaExplanation):
         if new_self is None:
             new_self = copy.copy(self)
         new_self._s = new_self._s.__getitem__(item)
-        new_self.op_history.append({"name": "__getitem__", "args": (item,), "prev_shape": self.shape})
+        new_self.op_history.append(OpHistoryItem(name="__getitem__", args=(item,), prev_shape=self.shape))
 
         return new_self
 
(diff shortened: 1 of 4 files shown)


shap/_explanation.py
```diff
@@ -1,6 +1,7 @@
 import copy
 import operator
-from typing import Any, Union
+from dataclasses import dataclass, field
+from typing import Any, Union, cast
```
connortann: Following the addition of the "future" imports in #3754 (which has led to a bit of a merge conflict), we should now be able to avoid having to import `Union` or `Optional`.

shap/_explanation.py
```diff
@@ -16,6 +17,15 @@
 op_chain_root = OpChain("shap.Explanation")
 
 
+@dataclass
+class OpHistoryItem:
```
connortann, line 23: It might help to add a docstring to explain the purpose of this object, and perhaps to contrast it with the `OpChain` object.

Is it correct to say that:
- The `OpChain` represents a set of dot-chained operations to be applied in future when plotting different subsets, whereas
- The `list[OpHistoryItem]` objects represents the ops that have already been applied, as is used to help provide context when figuring out plot labels
 
?

shap/plots/_utils.py
```diff
@@ -22,7 +22,7 @@ def convert_ordering(ordering, shap_values):
     if issubclass(type(ordering), OpChain):
         ordering = ordering.apply(Explanation(shap_values))
     if issubclass(type(ordering), Explanation):
-        if "argsort" in [op["name"] for op in ordering.op_history]:
+        if any("argsort" in op.name for op in ordering.op_history):
```
connortann: The previous implementation checked for a name that was equal to "argsort" rather than contains "argsort", so I think to be consistent with previous behaviour it would be:

```suggestion
        if any(op.name=="argsort" for op in ordering.op_history):
```

connortann:
Some minor suggestions above but otherwise looks great.

Two broader discussion points:

## Support for scipy.sparse

> Explanation.shape calculates the shape of the underlying Explanation.values array, which should never contain a string, nor is it ever sparse.

Are we confident that `Explanation.values` will never be sparse? I noticed a few indications of sparse support:
1.   `slicer`, whose main function seems to be supporting Explanation objects, supports scipy.sparse arrays
2.  I remember seeing slundberg comment somewhere that support for sparse arrays was at one point a work-in-progress (though I can't seem to find that thread unfortunately)
3. Finally I also noticed that `_compute_shape` has support for sparse arrays, for which the computed shape is `None`:

https://github.com/shap/shap/blob/07d9d2fb5c292afb14ea4a8d3dc8417ffbbf2076/shap/_explanation.py#L796-L800

Overall it seems like there is at least partial support, so perhaps we should design for the possibility of sparse arrays if it's not too difficult.

Overall I definitely think this refactor is an improvement for clarity, and it's a nice opportunity for us to pin down precisely what data types are expected.

## Test plan

Are there existing tests for the "ophistory" functionality or would it be worthwhile adding some further unit tests?

shap/_explanation.py
thatlittleboy: Noted, thanks, I've rebased on top of your changes.

shap/_explanation.py
thatlittleboy: I think your description is somewhat correct, but I don't think it's too helpful to make a comparison between these two. `OpChain` is representing a set of dot-chained operations that can be applied to some object in the future via the `.apply(obj)` method.  
`OpChain._ops` stores the list of such ops, to be applied.

And `Explanation().op_history` stores the list of ops that have already been applied to the said Explanation object.

Technically, we could use `OpHistoryItem` to describe the ops that are in `OpChain._ops` as well, but we'll need to set the `.prev_shape` to be optional since it doesn't make sense to speak of a previous shape when the object on which to apply the ops on hasn't been supplied yet. I decided against this for now though, I'm not sure whether this is something we want to do -- they should be separate objects. That one could be an `OpChainItem` for example.

I'll add docstrings to this dataclass in the meantime, but I'll leave out the comparison.

shap/plots/_utils.py
thatlittleboy: Thanks, good catch

thatlittleboy:
@connortann 

> "Are we confident that Explanation.values will never be sparse?"

Yea I think so. The 3 points you mentioned only point towards general arrays in Explanation possibly being sparse, which I think we should and are supporting. Here, more explicitly, we're talking about `Explanation.values` which correspond to shap values. I wouldn't expect these to be sparse, quite the opposite really. Are there any scenarios where shap values array should be sparse?

Though, the above discussion is a slight distraction. Circling back to the point that I am addressing in this PR: I hope we are of agreement that `Explanation.shape` / `Explanation.values.shape` is always a `tuple[int,...]`, since `Explanation.values` can't possibly be a **sparse array containing strings**.

> Are there existing tests for the "ophistory" functionality or would it be worthwhile adding some further unit tests?

Not that I'm aware. Sure I can add one.

shap/_explanation.py
connortann: Thanks for the detailed explanation, that's helpful to help me build confidence in my understanding :)

connortann (approved):
Looks great!

> I hope we are of agreement that Explanation.shape / Explanation.values.shape is always a tuple[int,...], since Explanation.values can't possibly be a sparse array containing strings.

Yep, I'm in agreement. Thanks for clearing that up for me.

PR #3779, opened 2024-08-04
refactor: Cohorts internals, use @property, add type hinting and expose it in docs

thatlittleboy:
## Overview

- Use `@property` for `Cohorts.cohorts` to enforce that the value passed in is always a dictionary of Explanation objects *before* the value is set. I believe this is the more pythonic way of enforcing this constraint, rather than doing it only in `__init__`. A side effect here of using the property descriptor is that users will no longer be able to `del` the cohorts attribute, but I guess that's not a "normal" thing to do..
- Added docstrings with examples to the Cohorts class. And made `Cohorts` a publicly documented feature (it's already in our public API being exposed in our `__init__.py`).
- <s>One benefit from doing this refactoring is that I spotted a bug in the definition of `Cohorts` (mypy, actually), in that the `Cohorts.__call__()` definition does not make sense. It calls `Explanation.__call__()` which does not exist. See: </s>

(minimized the image because it's not accurate, see follow up comment)

- EDIT: It turns out that `Cohorts.__call__` was actually included here to invoke methods on the individual Explanation objects, not for doing `Explanation.__call__` (which doesn't exist) directly. An example (also included in the newly added tests) would be the following:

```python
exp = shap.Explanation(
    values=rs.uniform(low=-1, high=1, size=e_size),
    data=rs.normal(loc=1, scale=3, size=e_size),
    feature_names=list("abcde"),
)
exp_neg = exp[exp[:, "a"].data < 0]
exp_pos = exp[exp[:, "a"].data >= 0]

ch = shap.Cohorts(col_a_neg=exp_neg, col_a_pos=exp_pos)

# ----------- here is the crux:
new_ch = ch.mean(axis=0)
# This does a `getattr(ch, "mean")` first on the Cohorts object, which results in a bunch of `Explanation.mean` bound methods;
# A Cohorts object is being returned by getattr, hence the subsequent `(axis=0)` is being invoked as a Cohorts.__call__
```

Thus, I've retained the `__call__` magic method on `Cohorts`, but storing the bound methods in a separate private variable, and these would be individually invoked during the `Cohorts.__call__`.

- Added tests to test all of the above mentioned intricacies.

## Checklist

### DIFF
--- a/shap/_explanation.py
+++ b/shap/_explanation.py
@@ -910,9 +910,10 @@ class Cohorts:
         if not isinstance(cval, dict):
             emsg = "self.cohorts must be a dictionary!"
             raise TypeError(emsg)
-        if not all(isinstance(exp, Explanation) for exp in cval.values()):
-            emsg = "All the arguments to a Cohorts set must be Explanation objects!"
-            raise TypeError(emsg)
+        for exp in cval.values():
+            if not isinstance(exp, Explanation):
+                emsg = f"Arguments to a Cohorts set must be Explanation objects, but found {type(exp)}"
+                raise TypeError(emsg)
 
         cast(dict[str, Explanation], cval)  # type narrowing for mypy
         self._cohorts: dict[str, Explanation] = cval
--- a/shap/_explanation.py
+++ b/shap/_explanation.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import copy
 import operator
 from dataclasses import dataclass, field
-from typing import Any, cast
+from typing import Any, Callable, cast
 
 import numpy as np
 import pandas as pd
@@ -898,7 +898,7 @@ class Cohorts:
 
     def __init__(self, **kwargs: Explanation) -> None:
         self.cohorts = kwargs
-        self._callables = {}
+        self._callables: dict[str, Callable] = {}
 
     @property
     def cohorts(self) -> dict[str, Explanation]:
--- a/shap/_explanation.py
+++ b/shap/_explanation.py
@@ -915,7 +915,6 @@ class Cohorts:
                 emsg = f"Arguments to a Cohorts set must be Explanation objects, but found {type(exp)}"
                 raise TypeError(emsg)
 
-        cast(dict[str, Explanation], cval)  # type narrowing for mypy
         self._cohorts: dict[str, Explanation] = cval
 
     def __getitem__(self, item) -> Cohorts:
--- a/shap/_explanation.py
+++ b/shap/_explanation.py
@@ -898,6 +898,7 @@ class Cohorts:
 
     def __init__(self, **kwargs: Explanation) -> None:
         self.cohorts = kwargs
+        self._callables = {}
 
     @property
     def cohorts(self) -> dict[str, Explanation]:
@@ -923,9 +924,24 @@ class Cohorts:
         return Cohorts(**new_cohorts)
 
     def __getattr__(self, name: str) -> Cohorts:
-        new_cohorts = {}
+        new_cohorts = Cohorts()
         for k in self._cohorts:
-            new_cohorts[k] = getattr(self._cohorts[k], name)
+            result = getattr(self._cohorts[k], name)
+            if callable(result):
+                new_cohorts._callables[k] = result  # bound methods like .mean, .sample
+            else:
+                new_cohorts._cohorts[k] = result
+        return new_cohorts
+
+    def __call__(self, *args, **kwargs) -> Cohorts:
+        """Call the bound methods on the Explanation objects retrieved during attribute access."""
+        if not self._callables:
+            emsg = "No methods to __call__!"
+            raise ValueError(emsg)
+
+        new_cohorts = {}
+        for k, bound_method in self._callables.items():
+            new_cohorts[k] = bound_method(*args, **kwargs)
         return Cohorts(**new_cohorts)
 
     def __repr__(self):
--- a/shap/_explanation.py
+++ b/shap/_explanation.py
@@ -917,16 +917,16 @@ class Cohorts:
         self._cohorts: dict[str, Explanation] = cval
 
     def __getitem__(self, item) -> Cohorts:
-        new_cohorts = Cohorts()
+        new_cohorts = {}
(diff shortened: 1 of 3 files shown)


shap/_explanation.py
```diff
@@ -826,33 +826,39 @@ def _compute_shape(x) -> tuple[Union[int, None], ...]:
 
 
 class Cohorts:
-    def __init__(self, **kwargs):
+    def __init__(self, **kwargs) -> None:
         self.cohorts = kwargs
-        for k in self.cohorts:
-            assert isinstance(
-                self.cohorts[k], Explanation
-            ), "All the arguments to a Cohorts set must be Explanation objects!"
```
thatlittleboy, line 860: This is now enforced in the property setter

shap/_explanation.py
```diff
@@ -826,33 +826,39 @@ def _compute_shape(x) -> tuple[Union[int, None], ...]:
...
         return new_cohorts
 
-    def __call__(self, *args, **kwargs):
+    def __getattr__(self, name) -> "Cohorts":
         new_cohorts = Cohorts()
-        for k in self.cohorts:
-            new_cohorts.cohorts[k] = self.cohorts[k].__call__(*args, **kwargs)
+        for k in self._cohorts:
+            new_cohorts.cohorts[k] = getattr(self._cohorts[k], name)
         return new_cohorts
```
thatlittleboy, line 934: `__call__()` is removed here, it makes no sense once we enforce that `self.cohorts[k]` must be an Explanation object.

shap/_explanation.py
```diff
@@ -826,33 +826,39 @@ def _compute_shape(x) -> tuple[Union[int, None], ...]:
 
 
 class Cohorts:
-    def __init__(self, **kwargs):
+    def __init__(self, **kwargs) -> None:
```
connortann: This might be a good opportunity to add a class docstring to explain what a cohort represents and how the class is intended to be used.

We don't need to go overboard, but a few docstrings might help future devs get up to speed faster.

connortann: PS, I think this `kwargs` could be type hinted as:

```suggestion
    def __init__(self, **kwargs: Explanation) -> None:
```

Using syntax explained here:
https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

shap/_explanation.py
```diff
@@ -826,33 +826,39 @@ def _compute_shape(x) -> tuple[Union[int, None], ...]:
...
+        return self._cohorts
+
+    @cohorts.setter
+    def cohorts(self, cval):
+        if not isinstance(cval, dict):
+            emsg = "self.cohorts must be a dictionary!"
+            raise TypeError(emsg)
+        if not all(isinstance(exp, Explanation) for exp in cval.values()):
+            emsg = "All the arguments to a Cohorts set must be Explanation objects!"
+            raise ValueError(emsg)
```
connortann: Should this be:
```suggestion
            raise TypeError(emsg)
```

shap/_explanation.py
thatlittleboy: Thanks, added docstring with examples, please have a read and see if it makes sense.

shap/_explanation.py
thatlittleboy: NOTE to self: The above claim is somewhat wrong. We still need to retain `__call__` here in order to invoke any methods that were retrieved during `__getattr__`.

E.g. `Cohorts(...).mean(axis=0)` would first run `__getattr__("mean")` and return a bound method `Explanation.mean` for all the Explanation objects inside the Cohorts. And return a Cohorts object.
Then the `(axis=0)` call would be executed on that returned Cohorts object, which is why we need `__call__`.

shap/_explanation.py
connortann: Looks great!

shap/_explanation.py
```diff
@@ -852,36 +856,99 @@ def _compute_shape(x) -> tuple[int | None, ...]:
...
+        return self._cohorts
+
+    @cohorts.setter
+    def cohorts(self, cval):
+        if not isinstance(cval, dict):
+            emsg = "self.cohorts must be a dictionary!"
+            raise TypeError(emsg)
+        if not all(isinstance(exp, Explanation) for exp in cval.values()):
+            emsg = "All the arguments to a Cohorts set must be Explanation objects!"
+            raise TypeError(emsg)
```
connortann: Nitpick: it might help to include the unexpected type in the exception message.

```suggestion
        for exp in cval.values():
            if not isinstance(exp, Explanation):
                emsg = f"Arguments to a Cohorts set must be Explanation objects, but found {type(exp)}"
                raise TypeError(emsg)
```

shap/_explanation.py
```diff
@@ -852,36 +856,99 @@ def _compute_shape(x) -> tuple[int | None, ...]:
...
+
+        cast(dict[str, Explanation], cval)  # type narrowing for mypy
+        self._cohorts: dict[str, Explanation] = cval
```
connortann: I think the type narrowing applies to the _return value_ of `typing.cast`, so I think it is typically used like this:

```suggestion
        self._cohorts: dict[str, Explanation] = cast(dict[str, Explanation], cval)  # type narrowing for mypy
```

However, it might be that the assert statements above provide sufficient type narrowing, so maybe we don't need the `cast` call?

shap/_explanation.py
connortann: Nice! To help out our future selves, I think it would help to include your explanation above as a comment in the `__call__` method, as it's quite subtle.

connortann (approved):
Some minor suggestions but otherwise looks great. Nice one!

shap/_explanation.py
thatlittleboy: Yeah, you're right. For some reason, I don't think mypy picked up that `cval` could only be a dict containing `Explanation`s when it was a generator comprehension (or I'm going crazy).

Anyway now it works after testing locally, so I've implemented your suggestion. Thanks~

shap/_explanation.py
thatlittleboy: done

PR #3785, opened 2024-08-08
FIX: Re-vendor colorconv from skimage 0.24.0

### LINKED ISSUE #3782: Deprecated numpy functions in vendored colorconv code
The `_colorconv` module (which has been vendored from scikit-image) contains some deprecated numpy calls, which are starting to throw deprecation warnings:

```
/shap/plots/colors/_colorconv.py:819: DeprecationWarning: Converting `np.inexact` or `np.floating` to a dtype is deprecated. The current result is `float64` which is not strictly correct.
  if np.issubdtype(dtype_in, np.dtype(dtype).type):
```

_Originally posted by @jonathanng in https://github.com/shap/shap/issues/3704#issuecomment-2267281086_
            

We could address this by perhaps re-vendoring the relevant functions from a recent release of scikit-image.

connortann:
## Overview

Closes #3782 , fixes numpy deprecation warnings.

Note to reviewers: this PR builds on #3786, which adds supporting tests to ensure the behaviour remains unchanged. ~That PR should probably be reviewed & merged first.~ EDIT: Done.

## Discussion:

Replaces the vendored `_colorconv.py` module from skimage with a fresh copy from a recent release `0.24.0`. This should fix the deprecation warnings from uses of deprecated numpy functions.

### How the code was vendored

I tried to vendor the code in a way that minimised changes from the original package, whilst only including the minimum set of functions that are required for the two functions that we use: `lab2rgb` and `lch2lab`.

I recursively included any objects which were required. I removed any code paths which were unused - for example, handing input data types other than floats.

From the commit history you can see this iterative simplification. First I vendored all relevant functions, and then subsequently removed the unused code paths. We are left with only ~200 LOC.

I decided to enable ruff on this module. It was previously disabled, but it passes on the new module. As we have extracted only a subset of scikit-image, I think the ruff linter is helpful to check that we don't have any undefined objects or other issues.

### Compatibility check

- [x] skimage supports python 3.9+
- [x] skimage supports numpy 1.23+, and the oldest we test against is 1.24

### Test plan

TODO:
- [x] ensure that shap's use of these two functions has appropriate unit tests. Done on #3786
- [x] ensure the numpy deprecation warnings are fixed (1586 warnings -> 960 warnings total! More work to do...)


connortann:
It strikes me that we could probably simplify this a lot further, as there is as lot of unused logic relating to handling different input types that we don't need. A simpler version would be a lot less code and more maintainable.

EDIT: this is now done, we are down to only about 200 LOC.

shap/plots/colors/_colorconv.py
```diff
@@ -1,475 +1,103 @@
...
 # This is a small chunk of code from the skimage package. It is reproduced
-# here because all we need is a couple color conversion routines, and adding
-# all of skimage as dependency is really heavy.
+# here because all we need is a couple color conversion routines (lab2rgb, lch2lab)
+# and adding all of skimage as dependency is really heavy.
 
-# Disable linting on vendored code
-# ruff: noqa
 
 # Copyright (C) 2019, the scikit-image team
```
thatlittleboy: Shall we follow their [LICENSE.txt](https://github.com/scikit-image/scikit-image/blob/main/LICENSE.txt) file?

```suggestion
Copyright: 2009-2022 the scikit-image team
License: BSD-3-Clause
```

thatlittleboy (approved):
Thanks, not much concerns on the vendoring process, just a small comment on the license declaration.

squidward-lacklustre-hooray

shap/plots/colors/_colorconv.py
connortann: Good spot, done. I committed separately to ensure the `#` is retained

DavidRavnsborg:
Really looking forward to this fix in the stable release!

PR #3787, opened 2024-08-09
MAINT: Switch run_notebooks.yml to use uv

connortann:
Brings this CI workflow in line with how dependencies are installed in the main test suite.


thatlittleboy (approved):
corgi-excited-tippytaps

PR #3377, opened 2023-11-02
add waterfall color config for waterfall plot

CloseChoice:
## Overview

Closes #3266 

Description of the changes proposed in this pull request:

## Checklist

- [x] All [pre-commit checks](https://pre-commit.com/#install) pass.
- [x] Unit tests added (if fixing a bug or adding a new feature)

PR #3788, opened 2024-08-09
ENH: New interface to customise visualisations

### LINKED ISSUE #3266: No native way to change colors for plots such as Waterfall plots
### Problem Description

Hello, i've been using shap plots for a while and i've always had the same problem. Most plots have cmap  parameter for changing the colors, but some of them , like `shap.plots.waterfall` and `shap.plots.bar` doesn't have support  and i've really tried to edit things, and to some extent i've figured out a work-around to change the main colors. But there's no way to change things like text-color, edge-color (Which is pretty relevant , cause even if you change the main color, the border is still red) and other details. I suggest that a new parameter is introduced and colors are parametrized. Thank you!

### Alternative Solutions

```python
import matplotlib

default_pos_color = "#ff0051"
default_neg_color = "#008bfb"
positive_color = "#ca0020"
negative_color = "#92c5de"
shap.plots.waterfall(shap_values[0], show = False)
for fc in plt.gcf().get_children():
    for fcc in fc.get_children():
        if (isinstance(fcc, matplotlib.patches.FancyArrow)):
            if (matplotlib.colors.to_hex(fcc.get_facecolor()) == default_pos_color):
                fcc.set_facecolor(positive_color)
            elif (matplotlib.colors.to_hex(fcc.get_facecolor()) == default_neg_color):
                fcc.set_color(negative_color)
        elif (isinstance(fcc, plt.Text)):
            if (matplotlib.colors.to_hex(fcc.get_color()) == default_pos_color):
                fcc.set_color(positive_color)
            elif (matplotlib.colors.to_hex(fcc.get_color()) == default_neg_color):
                fcc.set_color(negative_color)
plt.show()
```

This is a workaround to change colors, but it doesn't work if you change the positive colors to other color than red because the edgecolor is still red 

### Additional Context

_No response_

### Feature request checklist

connortann:
## Overview

Closes #3266

Draws heavily from work already done by @CloseChoice in #3377 , but with a slightly different proposed interface to modify options following our subsequent discussion.

## Discussion

This is a prototype of the "hybrid approach" we discussed here: https://github.com/shap/shap/issues/3266#issuecomment-1833550189

Following that PR, I've taken the "waterfall" plot to start with. 

### Proposed design

The idea is to have a module-level setting that can be changed at the global level, or used as a context manager:

```python
# Modify global settings
shap.plots.set_style(text_color="green")
shap.plots.waterfall(shap_values)

# Or as a context manager
with shap.plots.style_context(text_color="black"):
    shap.plots.waterfall(shap_values)
```

This has a few advantages over some of our initial ideas:
- Consistency with wider ecosystem (e.g. matplotlib, arviz, seaborn)
- Separation of concerns: "setting style options" vs "making plots"
- Ability to override the global defaults, e.g. potentially with a stylesheet
- Ability to share config settings between different plots (e.g. primary color, secondary color...)

Matplotlib's rcParams is sometimes hard to use, as discussed in https://github.com/shap/shap/issues/3266#issuecomment-1813203451 . So, this PR proposes using type hints to make it easier to discover the set of valid settings.

I've tried to name the parameters in a way that is generalisable to other plots, beyond just the waterfall plot.

### Proposed roll-out

I propose adding this as a private module. That means that initially it will not be part of the public API, giving us some time to refine the design and roll it out to all the plots.

Then, when we're ready to make the functionality public, we can expose a selection of functions in the public `shap.plots` namespace. The public functions could be:

```python
# Set global state
set_style()

# Context manager
style_context()

# Get current state
get_style()

# Get default options
load_default_style()


```

### TODO

- [x] Define public methods to modify the global singleton
- [x] Use styles in waterfall_plot
- [x] Add more mpl_image_compare tests


shap/plots/_style.py
```diff
@@ -0,0 +1,109 @@
...
+# Singleton instance that determines the current style.
+# CAREFUL! Only access this instance from the `_style` namespace, do not import it directly.
+# This is to ensure that the correct instance is picked up when the STYLE object is
+# reassigned in the _style module.
+#
+#     from shap.plots import _style
+#     color = _style.STYLE.text_color       # Correct
+#
+#     from shap.plots._style import STYLE   # Wrong!
+#     color = STYLE.text_color
```
thatlittleboy: Would the better advice here be to just use the `get_style()` helper function instead of accessing `STYLE` directly?

shap/plots/_style.py
```diff
@@ -0,0 +1,109 @@
...
+    Example
+    -------
+    with shap.plots.style_overrides(text="black"):
+        shap.plots.waterfall(...)
+    """
+    global STYLE
+    old_style = STYLE
+    STYLE = replace(old_style, **kwargs)
+    yield
+    STYLE = old_style
```
thatlittleboy: I suggest we create a specific exception for when an invalid option is passed in, similar to pandas [OptionError](https://github.com/pandas-dev/pandas/blob/d093fae3cd3466fc7a493f271535483f69fbbda1/pandas/_config/config.py#L102).

RIght now we'll get a weird TypeError with StyleConfig `__init__` if an unaccepted kwarg is passed in.

ps: if you add this, then we need an equivalent test for this exception to be thrown.

shap/plots/_style.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+@contextmanager
+def style_overrides(**kwargs):
+    """Context manager to temporarily override a subset of parameters.
+
+    NOTE: This is experimental and subject to change!
+
+    Example
```
thatlittleboy: This should be `Examples`, not `Example`, in order to render correctly in the docs.

Same comment for above `style_context`

tests/plots/test_style.py
```diff
@@ -0,0 +1,47 @@
...
+
+from shap.plots import _style
+
+
+def test_default_style():
+    new_style = _style.load_default_style()
+    assert configs_are_equal(_style.STYLE, new_style)
+
+    new_style.secondary_color_negative = "black"
+    assert not configs_are_equal(_style.STYLE, new_style)
```
thatlittleboy: Discussion: I'm not sure we want them to be able to do this(?)

Perhaps the `StyleConfig` dataclass should be `frozen=True`. Any overrides should be via `style_overrides` context manager, else if the user is simply modifying their own style dataclass, they should create a new one. WDYT?

tests/plots/test_style.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def test_style_context():
+    custom_style = _style.StyleConfig(text_color="green")
+    assert _style.STYLE.text_color == "white"
```
thatlittleboy: Here, we should test only the behaviour of the contextmanager, so we have less coupling to the actual style itself.

How about this instead?

```python
original_text_color = _style.STYLE.text_color
custom_style = _style.StyleConfig(text_color="green")
with _style.style_context(custom_style):
    assert _style.STYLE.text_color == "green"
assert _style.STYLE.text_color == original_text_color
```

And same for `test_style_overrides`.

shap/plots/_style.py
```diff
@@ -0,0 +1,109 @@
...
+    -------
+    with shap.plots.style_overrides(text="black"):
+        shap.plots.waterfall(...)
```
thatlittleboy: 1. This should be `text_color=` instead of `text`. :)
2. Code blocks need to be indented 4 spaces in order to render correctly in the docs.

shap/plots/_waterfall.py
```diff
@@ -127,7 +127,12 @@ def waterfall(shap_values, max_display=10, show=True):
             neg_lefts.append(loc)
         if num_individual != num_features or i + 4 < num_individual:
             plt.plot(
-                [loc, loc], [rng[i] - 1 - 0.4, rng[i] + 0.4], color="#bbbbbb", linestyle="--", linewidth=0.5, zorder=-1
+                [loc, loc],
+                [rng[i] - 1 - 0.4, rng[i] + 0.4],
+                color=_style.STYLE.vlines_color,
```
thatlittleboy: Right now, this `_style.STYLE.xxx` is a little long and verbose. I'll suggest defining

```python
def plot_xxx():
    style_cfg = get_style()
    # ...
```

at the start of the function definition, then use `style_cfg.primary_color_positive` etc. down below. Seems cleaner to me.

tests/plots/test_style.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+# Helper functions to compare equality of config dataclasses
+
+
+def configs_are_equal(config1: _style.StyleConfig, config2: _style.StyleConfig):
+    d1 = asdict(config1)
+    d2 = asdict(config2)
+    return all(_values_are_equivalent(d1[key], d2[key]) for key in d1.keys())
```
thatlittleboy, line 59: We might want to first check if d1 and d2 have the same keys, if not return False straight away. Otherwise it might blow up when it receives d2 without the `d1 key`, for example, and make it harder for us when debugging failing tests.

shap/plots/_style.py
```diff
@@ -0,0 +1,109 @@
...
+STYLE = load_default_style()
+
+
+def get_style() -> StyleConfig:
+    """Return the current style configuration."""
+    return STYLE
+
+
+def set_style(style: StyleConfig):
+    """Set the current style configuration."""
```
thatlittleboy: Let's also mention specifically that this is for setting the plot style globally? It wasn't immediately obvious to me (as a reader / user of the API)

thatlittleboy (changes requested):
Nice, thank you. I'm generally agreeable with the design patterns, but maybe a couple of things to iron out / discuss. Since this is a new API and sets the tone for future PRs, I hope you don't mind we be a bit more thorough in this PR.

shap/plots/_style.py
connortann: Yes great idea, that would be much neater. Updated

shap/plots/_waterfall.py
connortann: Great shout, done

connortann:
> Nice, thank you. I'm generally agreeable with the design patterns, but maybe a couple of things to iron out / discuss. Since this is a new API and sets the tone for future PRs, I hope you don't mind we be a bit more thorough in this PR.

Absolutely. Thanks for the detailed review!

shap/plots/_style.py
connortann: Done. I raised the new exception `from` the previous one, so the user still sees the inner exception message which includes specifically which kwarg is invalid.

tests/plots/test_style.py
connortann: Good shout. I suggested doing this as an `assert`, as presently we don't ever expect the keys to differ.

tests/plots/test_style.py
connortann: Done

tests/plots/test_style.py
connortann: Is there much harm in allowing users to mutate StyleConfig classes? This seems like something a user might want to do, and not everyone will be familiar with `dataclasses.replace`.

Are there any downsides in allowing mutation?

connortann: Having thought about this more, I think you make a good point.

In particular, it's probably not a good design to ask users to provide the entire "Config" object just to change one or two parameters. That has a few downsides:
- The user has to pass in more information than necessary: all they really need to pass is the changed params
- It becomes more sensitive to breaking changes. E.g. if we add a new style option, then any code that tries to load in a previous StyleConfig will presumably break.

So, I think it would be better to design the API around _overriding_ parameters, rather than having to provide the entire set.

connortann:
@thatlittleboy following your comments I updated the design. I'd be grateful for another review!

Here's a quick overview of the update...

### Updated Design

The `set_style()` and `style_context()` functions now accept a dictionary partial options (a `StyleOptions` TypedDict), rather than requiring a full dataclass of all settings.

I think this is a nice improvement, as users only have to provide the options they wish to change, rather than a comprehensive set. It also makes the code more robust to breaking changes: if we add a new style option, it won't break existing code.

There is now just one context manager, just the "overrides" one.

The active `StateConfig` remains a dataclass, for easy use in plotting functions. It is now immutable (frozen) for simplicity. The style state is accessed like:

```python
style =  get_style()
color = style.text_color
```

### Implementation 

A priority is providing helpful type hints for valid config options, and I think I have a good solution. The [Unpack](https://typing.readthedocs.io/en/latest/spec/callables.html#unpack-kwargs) type hint is exactly what we need:

```python
def set_style(**options: Unpack[StyleOptions]): ...
```

This means that valid options are type checked and can be autocompleted in an IDE.

This design means we have two separate but closely related objects:
- `StyleOptions` is a TypedDict, which represents zero or more changes.
- `StyleConfig` is an immutable dataclass, which represents the complete configuration state.

I think this is a helpful separation, as these are two logically distinct concepts.

However, presently this requires a little bit of code duplication as the options need to be defined in both classes (at least until `Unpack` supports dataclasses: https://github.com/python/typing/issues/1495 ). I don't think this is a big deal: I added a unit test to ensure both classes have the same options.

The `Unpack` hint is new in python 3.11, so requires `typing-extensions` to backport to previous python versions. This is already a transitive dependency, so in this PR I've added it as a direct dependency.

(Edited for clarity)

tests/plots/test_style.py
connortann: Done, the StyleConfig is now frozen.

connortann:
@thatlittleboy it would be great to have another review if you have time

connortann:
@CloseChoice @dsgibbons if you have time a code review for this would be great

shap/plots/_style.py
```diff
@@ -0,0 +1,123 @@
...
+    yield
+    _STYLE = old_style
+
+
+def _apply_options(style: StyleConfig, changes: StyleOptions) -> StyleConfig:
+    """Return a new StyleConfig with any changes applied, handling any invalid options."""
+    valid_keys = set(f.name for f in fields(StyleConfig))
+    for key in changes.keys():
+        if key not in valid_keys:
+            raise InvalidOptionError(f"Invalid style config option: {key}")
```
CloseChoice: When I first saw the error I would expect that the values are checked as well. The plethora of options here is not really checkable but could we somehow just check for the correct type?

IMO we don't need to implement this now, think this is good for now

CloseChoice (approved):
Nice work @connortann and @thatlittleboy. I tried the new interface myself and I really like it. I just have one question which do not have to tackle now.

shap/plots/_style.py
connortann: That's a good point, I hadn't really considered it. That might be something we can come back to in a future PR to improve :)

For now, hopefully the type hints (`TypedDict` & `Unpack`) will help users validate that their changes are valid. Also, if a user _does_ provide an invalid object in a style setting, hopefully they will get an informative error message from matplotlib when they actually create a plot. But yes, type checking the inputs would be nice.

CloseChoice: absolutely, let's go ahead with the current implementation

shap/plots/_style.py
```diff
@@ -0,0 +1,123 @@
...
+
+def load_default_style() -> StyleConfig:
+    """Load the default style configuration."""
+    # In future, this could allow reading from a persistent config file, like matplotlib rcParams.
+    return _shap_defaults
+
+
+# Singleton instance that determines the current style.
+# CAREFUL! To ensure the correct object is picked up, do not import this directly,
+# but intead access this at runtime with get_style().
```
thatlittleboy:
```suggestion
# but instead access this at runtime with get_style().
```

typo

shap/utils/_exceptions.py
```diff
@@ -38,3 +38,7 @@ class InvalidModelError(ValueError):
 
 class InvalidClusteringError(ValueError):
     pass
+
+
+class InvalidOptionError(ValueError):
```
thatlittleboy: How about `InvalidStyleOptionError` instead? invalid option seems too vague to me.

shap/plots/_style.py
```diff
@@ -0,0 +1,123 @@
...
+    To set a new style::
+
+        with shap.plots.style_context(text_color="black"):
+            shap.plots.waterfall(...)
+    """
+    global _STYLE
+    old_style = _STYLE
+    _STYLE = _apply_options(_STYLE, options)
+    yield
+    _STYLE = old_style
```
thatlittleboy: quick question for my understanding: why doesn't the implementation of `style_context` use the `get_style()` and `set_style()` functions?

As in:

```python
def style_context(**options):
    old_style = get_style()
    set_style(**options)
    yield
    set_style(**asdict(_style.load_default_style()))  # or even better, if set_style can support it, `set_style(old_style)`
```

This seems like it would be easier to understand and maintain,no? ^

shap/plots/_style.py
```diff
@@ -0,0 +1,123 @@
...
+def get_style() -> StyleConfig:
+    """Return all currently active global style configuration options."""
+    return _STYLE
+
+
+def set_style(**options: Unpack[StyleOptions]) -> None:
+    """Set options in the currently active global style configuration."""
+    global _STYLE
+    _STYLE = _apply_options(_STYLE, options)
+
```
thatlittleboy: Discussion point: I like the StyleOptions solution, but I still think it would be valuable to allow users to pass in their own StyleConfig to `set_style`, so that we can have an API that guarantees:

```python
t1 = get_style()
set_style(t1)
t2 = get_style()
assert t1 == t2
```

Also, since `get_style()` returns a StyleConfig, I (from the user's perspective) would expect the similarly-named `set_style` function would then accept a StyleConfig.

WDYT? @connortann @CloseChoice

---

To be clear: I'm proposing a signature something like this:

```python
def set_style(_style: StyleConfig | None = None, /, **options: Unpack[StyleOptions]) -> None:
    if isinstance(_style, StyleConfig):
        options |= **asdict(_style)
    ...
```
Note this merges the input StyleConfig with any additional options specified as kwargs. Though I would say that's abit of an unconventional use -- I would expect we use one or the other generally, but I don't really see a reason to prevent the user from doing that yet.  
And we might want to prefix the positional arg with an underscore to prevent any unlikely clashes with the options in the future.

shap/plots/_style.py
```diff
@@ -0,0 +1,123 @@
...
+@contextmanager
+def style_context(**options: Unpack[StyleOptions]):
+    """Context manager to temporarily change style options.
+
+    NOTE: This is experimental and subject to change!
+
+    Examples
+    --------
+    To set a new style::
```
thatlittleboy: Let's be abit clearer with our explanation/example here:

```suggestion
    To temporarily use black text color instead of the default (white)::
```

thatlittleboy (changes requested):
It's coming along really nicely, thanks for the good work; Just a few more comments / discussion points! :)

shap/utils/_exceptions.py
connortann: Done

shap/plots/_style.py
connortann: Yes, I think that would be simpler. Updated accordingly (although using `asdict(old_style)` rather than `asdict(_style.load_default_style())` , as I presume would be your suggestion)

shap/plots/_style.py
connortann: That's an interesting idea. I do see what you mean about `get_style()` and `set_style()` seeming like they should accept and return the same kind of object. I like your idea, although the type signature is getting a little complex.

How about this as an alternative way to achieve the same aim. We could make `get_style` and `set_style` both use `StyleOptions` objects, by introducing seperate `get_active_state()` function for use in plotting funcs:

```python
# Public API
def get_style() -> StyleOptions: ...
def set_style(**options: Unpack[StyleOptions]) -> None: ...

# For accessing the current runtime values
def get_active_state() -> StyleConfig: ... 
```

That way, the API is consistent in the sense you can do:
```python
t1 = get_style()
set_style(**t1)
t2 = get_style()
assert t1 == t2
```

connortann: I pushed a commit with this update, to explore how it looks. I think it's probably an improvement. What do you think @thatlittleboy ?

CloseChoice: I would prefer @thatlittleboy's proposal just because output and input parameters do not need an additional transformation. 

I would suggest to use `typing.overload` though, in order to keep this as simple as possible for the user:

```python
@overload
def set_style(style: StyleConfig) -> None:
    ...

@overload
def set_style(**options: Unpack[StyleOptions]) -> None:
    ...

# Implementation of the function
def set_style(style=None, **options):
    if isinstance(style, StyleObject):
        print(f"Setting style using StyleObject: font={style.font}, size={style.size}")
    elif isinstance(options, dict):
        print(f"Setting style using keyword arguments: {options}")
    else:
        raise TypeError("Invalid arguments provided")
```

connortann: One worry I have about accepting `StyleConfig` as an input is that it's quite susceptible to any breaking changes.

Specifically: if client code creates a `StyleConfig` object, and then we subsequently add a new style option, the client code will be broken as the definition of the dataclass will be incomplete.

So, I think it's safer to have the "set" API designed around the `dict` object which represents a possibility partial set of changes. Open to other ideas though

CloseChoice: Wouldn't we just need to set a default for the new option, so that it's backwards compatible? I wouldn't want code to break just to have a new option.

connortann: OK, I've updated `set_style` to allow passing in a dataclass as a positional parameter.

I still remain a little uncomfortable with this method of setting options, as it seems like a bad conceptual fit. `StyleOptions` represents a "possibly partial set of updates", whereas the dataclass represents a "state of all options" that is guaranteed to be total. The former object seems like the better choice for setting params. 

RE breaking changes: adding default values to all items in the dataclass would indeed help in this regard, but feels a bit messy to me. It is open to misinterpretation: a user might think that `set_style(StyleConfig(foo="bar"))` would simply update the option `foo`, whereas actually this would have the side-effect of setting all other style options to the library defaults, potentially overriding anything that was previous set from a user stylesheet. 

So, when we create documentation and examples for this new functionality, I do think we should encourage the setting of options via keyword arguments, and not by instantiating a dataclass directly. Or, put another way, I don't think the DataClass should ever be instantiated directly by a user.

connortann: PS - this can always be updated in future PRs. I would be keen to try and get a "good enough" first pass implementation merged in, and we are free to add improvements in future.

thatlittleboy: Thanks @connortann  and @CloseChoice ,

> I do think we should encourage the setting of options via keyword arguments, and not by instantiating a dataclass directly. Or, put another way, I don't think the DataClass should ever be instantiated directly by a user.

Agreed.

Let's roll with this implementation first, we can revisit this again in the future once we start playing around more with this (and hopefully also get some user feedback).

thatlittleboy (approved):
LGTM!

zuckerberg-nice-job-team

connortann:
Thanks both for the very detailed reviews on this one! Your feedback was really helpful, lots of good improvements compared to the first draft.

PR #3790, opened 2024-08-10
MAINT: Add timeout to run_tests

connortann:
I noticed one recent test run timeout after 6 hours (having seemingly hung), so I propose adding a timeout to fail after a shorter time.

At present the slowest job (Windows) takes ~18 minutes, so I think 40 mins should provide plenty of leeway.


connortann:
I will go ahead and merge this one, as I'm a bit worried about accidentally burning through lots of GitHub actions hours on current PRs. Each "hang" event currently uses 6 hours per job :S

Nonetheless I'd appreciate a quick review indeed retrospect :)

PR #3791, opened 2024-08-11
MAINT: Add tests for scatter plot

connortann:
## Overview

Adds unit tests for `shap.plots.scatter`

Supports #3804


tests/plots/test_scatter.py
```diff
@@ -0,0 +1,52 @@
...
+import shap
+
+
+@pytest.mark.mpl_image_compare
+def test_scatter_single(explainer):
+    explanation = explainer(explainer.data)
+    shap.plots.scatter(explanation[:, "Age"], show=False)
+    fig = plt.gcf()
+    plt.tight_layout()
+    return fig
```
thatlittleboy, line 13: Can you also add a test for the multiple columns variant of scatter plot?

tests/plots/test_summary.py
```diff
@@ -123,16 +123,18 @@ def test_random_summary_with_log_scale():
...
 
-def test_summary_plot_with_multiclass_model():
+@pytest.mark.parametrize("background", [True, False])
+def test_summary_binary_multiclass(background):
     # See GH #2893
     lightgbm = pytest.importorskip("lightgbm")
-    num_examples, num_features = 20, 3
-    X = np.random.uniform(size=[num_examples, num_features])
-    y = np.random.choice([0, 1], size=num_examples)
+    num_examples, num_features = 100, 3
```
connortann, line 130: To provide a bit of context: this change isn't strictly related to the scatterplot, but I noticed this test was throwing warnings due to zero SHAP values being plotted, so I fixed it up by increasing the number of examples slightly.

tests/plots/test_scatter.py
connortann: Good suggestion, done.

One bit that remains uncovered by tests is plotting interaction values, in this block:
https://github.com/shap/shap/blob/a82f014a5cbc08f9ae3ece6ad2b08a603dc2cb48/shap/plots/_scatter.py#L250-L251

 I couldn't work out the correct input data to invoke this functionality , so if you know how to do this that would make for a useful follow-up PR.

thatlittleboy: > I couldn't work out the correct input data to invoke this functionality

I took a glance, neither could I. Isn't it dead code though, since `ind=0` is set at the beginning, then `convert_name(ind,...)` will return 0 since it's not a string. It looks awfully like dead code, but let's leave this for a future PR.

---

ps: I just noticed in our docstrings, we only mentioned single column. Let's edit this as well while we're at it? We technically support multiple columns + the behaviour is tested now.

image

connortann: Good shout, done.

thatlittleboy (approved):
sorry this took a while, lgtm

seal-of-approval-seal

PR #3792, opened 2024-08-11
MAINT: Improved isolation and cleanup of matplotlib tests 

connortann:
Improved isolation of matplotlib tests:
- Call `plt.close()` _after_ each test, rather than before
- Moves the matplotlib fixture to the top level, as there are some plotting tests outside of `tests/plots`
- Set the matplotlib backend in the fixture, rather than as an import-level side-effect in various places.
- Ensures `show=False` is passed, to fix test warnings about plotting in a non-interactive environment

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,6 +8,7 @@ try:
 except ImportError:
     pass
 
+import matplotlib.pyplot as plt
 import numpy as np
 import pytest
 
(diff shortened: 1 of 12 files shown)


thatlittleboy (approved):
nice use of `plt.rc_context()`!

singapore-nsp-thumbsup-man

PR #3706, opened 2024-06-21
#3135: Adds categorical support for shap.plots.scatter

hypostulate:
## Overview

Closes #3135

Adds categorical scatter support. Requires the Explainer to be setup in a method compatible with true category dtypes.

Note: One failing test was found _before_ feature was created. That test is still failing on this commit.

## Checklist

- [X] All [pre-commit checks](https://pre-commit.com/#install) pass.
- [X] Unit tests added (if fixing a bug or adding a new feature)

PR #3811, opened 2024-08-13
ENH: Improved typing and docs for scatter plot

### LINKED ISSUE #3804: Plot overhaul: scatter

connortann:
## Overview

Closes #3804

## Discussion

A number of refactoring improvements to `shap.plots.scatter`:

- Adds type hints for most parameters, and gets it passing with mypy
- Updates docstring to describe some missing params  (e.g. `overlay`), and explain the range of supported options (e.g. the axis limits)
- Returns `ax` if `show=False`
- Creates a few helper functions to parse the axis limits, to reduce the size of the mammoth main function
- Removes a bit of dead code relating to calling `legacy_dependence`
- Renames matplotlib import from `pl` to the canonical `plt`
- Replaces some `assert` statements with raising an appropriate `Exception`

There is still more to be done, but I think this is a manageable start.


connortann:
@thatlittleboy  I have updated this PR to fix some conflicts that arose from the merging of #3706 .

shap/plots/_scatter.py
```diff
@@ -71,101 +84,107 @@ def scatter(
         increase plot readability when a feature is discrete. By default, ``x_jitter``
         is chosen based on auto-detection of categorical features.
 
+    title: Optional str
```
thatlittleboy: I believe the numpydoc spec is to use `str, optional`, [REF](https://numpydoc.readthedocs.io/en/latest/format.html#parameters)

shap/plots/_scatter.py
```diff
@@ -528,7 +426,106 @@ def scatter(
...
+    if ax_max is None:
+        ax_max = float(nan_max + buffer)
+    return ax_min, ax_max
+
+
+def _suggest_x_jitter(values: np.ndarray) -> float:
+    """Suggest a suitable x_jitter value based on the unique values in the feature"""
+    unique_vals = np.sort(np.unique(values))
+    try:
+        # Identify the smallest difference between uniqu evalues
```
thatlittleboy:
```suggestion
        # Identify the smallest difference between unique values
```
small typo here

shap/plots/_scatter.py
```diff
@@ -71,101 +84,107 @@ def scatter(
...
         if show:
-            pl.show()
+            plt.show()
         return
 
     if len(shap_values.shape) != 1:
-        raise Exception(
-            "The passed Explanation object has multiple columns, please pass a single feature column to "
-            "shap.plots.dependence like: shap_values[:,column]"
+        raise ValueError(
```
thatlittleboy: I think DimensionError is better suited here.

shap/plots/_scatter.py
```diff
@@ -47,17 +55,22 @@ def scatter(
...
-        Typically a single column of an :class:`.Explanation` object (i.e.
-        ``shap_values[:,"Feature A"]``).
-        Alternatively, pass multiple columns to create several subplots.
+        Typically a single column of an :class:`.Explanation` object
+        (i.e. ``shap_values[:, "Feature A"]``).
+
+        Alternatively, pass multiple columns to create several subplots
+        (i.e. ``shap_values[:, ["Feature A", "Feature B"]]``).
 
     color : string or shap.Explanation
```
thatlittleboy: color is optional (it can be None), in which case no interaction feature is not plotted

```
color : string or shap.Explanation, optional
```

shap/plots/_scatter.py
```diff
@@ -1,36 +1,44 @@
...
+    xmin: LimitSpec = None,
+    xmax: LimitSpec = None,
+    ymin: LimitSpec = None,
+    ymax: LimitSpec = None,
+    overlay: dict[str, Any] | None = None,
+    ax: plt.Axes | None = None,
+    ylabel: str = "SHAP value",
+    show: bool = True,
 ):
     """Create a SHAP dependence scatter plot, colored by an interaction feature.
```
thatlittleboy:
```suggestion
    """Create a SHAP dependence scatter plot, optionally colored by an interaction feature.
```

shap/plots/_scatter.py
```diff
@@ -528,7 +426,106 @@ def scatter(
...
+    else:
+        return ax
+
+
+def _parse_limit(ax_limit: LimitSpec, ax_values: np.ndarray, is_shap_axis: bool) -> float | None:
+    """Handle axis limits in "percentile(float)" format or from Explanation objects"""
+    if isinstance(ax_limit, str):
+        if not ax_limit.startswith("percentile(") or not ax_limit.endswith(")"):
+            raise ValueError("Only strings of the format `percentile(x)` are supported.")
+        return np.nanpercentile(ax_values, float(ax_limit[11:-1]))
```
thatlittleboy: I suggest just do a try-except here instead, and do a `raise from`. Because there's still a case which is not covered: when `ax_limit[11:-1]` is not a number, then `float()` would throw an error.

shap/plots/_scatter.py
```diff
@@ -71,101 +84,107 @@ def scatter(
...
+    xmin, xmax, ymin, ymax : float, string, aggregated Explanation or None
+        Desired axis limits. Can be a float to specify a fixed limit.
 
-    xmax : float or string
-        Represents the upper bound of the plot's x-axis. It can be a string of the format
-        "percentile(float)" to denote that percentile of the feature's value used on the x-axis.
+        It can be a string of the format ``"percentile(float)"`` to denote that
+        percentile of the feature's value.
+
+        It can also be an aggregated column of a single column of an Explanation,
```
thatlittleboy:
```suggestion
        It can also be an aggregated column of a single column of an :class:`.Explanation`,
```

shap/plots/_scatter.py
```diff
@@ -71,101 +84,107 @@ def scatter(
...
-            "shap.plots.dependence like: shap_values[:,column]"
+        raise ValueError(
+            "The passed Explanation object has multiple columns. Please pass a single feature column to "
+            "shap.plots.scatter like: shap_values[:, column]"
         )
 
     # this unpacks the explanation object for the code that was written earlier
     feature_names = [shap_values.feature_names]
-    ind = 0
+    ind: Any = 0
```
CloseChoice: `int` is not working here?

shap/plots/_scatter.py
```diff
@@ -248,49 +244,9 @@ def scatter(
     categorical_interaction = False
 
     # create a matplotlib figure, if `ax` hasn't been specified.
-    if not ax:
+    if ax is None:
```
CloseChoice, line 245: we are doing the opposite here than compared to the previous version. I don't see how the previous case (including the recursion) is handled. Unfortunately I believe the plots are not really well covered by tests.

shap/plots/_scatter.py
```diff
@@ -71,101 +84,107 @@ def scatter(
...
-        nan_min = np.nanmin(shap_values.values)
-        nan_max = np.nanmax(shap_values.values)
-        if ymin is None:
-            ymin = nan_min - (nan_max - nan_min) / 20
-        if ymax is None:
-            ymax = nan_max + (nan_max - nan_min) / 20
-        # FIXME: the following code ignores any passed in `ax`
-        _ = pl.subplots(1, len(inds), figsize=(min(6 * len(inds), 15), 5))
+        ymin = _parse_limit(ymin, shap_values.values, True)
+        ymax = _parse_limit(ymax, shap_values.values, True)
```
thatlittleboy: nit: I generally oppose specifying arguments like "True" as a positional arg because it is not clear what it refers to, and is particularly prone to bugs, e.g. if we change from `is_shap_axis` to `is_y_axis` but forget to change from True->False here.

So I would suggest:

```suggestion
        ymin = _parse_limit(ymin, shap_values.values, is_shap_axis=True)
        ymax = _parse_limit(ymax, shap_values.values, is_shap_axis=True)
```

and same for the other calls to `_parse_limit` below.

CloseChoice (changes requested):

shap/plots/_scatter.py
```diff
@@ -528,7 +426,106 @@ def scatter(
...
+    return ax_min, ax_max
+
+
+def _suggest_x_jitter(values: np.ndarray) -> float:
+    """Suggest a suitable x_jitter value based on the unique values in the feature"""
+    unique_vals = np.sort(np.unique(values))
+    try:
+        # Identify the smallest difference between uniqu evalues
+        diffs = np.diff(unique_vals)
+        min_dist = np.min(diffs[diffs > 1e-8])
```
thatlittleboy, line 466: cmiiw, but isn't it possible that `diffs` here will all contain `diffs > 1`?
In which case, your implementation would be different from the version before refactoring, which essentially does a `min(*diffs, 1)` 

image

shap/plots/_scatter.py
```diff
@@ -214,31 +233,8 @@ def scatter(
     ind = convert_name(ind, shap_values_arr, feature_names)
```
thatlittleboy: `ind` is set to 0 above 

image

so this call to `convert_name` is pretty much just dead code. (`convert_name` only does something if the input is a string)

shap/plots/_scatter.py
```diff
@@ -214,31 +233,8 @@ def scatter(
...
-            if x_jitter == "auto":
-                x_jitter = min_dist * 0.1
-        else:
-            # categorical = True
-            if x_jitter == "auto":
-                x_jitter = min_dist * 0.2
+        x_jitter = _suggest_x_jitter(features[:, ind])
 
     # guess what other feature as the stongest interaction with the plotted feature
     if not hasattr(ind, "__len__"):
```
thatlittleboy: similarly here: `ind` will never have the `__len__` attribute, it's the literal 0.

thatlittleboy (changes requested):

shap/plots/_scatter.py
connortann: Yes, I believe that is correct. However, I think this change is desirable.

Up until 4 days ago, the `min_dist` was simply the minimum distance between points:

```
min_dist = np.inf
for i in range(1, len(vals)):
    d = vals[i] - vals[i - 1]
    if d > 1e-8 and d < min_dist:
        min_dist = d
```

However, this was updated recently by #3706 to fix the plotting of categorical variables. The `min_dist=1` default is fairly arbitrary, intended for the case where there are no numeric values to compare. When reviewing that PR I missed that it would actually always limit `min_dist` to 1, so here I propose reverting to the prior behaviour :)

shap/plots/_scatter.py
connortann: Fixed. Thanks for spotting these details, very helpful to have nice docs :)

shap/plots/_scatter.py
connortann: Fixed

shap/plots/_scatter.py
connortann: Agreed!

shap/plots/_scatter.py
connortann: Good idea

shap/plots/_scatter.py
connortann: Good catch. I hadn't looked at the use of `ind` too closely, so I'll take another look and try to simplify that. It looks like this is just a hangover from code in `scatter_legacy` that was copied over, so it should be safe to remove now.

EDIT: Done

shap/plots/_scatter.py
connortann: Fixed! It wasn't working earlier, but it does now that the dead code below is removed :)

shap/plots/_scatter.py
connortann: Agreed, fixed. In future, we could consider adding this linting rule:
https://docs.astral.sh/ruff/rules/boolean-type-hint-positional-argument/

I found it interesting reading about the motivation behind it. Nice to learn something new!
https://adamj.eu/tech/2021/07/10/python-type-hints-how-to-avoid-the-boolean-trap/

shap/plots/_scatter.py
connortann: @closechoice what do you mean, what change is this the opposite behaviour?

I added some tests for the scatter plot in a recent PR so hopefully we have most cases covered. I _think_ that most of the remaining uncovered code is "dead code" so to speak.

Are there any particular cases that you think we should cover?

shap/plots/_scatter.py
thatlittleboy: got it, thanks for the explanation!

shap/plots/_scatter.py
```diff
@@ -71,101 +85,107 @@ def scatter(
...
+            raise ValueError("The ax parameter is not supported when plotting multiple features")
+        # Define order of columns (features) to plot based on average shap value
         inds = np.argsort(np.abs(shap_values.values).mean(0))
-        nan_min = np.nanmin(shap_values.values)
-        nan_max = np.nanmax(shap_values.values)
-        if ymin is None:
-            ymin = nan_min - (nan_max - nan_min) / 20
-        if ymax is None:
-            ymax = nan_max + (nan_max - nan_min) / 20
-        # FIXME: the following code ignores any passed in `ax`
```
thatlittleboy, line 113: has this FIXME been fixed?

shap/plots/_scatter.py
```diff
@@ -71,101 +85,107 @@ def scatter(
...
-        Optionally specify an existing matplotlib ``Axes`` object, into which the plot will be placed.
-        In this case, we do not create a ``Figure``, otherwise we do.
+        It can also be an aggregated column of a single column of an :class:`.Explanation`,
+        such as ``explanation[:, "feature_name"].percentile(20)``.
+
+    overlay: dict, optional
+        Optional dictionary of curves to overlay as line plots.
+
+        The dictionary maps a curve name to a list of (xvalues, yvalues) pairs, where
+        there is one pair for each feature to be plotted.
```
thatlittleboy: We might want to also specify that overlay only takes up to a maximum of 3 additional curves for plotting.

Another thing I noticed was that `overlay` is only used when we are plotting multiple columns.. Not sure why this was the design originally? But I think it's alright, let's leave it as is for now.  
Though I think it would be worth considering deprecating `overlay` parameter in the future, once we fully embrace the Axes API. Because then, the user would be able to customize the plot more freely, and we can simplify our scatter plots API and code.

If you agree on the latter point, I'll create a corresponding issue so that we can come back to this later.

shap/plots/_scatter.py
connortann: Yes, in a sense. When multiple columns are passed this function creates a figure with several subplots, so it's a figure-level function rather than an axes-level function and the `ax` parameter doesn't really make sense. So, now on Line ~137 we raise a ValueError if the `ax` parameter is passed when plotting multiple features.

shap/plots/_scatter.py
connortann: > specify that overlay only takes up to a maximum of 3 additional curves

Good idea, done.

> it would be worth considering deprecating overlay parameter in the future

Yeah, I think that's a good idea for a future change. It's becoming apparent that this function is doing a lot of work, and the overlay functionality seems a little niche and over-complex.

thatlittleboy (approved):
Thanks, LGTM

squidward-lacklustre-hooray

PR #3815, opened 2024-08-17
chore: Add type hints for datasets module and improve docs

thatlittleboy:
## Overview

- Adds some type hints for the datasets in `shap.datasets`
- Standardize / correct some of the output formatting of the docstrings. E.g. we had issues like this:

## Checklist

### DIFF
--- a/shap/datasets.py
+++ b/shap/datasets.py
@@ -10,7 +10,7 @@ import sklearn.datasets
 
 import shap
 
-if TYPE_CHECKING:
+if TYPE_CHECKING:  # pragma: no cover
     import scipy.sparse as ssp
 
 github_data_url: Final[str] = "https://github.com/shap/shap/raw/master/data/"
--- a/shap/datasets.py
+++ b/shap/datasets.py
@@ -1,4 +1,7 @@
+from __future__ import annotations
+
 import os
+from typing import TYPE_CHECKING, Final
 from urllib.request import urlretrieve
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


shap/datasets.py
```diff
@@ -84,10 +102,11 @@ def california(n_points=None):
     To get the processed data and target labels::
 
         data, target = shap.datasets.california()
+
     """
     d = sklearn.datasets.fetch_california_housing()
     df = pd.DataFrame(data=d.data, columns=d.feature_names)
-    target = d.target
+    target: np.ndarray = d.target
```
connortann, line 109: Just to check, should this be `pd.Series`?

connortann (approved):

connortann:
Looks great! The rendered docs look much better.

CloseChoice (approved):
Nice work, looks great

shap/datasets.py
CloseChoice: yes, I think it should be a series

thatlittleboy: @connortann @CloseChoice hi folks, sorry, just to check, do you mean we want to return a pandas series here, or you mean `d.target` is a Series?

Cuz I checked, it's a numpy array:

image

CloseChoice: I thought `d` is a `pd.Series`. Seems not to be the case so everything's settled in here. Ready to merge IMO.

PR #3822, opened 2024-08-22
MAINT: Fixup unused variables from flake8-bugbear

connortann:
Fixes a few minor issues, as caught by a miscellaneous selection of linting rules such as flake8-bugbear.

Does not actually enforce these rules for now on CI, as I found a few false positives: just includes the fixes which I think lead to simpler code.

- Removes unused loop variables (e.g. `enumerate` statements)
- Simplify some repeated comparisons
- Break down `assert x and y` statements into multiple asserts
- Replaces redundant use of `setattr` where simple attribute access will do


shap/explainers/_explainer.py
```diff
@@ -391,7 +391,7 @@ def __call__(
 
         # split the values up according to each input
         arg_values = [[] for a in args]
-        for i, v in enumerate(values):
+        for i in range(len(values)):
```
CloseChoice, line 394: this is a nice catch

shap/plots/_force_matplotlib.py
```diff
@@ -13,9 +13,9 @@ def draw_bars(out_value, features, feature_type, width_separators, width_bar):
     separator_list = []
 
     pre_val = out_value
-    for index, features in zip(range(len(features)), features):
+    for index, feature_values in enumerate(features):
```
CloseChoice, line 16: okay, I am getting impressed by this tool

CloseChoice (approved):
Looks neat.

Fixes a lot of things that are caught by the eye when reading the code. Think having this tool (even in CI at some point) is a huge improvement.

PR #3833, opened 2024-08-30
refactor: plot utils and colors

thatlittleboy:
## Overview

Just some tidying up before I start work proper on the overhaul of the `partial_dependence_plot` plotting function (as part of the visualization overhaul).

Changes:

* Add some type hints to the `colors` module
* Migrate the recently added `LimitSpec` type and `_parse_limit` helper functions (added in #3811) to the `shap/plots/_utils.py` module. (Discussion below)
* Renamed the `LimitSpec` -> `AxisLimitSpec`, and `_parse_limit` -> `parse_axis_limit` (only because it is now in the utils module instead of only in `_scatter.py`, and this naming makes it clearer what these are used for).
* Made the `is_shap_axis` parameter a kw-only. Discussion for why is in the previous [PR review](https://github.com/shap/shap/pull/3811#discussion_r1729783962).

## Discussion

I'm proposing to migrate the recently added `LimitSpec` type and `_parse_limit` helper functions (added in #3811) to the `shap/plots/_utils.py` module. Reason is because the same functionality is also used in `partial_dependence_plot`, so thought we could refactor this out into the general utils module. (ps: I'll work on the pdp overhaul next, just wanted to get this out of the way)

In light of this, I've renamed the 2 objects, & added some docstrings. To be kinder to our future selfs :)

@connortann let me know what you think since these 2 were introduced by you. Open for discussion, because FWIW: I didn't see similar parsing axis-limit functionality being required in other plotting functions (but I only glanced briefly), so I guess it would also be okay duplicating the functions in `_scatter.py` and `_partial_dependence.py` respectively.

### DIFF
--- a/shap/plots/_utils.py
+++ b/shap/plots/_utils.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Literal, overload
+from typing import TYPE_CHECKING, Literal, Union, overload
 
 import matplotlib.pyplot as plt
 import numpy as np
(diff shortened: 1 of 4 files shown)


shap/plots/_utils.py
```diff
@@ -1,19 +1,32 @@
...
 from . import colors
 
+if TYPE_CHECKING:
+    from matplotlib.colors import Colormap
+
+
+@overload
+def convert_color(color: Literal["shap_red", "shap_blue"]) -> np.ndarray: ...
+@overload
+def convert_color(color: str) -> np.ndarray | str | Colormap: ...
```
connortann: Nit:  I feel that the `@overload` decorator is quite advanced python, and quite verbose, so we should be judicious in choosing where to use it. 

In this case I feel the different overloads are essentially very similar so I would question whether the type narrowing adds much benefit - the return object will be used in exactly the same way in each case.

shap/plots/_utils.py
```diff
@@ -226,3 +239,40 @@ def sort_inds(partition_tree, leaf_values, pos=None, inds=None):
...
+        Can be in "percentile(float)" format, a float or an :class:`.Explanation` object that has been aggregated
+        into a single value, e.g. ``explanation[:, "feature_name"].percentile(20)``.
+
+    ax_values : np.ndarray
+        The values represented by the axis in question. Usually the SHAP values or the feature values. Only used if
+        ``ax_limit`` is a string of the "percentile(float)" form.
+
+    is_shap_axis : bool
+        Whether the ``ax_limit`` is describing the axis representing SHAP values, or not. Only relevant when
+        ``ax_limit`` is an :class:`.Explanation` object. It is assumed that when False, the axis is representing
```
connortann, line 260: Nit: as this function is for internal use only, there probably not any benefit in RST-style hyperlinks as they will never be rendered in Sphinx.

connortann (approved):
A few optional nitpick suggestions (all of which are you to you), and otherwise LGTM! Yeah, I agree it makes sense to refactor those helper functions to a shared module.

shap/plots/_utils.py
```diff
@@ -226,3 +239,40 @@ def sort_inds(partition_tree, leaf_values, pos=None, inds=None):
     sort_inds(partition_tree, leaf_values, right, inds)
 
     return inds
+
+
+# Various ways to specify a desired axis limit in plots
+AxisLimitSpec = Union[Explanation, str, float, None]
```
connortann: Nit: Now that we have `typing_extensions` as a dependency, we could use `TypeAlias` here to make it explicit that this is an Alias and not a regular variable assignment.

```suggestion
AxisLimitSpec: TypeAlias = Union[Explanation, str, float, None]
```

shap/plots/_utils.py
thatlittleboy: Sure, I'll remove it.

shap/plots/_utils.py
thatlittleboy: I see what you mean. Though I rather we maintain a consistent docstyle format throughout. I don't think readability is affected that much (after all, for public API docstrings we've gotten used to reading these markup in the source code).

PR #3834, opened 2024-08-31
FIX: Improve handling of the `approximate` parameter to TreeExplainer for consistency, and deprecate the argument in the explainer's init method

### LINKED ISSUE #3764: BUG: correct "approximate" parameter description in TreeExplainer
### Issue Description

It seems like in the recent version of shap "approximate" parameter in TreeExplainer does nothing.

I tried setting it to both True and False and did not notice any difference in runtime, + outputs were exactly the same.
This is consistent across all possible values of feature_perturbation parameter, be it `interventional` or `tree_path_dependent.`

This experience is not aligned with the docs:

> approximate bool
> Run fast, but only roughly approximate the Tree SHAP values. This runs a method previously proposed by Saabas which only considers a single feature ordering. Take care since this does not have the consistency guarantees of Shapley values and places too much weight on lower splits in the tree.

```python
for explainer in tqdmu([shap.explainers.TreeExplainer(full_model,X_val,feature_perturbation= "interventional",approximate=False),
                        shap.explainers.TreeExplainer(full_model,X_val,feature_perturbation= "interventional",approximate=True),
                        shap.explainers.TreeExplainer(full_model,feature_perturbation= "tree_path_dependent",approximate=False),
                        shap.explainers.TreeExplainer(full_model,feature_perturbation= "tree_path_dependent",approximate=True),]):
    
    start=timer()
    shap_values=explainer.shap_values(X_val)
    print(timer()-start,shap_values[:5,0])
```

### Traceback

```shell
100%|===================| 9997/9999 [09:10<00:00]        
550.7331977249996 [ 0.00150003  0.00044561  0.00491552 -0.00327678  0.00423344]
100%|===================| 9983/9999 [09:19<00:00]        
559.8140876140023 [ 0.00150003  0.00044561  0.00491552 -0.00327678  0.00423344]

0.7823838650001562 [ 0.00113404  0.00081992  0.0052229  -0.00292402  0.00476501]
0.8053742159972899 [ 0.00113404  0.00081992  0.0052229  -0.00292402  0.00476501]
```

### Expected Behavior

I'd expect timing and shap values to be different.

### Bug report checklist

### Installed Versions

0.46.0

CloseChoice:
## Overview

Closes #3764 

 - make use of parameter `approximate` consistent for the deepexplainer, which means deprecating `approximate` in the constructor and only supporting it in the `shap_values` and `__call__` functions
 - adding type hints to various methods in the treeexplainer

## Checklist

### DIFF
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -297,7 +297,7 @@ class TreeExplainer(Explainer):
         """
         start_time = time.time()
 
-        feature_names: Any | list[Any] | None
+        feature_names: Any
         if isinstance(X, pd.DataFrame):
             feature_names = list(X.columns)
         else:
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -5,6 +5,7 @@ import json
 import os
 import time
 import warnings
+from typing import Any
 
 import numpy as np
 import pandas as pd
@@ -35,6 +36,8 @@ try:
     import pyspark  # noqa
 except ImportError as e:
     record_import_error("pyspark", "PySpark could not be imported!", e)
+
+DEPRECATED_APPROX = object()
 
 output_transform_codes = {
     "identity": 0,
@@ -92,7 +95,7 @@ class TreeExplainer(Explainer):
         model_output="raw",
         feature_perturbation="interventional",
         feature_names=None,
-        approximate=False,
+        approximate=DEPRECATED_APPROX,
         # FIXME: The `link` and `linearize_link` arguments are ignored. GH #3513
         link=None,
         linearize_link=None,
@@ -155,7 +158,17 @@ class TreeExplainer(Explainer):
             Currently the "probability" and "log_loss" options are only
             supported when ``feature_perturbation="interventional"``.
 
+        approximate : bool
+            Deprecated, will be removed in a version v.0.48.0. Please use the approximate argument in the
+            ``shap_values`` or ```__call__`` methods instead.
+
         """
+        if approximate is not DEPRECATED_APPROX:
+            warnings.warn(
+                "The approximate argument has been deprecated and will be removed in a future version. "
+                "Please use the approximate argument in the TreeExplainer constructor instead.",
+                DeprecationWarning,
+            )
         if feature_names is not None:
             self.data_feature_names = feature_names
         elif isinstance(data, pd.DataFrame):
@@ -196,7 +209,6 @@ class TreeExplainer(Explainer):
         self.model_output = model_output
         # self.model_output = self.model.model_output # this allows the TreeEnsemble to translate model outputs types by how it loads the model
 
-        self.approximate = approximate
         if feature_perturbation not in feature_perturbation_codes:
             raise InvalidFeaturePerturbationError("Invalid feature_perturbation option!")
 
@@ -254,7 +266,35 @@ class TreeExplainer(Explainer):
         """This computes the expected value conditioned on the given label value."""
         return self.model.predict(self.data, np.ones(self.data.shape[0]) * y).mean(0)
 
-    def __call__(self, X, y=None, interactions=False, check_additivity=True):
+    def __call__(
+        self,
+        X: Any,
+        y: np.ndarray | pd.Series | None = None,
+        interactions: bool = False,
+        check_additivity: bool = True,
+        approximate: bool = False,
+    ) -> Explanation:
+        """Calculate the SHAP values for the model applied to the data.
+
+        Parameters
+        ----------
+        X : Any
+            Can be a dataframe like object e.g. numpy.array, pandas.DataFrame or catboost.Pool (for catboost).
+            A matrix of samples (# samples x # features) on which to explain the model's output.
+
+        y : numpy.array
+            An array of label values for each sample. Used when explaining loss functions.
+
+        approximate : bool
+            Run fast, but only roughly approximate the Tree SHAP values. This runs a method
+            previously proposed by Saabas which only considers a single feature ordering. Take care
+            since this does not have the consistency guarantees of Shapley values and places too
+            much weight on lower splits in the tree.
+
+        Returns
+        -------
+            shap.Explanation object containing the given data and the SHAP values.
+        """
         start_time = time.time()
 
         if isinstance(X, pd.DataFrame):
@@ -263,13 +303,11 @@ class TreeExplainer(Explainer):
             feature_names = getattr(self, "data_feature_names", None)
 
         if not interactions:
-            v = self.shap_values(
-                X, y=y, from_call=True, check_additivity=check_additivity, approximate=self.approximate
-            )
+            v = self.shap_values(X, y=y, from_call=True, check_additivity=check_additivity, approximate=approximate)
             if isinstance(v, list):
                 v = np.stack(v, axis=-1)  # put outputs at the end
         else:
-            assert not self.approximate, "Approximate computation not yet supported for interaction effects!"
+            assert not approximate, "Approximate computation not yet supported for interaction effects!"
             v = self.shap_interaction_values(X)
 
         # the Explanation object expects an `expected_value` for each row
@@ -373,12 +411,21 @@ class TreeExplainer(Explainer):
 
         return X, y, X_missing, flat_output, tree_limit, check_additivity
 
-    def shap_values(self, X, y=None, tree_limit=None, approximate=False, check_additivity=True, from_call=False):
+    def shap_values(
+        self,
+        X: Any,
+        y: np.ndarray | pd.Series | None = None,
+        tree_limit: int | None = None,
+        approximate: bool = False,
+        check_additivity: bool = True,
+        from_call: bool = False,
+    ):
         """Estimate the SHAP values for a set of samples.
 
         Parameters
         ----------
-        X : numpy.array, pandas.DataFrame or catboost.Pool (for catboost)
+        X : Any
+            Can be a dataframe like object, e.g. numpy.array, pandas.DataFrame or catboost.Pool (for catboost).
             A matrix of samples (# samples x # features) on which to explain the model's output.
 
         y : numpy.array
@@ -389,8 +436,10 @@ class TreeExplainer(Explainer):
             is used (``None``). ``-1`` means no limit.
 
         approximate : bool
-            Deprecated, will be removed in a future version. Please use the approximate argument in the
-            TreeExplainer constructor instead.
+            Run fast, but only roughly approximate the Tree SHAP values. This runs a method
+            previously proposed by Saabas which only considers a single feature ordering. Take care
+            since this does not have the consistency guarantees of Shapley values and places too
+            much weight on lower splits in the tree.
 
         check_additivity : bool
             Run a validation check that the sum of the SHAP values equals the output of the model. This
@@ -416,12 +465,6 @@ class TreeExplainer(Explainer):
                 Return type for models with multiple outputs changed from list to np.ndarray.
 
         """
-        if approximate is not False:
-            warnings.warn(
-                "The approximate argument has been deprecated and will be removed in a future version. "
-                "Please use the approximate argument in the TreeExplainer constructor instead.",
-                DeprecationWarning,
-            )
         # see if we have a default tree_limit in place.
         if tree_limit is None:
             tree_limit = -1 if self.model.tree_limit is None else self.model.tree_limit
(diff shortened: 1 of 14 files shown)


shap/explainers/_tree.py
```diff
@@ -389,10 +389,8 @@ def shap_values(self, X, y=None, tree_limit=None, approximate=False, check_addit
             is used (``None``). ``-1`` means no limit.
 
         approximate : bool
-            Run fast, but only roughly approximate the Tree SHAP values. This runs a method
-            previously proposed by Saabas which only considers a single feature ordering. Take care
-            since this does not have the consistency guarantees of Shapley values and places too
-            much weight on lower splits in the tree.
+            Deprecated, will be removed in a future version. Please use the approximate argument in the
```
thatlittleboy: 1. Shall we specify the version here when we will remove the parameter? Probably v0.48 ? Then we can update the meta tracker https://github.com/shap/shap/issues/3507 accordingly.

1. Also, we'll lose the documentation for `approximate` if we do this (i notice there's no docstring for `approximate` on the `TreeExplainer` init. Can you add the `approximate` docstring there please?

1. I also suppose it does mean that anyone who uses the general `shap.Explainer` API won't have access to approximate anymore. Though that's probably another issue for another day.

shap/explainers/_tree.py
```diff
@@ -418,6 +416,12 @@ def shap_values(self, X, y=None, tree_limit=None, approximate=False, check_addit
                 Return type for models with multiple outputs changed from list to np.ndarray.
 
         """
+        if approximate is not False:
+            warnings.warn(
+                "The approximate argument has been deprecated and will be removed in a future version. "
+                "Please use the approximate argument in the TreeExplainer constructor instead.",
+                DeprecationWarning,
+            )
```
thatlittleboy: Same comments here regarding specifying the version number explicitly.

---

Also: a thought regarding the method of warning the users: I think this method misses out users who have code lying around looking like this:

```python
shap_values = explainer.shap_values(..., approximate=False)
```
I.e., they manually specify the default value of the deprecated argument. These users won't get a warning, yet their code will break when we eventually remove `approximate`..

Instead, I think it's better to:

1. create a singleton `DEPRECATED_PARAM = object()` somewhere above.
2. change the default value of `approximate` to `DEPRECATED_PARAM` in this PR
3. check if `approximate is not DEPRECATED_PARAM` in the function body. If so, it means the user has used the `approximate` parameter somewhere, and we should warn them that their code will break in a future version.

In short

image

tests/explainers/test_tree.py
```diff
@@ -1841,3 +1841,34 @@ def test_xgboost_dart_regression():
     explainer = shap.TreeExplainer(model)
     shap_values = explainer.shap_values(X)
     assert np.allclose(shap_values.sum(1) + explainer.expected_value, model.predict(X), atol=1e-4)
+
+
+def test_consistency_approximate():
+    """GH 3764. Test that the call interface and shap_values interface are consistent when called with `approximate=True`."""
```
thatlittleboy: nit: may I suggest, let's standardize the naming of these kind of variables to `GH #3764`

so it's easier to grep for this in the future.

tests/explainers/test_tree.py
```diff
@@ -1841,3 +1841,34 @@ def test_xgboost_dart_regression():
...
+    dtc = sklearn.tree.DecisionTreeRegressor(max_depth=2)
+    arr = np.array([[0, 0], [0, 1], [1, 0], [1, 1]])
+    target = np.array([0, 0, 0, 80])
+
+    dtc.fit(arr, target)
+    # these explanations can be found in the paper "Consistent Individualized Feature Attribution for Tree Ensembles",
+    # https://arxiv.org/abs/1802.03888
+    expected_approx_explanations = np.array([[0.0, -20.0], [-40.0, 20.0], [0.0, -20.0], [40.0, 20.0]])
+
+    exp = shap.explainers.TreeExplainer(dtc, approximate=True)
```
thatlittleboy: small suggestion, shall we just parametrize the `approximate` input and expected output? Seems like a lot of repeated code

thatlittleboy (changes requested):
LGTM in general, though there are a few items need addressing / discussing 😃

shap/explainers/_tree.py
CloseChoice: about 3. how about adding `approximate` to the `__call__` and `shap_values` function so that it can be changed on different calls and is not fixed on an instance of the explainer? This way the general API is not limited to non-approximate executions.

thatlittleboy: Yeah, that seems reasonable, you may assess whether that would have any unintended consequences (I'll do so as well, on the second pass of the code review~)

CloseChoice: implemented. Ready for the next round.

shap/explainers/_tree.py
```diff
@@ -155,7 +158,17 @@ def __init__(
             Currently the "probability" and "log_loss" options are only
             supported when ``feature_perturbation="interventional"``.
 
+        approximate : bool
+            Deprecated, will be removed in a version v.0.48.0. Please use the approximate argument in the
+            ``shap_values`` or ```__call__`` methods instead.
+
         """
+        if approximate is not DEPRECATED_APPROX:
```
thatlittleboy:
```suggestion
        approximate : bool
            Deprecated, will be removed in version v0.48.0. Please use the ``approximate`` argument in the
            :meth:`.shap_values` or ``__call__`` methods instead.

        """
        if approximate is not DEPRECATED_APPROX:
```

I don't think it's conventional to put a period after the `v`

shap/explainers/_tree.py
```diff
@@ -155,7 +158,17 @@ def __init__(
...
+            warnings.warn(
+                "The approximate argument has been deprecated and will be removed in a future version. "
+                "Please use the approximate argument in the TreeExplainer constructor instead.",
```
thatlittleboy: They should be using the approximate argument in the shap_values or __call__ methods, not the constructor.

Let's put the v0.48.0 version in the warning message as well.

shap/explainers/_tree.py
```diff
@@ -254,22 +266,49 @@ def __dynamic_expected_value(self, y):
...
+    ) -> Explanation:
+        """Calculate the SHAP values for the model applied to the data.
+
+        Parameters
+        ----------
+        X : Any
+            Can be a dataframe like object e.g. numpy.array, pandas.DataFrame or catboost.Pool (for catboost).
+            A matrix of samples (# samples x # features) on which to explain the model's output.
+
+        y : numpy.array
```
thatlittleboy:
```suggestion
        y : numpy.array, optional
```

shap/explainers/_tree.py
```diff
@@ -254,22 +266,49 @@ def __dynamic_expected_value(self, y):
...
+            since this does not have the consistency guarantees of Shapley values and places too
+            much weight on lower splits in the tree.
+
+        Returns
+        -------
+            shap.Explanation object containing the given data and the SHAP values.
+        """
         start_time = time.time()
 
+        feature_names: Any | list[Any] | None
```
thatlittleboy: is this type hint necessary? If it's already an Any.

thatlittleboy (changes requested):

shap/explainers/_tree.py
CloseChoice: changed

shap/explainers/_tree.py
CloseChoice: done

shap/explainers/_tree.py
```diff
@@ -155,7 +158,17 @@ def __init__(
             Currently the "probability" and "log_loss" options are only
             supported when ``feature_perturbation="interventional"``.
 
+        approximate : bool
+            Deprecated, will be removed in version v0.48.0. Please use the ``approximate`` argument in the
```
connortann: May I suggest a slightly longer deprecation period - how about giving 2 versions' notice (`v0.49.0`), which I think is somewhat standard?

Also, may I suggest adding two entries to our tracker #3507:
- Deprecating the param in 0.47
- Removing the param in 0.49 (?)

Feel free to edit the description in that issue directly

shap/explainers/_tree.py
```diff
@@ -254,22 +266,49 @@ def __dynamic_expected_value(self, y):
...
-    def __call__(self, X, y=None, interactions=False, check_additivity=True):
+    def __call__(  # type: ignore  # noqa: F821
+        self,
+        X: Any,
+        y: np.ndarray | pd.Series | None = None,
+        interactions: bool = False,
+        check_additivity: bool = True,
+        approximate: bool = False,
+    ) -> Explanation:
+        """Calculate the SHAP values for the model applied to the data.
```
connortann, line 277: A quick note of appreciation: it's marvellous that you are adding type hints and useful docstrings to the key methods like this. It's a huge improvement!

Two small nitpicks:
1. `interactions` and `check_additivity` are missing from the docstring. In future, I think we should aim to get ruff rule [D417 Undocumented param](https://docs.astral.sh/ruff/rules/undocumented-param/) passing.
2. What is the `noqa: F821` for? Can this be fixed?

shap/explainers/_tree.py
```diff
@@ -254,22 +266,49 @@ def __dynamic_expected_value(self, y):
...
         if not interactions:
-            v = self.shap_values(
-                X, y=y, from_call=True, check_additivity=check_additivity, approximate=self.approximate
-            )
+            v = self.shap_values(X, y=y, from_call=True, check_additivity=check_additivity, approximate=approximate)
             if isinstance(v, list):
                 v = np.stack(v, axis=-1)  # put outputs at the end
         else:
-            assert not self.approximate, "Approximate computation not yet supported for interaction effects!"
+            assert not approximate, "Approximate computation not yet supported for interaction effects!"
```
connortann:
```suggestion
            if approximate:
                raise NotImplementedError("Approximate computation not yet supported for interaction effects!")
```

tests/explainers/test_tree.py
```diff
@@ -1841,3 +1841,33 @@ def test_xgboost_dart_regression():
...
+    arr = np.array([[0, 0], [0, 1], [1, 0], [1, 1]])
+    target = np.array([0, 0, 0, 80])
+
+    dtc.fit(arr, target)
+
+    exp = shap.explainers.TreeExplainer(dtc)
+    explanations_call_approx = exp(arr, approximate=approximate)
+    explanations_shap_values_approx = exp.shap_values(arr, approximate=approximate)
+    assert (explanations_call_approx.values == explanations_shap_values_approx).all()
+    assert (explanations_call_approx.values == expected_result).all()
```
connortann: Would `np.testing.assert_allclose` or `np.testing.assert_equal` be more appropriate here?

Those are generally better for use in tests:
- Error messages are more informative
- Floating point tolerances are handled
- `nan` values are treated as equal

tests/explainers/test_tree.py
```diff
@@ -1841,3 +1841,33 @@ def test_xgboost_dart_regression():
...
+    ],
+)
+def test_consistency_approximate(expected_result, approximate):
+    """GH #3764.
+    Test that the call interface and shap_values interface are consistent when called with `approximate=True`."""
+
+    import numpy as np
+
+    import shap
+
```
connortann: Are these imports needed?

```suggestion

```

connortann:
I'll edit the title to something more descriptive, to help with generating release notes.

thatlittleboy (approved):
All good on my end, nothing more to add on-top of what @connortann  has also mentioned in his review.

sled-dog-puppy-air-swimming

shap/explainers/_tree.py
CloseChoice: done

shap/explainers/_tree.py
CloseChoice: removed the noqa, that was a copy paste thing

tests/explainers/test_tree.py
CloseChoice: replaced with `np.testing.assert_allclose`

CloseChoice:
@connortann implemented your suggestions. Note that this is not breaking, just if we remove the parameter in an upcoming version it will be

connortann (approved):

connortann:
@CloseChoice looking back at this, I noticed the `approximate` argument to `TreeExplainer.__init()` now has no effect. Usually I would understand "Deprecated" to mean "it still works but will not work in future", but in this case it looks like this parameter is already ignored. 

Do we need to:
- Update the warning to clarify the passed parameter has no effect, or
- Update the code to still use the parameter for now?

Apologies for not picking up on this when I reviewed the PR!

PR #3837, opened 2024-09-01
refactor: Cleanup explanation ops

thatlittleboy:
## Overview

A general clean-up of the Explanation file, see details below.

1. Probably the most significant proposed change here is to group up the methods/properties of the `Explanation` class into several logical sections. I propose the following: 1) **Slicer passthrough** (~ all attributes of the underlying slicer object), 2) **Data model** (basically most of the dunder methods; excluding the binary operations which are parked under Operations), 3) **Operations** (generally speaking, within the scope of OpChain's), and 4) **Utilities** (for everything else, helpful methods/functions).
    - This is done, for example, in the pandas DataFrame definition, see here: [LINK to pandas source](https://github.com/pandas-dev/pandas/blob/078e11fbcd00b71790f7ea8d87f04041247c89fa/pandas/core/frame.py#L3863). I find the definition of the Explanation class to be a bit unwieldy, so some logical grouping should help here.

1. Use kwargs to instantiate the `Explanation` objects within this file, instead of using positional arguments. For example, in the `Explanation.__getitem__()` definition. This is, in general, good practice and helps to prevent future bugs.
    - A note here: I noticed a potential **bug** which I highlighted with a FIXME (resolving it is outside the scope of this PR).
    - I'll be grateful if someone can help to look into this.

1. Removed the `Op` and `Percentile` classes. I don't think these are used anywhere, nor are they part of the public API.

1. Change `issubclass` -> `isinstance` calls.

## Checklist


shap/_explanation.py
```diff
@@ -803,56 +815,42 @@ def is_1d(val):
...
-class Op:
-    pass
-
-
-class Percentile(Op):
-    def __init__(self, percentile):
-        self.percentile = percentile
-
-    def add_repr(self, s, verbose=False):
-        return "percentile(" + s + ", " + str(self.percentile) + ")"
```
thatlittleboy, line 815: Note: these 2 classes are removed in this PR, I don't think this is used anywhere.

shap/_explanation.py
```diff
@@ -378,19 +358,19 @@ def __getitem__(self, item) -> Explanation:
...
-                            self.upper_bounds,
-                            self.error_std,
-                            self.main_effects,
-                            self.hierarchical_values,
-                            self.clustering,
+                            base_values=np.array(new_base_values),
+                            data=np.array(new_data),
+                            display_data=self.display_data,
+                            instance_names=self.instance_names,
+                            feature_names=np.array(new_data),  # FIXME: this is probably a bug
```
thatlittleboy, line 365: Here, this `feature_names=np.array(new_data)` line looks weird, hence flagged out as a potential bug.

connortann: Good spot. A great example of why the new code using kwargs is much more readable than the previous style!

I agree it looks like a bug 🤔

shap/_explanation.py
connortann: Perhaps this class was intended as a better way of representing limits in plots. In `plots._utils.parse_axis_limit()`, we currently accept a string like `"percentile(20)"`. I presume the intention was to replace that with a python object.

In any case, I agree it's fine to remove this class for now as it's unused.

connortann (approved):
LGTM!

In an ideal world it would be nice to have some extra unit tests to cover all the changed lines - up to you. I appreciate many of the changes are just rearranging methods with no functional changes.

PR #3838, opened 2024-09-01
fix: OpChain repr for kwargs-only operations

thatlittleboy:
## Overview

Previously, if an `OpChain` Op had only kwargs, then the repr would be wrong, as in `Explanation.mean(axis=0)` would get rendered as `shap.Explanation.mean(, axis=0)`

image

This PR fixes the above bug, adds tests to prevent regression. And also some general refactor to be more pythonic. 

## Checklist

### DIFF
--- a/shap/utils/_general.py
+++ b/shap/utils/_general.py
@@ -319,12 +319,12 @@ class OpChain:
 
     def __repr__(self):
         out = self._root_name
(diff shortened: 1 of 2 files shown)


connortann (approved):
Looks good, a nice improvement

PR #3818, opened 2024-08-19
MAINT: Add numba lower version bound

connortann:
Fixes build, closes #3817

Discussion of the root cause of the issue: https://github.com/numba/numba/issues/9708#issuecomment-2296824946

Issue #3817, opened 2024-08-19
CI: Build failure with `numpy==2.1.0`, `llvmlite==0.34.0`


connortann:
Today's release of numpy `2.1.0` seems to have broken some of the tests, as it appears to downgrade the version of llvmlite installed to one that does not build.

Before (working):

```
llvmlite==0.43.0
numba==0.60.0
numpy==2.0.1
```

After (not working):

```
llvmlite==0.34.0
numba==0.51.2
numpy==2.1.0
```

Example build error:

```
RuntimeError: llvm-config failed executing, please point LLVM_CONFIG to the path for llvm-config
```

<details>

```
Resolved 51 packages in 6.21s
error: Failed to prepare distributions
  Caused by: Failed to fetch wheel: llvmlite==0.34.0
  Caused by: Build backend failed to build wheel through `build_wheel()` with exit status: 1
--- stdout:
running bdist_wheel
/home/runner/.cache/uv/builds-v0/.tmpCfCZXK/bin/python /home/runner/.cache/uv/built-wheels-v3/pypi/llvmlite/0.34.0/pNNRQw99Zzoo5oeqd06tk/llvmlite-0.34.0.tar.gz/ffi/build.py
LLVM version...
--- stderr:
Traceback (most recent call last):
  File "/home/runner/.cache/uv/built-wheels-v3/pypi/llvmlite/0.34.0/pNNRQw99Zzoo5oeqd06tk/llvmlite-0.34.0.tar.gz/ffi/build.py", line 105, in main_posix
    out = subprocess.check_output([llvm_config, '--version'])
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.4/x64/lib/python3.12/subprocess.py", line 466, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.4/x64/lib/python3.12/subprocess.py", line 548, in run
    with Popen(*popenargs, **kwargs) as process:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.4/x64/lib/python3.12/subprocess.py", line 1026, in __init__
[...]

PR #3843, opened 2024-09-03
Fix failing CI on python 3.12 due to outdated numba & llvmlite

thatlittleboy:


thatlittleboy:
Local copy is python 3.12.1, installing from latest master from github (with the failing CI)

image

---

image

thatlittleboy:
Numpy and numba had releases earlier (the CI run from yesterday ran successfully), so it can't be due to these two.


image

Setuptools has had a new release yesterday.

connortann:
This looks like an issue with numba - I believe I encountered it earlier. I'll dig up the relevant PR / issue....

connortann:
@thatlittleboy I think this could be another occurrence of a bug we encountered a few weeks ago: https://github.com/numba/numba/issues/9708#issuecomment-2296824946

## What's the cause?

Modern `numba` releases include an upper bound on numpy. However, older versions of `numba` do not include this upper bound, which can lead a package manager (`uv` in our case) to resolve a much older version of numba, in order to allow a recent version of numpy.

This is erroneous, as the older version of numba is not actually compatible with the latest numpy. That older numba uses a corresponding older version of `llvmlite` which fails to build on python 3.12.

This is not the package manager's fault but rather a fault with incomplete & inconsistent version bounds. For a 10,000 word essay on the problems that upper versions bounds cause in python, this [blog post](https://iscinumpy.dev/post/bound-version-constraints/) makes for interesting reading.

We encountered a similar error previously on #3817, which I fixed in #3818 by increasing the numba lower bound to `0.52`.

## Proposed fix

I think the cleanest fix is to bump the lower bound for numba to `numba>=0.54`, which seems to be the first release which includes the numpy upper bound.  `0.54` was released on 19 August 2021, which is well over the two-year window we should support according to [SPEC 0](https://scientific-python.org/specs/spec-0000/).

thatlittleboy:
hey @connortann thanks for the writeup, feel free to make the relevant fixes, i'll probably only have time to come back to this later during the week; abit busy with work now.

connortann:
OK, I'll push to this branch..

connortann (approved):
That pin has fixed it. Back over to you @thatlittleboy to approve & merge if you're happy :)

Issue #3844, opened 2024-09-03
ENH: Saliency Maps


eaguaida:
### Problem Description

I've realized that SHAP doesn't have an implementation for Saliency Maps. Even though image_plots lets us compute the Shapley values for an image, it doesn't provide an attention map.

By creating this implementation, SHAP can be compared against other methods like Grad-CAM or RISE, which SHAP outperforms anyway. At the same time, this saliency map can be benchmarked using Causal Metrics as described in https://arxiv.org/pdf/2311.17876.

### Feature request checklist

- [X] I have checked the issue tracker for duplicate issues.
- [X] I'd be interested in making a PR to implement this feature

connortann:
Thanks for creating the issue and the PR! It's great to have new contributors on board.

Please could you provide a bit more detail on the differences between `image_plot` and the proposed new `saliency_plot`? Are they both fundamentally a grayscale image with `ax.imshow(shap_values)` superposed on top?

It would help if you were able to provide an example of both plots for the same `Explanation` object

eaguaida:
Hey Connor! You are welcome, the pleasure is mine :)

While they are both grayscale with values superposed on top, my method extracts the values from the heatmap and uses bilinear interpolation to produce an attention map

I was doing some experimentation and comparisons between RISE[[1]](https://arxiv.org/pdf/1806.07421) and Grad-CAM[[2]](https://arxiv.org/pdf/1610.02391) as they both display a Saliency Map, and thought that this could be a nice and complementary feature for SHAP to have

For example, in Christopher Molnar book about Interpretable ML[[3]](https://christophm.github.io/interpretable-ml-book/pixel-attribution.html), SHAP is not mentioned in the Pixel Attribution (Saliency Maps) section

A few papers [[5]](https://arxiv.org/pdf/1806.07421), [[6]](https://www.researchgate.net/publication/358261598_Metrics_for_saliency_map_evaluation_of_deep_learning_explanation_methods) discuss about creating techniques to compare these Saliency Maps, and I've seen Saliency Maps mainly being used in Medical Settings [[5]](https://pubs.rsna.org/doi/10.1148/ryai.2021200267),[[6]](https://www.nature.com/articles/s42256-022-00536-x),[[7]](https://www.researchgate.net/figure/Representative-original-images-and-saliency-maps-of-WA-F-and-AM-F-The-regions-of_fig5_351914648)

[...]

PR #3845, opened 2024-09-03
ENH: Adding Saliency Map plot to _image.py

eaguaida:
Closes ##3844 

Added a new feature for plotting saliency maps using the shap values

Use:
```python
shap.saliency_plot(shap_values)
```
SHAP_plots

## Checklist

### DIFF
--- a/shap/plots/_image.py
+++ b/shap/plots/_image.py
@@ -7,6 +7,8 @@ from typing import TYPE_CHECKING, Literal, cast
 
 import matplotlib.pyplot as pl
 import numpy as np
+import cv2
+import keras
 
 try:
     from IPython.display import HTML, display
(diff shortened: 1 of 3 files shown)


connortann:
Thanks for the PR! We had a minor CI issue which was fixed by #3843, so I've merged the latest `master` into your branch to enable you to run the tests..

connortann:
I'd like to understand the proposed enhancement it a bit more detail, so let's discuss that first over at #3844. Once I have a clearer understanding of the purpose new plot, I'd be happy to give a detailed code review :)

That said, a quick head's up on two things that we'd need to address:
1. We would need at least one passing unit test for the plot.
   - Have a look in `tests/plots/test_image.py` as an example: we could create a new test `test_saliency_single()`, as a counterpart to the existing test for the image plot `test_image_single()`.
   - Nb. once the test has been written, the command to generate the baseline image would be:
   `pytest test/plots/test_image.py -k test_saliency_single --mpl-generate-path=tests/plots/baseline`
3. The shap package does not have `cv2` or `keras` as mandatory dependencies, so we cannot assume they are installed. Ideally we would make plots without these packages

shap/plots/_image.py
```diff
@@ -5,6 +5,8 @@
 import string
 from typing import TYPE_CHECKING, Literal, cast
 
+import cv2
```
CloseChoice, line 8: could we please move the imports of `keras` and `cv2` in the `saliency_map` function? I would like to avoid them being necessary for an import of `shap.plots`.

shap/plots/_image.py
```diff
@@ -196,6 +198,112 @@ def image(
...
+    shap_values: Explanation | np.ndarray | list[np.ndarray],
+    pixel_values: np.ndarray | None = None,
+    labels: list[str] | np.ndarray | None = None,
+    cmap: str | Colormap | None = "jet",
+):
+    """Plots Saliency Map using SHAP values.
+
+    Parameters
+    ----------
+    shap_values : [numpy.array]
```
CloseChoice, line 211:
```suggestion
    shap_values : shap.Explanation or numpy.ndarray or list[numpy.ndarray]
       List of arrays, a single array or an shap.Explanation object, containing SHAP Values. 
       If given arrays, they have the shape (# samples x width x height x channels), and the
       length of the list is equal to the number of model outputs that are being
       explained. If just one array is supplied, the model is expected to have a scalar output.
```

shap/plots/_image.py
```diff
@@ -196,6 +198,112 @@ def image(
...
+        List of arrays of SHAP values. Each array has the shape
+        (# samples x width x height x channels), and the
+        length of the list is equal to the number of model outputs that are being
+        explained.
+
+    pixel_values : numpy.array
```
CloseChoice, line 217:
```suggestion
    pixel_values : numpy.ndarray, optional
```

shap/plots/_image.py
```diff
@@ -196,6 +198,112 @@ def image(
...
+        Matrix of pixel values (# samples x width x height x channels) for each image.
+        It should be the same
+        shape as each array in the ``shap_values`` list of arrays.
+
+    labels : list or np.ndarray
+        List or ``np.ndarray`` (# samples x top_k classes) of names for each of the
+        model outputs that are being explained.
+
+    cmap: str or matplotlib.colors.Colormap
```
CloseChoice, line 226:
```suggestion
    cmap: str or matplotlib.colors.Colormap, optional
```

shap/plots/_image.py
```diff
@@ -196,6 +198,112 @@ def image(
...
+
+    for sample in range(rows):
+        # Plot original image
+        axes[sample, 0].imshow(pixel_values[sample], cmap="gray")
+        axes[sample, 0].axis("off")
+        axes[sample, 0].set_title("Original", fontsize=10)
+
+        # Generate and plot saliency maps
+        for i in range(len(shap_values)):
+            extracted_values = shap_exp.values
```
CloseChoice, line 269: If one hands over an array as shap_values, `shap_exp` is not defined.

Is it possible to add a test for this?

CloseChoice (changes requested):

CloseChoice:
@eaguaida : is it possible for you to provide an example I can run? Just tried to run it on the `notebooks/image_examples/image_classification/Explain MobilenetV2 using the Partition explainer (PyTorch).ipynb` notebook but failed to make it work.

shap/plots/_image.py
eaguaida: @connortann @CloseChoice Can you confirm if you would like me to create the plots without using ```keras``` and ```cv2```? Also, since ```torchvision``` is a dependency we could simply just use it, right?

eaguaida:
@CloseChoice Sorry I got extremely busy with work these weeks, thanks for the review! I improve the code next week, I also want to create a few examples and some extra documentation for use cases

CloseChoice:
> @CloseChoice Sorry I got extremely busy with work these weeks, thanks for the review! I improve the code next week, I also want to create a few examples and some extra documentation for use cases

Thanks for coming back to this. Please don't feel rushed, we are all doing this besides work so take all the time you need. Will be there when you need another review, feel free to ping me then.

Am still very much looking forward to have this in shap, but as said before, I just mean this in a welcoming sense to ask for help and raise questions when needed, I don't mean to make pressure.

PR #3846, opened 2024-09-04
Error in documentation of shap.datasets.communitiesandcrime

TommyGiak:
The target data (y) returned is the "violent crimes per 100k population", up to now is written "non-violent" which is wrong.


TommyGiak:
Alternatively, if you prefer to have as target the non-violent crimes you should select the last column in the dataset to you must correct line 242 of the shap/dataset.py file as:
```python
y = np.array(raw_data.iloc[valid_inds, -1], dtype=float)
```

connortann (approved):
Thanks - good spot!

PR #3850, opened 2024-09-07
refactor: Optimize explanation ops

thatlittleboy:
## Overview

A follow-up from the previous cleanup PR #3837 . Here we focus on a few more optimizations related to the Explanation class:

- Use `shap.utils.hclust_ordering` utility function in `Explanation.hclust` op, instead of duplicating the implementation logic.
- Remove extraneous copying of the `op_history` object during `Explanation._apply_binary_operator` because it's already done in `__copy__`
- Use numpy RandomState instead of setting seeds in the `sample` method
- A few type hints and docstring changes.

## Checklist

### DIFF
--- a/shap/_explanation.py
+++ b/shap/_explanation.py
@@ -455,8 +455,8 @@ class Explanation(metaclass=MetaExplanation):
 
     def _apply_binary_operator(self, other, binary_op, op_name):
         new_exp = self.__copy__()
-        new_exp.op_history = copy.copy(self.op_history)
         new_exp.op_history.append(OpHistoryItem(name=op_name, args=(other,), prev_shape=self.shape))
+
         if isinstance(other, Explanation):
             new_exp.values = binary_op(new_exp.values, other.values)
             if new_exp.data is not None:
--- a/shap/_explanation.py
+++ b/shap/_explanation.py
@@ -13,6 +13,7 @@ import scipy.spatial
 import sklearn
 from slicer import Alias, Obj, Slicer
 
+from .utils._clustering import hclust_ordering
 from .utils._exceptions import DimensionError
 from .utils._general import OpChain
 
@@ -641,11 +642,7 @@ class Explanation(metaclass=MetaExplanation):
         if axis == 1:
             values = values.T
 
-        # compute a hierarchical clustering and return the optimal leaf ordering
-        D = scipy.spatial.distance.pdist(values, metric)
-        cluster_matrix = scipy.cluster.hierarchy.complete(D)
-        inds = scipy.cluster.hierarchy.leaves_list(scipy.cluster.hierarchy.optimal_leaf_ordering(cluster_matrix, D))
-        return inds
+        return hclust_ordering(X=values, metric=metric)
 
     # =================== Utilities ===================
 
--- a/shap/_explanation.py
+++ b/shap/_explanation.py
@@ -597,7 +597,7 @@ class Explanation(metaclass=MetaExplanation):
         )
         return new_self
 
-    def sample(self, max_samples, replace=False, random_state=0) -> Explanation:
+    def sample(self, max_samples: int, replace: bool = False, random_state: int = 0) -> Explanation:
         """Randomly samples the instances (rows) of the Explanation object.
 
         Parameters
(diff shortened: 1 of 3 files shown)


shap/_explanation.py
```diff
@@ -613,11 +609,10 @@ def sample(self, max_samples, replace=False, random_state=0) -> Explanation:
             Random seed to use for sampling, defaults to 0.
 
         """
-        prev_seed = np.random.seed(random_state)
+        rng = np.random.RandomState(random_state)
```
connortann, line 612: The numpy docs suggest `numpy.random.default_rng()` as the new-style way to generate random numbers. Shall we use that instead?

Nb. the numpy docs and [NEP19](https://numpy.org/neps/nep-0019-rng-policy.html#nep19) say the legacy `RandomState` is appropriate for generating data for unit tests, but I don't think this applies here.

shap/_explanation.py
```diff
@@ -665,8 +656,8 @@ def hstack(self, other: Explanation) -> Explanation:
 
         """
         assert self.shape[0] == other.shape[0], "Can't hstack explanations with different numbers of rows!"
-        assert (
-            np.max(np.abs(self.base_values - other.base_values)) < 1e-6
+        assert np.allclose(
+            self.base_values, other.base_values, atol=1e-6
         ), "Can't hstack explanations with different base values!"
```
connortann: Would an Exception be better here?

```suggestion
        if not np.allclose(self.base_values, other.base_values, atol=1e-6):
            raise ValueError("Can't hstack explanations with different base values!")
```

shap/_explanation.py
```diff
@@ -302,11 +303,11 @@ def clustering(self, new_clustering):
 
     def __repr__(self):
         """Display some basic printable info, but not everything."""
-        out = ".values =\n" + self.values.__repr__()
+        out = f".values =\n{self.values!r}"
```
connortann, line 306: Can we add a unit test for this method?

connortann (approved):
A few small nitpicks as commented above, but otherwise looks good to me. The suggestions are up to you, so approving in advance :)

shap/_explanation.py
thatlittleboy: There's one very minor caveat that I don't see highlighted in the docs all that much, which is that `np.random.default_rng()` / `np.random.Generator` uses a different BitGenerator and SeedSequence to `RandomState`.

```python
[ins] In [1]: import numpy as np

[ins] In [2]: rng1 = np.random.default_rng(12345); print(rng1, "=>", rng1.normal())
Generator(PCG64) => -1.4238250364546312

[ins] In [3]: rng2 = np.random.RandomState(12345); print(rng2, "=>", rng2.normal())
RandomState(MT19937) => -0.20470765948471295

[ins] In [4]: rng3 = np.random.Generator(np.random.MT19937(12345)); print(rng3, "=>", rng3.normal())
Generator(MT19937) => -0.05103780855384499

[nav] In [6]: # original
         ...: prev_seed=np.random.seed(12345); print(f"original {np.random.randn() =}"); np.random.seed(prev_seed);
original np.random.randn() =-0.20470765948471295
```

As you see, only using the legacy RandomState (`rng2`) generates the exact same output as the original approach prior to this PR.
That was my main concern anyway, I recall we had a similar problem / issue previously regarding changing random generation in `shap.datasets` previously as well.

Some other refs: [NetworkX](https://networkx.org/documentation/stable/developer/nxeps/nxep-0004.html#backward-compatibility) specifically calls out this backwards incompatibility, and advocates for a major release before  this behaviour is changed as the default.

connortann: NEP 19 (linked above) has a detailed discussion about this, which I think boils down to the fact that the reproducibility guarantee between versions has become less relevant:

> The standard practice now for bit-for-bit reproducible research is to pin all of the versions of code of your software stack, possibly down to the OS itself. The landscape for accomplishing this is much easier today than it was in 2008. We now have pip. We now have virtual machines. Those who need to reproduce simulations exactly now can (and ought to) do so by using the exact same version of numpy. We do not need to maintain stream-compatibility across numpy versions to help them.

Their recommendation is to use `default_rng` in new code, but in any case I'd be happy either way.

shap/_explanation.py
thatlittleboy: sure, added

shap/_explanation.py
thatlittleboy: Totally agree, let's use `default_rng` for new code. I'll leave it as is then (with `RandomState`), I don't see that much of a downside to keeping status quo.

Thanks for raising this regardless.

PR #3851, opened 2024-09-08
docs: use intersphinx for some external links 

thatlittleboy:
## Overview

- Uses `intersphinx` sphinx extension for providing some of our external links, e.g. to scipy or to scikit-learn. This way we also ensure that the links are not outdated (else sphinx would complain during build), and the links are not hardcoded.
- fix some minor documentation build errors (such as improper indentation)


shap/plots/_beeswarm.py
```diff
@@ -62,23 +62,25 @@ def beeswarm(
...
 
     plot_size : "auto" (default), float, (float, float), or None
         What size to make the plot. By default, the size is auto-scaled based on the
         number of features that are being displayed. Passing a single float will cause
         each row to be that many inches high. Passing a pair of floats will scale the
         plot by that number of inches. If ``None`` is passed, then the size of the
-        current figure will be left unchanged. If ax is not ``None``, then passing
-        plot_size will raise a Value Error.
+        current figure will be left unchanged. If ``ax`` is not ``None``, then passing
+        ``plot_size`` will raise a :external+stdlib:exc:`ValueError`.
```
connortann: Nit: do we really need to point to the docs for simple builtin objects like these?

A slight drawback is that the raw source is a bit less readable, and I don't think it adds much value in this particular case

connortann (approved):
I'm okay with these changes, and I appreciate your ongoing efforts to improve the docs.

I am a little hesitant about the switch from full URLs to the intersphinx style. The full URLs are readable and useful when browsing the source code, whereas the intersphinx links are quite opaque in the source code but are better in the rendered docs. Overall I'm not sure if the improved docs are worth the more opaque source.

That said, I don't want to hold up this PR on something which is quite subjective so happy to approve.

shap/plots/_beeswarm.py
thatlittleboy: Removed

connortann:
> I'm okay with these changes, and I appreciate your ongoing efforts to improve the docs.
> 
> I am a little hesitant about the switch from full URLs to the intersphinx style. The full URLs are readable and useful when browsing the source code, whereas the intersphinx links are quite opaque in the source code but are better in the rendered docs. Overall I'm not sure if the improved docs are worth the more opaque source.
> 
> That said, I don't want to hold up this PR on something which is quite subjective so happy to approve.

I've thought a bit more about this, and I think for a large public package like this the rendered docs are probably read a great deal more than the source code of the docs. So, the tradeoff makes sense and I wholeheartedly support this PR :)

PR #3852, opened 2024-09-08
docs: document important APIs in Explanation and Explainers

thatlittleboy:
## Overview

Our `Explanation` and various `Explainer` classes make use of the magic methods in the python data model to expose various functionalities to the user. E.g. `Explanation.__getitem__` performs special indexing operations to return a new Explanation object, `Explainer.__call__` returns an Explanation object containing the shap values).

I propose to make this more explicit in our API documentation, not just demonstrate them in our tutorial notebooks.

Precedence: `numpy` documents down their [`ndarray.__getitem__`](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.__getitem__.html) API, Many others, including dask and pandas, do the same.

### Discussion

I also split up the Explainers section in the API documentation into subsections, otherwise the summary table gets a bit unwieldy. What it looks like now:

image

Please help to double check if the categorization is appropriate. Thanks~


connortann:
I think documenting the `__call__` methods is a great idea!

## Organising the explainers

I like your categorisation, that's a nice improvement.

## Where to put the docs

I think there might be a more appropriate way to organise these pages. The current proposal deviates a bit from the usual layout: each class being on one page, and all methods and attributes listed under that page.

Taking `TreeExplainer` as an example, at present the class is documented on [this page](https://shap.readthedocs.io/en/latest/generated/shap.TreeExplainer.html), and all the methods are tabulated there. By default the only dunder method included is `__init__`:

image

So, would it be better to **include the `__call__` method here along with all the other methods**, rather than pulling it out separately from the class docs?

I believe this can be done with the `:special-members:` audodoc setting, as described [here](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-option-automodule-special-members). Specifically, I think we would need to add this option in `conf.py`:

```python
autodoc_default_options = {..., "special-members": ["__init__", "__call__"]}
```

connortann (changes requested):
As above, suggest reorganising slightly

connortann:
I think the Sphinx autodoc settings are not quite working as hoped. Taking as an example: <https://shap--3852.org.readthedocs.build/en/3852/generated/shap.Explainer.html> 

1. The `__init__` method is duplicated: it appears above the table of methods, and also below

2. The `__call__` method doesn't appear in the table:

image

3. As a nitpick, it might be nicer to capitalise the `explainers` header:

image

connortann (changes requested):
A few suggestions as above

thatlittleboy:
@connortann so shall we go back to my original proposal, or do you have any other ideas?

I think the important thing is to get the documentation of `__call__` and `__getitem__` out there first, I intend to overhaul the documentation formatting in the future~

connortann:
I would prefer to see if the issues above can be sorted, so we have the methods organised in the usual manner (i.e. in the table of methods on each class's page).

There are lots of config options to the various Sphinx extensions (autosummary, autodoc, autoclass) but I don't know the right settings off the top of my head. This might help:

https://stackoverflow.com/a/62161540

connortann:
I came across the issue of duplicated method tables on a different PR, over at #3886 . I believe it's due to numpydoc and autosummary not playing nicely together. This should be fixed on that PR  together with #3885 , which will drop numpydoc.

To get the special methods included in the table of class methods, I came across this post which is quite helpful:
https://stackoverflow.com/a/62161540

PR #3853, opened 2024-09-08
docs: Use uv in readthedocs, RtD AddOn migration

thatlittleboy:
## Overview

Use `uv` in a custom RtD build (at least, until it's available as a [standard option](https://docs.readthedocs.io/en/stable/config-file/v2.html#packages) in RtD). This is noted in the RtD documentation as a standard recipe: [LINK](https://docs.readthedocs.io/en/stable/config-file/v2.html#packages).

**Impact**: build times are now reduced by 50s, or 26% faster than before.

Before:  
image

After:  
image

### Discussion

**Side effect**: ReadtheDocs addon will be enabled by default as a result of using build commands (though, it's going to be the default for ALL sphinx projects from 7th October 2024 anyway, even without build command being used). Read more about the migration plan [here](https://about.readthedocs.com/blog/2024/07/addons-by-default).

There is a note there about having to update `conf.py` in the above migration plan (Section [LINK](https://about.readthedocs.com/blog/2024/07/addons-by-default/#how-does-it-affect-my-projects)), I'm not sure whether it applies to us or not, but without it, the doc build still seems to run fine and produce the correct output (?)  
So I've left out that change for now.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -5,11 +5,13 @@ build:
   os: ubuntu-22.04
   tools:
     python: "3.11"
-
-# Optionally declare the Python requirements required to build your docs
-python:
-  install:
-    - method: pip
-      path: .
-      extra_requirements:
-        - docs
+  commands:
+    # cf. https://docs.readthedocs.io/en/stable/build-customization.html#install-dependencies-with-uv
+    - asdf plugin add uv
+    - asdf install uv latest
+    - asdf global uv latest
+    - uv venv
+    - uv pip install .[docs]
+    - >-
+      .venv/bin/python -m sphinx -T -b html -d docs/_build/doctrees
+      -D language=en docs $READTHEDOCS_OUTPUT/html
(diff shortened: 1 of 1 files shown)


connortann (approved):
Looks great!

connortann:
Going further: should we consider using a lockfile to pin all dependencies and transitive dependencies, as recommended by the ReadTheDocs guide to [reproducible builds](https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html)?

With uv, one can create such a lockfile with:

```bash
uv pip compile pyproject.toml --extra docs --python-version 3.11 -o docs/requirements-docs.txt
```

We would also want to remove exact dependency pins in `pyproject.toml`, as these would now represent constraints rather than exact pins.

thatlittleboy:
> Going further: should we consider using a lockfile to pin all dependencies and transitive dependencies, as recommended by the ReadTheDocs guide to [reproducible builds](https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html)?

Sure, I think that's a good idea! (We can do that in a separate PR)

PR #3854, opened 2024-09-09
CI: Test against lowest supported versions of key dependencies

connortann:
## Overview

Test against the lowest supported versions of several pacakages, not just the oldest supported numpy.

This provides us a mechanism to check that we retain compatibility with a reasonable range of supported versions, and will ensure that users do not install an incompatible set of packages.

### Philosophy of how to pin versions

At present, we have no bounds on most dependencies. This is not ideal, as we know for a fact that the library is incompatible with particular older versions of some libraries. For example, we require `matplotlib>=3.6`, as we use `plt.Figure.set_layout_engine` which was introduced in `v3.6`.  So, we should add lower bounds **where we know for a fact that this will cause an error**. 

However, I don't think we necessarily need to add lower bounds just for the sake of it, unless we know that the older versions will not work or incur a higher maintenance burden. So, overall I propose we only bump the lower supported versions if and when we find issues, whilst keeping support for at least 2 years as mandated by [SPEC 0](https://scientific-python.org/specs/spec-0000/) 

### How should we determine which versions to test?

We only really care about the packages with which we integrate tightly, like `numpy`, `pandas`, `pytorch` and so on. We don't want to have to care about defining the lowest supported versions of dev packages like `pytest`, or packages we barely touch like `ipython`, `requests` or `cloudpickle`.

I initially tried the `--resolution-strategy=lowest` feature of `uv`, which installs the lowest supported versions of _all_ dependencies. This is a bit burdensome, as we'd need to add lower bounds for all pacakges: https://github.com/astral-sh/uv/issues/5492

So, the proposal is to only test compatibility for **packages that have a lower bound constraint** declared in pyproject.toml. For example, the package requires `matplotlib>=3.6`, so we test against `matplotlib==3.6`. 

### Which packages should we test this way?

The new job determines versions to test, as shown in the `Extract lower bounds` step of the `run tests (oldest supported)` job on GH actions. To start with we have:

```
numpy==1.20
scipy==1.6.0
scikit-learn==1.0
pandas==1.2.0
tqdm==4.27
numba==0.54
matplotlib==3.6
Pillow==9.0.1
```

In future, as a follow-up PR, I think we could aim to also include older versions of xgboost, pytorch and tensorflow.

### Why we do *not* include any upper bound constraints

Whilst tempting at first, it seems to be a very bad idea to add upper bound constraints. This 10,000 word post gives a great  explanation as to why: https://iscinumpy.dev/post/bound-version-constraints/


.github/workflows/run_tests.yml
```diff
@@ -80,23 +80,33 @@ jobs:
...
       - name: Set up Python 3.9
         uses: actions/setup-python@v5
         with:
           python-version: 3.9
-      - name: Install dependencies
+      - name: Extract lower bounds
         run: |
           python -m pip install --upgrade uv
-          uv pip install --system numpy==1.24.0 '.[test-core,plots]'
+          uv run scripts/extract_oldest_supported.py > oldest_supported.txt
```
connortann, line 105: Side comment: I was quite impressed with how `uv run` works. It will read the dependencies from the script metadata (in this case `packaging` and `tomli`), and execute the script in a temporary virtual encironment with those dependencies installed!

It is super fast, and quite neat. In particular, it's nice that this is completely isolated from the python environment that is used to run the actual tests.

CloseChoice (approved):
Sorry for the late review, adjusted a couple things so that it fits to the latest master changes. LGTM

PR #3858, opened 2024-09-12
MAINT: Change default KernelExplainer l1reg to `numfeatures(10)`

connortann:
## Overview

Supports #3507: changes the default value of `l1reg` away from `auto` to `numfeatures(10)`.

### DIFF
--- a/shap/explainers/_kernel.py
+++ b/shap/explainers/_kernel.py
@@ -155,7 +155,7 @@ class KernelExplainer(Explainer):
                 tensor_as_np_array = sess.run(symbolic_tensor)
         return tensor_as_np_array
 
-    def __call__(self, X, l1_reg="auto", silent=False):
+    def __call__(self, X, l1_reg="num_features(10)", silent=False):
         start_time = time.time()
 
         if isinstance(X, pd.DataFrame):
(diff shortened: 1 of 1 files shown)


shap/explainers/_kernel.py
```diff
@@ -640,12 +640,7 @@ def solve(self, fraction_evaluated, dim):
         nonzero_inds = np.arange(self.M)
         log.debug(f"{fraction_evaluated = }")
         if self.l1_reg == "auto":
-            warnings.warn(
-                "l1_reg='auto' is deprecated and in a future version the behavior will change from a "
-                "conditional use of AIC to simply a fixed number of top features. "
-                "Pass l1_reg='num_features(10)' to opt-in to the new default behaviour.",
-                DeprecationWarning,
-            )
+            warnings.warn("l1_reg='auto' is deprecated and will be removed in a future version.", DeprecationWarning)
```
thatlittleboy, line 643: shall we maybe be a bit more specific on which version the removal will happen (I presume it's 0.48)?

thatlittleboy (approved):
A minor suggestion but otherwise lgtm

cat-standing-thankyou

shap/explainers/_kernel.py
connortann: Whilst we should probably aim for 0.48, it's hard to know precisely which version the removal will actually happen as we might desire to make releases for any number of unforeseen reasons. The majority of Deprecation warnings I've seen from other libraries don't include this particular promise, so I'd probably prefer to avoid mentioning a specific future version just to retain flexibility.

PR #3859, opened 2024-09-12
MAINT: Fix various test suite warnings, update deprecated uses of feature_perturbation

connortann:
Updates the LinearExplainer test suite to move away from the deprecated "feature_perturbation" syntax, and updates the docs a little to make the deprecation warning a little more prominent.

### DIFF
--- a/shap/explainers/_linear.py
+++ b/shap/explainers/_linear.py
@@ -56,25 +56,29 @@ class LinearExplainer(Explainer):
         Number of samples to use when estimating the transformation matrix used
         to account for feature correlations.
 
-    feature_perturbation : "interventional" (default) or "correlation_dependent"
+    feature_perturbation : None (default), "interventional" or "correlation_dependent"
+
+        DEPRECATED: this option is now deprecated in favor of using the appropriate
+        tabular masker and will be removed in a future release.
+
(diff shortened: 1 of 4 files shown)


shap/explainers/_linear.py
```diff
@@ -56,25 +56,29 @@ class LinearExplainer(Explainer):
         Number of samples to use when estimating the transformation matrix used
         to account for feature correlations.
 
-    feature_perturbation : "interventional" (default) or "correlation_dependent"
+    feature_perturbation : None (default), "interventional" or "correlation_dependent"
+
+        DEPRECATED: this option is now deprecated in favor of using the appropriate
```
CloseChoice, line 61: I would expect a deprecated warning if something other than the default is used. Or is there a reason why do not do that?

CloseChoice (changes requested):

shap/explainers/_linear.py
connortann: I think this is already handled on line 94, no?

CloseChoice: ah, sorry. I thought that's also part of this PR.

CloseChoice (approved):
LGTM

Issue #3663, opened 2024-05-17
ENH: Python 3.13 Support for upstream Packages


CloseChoice:
## Problem Description

Python 3.13b1 is out, which means its time to start looking into Python 3.13 support!

Release wheels for python 3.13.

## Which dependencies support 3.13 ?

Core dependencies:
 - [x] numba: https://github.com/numba/numba/issues/9413
 - [x] pandas: https://github.com/pandas-dev/pandas/issues/58734
 - [x] numpy

Test dependencies:
 - [x] xgboost: https://github.com/dmlc/xgboost/pull/11028
 - [x] pytorch: should be supported, as of https://github.com/pytorch/pytorch/issues/130249
 - [x] lightgbm: https://github.com/microsoft/LightGBM/pull/6668
 - [x] catboost: https://github.com/catboost/catboost/issues/2748
 - [x] tensorflow: not supported (as of 2025-03-23), https://github.com/tensorflow/tensorflow/issues/78774
 - [x] ngboost: not supported (as of 2025-05-06), https://github.com/stanfordmlgroup/ngboost/blob/master/pyproject.toml#L16

jameslamb:
> lightgbm: https://github.com/microsoft/LightGBM/pull/6668

To clarify.... `conda install -c conda-forge lightgbm` and `pip install lightgbm` will both totally work in Python 3.13 environments right now. That PR you've linked is just about adding more test coverage and a trove classifier.

For `shap`'s purposes, `lightgbm` already supports Python 3.13. The same is probably true for `xgboost` as well.

CloseChoice:
@jameslamb you're right. I think we can start supporting python 3.13 as soon as #3861 passes.

tr01-crypto:
any news?

btat-qc:
@CloseChoice it looks like #3861 is passing according to your comment https://github.com/shap/shap/pull/3861#issuecomment-2854811847 Just wanted to ask what the status of 3.13 support is?

CloseChoice:
@btat-qc yes, tests are passing and we support python 3.13 on master. I renamed this ticket to track the progress on Python 3.13 support on the remaining dependencies.

We can schedule a release quite soon. Would just like to fix a regression bug.

CloseChoice:
All packages support python 3.13!

PR #3861, opened 2024-09-13
WIP: Test against 3.13

connortann:
A draft PR to test against release candidates of 3.13.

Progress towards #3663

Note: this branch is probably going to be subject to frequent rebases.


.github/workflows/run_tests.yml
```diff
@@ -26,22 +26,22 @@ jobs:
       matrix:
         os: ["ubuntu-latest"]
         # The total number of matrix jobs should match codecov.yml `after_n_builds`.
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
```
CloseChoice, line 30: I would suggest getting rid of 3.9

CloseChoice (dismissed):
LGTM

Edit: seems like we need to remove the torchvision dependency as well for python 3.13

connortann:
@CloseChoice I saw you approved, but this PR isn't ready yet as the tests fail, and the `transformers` dep is commented out entirely (for testing purposes). I'll see if I can get it to a passable state and re-request a review then.

However, I think we should aim to make a release first, and then save this for the next release (which might also drop 3.9 as you suggest)

CloseChoice:
> @CloseChoice I saw you approved, but this PR isn't ready yet as the tests fail, and the `transformers` dep is commented out entirely (for testing purposes). I'll see if I can get it to a passable state and re-request a review then.
> 
> However, I think we should aim to make a release first, and then save this for the next release (which might also drop 3.9 as you suggest)

I agree, let's get the release out first

CloseChoice:
Note, we first should get #4021 passing before merging this

CloseChoice:
tests are passing, we are good to merge!

PR #3862, opened 2024-09-15
Ci fix feedstock build

CloseChoice:
## Overview

This fixed some issues in our feedstock repository (see https://github.com/conda-forge/shap-feedstock/pull/85) when I tested this locally. Could not get a local test to succeed since the feedstock repo is a nightmare to test due to all the dependencies. But this increases the chance that we can get the conda builds to pass

## Checklist

### DIFF
--- a/shap/benchmark/models.py
+++ b/shap/benchmark/models.py
@@ -64,14 +64,13 @@ def corrgroups60__gbm():
 
 def corrgroups60__ffnn():
     """4-Layer Neural Network"""
-    from tensorflow.keras.layers import Dense
-    from tensorflow.keras.models import Sequential
+    import tensorflow as tf
 
-    model = Sequential()
-    model.add(Dense(32, activation="relu", input_dim=60))
-    model.add(Dense(20, activation="relu"))
-    model.add(Dense(20, activation="relu"))
-    model.add(Dense(1))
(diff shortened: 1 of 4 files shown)


connortann:
Thanks for looking into this @CloseChoice !

I must admit this fix is quite mysterious to me, as I don't see what is different compared to the existing code. The fix seems to be changing:

```
# Before
from foo.bar import Baz
x = Baz()

# After
import foo
x = foo.bar.Baz()
```

Aren't these two styles identical in their behaviour?

Could you shed any light on what is the root cause of the issue, and how we could verify that shap uses tensorflow correctly?
 
In any case, in the short term I'm happy to approve as I don't see any problem with the new code and I don't want to hold up the conda fix.

connortann (approved):
Approve as it looks fine, although as above I'd love to understand the root cause of the issue

CloseChoice:
> Approve as it looks fine, although as above I'd love to understand the root cause of the issue

I can't explain why this happened, just saw the import errors locally and got rid of them with this branch. But then some follow-up dependency issues appeared so I couldn't get a complete test run through. Will still merge this, since it won't hurt. 

I thought it might be some conda issue with tensorflow dependencies?

PR #3880, opened 2024-10-04
pin uv version to download it from cache

CloseChoice:
## Overview

I just saw [this action](https://github.com/astral-sh/setup-uv) and this makes it more clear that we use uv in our pipelines, and we do not need to redownload uv on every run. Might also be worth to look into the caching that they provide, see [here](https://github.com/astral-sh/setup-uv?tab=readme-ov-file#install-the-latest-version-default)


.github/workflows/run_tests.yml
connortann, line 1: We install `uv` in two other jobs in this workflow, so for consistency I think these install steps should also be updated:

1. `test_oldest_supported_numpy`, around line 98
2. `run_mypy`, around line 116

If we do decide to pin the version number (see other comment), then given that we will be using the uv version number in multiple places, it would probably be help to define it as a top-level variable, something like:

```yml
env:
  UV_VERSION: "0.4.x"
```

connortann (changes requested):
I think changing the installation to use the action is a great idea. I don't think the speedup will be significant, but a big advantage is having `uv` installed in an environment that is isolated from the main project python env.

However, I think we should only pin the **major** number for uv, e.g. `0.4.x`. Or even just use `latest`.

Rationale: If we pinned the exact version, we would have to manually keep bumping it with new releases, especially as uv is under rapid development. I'm not sure this hassle is worth the marginal speedup, which I believe would be less than a second and so pretty inconsequential. I also do think the determinism of having the exact uv version is particularly important at the moment, given we don't pin our python dependencies anyway.

CloseChoice:
close for now

PR #3885, opened 2024-10-09
DOCS: Pin docs dependencies for reproducibility

connortann:
Pins the documentation dependencies with a lockfile, for reproducible builds. Follows on from: https://github.com/shap/shap/pull/3853#issuecomment-2339113730

Supporting changes:
- Update the maintainer guide with instructions on how to update the pins
- Remove some redundant pins in pyproject.toml : these are now constraints, not pins
- Update our versions of Sphinx and sphinx-rtd-theme, and check built docs still look good

PR #3886, opened 2024-10-09
DOCS: switch from numpydoc to napoleon, improve display of type hints

### LINKED ISSUE #3883: DOCS: move rendered type hints to parameter description, away from title
At present our type hints look rather ugly in the Sphinx docs, obfuscating the call signatures. It would be useful to clean up this display so the function titles are a more readable.

It looks like this could be improved by following in the footsteps of optuna and XGBoost:

- https://github.com/optuna/optuna/pull/2147
- https://github.com/dmlc/xgboost/pull/6519#issuecomment-751942569

Screenshot of current docs:

image

connortann:
## Overview

numpydoc and napoleon are tools to parse numpy-style docstrings into Sphinx documentation. This PR switches us to napoleon.

Note to reviewer: I'd appreciate a review of the new rendered API docs, to check I haven't missed anything.

## Discussion

This gives us several improvements:

- Napolean comes with Sphinx as a builtin extension, so switching to this will allow us to drop a dependency.
- The output is a fair bit prettier:
  - Allows us to simplify the titles of functions by moving the type hints to the parameter descriptions
  - Types of inputs will be automatically hyperlinked to the relevant class
- Works more nicely with autosummary templates. With numpydoc, we sometimes get duplicate tables of classes (as experienced in #3852 )

Closes #3883

## Comparison

Current master:

<details>

image

</details>

This PR:

<details>

image

</details>

## TODO

- [x] Fix duplicate table of methods, e.g. <https://shap--3886.org.readthedocs.build/en/3886/generated/shap.Explainer.html>
- [x] Drop numpydoc from deps, once #3885 is merged


CloseChoice:
Am I missing something here? On [prod](https://shap.readthedocs.io/en/latest/generated/shap.Explanation.html#shap.Explanation) I see:
image

while on [this branch's docs](https://shap--3886.org.readthedocs.build/en/3886/generated/shap.Explanation.html#shap.Explanation) I just find:
image

I guess this branch is just a few commits behind. But could we update it so that it gets easier to compare?

connortann:
> Am I missing something here?

I don't think so. The change in the rendering style is to be expected, as napolean and numpydoc render in slightly different styles.

> I guess this branch is just a few commits behind. But could we update it so that it gets easier to compare?

It's up to date with master, as of about an hour ago :)

CloseChoice:
> > Am I missing something here?
> 
> I don't think so. The change in the rendering style is to be expected, as napolean and numpydoc render in slightly different styles.

does that mean that napoleon does not take into account type hints?

EDIT: maybe we should add a linter rule that checks that type hints + docs align then.

connortann:
> does that mean that napoleon does not take into account type hints?

I think it's flexible and can behave as we wish. The PR includes a config change that is relevant, moving the typehints from signatures to the Parameters section:

```python
autodoc_typehints = "description"
```

More info on that setting [here](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_typehints). I was inspired by: https://github.com/optuna/optuna/pull/2147

I think this setting makes the display of longer functions (e.g. scatter plot) more tidy.

connortann:
We have a few options for type hints. At present we typically have the hints in two places, both in the function signature and also in the docstring:

```python
def foo(x: Bar):
    """Calculate foo

    Parameters
    ----------
    x : Baz
        The input
    """
```

The type hints can be put in the signature (`Bar`) and/or in the docstring (`Baz`), and in general they might be different. I believe the napolean docs will use `Bar` if provided, or else it will use `Baz` if provided.

EDIT: actually the order of priority is the other way around, `Baz` takes precedence over `Bar`.

CloseChoice:
> We have a few options for type hints. At present we typically have the hints in two places, both in the function signature and also in the docstring:
> 
> ```python
> def foo(x: Bar):
>     """Calculate foo
> 
>     Parameters
>     ----------
>     x : Baz
>         The input
>     """
> ```
> 
> The type hints can be put in the signature (`Bar`) and/or in the docstring (`Baz`), and in general they might be different. I believe the napolean docs will use `Bar` if provided, or else it will use `Baz` if provided.

Thanks for the clarification. I'll have a deeper look into this PR later today.

connortann:
A quick correction to my earlier comment: sphinx-napoleon will use the type in the docstring (`Baz`) if provided, or else it will use the type hint in the function signature (`Bar`).

As a quick demo, I experimented with `shap.plots.scatter()`. Here are some screenshots to show how it seems to work:

## Option A: using the docstring types

At present the types are both in the signature and the docstring, like this:

```python
def scatter(
    x_jitter: float | Literal["auto"] = "auto", title: str | None = None, alpha: float = 1.0, ...
):
    """
    x_jitter : 'auto' or float
        Adds random jitter...

    title: str, optional
        Plot title.

    alpha : float
        The transparency of the data points...
    """
```

The rendered docs look like:

image


## Option B: Using the function signature types

Removing the types from the docstring, leaving them just in the signature:

```python
def scatter(
    x_jitter: float | Literal["auto"] = "auto", title: str | None = None, alpha: float = 1.0, ...
):
    """
    x_jitter
        Adds random jitter...

    title
        Plot title.

    alpha
        The transparency of the data points...
    """
```

The rendered docs like this:

image

This seems to work nicely with type aliases. For example, the signature type hint `xmin: AxisLimitSpec = None` gets rendered as:

image

CloseChoice:
Here are a couple things:
 - when multiple `Added/Changed in version x` appear, I think the old layout is nicer, since the new layout is IMO too dense, e.g. [here](https://shap--3886.org.readthedocs.build/en/3886/generated/shap.GPUTreeExplainer.html)
grafik
(left new - right old)
 - I consider the positioning of the `Changed` note weird in the new version:
 
grafik
Would expect that this appears after the return type. I think how the return type is handled pretty inconsistent altogether. Any other parameter has it's type listed right next to the parameter name, but the return type comes just after the description of what is returned. I consider the old layout better here.
- what I consider better in the new version is the automatic reference to another class, e.g. on masker [here](https://shap--3886.org.readthedocs.build/en/3886/generated/shap.LinearExplainer.html)
grafik
(left new - right old)

So I think this is it. Altogether I noted a couple of negative things about the new layout but I think the last point outweighs the negatives slightly. How do you see these points?

connortann:
Thanks for the analysis! You spotted some interesting things that I'd missed.

Yes, it's a shame the `versionadded` and `versionchanged` directives don't look as tidy. I'll see if I can improve things. As it stands I think this is probably a slight marginal improvement.

I think the main benefit of switching to napoleon is having everything inside the Sphinx builtin ecosystem, and hopefully enabling us to document the `__call__` methods. I was planning on doing that in a follow-up PR, but I'll have a go here to check it works. That will hopefully make this PR a suitably significant improvement to be worthwhile.

CloseChoice:
> Thanks for the analysis! You spotted some interesting things that I'd missed.
> 
> Yes, it's a shame the `versionadded` and `versionchanged` directives don't look as tidy. I'll see if I can improve things. As it stands I think this is probably a slight marginal improvement.
> 
> I think the main benefit of switching to napoleon is having everything inside the Sphinx builtin ecosystem, and hopefully enabling us to document the `__call__` methods. I was planning on doing that in a follow-up PR, but I'll have a go here to check it works. That will hopefully make this PR a suitably significant improvement to be worthwhile.

What's the status here? Can we merge or would you like to check out if the follow up works?

PR #3890, opened 2024-10-15
Add tests for layernorm and add op_handler functions

### LINKED ISSUE #3438: Add Support for More nn.Modules Layers.
### Problem Description

I have tried SHAP on my Vision Transformer Model, and it said that it does not recognize the following layers:
- nn.GELU
- nn.Identity
-  nn.LayerNorm

Could you please add support for those layers, and maybe for many others. Thank you kindly for your consideration!

### Alternative Solutions

_No response_

### Additional Context

_No response_

### Feature request checklist

JulioHC00:
## Overview

Partially closes #3438  by adding support for LayerNorm and testing it

Adds LayerNorm to the op_handler dictionary with nonlinear_1d as well as Identity as passthrough. I've kept the structure of the tests as similar as possible to previous ones for consistency.

## Checklist

- [ x] Unit tests added (if fixing a bug or adding a new feature)

### DIFF
--- a/shap/explainers/_deep/deep_pytorch.py
+++ b/shap/explainers/_deep/deep_pytorch.py
@@ -382,6 +382,7 @@ op_handler["Dropout3d"] = passthrough
 op_handler["Dropout2d"] = passthrough
 op_handler["Dropout"] = passthrough
 op_handler["AlphaDropout"] = passthrough
+op_handler["Identity"] = passthrough
 
 op_handler["Conv1d"] = linear_1d
 op_handler["Conv2d"] = linear_1d
@@ -400,6 +401,7 @@ op_handler["BatchNorm1d"] = linear_1d
 op_handler["BatchNorm2d"] = linear_1d
 op_handler["BatchNorm3d"] = linear_1d
 
+op_handler["LayerNorm"] = nonlinear_1d
 op_handler["LeakyReLU"] = nonlinear_1d
 op_handler["ReLU"] = nonlinear_1d
 op_handler["ELU"] = nonlinear_1d
--- a/tests/explainers/test_deep.py
+++ b/tests/explainers/test_deep.py
@@ -757,3 +757,123 @@ def test_pytorch_multiple_inputs(torch_device, disconnected, activation):
         np.testing.assert_allclose(sums + e.expected_value, outputs, atol=1e-3),
         "Sum of SHAP values does not match difference!",
     )
+
+
+@pytest.mark.parametrize("layernorm_first_layer", [True, False])
+@pytest.mark.parametrize("torch_device", TORCH_DEVICES)
+@pytest.mark.parametrize("baseline_equal_test", [True, False])
+def test_pytorch_layernorm(layernorm_first_layer, torch_device, baseline_equal_test):
+    """Test LayerNorm layer"""
+    torch = pytest.importorskip("torch")
+
+    from sklearn.datasets import fetch_california_housing
+    from torch import nn
+    from torch.nn import functional as F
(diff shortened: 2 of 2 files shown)


shap/explainers/_deep/deep_pytorch.py
```diff
@@ -400,6 +401,7 @@ def nonlinear_1d(module, grad_input, grad_output):
 op_handler["BatchNorm2d"] = linear_1d
 op_handler["BatchNorm3d"] = linear_1d
 
+op_handler["LayerNorm"] = nonlinear_1d
```
CloseChoice, line 404: Hey, thanks for the update. Unfortunately this is wrong. Sorry, I had a thought about this and the problem here is that multiple inputs (given one output) vary. So for instance, let's have a look at the [tensorflow implementation for multiplicative attribution](https://github.com/shap/shap/blob/master/shap/explainers/_deep/deep_tf.py#L782) and deduct it from the definition of shapley values:

$\phi_{i} = \sum_{S \subseteq N \setminus \{i\}} \frac{|S|!(|N|-|S|-1)!}{|N|!} \left( f(S \cup \{i\}) - f(S) \right)$

Now let's assume we have just $x, y$ (and therefore $b_{x}, b_{y}$ as baselines). 
We are looking for a multiplication (or any function that takes two arguments):
$f(x, y) = x \cdot y$
Then this formula becomes:
$\phi_{x} = \frac{1}{2} (f(x, y) - f(b_{x}, y)) + \frac{1}{2} (f(x, b_{y}) - f(b_{x}, b_{y}))$
We would need to do something similar for the layer norm BUT this is computationally expensive, so we might be able to use the approach that the tensorflow implementation takes for [`softmax`](https://github.com/shap/shap/blob/master/shap/explainers/_deep/deep_tf.py#L488) (yes, pytorch impl is wrong here too!). I don't fully understand this yet, but doesn't seem too complicated.

Sorry, I just figured this out too late. Really appreciate your effort. Let me know how you want to proceed with this.

CloseChoice (changes requested):

shap/explainers/_deep/deep_pytorch.py
JulioHC00: Ah! It was too easy to be true. So, if I understand this correctly, in the original DeepLIFT paper they have the rescale rule which actually only applies to a **single input**. That's what's implemented in the `nonlinear_1d` function? And that's why this isn't valid for `LayerNorm` which takes several inputs. So we need to use the full equation to derive the proper function. Is this right? If so I think I understand and can try to see how to translate the way it's done in tensorflow to pytorch.

P.S. Does this mean that the `softmax` implementation for pytorch also needs to be fixed?

CloseChoice: @JulioHC00 yes, that's right. And also we would need to fix the implementation for pytorch's `softmax` attribution. I believe that once we understand how the tensorflow softmax is implemented, we can apply this to pytorch LayerNorm and softmax as well.

JulioHC00: @CloseChoice I guess the only part I'm confused about is how does this fit with the op_handlers? Don't these handle the gradients in the backward propagation? How does this relate to implementing the shap values 
equation?

rjbruin: @CloseChoice When I integrate LayerNorm into a testcase and use the `op_handler['LayerNorm'] = softmax`, the testcase passes. Do we need to create a testcase that *does* show how Softmax/LayerNorm are broken? What would be in such a testcase?

PR #3900, opened 2024-10-28
Fix typo in beeswarm.ipynb

kamurani:
fow -> row


CloseChoice (approved):
Thanks for the PR @kamurani. LGTM

PR #3906, opened 2024-11-05
Refactored explainer to use python switch

max1e:
## Overview

Refactored the Explainer class to use a switch statement for algorithm selection.

Python match statements were introduced in python 3.10. Its use here improves readability by emphasising conditioning based on the selected algorithm, clearly separating the different algorithms in switch cases.

### DIFF
--- a/shap/explainers/_explainer.py
+++ b/shap/explainers/_explainer.py
@@ -171,125 +171,126 @@ class Explainer(Serializable):
         if self.__class__ is Explainer:
             # do automatic algorithm selection
             # from .. import explainers
-            if algorithm == "auto":
-                # use implementation-aware methods if possible
-                if explainers.LinearExplainer.supports_model_with_masker(model, self.masker):
-                    algorithm = "linear"
-                elif explainers.TreeExplainer.supports_model_with_masker(
-                    model, self.masker
-                ):  # TODO: check for Partition?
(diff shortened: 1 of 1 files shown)


shap/explainers/_explainer.py
```diff
@@ -171,125 +171,126 @@ def __init__(
...
-                    algorithm = "tree"
-                elif explainers.AdditiveExplainer.supports_model_with_masker(model, self.masker):
-                    algorithm = "additive"
-
-                # otherwise use a model agnostic method
-                elif callable(self.model):
-                    if issubclass(type(self.masker), maskers.Independent):
-                        if self.masker.shape[1] <= 10:
-                            algorithm = "exact"
+            match algorithm:
```
CloseChoice, line 174: Thanks for the PR and would support refactoring this, but I think we first need to remove support for Python 3.9 since this pattern matching is only available from 3.10 onwards.

max1e: Ah I see. Thank you for the feedback! Should I keep this PR open as draft for when Python 3.9 support is removed, or would you prefer I close it for now?

CloseChoice: No, let's keep this open.

CloseChoice:
Thanks for the PR, due to faded activity I'll close this one.

Issue #3914, opened 2024-11-26
ENH: Causal Shapley values


max1e:
### Problem Description

Traditional Shapley values are computed by integrating over the marginal distribution of a data set, introducing a naïve assumption that all features are independent. This can lead to misleading explanations, especially when features are highly correlated.

[Heskes et al. (2020)](https://proceedings.neurips.cc/paper/2020/file/32e54441e6382a7fbacbbbaf3c450059-Paper.pdf)  introduced causal Shapley values. They incorporate causal relationships in the computation of Shapley values by including a (partial) causal graph and performing intervention. This allows for explaining the total combined direct and indirect effects of a feature on the outcome.

### Additional Context

Causal Shapley values were already requested in a comment on issue #1912, but deserve their own issue as supporting them is a feature on its own.

### Feature request checklist

- [X] I have checked the issue tracker for duplicate issues.
- [X] I'd be interested in making a PR to implement this feature

PR #3915, opened 2024-11-28
ENH: Causal Shapley values for Masker based explainers

max1e:
## Overview

Based on issue #3914 

Added support for causal Shapley values to all explainers that support tabular maskers by creating a new masker, which samples from the interventional distribution as implemented in [Heskes et al. (2020)](https://proceedings.neurips.cc/paper/2020/file/32e54441e6382a7fbacbbbaf3c450059-Paper.pdf).

## Implementation Rationale

The key distinction between traditional Shapley values and causal Shapley values lies in the distributions they integrate over. Traditional Shapley values are computed using the marginal data distribution, while causal Shapley values are based on the interventional distribution. By designing a Masker that samples from the causal distribution, we enable causal Shapley values for a wide range of explainers (those utilising Maskers) without changing any of the existing codebase.

## Implementation details

We implement a causal masker, which samples from the interventional distribution. Explainers that support tabular maskers can use this masker to compute causal Shapley values. The implementation is inspired by [Heskes et al. (2020)](https://proceedings.neurips.cc/paper/2020/file/32e54441e6382a7fbacbbbaf3c450059-Paper.pdf) (Section 5) and the corresponding R implementation for [shapr](https://gitlab.science.ru.nl/gbucur/shapr/-/tree/master/R?ref_type=heads).

The causal masker takes in
- Ordering: a causal ordering in the form of a causal chain graph
- Confounding: a boolean array, marking causal ordering components that contain a confounding variable

In causal chain graphs, interventions can be simulated through conditioning, as described in [Heskes et al. (2020)](https://proceedings.neurips.cc/paper/2020/file/32e54441e6382a7fbacbbbaf3c450059-Paper.pdf) (Section 5, Theorem 1). Features are always conditioned on their upstream (parent) features. Furthermore, features within a causal ordering component, are also conditioned on one another when their component does not include a confounder. As when a confounding variable is present, an intervention on a specific feature breaks its dependency with other features in the same component.

## Limitations

Gaussian sampling is relatively quick and simple. However, when features have a different distribution than Gaussian, this approach can yield skewed results. This reduces the reliability of the computed Causal Shapley values. Open to other suggestions about how to develop a distribution-agnostic implementation.

The masker is currently only compatible with PermutationExplainer and ExactExplainer, since LinearExplainer, PartitionExplainer, and AdditiveExplainer all contain built-in constraints for their maskers.

## To be done
- [x] Finish checks for parameters
- [x] Add unit tests
- [x] Add example usage in a new notebook

## Out of scope
- Causal Shapley value support for explainers that do not utilise maskers, and PartitionExplainer, AdditiveExplainer, and LinearExplainer
- Causal Shapley value support for non-tabular data
- Separately retrieving the direct and indirect effects of features (see [Heskes et al. (2020)](https://proceedings.neurips.cc/paper/2020/file/32e54441e6382a7fbacbbbaf3c450059-Paper.pdf) section 3).
- Add compatibility for delta masks

## Checklist

- [x] Unit tests added
- [x] Usage example added


shap/maskers/_tabular.py
```diff
@@ -335,3 +335,182 @@ def __init__(self, data, method="linear"):
...
+            # Sample
+            new_samples = np.random.multivariate_normal(
+                mean=np.zeros(len(to_be_sampled)), cov=conditional_covariance, size=self.n_samples
+            )
+            new_samples += conditional_mean
+
+        return new_samples
+
+    @staticmethod
+    def _is_positive_definitive(matrix):
```
CloseChoice: I think it's faster to check whether `matrix - matrix.mean(0)` has `rank == matrix.shape[0]` as per:

https://stats.stackexchange.com/a/53105/206297

But haven't checked.

max1e: That could be, but since this method is only called in masker initialisation, performance should not really be an issue here. I find the eigenvalue approach clearer, since it is more closely tied to the definition of positive definiteness.

CloseChoice:
@max1e: ping me when you feel this is ready for a review or you want the tests to run.

max1e:
@CloseChoice, it should now be ready for review.

max1e:
Peculiar, tests do run locally on macOS (15.1.1) with Python 3.11.10. Maybe this commit will fix it.

shap/maskers/_tabular.py
```diff
@@ -335,3 +335,327 @@ def __init__(self, data, method="linear"):
...
+            groups containing confounding factors (i.e., where the value is `True`), features within the group are
+            not conditioned on each other, but only on the preceding groups.
+            Example: [True, False, False]
+
+        max_samples : int
+            The maximum number of samples to use from the passed background data. If data has more
+            than max_samples then shap.utils.sample is used to subsample the dataset. The number of
+            samples coming out of the masker (to be integrated over) matches the number of samples in
+            the background dataset. This means larger background dataset cause longer runtimes. Normally
+            about 1, 10, 100, or 1000 background samples are reasonable choices.
```
CloseChoice, line 381: Is there an argument why the range from 1-1000 background samples are reasonable? I understand that we have a trade-off between accuracy and performance but I think we generally advise not to use more than 1k background samples. But why would 1 be enough?

shap/maskers/_tabular.py
```diff
@@ -335,3 +335,327 @@ def __init__(self, data, method="linear"):
...
+            feature_names=getattr(self, "feature_names", None),
+        )
+
+        # TODO make Causal masker abstract and create GaussianCausalMasker that inherits from Causal masker
+        # Mean and covariance matrix are for the gaussian sampling approach and should be moved to a subclass in the future, making this class abstract
+        if isinstance(data, pd.DataFrame):
+            # Case 1. Pandas DataFrame
+            self.mean = data.mean().values
+            self.covariance_matrix = data.cov().values
+        else:
```
CloseChoice: could we please check explicitly that this is a numpy array and raise a NotImplementedError otherwise?

CloseChoice (changes requested):
Sorry, I added this review months ago but somehow forgot to send it.

shap/maskers/_tabular.py
max1e: I copied this docstring from the Tabular(Masker) superclass, so I don’t have original reasoning for the exact numbers. It seems odd to use 1 sample.

That said, using a single background sample seems to yield reasonable results in simple test cases. I assume this works when the model has a smooth decision boundary and the sample is representative of the dataset.

I haven’t tested it on more complex models or datasets.

bitbalancer:
@max1e Thanks for the work on this! Does the causal SHAP implementation support feature groups as described in the [shapr vignette](https://cran.r-project.org/web/packages/shapr/vignettes/asymmetric_causal.html#groups-of-features)?

max1e:
> @max1e Thanks for the work on this! Does the causal SHAP implementation support feature groups as described in the [shapr vignette](https://cran.r-project.org/web/packages/shapr/vignettes/asymmetric_causal.html#groups-of-features)?

The current implementation does not support this. However, it seems like a valuable feature and should be doable to incorporate. I currently don’t have the time to work on this, but it might be worth creating an issue to explore this for future work.

PR #3927, opened 2024-12-10
MAINT: Fix Ruff UP031, bump ruff

connortann:
## Overview

Fixes new Ruff rule [UP031](https://docs.astral.sh/ruff/rules/printf-string-formatting/) (printf-string-formatting).

I noticed our pre-commit auto-update #3919 was failing, as this new rule is enabled on the latest ruff linter. So, added these fixes manually to get precommit passing.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ exclude: '.*tree_shap_paper.*|.*user_studies.*'
 
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.7.2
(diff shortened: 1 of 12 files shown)


connortann:
Hi @fredheda , would you like to give this one a review? Feel free to leave comments if you have any questions on any of the changes.

I acknowledge that two of the "Checks" have failed. This is to be expected in this particular case, as rather unusually the `_benchmark` module is excluded from the test suite. We can talk about the rationale for that if you're interested.

Fredheda (approved):
Looks good to me.

PR #3937, opened 2024-12-29
GPUTreeExplainer fixes

### LINKED ISSUE #3936: BUG: GPUTreeExplainer issues
### Issue Description

The GPUTreeShap implementation has several issues:
1. SHAP values are incorrectly calculated when there are nan values present in the input. 
2. The expected value is not updated from phi after shap values have been calculated.

```python
import numpy as np
import shap
from xgboost import XGBRegressor

X, y = shap.datasets.diabetes()

model = XGBRegressor()
model.fit(X, y)

tree_explainer = shap.TreeExplainer(model)
print(tree_explainer.expected_value)
shap_values = tree_explainer.shap_values(X)
print(tree_explainer.expected_value)

gpu_explainer = shap.GPUTreeExplainer(model)
print(gpu_explainer.expected_value)
gpu_shap_values = gpu_explainer.shap_values(X)
print(gpu_explainer.expected_value)

assert shap_values.shape == gpu_shap_values.shape
assert np.allclose(shap_values, gpu_shap_values, 1e-3, 1e-3)

X_nans = X.copy()
np.random.seed(0)
X_nans.loc[X_nans.sample(frac=0.1).index, 'bmi'] = np.nan
shap_values_nans = tree_explainer.shap_values(X_nans)
gpu_shap_values_nans = gpu_explainer.shap_values(X_nans, check_additivity=False)

assert shap_values_nans.shape == gpu_shap_values_nans.shape
print(np.allclose(shap_values_nans, gpu_shap_values_nans, 1e-3, 1e-3))

rows_info = X_nans['bmi'].isnull()
rows_info_df = rows_info.to_frame(name="bmi_has_nan")
rows_info_df["has_diff"] = np.any(np.abs(shap_values_nans - gpu_shap_values_nans) > 1e-3, axis=1)
print(rows_info_df.head())
print(rows_info_df[rows_info_df['has_diff']]["bmi_has_nan"].all())
```

### Traceback

```shell
[152.13348]
152.13225
[152.13348]
[152.13348]
False
   bmi_has_nan  has_diff
0        False     False
1         True      True
2        False     False
3        False     False
4        False     False
True
```

### Expected Behavior

Please note that the code above doesn't pass when check_additivity isn't set to False in the last GPUTreeExplorer call. 
As evident from the output neither the SHAP values nor the expected value is the same between the CPU and the GPU version (and also between this GPU implementation and the one provided in the rapids library).

The expected result here would be to have the additivity checks pass and all values be the same (within some epsilon distance).

I have identified and fixed these issues and will open a PR shortly. However, I think that perhaps some new tests should be added that capture issues like these. The current ones can be extended to also include checks similar to what I have done.
Also I haven't tested the behaviour when `feature_perturbation=interventional`, which is related to #3655, which doesn't get entirely fixed -- the issue is present when the model_output='log_loss' option is specified, but values are the same otherwise (screenshow below). 
image

### Bug report checklist

### Installed Versions

python==3.11.11
gcc==11.4
xgboost==2.1.3
pandas==2.2.3
numpy==1.26.4
cuda==12.4
shap:[ from master branch ](https://github.com/shap/shap/tree/8f04ad2a379a5e46c535afa6d14a1aebaa528c78)

RostislavStoyanov:
## Overview

Closes #3936 

1. Correctly assign branches for missing values in ShapSplitCondition.
2. Update expected value similarly to the way its done when calling other solutions:
https://github.com/shap/shap/blob/8f04ad2a379a5e46c535afa6d14a1aebaa528c78/shap/explainers/_tree.py#L567-L571

## Checklist


RostislavStoyanov:
Problems mentioned in the PR get solved:
image

CloseChoice:
Thanks for the PR. Could you please add a test for this? Having the case from the issue would help a lot here. Also, have you investigated whether the issue only appears when one uses the gpu tree explainer or if it is present in the non-gpu code as well?

RostislavStoyanov:
Hi, thanks for the input. I've included a test that simulates the original issue.
You were right, I hadn't checked if expected value gets updated for the non-gpu tree, so I've adjusted the code accordingly.
This, however, leads to more issues that can be found in the failing tests:
 1. The first issue is additivity checks failure for XGBoostClassifier. This is due to the expected value being very different:
image
Not yet sure why it happens only to XGBoostClassifier and not to Regressor, probably something to do with model loading?
2. Two new tests fail in `tests/explainers/test_tree.py`:
- `test_HistGradientBoostingClassifier_proba`: Some shape assertion fails here should be easy to fix` TestExplainerSklearn::test_HistGradientBoostingClassifier_proba - assert (32561,) == (32561, 2).`
- `test_consistency_approximate `: another additivity check fail.

I will get more into these in the next days, might need some guidance on the XGBoost part. Not sure if this should be still standing as a draft PR considering you've already taken a look.

PR #3939, opened 2024-12-30
Add Interaction Order to Explanation Class

### LINKED ISSUE #3874: BUG: shap.image_plot bug when using DeepExplainer
### Issue Description

I encounter this bug when trying to visualize the plot create by shap.image_plot. In this plot all of the images in shap_values are very small and are unable to view

```python
import shap
import numpy as np
import tensorflow as tf
from tensorflow import keras
from keras.models import Sequential
import ssl

ssl._create_default_https_context = ssl._create_unverified_context
(x_train, y_train), (x_test, y_test) = keras.datasets.cifar10.load_data()
# reshape and normalize data
x_train = x_train.reshape(50000, 32, 32, 3).astype("float32") / 255
x_test = x_test.reshape(10000, 32, 32, 3).astype("float32") / 255
y_train = y_train.reshape(50000,)
y_test = y_test.reshape(10000,)
# define the model architecture
inputs = keras.Input(shape=(32, 32, 3))
x = keras.layers.Conv2D(32, (3, 3), activation='relu')(inputs)
x = keras.layers.MaxPooling2D((2, 2))(x)
x = keras.layers.Conv2D(128, (3, 3), activation='relu')(inputs)
x = keras.layers.MaxPooling2D((2, 2))(x)
x = keras.layers.Conv2D(64, (3, 3), activation='relu')(inputs)
x = keras.layers.MaxPooling2D((2, 2))(x)
x = keras.layers.Conv2D(32, (3, 3), activation='relu')(inputs)
x = keras.layers.MaxPooling2D((2, 2))(x)
x = keras.layers.Flatten()(x)
x = keras.layers.Dense(256, activation='relu')(x)
x = keras.layers.Dense(64, activation='relu')(x)
outputs = keras.layers.Dense(10, activation='softmax')(x)
# inputs and outputs
model = keras.Model(inputs=inputs, outputs=outputs, name="test_for_shap")
    # compile the model
model.compile(
loss=tf.keras.losses.SparseCategoricalCrossentropy(),
optimizer=keras.optimizers.Adam(),
metrics=[tf.keras.metrics.SparseCategoricalAccuracy()])
model.fit(x_train, y_train, validation_data=(x_test, y_test), epochs = 10)

# class label list
class_names = ['airplane', 'automobile', 'bird', 'cat', 'deer',
               'dog', 'frog', 'horse', 'ship', 'truck']
# example image for each class
images_dict = dict()
for i, l in enumerate(y_train):
  if len(images_dict)==10:
    break
  if l not in images_dict.keys():
    images_dict[l] = x_train[i].reshape((32, 32,3))
images_dict = dict(sorted(images_dict.items()))

# example image for each class for test set
x_test_dict = dict()
for i, l in enumerate(y_test):
  if len(x_test_dict)==10:
    break
  if l not in x_test_dict.keys():
    x_test_dict[l] = x_test[i]
# order by class
x_test_each_class = [x_test_dict[i] for i in sorted(x_test_dict)]
x_test_each_class = np.asarray(x_test_each_class)
# Compute predictions
predictions = model.predict(x_test_each_class)
predicted_class = np.argmax(predictions, axis=1)

# plot actual and predicted class
def plot_actual_predicted(images, pred_classes):
  fig, axes = plt.subplots(1, 11, figsize=(16, 15))
  axes = axes.flatten()

  # plot
  ax = axes[0]
  dummy_array = np.array([[[0, 0, 0, 0]]], dtype='uint8')
  ax.set_title("Base reference")
  ax.set_axis_off()
  ax.imshow(dummy_array, interpolation='nearest')
  # plot image
  for k,v in images.items():
    ax = axes[k+1]
    ax.imshow(v, cmap=plt.cm.binary)
    ax.set_title(f"True: %s \nPredict: %s" % (class_names[k], class_names[pred_classes[k]]))
    ax.set_axis_off()
  plt.tight_layout()
  plt.show()

# select backgroud for shap
background = x_train[np.random.choice(x_train.shape[0], 1000, replace=False)]
# DeepExplainer to explain predictions of the model
explainer = shap.DeepExplainer(model, background)
# compute shap values
shap_values = explainer.shap_values(x_test_each_class)

shap_values = [np.clip(val, 0, 1) for val in shap_values]

# plot SHAP values
plot_actual_predicted(images_dict, predicted_class)
print()
shap.image_plot(shap_values,  x_test_each_class, show=False)
#This is where the bug happened
```

### Traceback

_No response_

### Expected Behavior

The expected ouput:
image
Bug output
Screenshot (390)

### Bug report checklist

### Installed Versions

0.46.0

CloseChoice:
## Overview

Closes #3874  

- the issue here was, that neither `base_values` nor `output_names` was provided when creating `Explanation` instances from the DeepExplainer which resulted in an empty tuple as `Explanation.output_dims`, which then caused problems in the `image_plot`. We solve this here in two different fashions: 1. set the base_values correctly, 2. add interaction_order. Each of these solutions would be fine by itself, but I think, the interaction_order can help to track down if everything was done correctly and helps to understand the `Explanation`'s shap_values better.
- add `interaction_order` to the explanation class. This is either `0` for no interactions or `1` if interactions are present. I also added this to a couple of tests.

## Checklist


tests/plots/test_image.py
```diff
@@ -43,6 +43,61 @@ def test_image_multi(imagenet50_example):
...
+    # Reshape the labels
+    y_train_subset = y_train_subset.reshape(
+        subset_size,
+    )
+    y_test_subset = y_test_subset.reshape(
+        subset_size,
+    )
+
+    # define the model architecture
+    inputs = tf.keras.Input(shape=(32, 32, 3))
```
CloseChoice: we have a problem here. Retraining this won't reproduce the image. Any ideas how we can get this to work? I would like to reproduce the issue from the ticket

CloseChoice (commented):
I don't understand why the tests fail. Just from the files that differ it looks like my test has something to do with it but when looking at the diffs I cannot understand why. I also cannot reproduce the issue locally.

CloseChoice:
Okay, just found that exactly these tests fail in other pipelines as well. Will try to reproduce and create a fix

CloseChoice:
I think this was closed in the mean time

CloseChoice:
is it really closed, take a deeper look

CloseChoice:
Note that this PR solves the issue only for the `__call__` method, `.shap_values` is left unchanged

PR #3945, opened 2025-01-03
ENH: support rng for summary_plot

tylerjereddy:
* Some folks on my team are having issues writing tests downstream of `shap` because `summary_plot` mutates the NumPy legacy random machinery global state, which is hard to work around in a large testsuite. It would be great if a heavily-used project in the ML space like `shap` could start to adopt support for local random `Generator` objects per the community document at https://scientific-python.org/specs/spec-0007/. This PR is a lot cruder than that approach, but appears to allow preservation of the current global state behavior while allowing the option of using the modern non-global approach for downstream developers that need it.

* This patch adds a regression test that fails when not using the optional `rng` argument, and passes when it is used to scope to a local random state. If there is a preference for using the full SPEC 7 approach (and that may very well be the best idea), I may ask one of my team members to help out a bit, since they will benefit from it.

* The full testsuite passed locally via `pytest --import-mode=append`

Maybe I'll cc @mdhaber @tupui -- not expecting a review, but because they also work in stats/ML space and were involved in the SPEC.

## Checklist

### DIFF
--- a/shap/plots/_beeswarm.py
+++ b/shap/plots/_beeswarm.py
@@ -531,7 +531,7 @@ def summary_legacy(
     cmap=colors.red_blue,
     show_values_in_legend=False,
     use_log_scale=False,
-    rng=None,
+    seed=None,
 ):
     """Create a SHAP beeswarm plot, colored by feature values when they are provided.
 
@@ -766,10 +766,10 @@ def summary_legacy(
             shaps = shap_values[:, i]
             values = None if features is None else features[:, i]
             inds = np.arange(len(shaps))
-            if rng is None:
+            if seed is None:
                 np.random.shuffle(inds)
             else:
-                rng.shuffle(inds)
+                seed.shuffle(inds)
             if values is not None:
                 values = values[inds]
             shaps = shaps[inds]
@@ -786,10 +786,10 @@ def summary_legacy(
             # curr_bin = []
             nbins = 100
             quant = np.round(nbins * (shaps - np.min(shaps)) / (np.max(shaps) - np.min(shaps) + 1e-8))
-            if rng is None:
+            if seed is None:
                 tmp_x = np.random.randn(N)
             else:
-                tmp_x = rng.standard_normal(N)
+                tmp_x = seed.standard_normal(N)
             inds = np.argsort(quant + tmp_x * 1e-6)
             layer = 0
             last_bin = -1
@@ -874,10 +874,10 @@ def summary_legacy(
                 rng = shap_max - shap_min
                 xs = np.linspace(np.min(shaps) - rng * 0.2, np.max(shaps) + rng * 0.2, 100)
                 if np.std(shaps) < (global_high - global_low) / 100:
-                    if rng is None:
+                    if seed is None:
                         tmp_y = np.random.randn(len(shaps))
                     else:
-                        tmp_y = rng.standard_normal(len(shaps))
+                        tmp_y = seed.standard_normal(len(shaps))
                     ds = gaussian_kde(shaps + tmp_y * (global_high - global_low) / 100)(xs)
                 else:
                     ds = gaussian_kde(shaps)(xs)
@@ -1017,10 +1017,10 @@ def summary_legacy(
                         ys[i, :] = ys[i - 1, :]
                     continue
                 # save kde of them: note that we add a tiny bit of gaussian noise to avoid singular matrix errors
-                if rng is None:
+                if seed is None:
                     tmp_z = np.random.normal(loc=0, scale=0.001, size=shaps.shape[0])
                 else:
-                    tmp_z = rng.normal(loc=0, scale=0.001, size=shaps.shape[0])
+                    tmp_z = seed.normal(loc=0, scale=0.001, size=shaps.shape[0])
                 ys[i, :] = gaussian_kde(shaps + tmp_z)(x_points)
                 # scale it up so that the 'size' of each y represents the size of the bin. For continuous data this will
                 # do nothing, but when we've gone with the unqique option, this will matter - e.g. if 99% are male and 1%
--- a/shap/plots/_beeswarm.py
+++ b/shap/plots/_beeswarm.py
@@ -531,6 +531,7 @@ def summary_legacy(
     cmap=colors.red_blue,
     show_values_in_legend=False,
     use_log_scale=False,
+    rng=None,
 ):
     """Create a SHAP beeswarm plot, colored by feature values when they are provided.
 
@@ -765,7 +766,10 @@ def summary_legacy(
             shaps = shap_values[:, i]
             values = None if features is None else features[:, i]
             inds = np.arange(len(shaps))
-            np.random.shuffle(inds)
+            if rng is None:
+                np.random.shuffle(inds)
+            else:
+                rng.shuffle(inds)
             if values is not None:
                 values = values[inds]
             shaps = shaps[inds]
@@ -782,7 +786,11 @@ def summary_legacy(
             # curr_bin = []
             nbins = 100
             quant = np.round(nbins * (shaps - np.min(shaps)) / (np.max(shaps) - np.min(shaps) + 1e-8))
-            inds = np.argsort(quant + np.random.randn(N) * 1e-6)
+            if rng is None:
+                tmp_x = np.random.randn(N)
+            else:
+                tmp_x = rng.standard_normal(N)
+            inds = np.argsort(quant + tmp_x * 1e-6)
             layer = 0
             last_bin = -1
             ys = np.zeros(N)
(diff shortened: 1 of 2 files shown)


CloseChoice (commented):
Thanks a lot for the PR, this looks sensible to me. I just need to think whether we'll actually want to introduce the keyword that might become redundant once we implement a Generator version for the rng. On the other hand, this is just an optional parameter but please give me some time to read into this and come back to you.

Also about the tests: these fail in all pipelines and are not caused by your changes.

mdhaber:
Thanks for the ping. I suppose if I'm here, I'd suggest using one of the legacy names for the keyword unless it is going to follow the SPEC 7 plan. If this keyword is intended to be permanent, a different name might be helpful to signal different behavior from the ecosystem standard; if not (if the library wants to adopt the standard some day, but not yet), it's easier to replace the keyword than to change an existing keyword's behavior in backyard compatible way.

tylerjereddy:
@CloseChoice @connortann I've revised in the following ways, let me know if you'd like any other changes:

- I ignored two testsuite `mypy` complaints related to NumPy `get_state()`, which may have caused confusion because it has a return `type` that is variable based on a modern or legacy mode, and `mypy` picked the incorrect one for whatever reason
- Based on Matt's feedback, I switched away from the `rng` keyword to make it easier for the `shap` team to properly adopt SPEC 7 in the future if that ever becomes desirable. I went with `seed` instead, though it is admittedly slightly annoying to have an argument called `seed` that doesn't accept an integer and only a `Generator`. We could try to add support for integer as well, although scope creep perhaps...
- I rebased on latest `master` and `pytest --import-mode=append` still seems happy locally at least

connortann:
Thank you @tylerjereddy  for the PR and for bringing SPEC 7 to our attention. It is fantastic to have you as a contributor.

I think it would make sense for `shap` to move towards adopting SPEC 7, and I'm happy to work on making that happen (related: #3980). The good news is that we don't have an existing `seed` argument to handle (in this case), so the transition should be a bit simpler than in the reference implementation in the SPEC.

I think we should introduce the new `rng` parameter immediately, rather than `seed` which to me suggests an integer. I like your suggestion of maintaining backwards-compatibility for now. How is this as a transition plan:

Initially:

- Add the new `rng` parameter, which if provided will be normalised as per SPEC 7.
- If the global seed is set, raise a FutureWarning that the future behaviour will change

Then in future:

- Remove use of the global state, and just use `rng = np.random.default_rng(rng)`.

The initial change could be implemented as:

```python
def my_func(*, rng=None):

    if rng is not None:
        # If rng argument is provided, normalise as per SPEC 7
        rng = np.random.default_rng(rng)
    else:
        # Otherwise, maintain backwards compatibility for now, raising a warning if the global seed was set
        global_seed_set = np.random.mtrand._rand._bit_generator._seed_seq is None
        if global_seed_set:
            msg = (
                "The NumPy global RNG was seeded by calling `np.random.seed`. "
                "In a future version this function will no longer use the global RNG. "
                "Pass `rng` explicitly to opt-in to the new behaviour and silence this warning."
            )
            warnings.warn(msg, FutureWarning, stacklevel=2)

    # Example use
    inds = np.arange(10)
    if rng is None:
        # For now, maintain backwards compatibility.
        np.random.shuffle(inds)
    else:
        rng.shuffle(inds)
   
```

Then at some point in future we'd just have:

```python
def my_func(*, rng=None):
    rng = np.random.default_rng(rng)
    
    inds = np.arange(10)
    rng.shuffle(inds)
```


PS  - @tylerjereddy you might be interested in the contributor discussion https://github.com/shap/shap/discussions/3559

shap/plots/_beeswarm.py
```diff
@@ -566,6 +567,19 @@ def summary_legacy(
...
+        rng = np.random.default_rng(rng)
+    else:
+        global_seed_set = np.random.mtrand._rand._bit_generator._seed_seq is None
+        if global_seed_set:
+            msg = (
+                "The NumPy global RNG was seeded by calling `np.random.seed`. "
+                "In a future version this function will no longer use the global RNG. "
+                "Pass `rng` explicitly to opt-in to the new behaviour and silence this warning."
+            )
+            warnings.warn(msg, FutureWarning, stacklevel=2)
```
tylerjereddy, line 586: @connortann ok, I've attempted to revise to embrace SPEC 7 per your review comments.

shap/plots/_beeswarm.py
```diff
@@ -566,6 +567,19 @@ def summary_legacy(
         by default.
 
     """
+    # handle randomization machinery in conformance with SPEC 7
+    if rng is not None:
+        rng = np.random.default_rng(rng)
+    else:
+        global_seed_set = np.random.mtrand._rand._bit_generator._seed_seq is None
```
tylerjereddy: `mypy` wasn't happy with this locally, I guess we'll see if CI complains and perhaps ignore typing on this line; there's a new test for proper issuing of the `FutureWarning`, so the machinery works "ok" at runtime it seems.

tests/plots/test_summary_plot.py
```diff
@@ -34,3 +35,31 @@ def test_summary_plot(explainer):
...
     return fig
+
+
+@pytest.mark.parametrize(
+    "rng",
+    [
+        np.random.default_rng(167089660),
+        17,
+        np.random.SeedSequence(entropy=60767),
+    ],
```
tylerjereddy, line 46: I've expanded the `rng` `type`s a bit to reflect the flexibility of SPEC 7, though there are even more possibilities than this.

tests/plots/test_summary_plot.py
```diff
@@ -34,3 +35,31 @@ def test_summary_plot(explainer):
...
+    state_after = np.random.get_state()[1]  # type: ignore[index]
+    assert_array_equal(state_after, state_before)
+
+
+def test_summary_plot_warning(explainer):
+    # enforce FutureWarning for usage of global random
+    # state as we prepare for SPEC 7 adoption
+    shap_values = explainer(explainer.data)
+    with pytest.warns(FutureWarning, match="NumPy global RNG"):
+        shap.summary_plot(shap_values, show=False)
```
tylerjereddy, line 65: The `FutureWarning `is correctly issued here it seems. I did wonder if it was not ideal that we may be depending on the fact that (I think) the `shap` testsuite as a whole does a bunch of global random seed setting. If that gets cleaned up at some point I suppose this test might need to be adjusted, but at least the test should be noisy about it, and by then we might be closer to ready to avoid global seed usage anyway.

tylerjereddy:
It also occurred to me that we might want a docstring entry for `rng` argument, and indeed the SPEC even provides a draft one we could copy/paste. My main justification for laziness on that is that the function currently doesn't have docstring entries for many other parameters either, but wouldn't be hard to add it of course.

tests/plots/test_summary_plot.py
connortann: Thanks for adding the tests. On your point about the global random seed setting - you're right, there is a fixture that resets the global random seed before each test:

https://github.com/shap/shap/blob/baace0f6116c6115a36df0a829fb21c9d73addfa/tests/conftest.py#L59-L66

I think the PR looks good. As you say, if that fixture is removed for some reason, this test should fail loudly.

connortann:
> It also occurred to me that we might want a docstring entry for rng argument

That would be fantastic. One of many things about this function in need of improvement!

> mypy wasn't happy with this locally, I guess we'll see if CI complains and perhaps ignore typing on this line

It seems to fail on CI. That line I copied-and-pasted from the example implementation of SPEC 7, but it strikes me as a bit risky: we are accessing private attributes, which are presumably not part of the public numpy API and may break in future versions without warning. Maybe a safer approach would be to wrap this in a helper function as:

```python
def global_seed_is_set() -> bool:
    try:
        return np.random.mtrand._rand._bit_generator._seed_seq is None  # type: ignore
    except AttributeError:
        return False
```

That's just a suggestion: I'm not familiar with the internals of the numpy bit generator or whether it's safe to use.

Other than that, this PR looks good to go!

tylerjereddy:
Ok, I revised to:

1) Add a docstring entry for the new `rng` parameter of `summary_legacy()`.
2) Suppress the static analysis error for `..._bit_generator._seed_seq`. The reason I didn't add the guard you mention is because the same functionality is used unguarded upstream at i.e., https://github.com/scipy/scipy/blob/main/scipy/_lib/_util.py#L388. Looks like `skimage` uses it unguarded here as well: https://github.com/scikit-image/scikit-image/blob/main/skimage/feature/brief.py#L143. I believe NumPy is pretty cautious about making changes to the random infra, even the legacy random infra, because it is used so heavily in ecosystem tests. We could ping Robert Kern if you want though, or I can just guard it if you're still nervous.

connortann (approved):
Thanks again for the PR and for making those edits!

tylerjereddy:
@connortann anything else you want from me here before a merge?

connortann:
Thanks again for the PR!

PR #3947, opened 2025-01-03
fix mypy and mpl comparison

### LINKED ISSUE #3946: CI: Pipelines fail due to mypy and image comparison issue
### Issue Description

See the following pipelines:
 - https://github.com/shap/shap/actions/runs/12603222076/job/35132903948?pr=3945
 - https://github.com/shap/shap/actions/runs/12600404503/job/35132662328?pr=3944
Comparison old vs. new images can be downloaded from e.g. https://github.com/shap/shap/actions/runs/12600404503/artifacts/2384677696

```python
see above
```

### Traceback

_No response_

### Expected Behavior

CI runs through

### Bug report checklist

### Installed Versions

master

CloseChoice:
## Overview

Closes #3946  

This PR fixes mypy + the image issues. I did not put much effort in recreating the old issues, just recreated them, the differences look like some matplotlib defaults changed.

## Checklist


shap/plots/_beeswarm.py
```diff
@@ -188,7 +189,7 @@ def beeswarm(
     if ax is None:
         ax = pl.gca()
     fig = ax.get_figure()
-    assert fig is not None  # type narrowing for mypy
+    assert isinstance(fig, Figure)  # type narrowing for mypy
```
CloseChoice, line 192: It seems like in matplotlib 3.10 `ax.get_figure()` can return a subfigure which does not have a "set_size_inches" method, that we'll use in lines 362 cf. Therefore we check here directly if `fig` has the correct type.

connortann:
Unfortunately the tests still fail. It seems that re-creating the images with the latest matplotlib seems to break the tests for the older versions.

As a short-term patch to fix the tests, shall we just add a new constraint to the `test` and `test-core` dependency extra in `pyproject.toml`, to use the older version of matplotlib?

```toml
test = [
   # ... ,
  "matplotlib<3.10",  # TODO: get tests passing with latest matplotlib 
]
```

Then once we figure out what's causing the difference, hopefully we can find a way of making the tests consistent between versions

connortann:
I've done some further analysis, and I think I might have identified the root cause of part of the issue. I'll put a write-up on #3946

connortann:
The CI is passing so closing this as not needed for now.

PR #3976, opened 2025-01-27
MAINT: Disable conflicting lint rules

connortann:
A small housekeeping PR to ensure our ruff linter settings are up to date.

Disable ruff linting rules which may conflict with the formatter, as recommended in the ruff docs:

https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules


pyproject.toml
```diff
@@ -228,8 +228,15 @@ select = [
...
   # Recommended rules to disable when using ruff formatter:
-  "E117",  # Over-indented
-  "E501",  # Line too long
+  # https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules
+  "D206",
+  "D300",
+  "E111",
+  "E114",
+  "E117",
+  "E117",
```
Fredheda: @connortann Is this line a duplicate?

Fredheda: Just for my understanding: Is the reason the remaining rules from the "recommended to avoid" list (Q000, Q001, Q002, COM812, COM819) are not included because "Q" and "Com" are also not included on lines 214-228?

connortann: > Is this line a duplicate?

Yep! You spotted the, ahem, deliberate error I made to check you were paying attention. Nice one

> Is the reason the remaining rules from the "recommended to avoid" list (Q000, Q001, Q002, COM812, COM819) are not included because "Q" and "Com" are also not included on lines 214-228?

That's correct. In this case we select a few broad category of rules from the [available list](https://docs.astral.sh/ruff/rules/) ; and then we only need to disable specific rules from within the catergories that are in use.

Fredheda (approved):

PR #3982, opened 2025-01-29
MAINT: use "default" matplotlib style for tests

### LINKED ISSUE #3954: MAINT: use "default" matplotlib style for tests
## Proposal

## Rationale

## Implementation

```
[tool.pytest.ini_options]
mpl-default-style = "default"
```

hachoj:
## Overview

Closes #3954 

By default, pytest-mpl uses the "classic" matplotlib style, but modern matplotlib versions use the "default" style. This change updates `pyproject.toml` to explicitly use "default" so that test plots better reflect user expectations.

- Updated `mpl-default-style = "default"` in `pyproject.toml`
- Regenerated baseline test images using pytest-mpl
- Ensured tests pass with the new style

## Checklist


CloseChoice (approved):
LGTM, thanks for updating this. Would merge this on green pipelines

PR #3944, opened 2025-01-03
ENH: faster non-tree KernelExplainer

tylerjereddy:
* Fixes gh-3943.

* Cythonizing one of the inner loops used by a non-tree `KernelExplainer` is about 30 % faster on the small `sklearn` `SVC` test case in the matching issue. I'm not sure if you'll accept a new dependency on Cython or ask for a rewrite using the Python C-API (looks like you have some limited use of the Python C-API at the moment).

* A standard `pip install -v .` followed by `pytest --import-mode=append` (based on your apparent preferred test incantation) does pass the test suite locally on this branch.

* I don't think I saw standard `asv` performance benchmarks in the codebase, but let me know if I've missed them.

* This is also testing the waters to see if you'll actually accept performance patches of this nature, since this isn't even the largest performance bottleneck.

## Checklist

- [ ] All [pre-commit checks](https://pre-commit.com/#install) pass.
- [ ] Unit tests added (if fixing a bug or adding a new feature)

Issue #3943, opened 2025-01-02
Query, ENH: faster KernelExplainer


tylerjereddy:
Someone on my research team was having issues with very slow performance of `KernelExplainer` with i.e., `SVC` from `sklearn` (i.e., non-tree model SHAP performance). A representative code snippet is below the fold.

<details>

```python
import pandas as pd
from sklearn.svm import SVC
import numpy as np
from sklearn.pipeline import make_pipeline, Pipeline
from sklearn.preprocessing import StandardScaler
from sklearn.datasets import make_classification
from difflib import SequenceMatcher
import shap

X,y = make_classification(n_features=100, random_state=45479, class_sep=0.1)
clf = make_pipeline(StandardScaler(), SVC(probability=True, kernel='linear', gamma='auto', random_state=131231))
clf.fit(X, y)

X_train_trans = pd.DataFrame(clf['standardscaler'].transform(X))

# use all data for background
explainer = shap.KernelExplainer(clf['svc'].predict_proba, X_train_trans, random_state=564978)
shap_values_all = explainer(X_train_trans)
```

</details>

Just using `line_profiler` I noticed some low hanging fruit, basically a core/inner loop that is pure CPython. I went ahead and just hacked in a quick Cython replacement in feature branch `treddy_faster_non_tree_explainer` on my fork, and it was about 32 % faster even for that small example scenario. It would obviously need testing/checking/refinement, but is this something you'd be interested in? Is Cython "ok" or some other preferred compiled backend approach? Or do you prefer folks just maintain external "plugins" or something like that for custom explainers that are faster performing?

PR #3983, opened 2025-01-30
ENH: minor speedups in non-tree KernelExplainer

tylerjereddy:
* A few minor simplifications/adjustments that seem to give about 5-10% better performance for non-tree `KernelExplainer` for the scenario described at gh-3943.

* I don't think we have formal `asv`-style benchmarks in this project, but informal testing with 5 trials each (times in seconds) for the reproducer in the above issue (on an ARM Mac):
  - `master`:     52.803, 50.711, 51.503, 50.613, 51.348
  - this branch:  47.557, 47.128, 48.197, 47.651, 47.216

* This is less impactful than then improvements at gh-3944, but also less intrusive, and the improvements should compound with each other.

* I added postdoc @arhall0 as a co-author on the commit, since they did the initial profiling work to find the ~10% bottleneck.

* Since it appears that we only really need to identify the first point of difference between the arrays, it is probably possible to write an early-break algorithm that is more efficient, but that would probably increase complexity/need for compiled backend, so this is just a start to bump things a little bit.

## Checklist

### DIFF
--- a/shap/explainers/_kernel.py
+++ b/shap/explainers/_kernel.py
@@ -499,6 +499,9 @@ class KernelExplainer(Explainer):
         elif hasattr(i, "dtype") and hasattr(j, "dtype"):
             if np.issubdtype(i.dtype, np.number) and np.issubdtype(j.dtype, np.number):
                 return 0 if np.allclose(i, j, equal_nan=True) else 1
+            if np.issubdtype(i.dtype, np.bool_) and np.issubdtype(j.dtype, np.bool_):
+                return 0 if np.allclose(i, j, equal_nan=True) else 1
+            return 0 if all(i == j) else 1
         else:
             return 0 if i == j else 1
 
--- a/shap/explainers/_kernel.py
+++ b/shap/explainers/_kernel.py
@@ -495,7 +495,10 @@ class KernelExplainer(Explainer):
     def not_equal(i, j):
         number_types = (int, float, np.number)
         if isinstance(i, number_types) and isinstance(j, number_types):
-            return 0 if np.isclose(i, j, equal_nan=True) else 1
+            return 0 if np.allclose(i, j, equal_nan=True) else 1
+        elif hasattr(i, "dtype") and hasattr(j, "dtype"):
+            if np.issubdtype(i.dtype, np.number) and np.issubdtype(j.dtype, np.number):
+                return 0 if np.allclose(i, j, equal_nan=True) else 1
         else:
             return 0 if i == j else 1
 
@@ -510,8 +513,7 @@ class KernelExplainer(Explainer):
                         varying[i] = False
                         continue
                     x_group = x_group.todense()
-                num_mismatches = np.sum(np.frompyfunc(self.not_equal, 2, 1)(x_group, self.data.data[:, inds]))
-                varying[i] = num_mismatches > 0
+                varying[i] = self.not_equal(x_group, self.data.data[:, inds])
             varying_indices = np.nonzero(varying)[0]
             return varying_indices
         else:
--- a/tests/explainers/test_kernel.py
+++ b/tests/explainers/test_kernel.py
@@ -360,3 +360,19 @@ def test_kernel_logits_zeros_ones_probs(nsamples):
     pred = rf.predict_proba(X_test_sampled)
 
     np.testing.assert_allclose(sigm(shap_values.values.sum(1) + explainer.expected_value), pred, atol=1e-04)
+
+
+@pytest.mark.parametrize("dt", [np.bool_, np.object_])
+def test_explainer_non_number_dtype(dt):
+    seed = 45479
+    rng = np.random.default_rng(seed)
+    X = rng.choice([True, False], size=(15, 8)).astype(dt)
+    y = rng.choice([True, False], size=(15,)).astype(float)
+    rf = sklearn.ensemble.RandomForestClassifier(random_state=seed)
+    rf.fit(X, y)
+    explainer = shap.KernelExplainer(
+        model=rf.predict_proba,
+        data=X,
+        random_state=seed)
+    shap_values = explainer(X)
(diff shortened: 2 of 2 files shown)


shap/explainers/_kernel.py
```diff
@@ -495,7 +495,10 @@ def explain(self, incoming_instance, **kwargs):
     def not_equal(i, j):
         number_types = (int, float, np.number)
         if isinstance(i, number_types) and isinstance(j, number_types):
-            return 0 if np.isclose(i, j, equal_nan=True) else 1
+            return 0 if np.allclose(i, j, equal_nan=True) else 1
+        elif hasattr(i, "dtype") and hasattr(j, "dtype"):
+            if np.issubdtype(i.dtype, np.number) and np.issubdtype(j.dtype, np.number):
+                return 0 if np.allclose(i, j, equal_nan=True) else 1
```
tylerjereddy, line 501: This part of the changes was needed because some "perfectly-normal" NumPy arrays in the testsuite were not caught by the condition above after removing the `frompyfunc` business (which, of course, doesn't actually make CPython code any faster).

Obviously, make sure you/we are comfortable that the current testsuite is robust for this codepath. It did seem to have many failures without this shim, so possibly coverage is "ok."

CloseChoice, line 501: Thanks for the comment. I started a new conversation on these lines 'cause I somehow couldn't comment + suggest on the 4 lines on the previous one. Isn't there also a case where i and j are not a subtype of `np.number`? Is there a chance that we return `None` here? I can't think of one but just to be sure could we do:

```suggestion
            return 0 if np.allclose(i, j, equal_nan=True) else 1
        elif hasattr(i, "dtype") and hasattr(j, "dtype"):
            if np.issubdtype(i.dtype, np.number) and np.issubdtype(j.dtype, np.number):
                return 0 if np.allclose(i, j, equal_nan=True) else 1
        return 0 if i == j else 1
```

To have the general comparison as a fallback for all cases where if/else branches might be missing?

shap/explainers/_kernel.py
```diff
@@ -510,8 +513,7 @@ def varying_groups(self, x):
                         varying[i] = False
                         continue
                     x_group = x_group.todense()
-                num_mismatches = np.sum(np.frompyfunc(self.not_equal, 2, 1)(x_group, self.data.data[:, inds]))
-                varying[i] = num_mismatches > 0
+                varying[i] = self.not_equal(x_group, self.data.data[:, inds])
```
CloseChoice, line 519: nice improvement

CloseChoice (changes requested):

shap/explainers/_kernel.py
```diff
@@ -495,7 +495,13 @@ def explain(self, incoming_instance, **kwargs):
...
         number_types = (int, float, np.number)
         if isinstance(i, number_types) and isinstance(j, number_types):
-            return 0 if np.isclose(i, j, equal_nan=True) else 1
+            return 0 if np.allclose(i, j, equal_nan=True) else 1
+        elif hasattr(i, "dtype") and hasattr(j, "dtype"):
+            if np.issubdtype(i.dtype, np.number) and np.issubdtype(j.dtype, np.number):
+                return 0 if np.allclose(i, j, equal_nan=True) else 1
+            if np.issubdtype(i.dtype, np.bool_) and np.issubdtype(j.dtype, np.bool_):
+                return 0 if np.allclose(i, j, equal_nan=True) else 1
+            return 0 if all(i == j) else 1
```
tylerjereddy, line 504: I tried to address https://github.com/shap/shap/pull/3983#discussion_r1947839307, but found that I actually needed a slight adjustment to the fallback for array-like inputs with `all()`. I also added custom handling for `bool_`, which seems "ok" with `np.allclose()`.

tests/explainers/test_kernel.py
```diff
@@ -360,3 +360,16 @@ def test_kernel_logits_zeros_ones_probs(nsamples):
     pred = rf.predict_proba(X_test_sampled)
 
     np.testing.assert_allclose(sigm(shap_values.values.sum(1) + explainer.expected_value), pred, atol=1e-04)
+
+
+@pytest.mark.parametrize("dt", [np.bool_, np.object_])
```
tylerjereddy, line 365: This test is a bit weird in the sense that it doesn't really "fail before" and "pass after," but it does fail if the fallback I recently added uses `return 0 if i == j else 1` instead of `return 0 if all(i == j) else 1` because for the `np.object_` `dtype` you'll get the typical `ValueError: The truth value of an array with more than one element is ambiguous`.

So it is more of a sniff test for things slipping through the cracks, but it isn't super picky beyond that, and in fact the original returning of `None` noted above in some cases still allows the test to pass. It may be possible to cook the test up in a way that is more stringent than that, but it is at least doing *something*.

tests/explainers/test_kernel.py
```diff
@@ -360,3 +360,16 @@ def test_kernel_logits_zeros_ones_probs(nsamples):
...
+def test_explainer_non_number_dtype(dt):
+    seed = 45479
+    rng = np.random.default_rng(seed)
+    X = rng.choice([True, False], size=(15, 8)).astype(dt)
+    y = rng.choice([True, False], size=(15,)).astype(float)
+    rf = sklearn.ensemble.RandomForestClassifier(random_state=seed)
+    rf.fit(X, y)
+    explainer = shap.KernelExplainer(model=rf.predict_proba, data=X, random_state=seed)
+    shap_values = explainer(X)
+    np.testing.assert_allclose(shap_values.values.max(), 0.26548, rtol=1e-2)
```
tylerjereddy, line 375: the `rtol` needed for CI to be happy is rather high; conversely, the assertion here isn't really needed for the purpose of the test, which is mostly to fail if there is fundamental logic issue in `not_equal` like using `i == j` when `all(i == j)` is needed.

CloseChoice (approved):
LGMT, very nice.

PR #3985, opened 2025-02-01
FIX: Resolve numpy array comparison error in SHAP's KernelExplainer

### LINKED ISSUE #3899: BUG: Support for array-type features in Kernel Explainer with PyTorch model
### Issue Description

Does SHAP not support features in the form of arrays? I am attempting to extract Shapley values using the KernelExplainer for a PyTorch model. However, an error arises because our dataset contains features of various types, including some in numpy array format.

The error occurs specifically with features that are in numpy array form. In the _kernel module, the not_equal method at line 498 throws an error at the following line:

`return 0 if i == j else 1
`
The reason for this error is that two numpy arrays cannot be directly compared using ==. This results in the following error message:

`ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
`

A solution or modification seems necessary to handle cases like this.

Let me know if there’s anything more you’d like to add.

```python
import shap
    import numpy as np

    value = list(np.asarray([[0] * 100], dtype=np.float32))
    batch = pd.DataFrame({'feature_1': value})

    # make simple pytorch model
    class Model(torch.nn.Module):
        def __init__(self):
            super(Model, self).__init__()
            self.linear = torch.nn.Linear(100, 1)

        def forward(self, x):
            data = torch.from_numpy(x[0][0])
            return self.linear(data).detach().numpy()

    model = Model()

    explainer = shap.KernelExplainer(model, batch)
    shap_values = explainer.shap_values(batch)
```

### Traceback

  File "/home/bc-user/torch-dnn/data_loader/online/shap_dataset.py", line 97, in <module>
    shap_values = explainer.shap_values(batch)
  File "/home/bc-user/.pyenv/versions/torch-dnn/lib/python3.10/site-packages/shap/explainers/_kernel.py", line 271, in shap_values
    explanations.append(self.explain(data, **kwargs))
  File "/home/bc-user/.pyenv/versions/torch-dnn/lib/python3.10/site-packages/shap/explainers/_kernel.py", line 303, in explain
    self.varyingInds = self.varying_groups(instance.x)
  File "/home/bc-user/.pyenv/versions/torch-dnn/lib/python3.10/site-packages/shap/explainers/_kernel.py", line 511, in varying_groups
    num_mismatches = np.sum(np.frompyfunc(self.not_equal, 2, 1)(x_group, self.data.data[:, inds]))
  File "/home/bc-user/.pyenv/versions/torch-dnn/lib/python3.10/site-packages/shap/explainers/_kernel.py", line 498, in not_equal
    return 0 if i == j else 1
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

### Bug report checklist

### Installed Versions

0.46.0

bewygs:
## Overview

Closes #3899  

Fixes an error in `KernelExplainer.not_equal` when comparing `np.array`. We ran into a `ValueError`; I replaced `i == j` with `np.array_equal(i, j)`. I've also added unit tests.

# AI Disclosure for branch: fix-array-type

Since this PR had been dormant for a long time, I used Claude to help me get back
into it, checking whether the bug was still present after #3983 was merged, understanding how
`not_equal` had changed since, and updating the fix and tests accordingly. I reviewed and
understand every line.

## Checklist


CloseChoice:
Thanks for the fix. We do have 2 open PRs that fix the issue: https://github.com/shap/shap/pull/3983/files#diff-7ac25c0cf4216cbad9d7079e83c9c36fea1de8413f8fc7460d1b32af21d1fbc2R498-R501. Could we wait for the other PR to be merged and take your test cases? Or am I missing something here and you fix something different?

bewygs:
Sorry for the very late reply and leaving the PR hanging for so long. 

I've checked after #3983 was merged and the issue from #3899 is still there, the crash just moved to a different line. Feature values that are numpy arrays still end up compared with `==`, which returns an array of booleans instead of a single one, so `all(i == j)` now raises the same "truth value is ambiguous" error.

Minimal repro :

```python
import numpy as np, pandas as pd, shap

batch = pd.DataFrame({"feature_1": list(np.random.randn(3, 4).astype(np.float32))})
model = lambda x: np.array([np.sum(row[0]) for row in x])
shap.KernelExplainer(model, batch).shap_values(batch.iloc[:1])  # ValueError
```

I've rebased and updated the PR. Object arrays are now compared element-wise with `np.array_equal`. I kept the test cases and added another test where feature values are numpy arrays.

PR #3990, opened 2025-02-07
BUG: TreeExplainer numerical sensitivity

### LINKED ISSUE #3948: BUG: shap.TreeExplainer numerical sensitivity
I can no longer reproduce gh-1071, but a similar issue is a blocker for research progress for one of my postdocs. I was able to create a reproducer that can both trigger a similar thresholding error and be adjusted slightly to cause a segfault on latest `master` branch.

If you run the below code as-is on latest `shap` master, you should see the usual `shap.utils._exceptions.ExplainerError: Additivity check failed..`. If you adjust with `n_rows=3`, you should see `Segmentation fault (core dumped)`, or at least I do on x86_64 Linux, at the `exp.shap_values(X)` line.

```python
import numpy as np
from sklearn.ensemble import RandomForestClassifier
import shap

seed = 0
n_rows = 5
rng = np.random.default_rng(seed)
X = rng.integers(low=0, high=2, size=(n_rows, 90_000)).astype(np.float64)
y = rng.integers(low=0, high=2, size=n_rows)

clf = RandomForestClassifier(random_state=seed)
clf.fit(X, y)
clf.predict_proba(X)
exp = shap.TreeExplainer(clf, X)
exp.shap_values(X)
```

The segfault happens in `_cext.dense_tree_shap` inside `TreeExplainer.shap_values()`. The last C extension call before the other error scenario is `_cext.dense_tree_predict()`. I haven't had a chance to check the source of that for numerical instability yet. Our in-house use case is much messier, with a massive number of columns, design matrix has shape `(1585, 1404325)`. In that case, I plotted the distribution of thresholds that violate from:

```diff
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -795,6 +795,10 @@ class TreeExplainer(Explainer):
     def assert_additivity(self, phi, model_output):
         def check_sum(sum_val, model_output):
             diff = np.abs(sum_val - model_output)
+            thresholds = diff / (np.abs(sum_val) + 1e-2)
+            import pickle
+            with open("thresh.p", "wb") as f:
+                pickle.dump(thresholds, f)
             if np.max(diff / (np.abs(sum_val) + 1e-2)) > 1e-2:
```

and observed that it wasn't just a fluke--many records had violating thresholds:

image

tylerjereddy:
* Fixes gh-3948

* I drafted the regression test in the matching issue which fails before and passes after the patch discovered by @arhall0 (added as co-author here), that seems to indicate some types were set a bit too small. `pytest --import-mode=append` (full suite) also passes locally on this branch.

## Checklist

### DIFF
--- a/tests/explainers/test_tree.py
+++ b/tests/explainers/test_tree.py
@@ -1898,3 +1898,15 @@ def test_consistency_approximate(expected_result, approximate):
     explanations_shap_values_approx = exp.shap_values(arr, approximate=approximate)
     np.testing.assert_allclose(explanations_call_approx.values, explanations_shap_values_approx)
     np.testing.assert_allclose(explanations_call_approx.values, expected_result)
+
+
+@pytest.mark.parametrize("n_rows", [3, 5])
+def test_gh_3948(n_rows):
+    rng = np.random.default_rng(0)
+    X = rng.integers(low=0, high=2, size=(n_rows, 90_000)).astype(np.float64)
+    y = rng.integers(low=0, high=2, size=n_rows)
+    clf = sklearn.ensemble.RandomForestClassifier(n_estimators=1, random_state=0)
+    clf.fit(X, y)
+    clf.predict_proba(X)
+    exp = shap.TreeExplainer(clf, X)
+    exp.shap_values(X)
--- a/tests/explainers/test_tree.py
+++ b/tests/explainers/test_tree.py
@@ -1901,11 +1901,12 @@ def test_consistency_approximate(expected_result, approximate):
 
 
 @pytest.mark.parametrize("n_rows", [3, 5])
-def test_gh_3948(n_rows):
+@pytest.mark.parametrize("n_estimators", [1, 100])
+def test_gh_3948(n_rows, n_estimators):
     rng = np.random.default_rng(0)
     X = rng.integers(low=0, high=2, size=(n_rows, 90_000)).astype(np.float64)
     y = rng.integers(low=0, high=2, size=n_rows)
-    clf = sklearn.ensemble.RandomForestClassifier(n_estimators=1, random_state=0)
+    clf = sklearn.ensemble.RandomForestClassifier(n_estimators=n_estimators, random_state=0)
     clf.fit(X, y)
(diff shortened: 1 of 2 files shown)


tests/explainers/test_tree.py
```diff
@@ -1898,3 +1898,15 @@ def test_consistency_approximate(expected_result, approximate):
...
     np.testing.assert_allclose(explanations_call_approx.values, explanations_shap_values_approx)
     np.testing.assert_allclose(explanations_call_approx.values, expected_result)
+
+
+@pytest.mark.parametrize("n_rows", [3, 5])
+def test_gh_3948(n_rows):
+    rng = np.random.default_rng(0)
+    X = rng.integers(low=0, high=2, size=(n_rows, 90_000)).astype(np.float64)
+    y = rng.integers(low=0, high=2, size=n_rows)
+    clf = sklearn.ensemble.RandomForestClassifier(n_estimators=1, random_state=0)
```
tylerjereddy: May want to consider parametrizing `n_estimators` over `[1, 100]` -- both caused problems and fail before/pass after--I believe one of the two actually caused a segfault on x86_64 Linux.

There are quite a few issues currently open about additivity check failures, some for 5 years or more, so thorough testing seems worth

CloseChoice: I agree, would be great to have the test parametrized by `n_estimators` as well

CloseChoice (approved):
Amazing work, thanks for digging so deep and fixing this issue. I leave it to you if you want to parametrize the test.

tylerjereddy:
I added in the additional test parametrization over `n_estimators`, and confirmed that we see failures with the test on x86_64 Linux before the patch is applied. If you run the test more than once on that platform without the patch, you should start seeing segfaults as well.

CloseChoice:
> I added in the additional test parametrization over `n_estimators`, and confirmed that we see failures with the test on x86_64 Linux before the patch is applied. If you run the test more than once on that platform without the patch, you should start seeing segfaults as well.

Thanks for adding.

PR #3993, opened 2025-02-14
BUG: fix issue with tight tolerances when calling check_additivity

### LINKED ISSUE #3992: BUG: more cases with additivity check failures
Although gh-3990 is clearly an improvement on the additivity check issues, we do still have occasional in-house additivity check failures even with that branch, for other reasons. It may be the case that there are 1-2 small bugs left, or it may even be the case that being able to pass in custom relative and absolute tolerances for the checks may be useful (and/or allowing the option of a more accurate but slower summation algorithm).

I've had some success flushing out corner cases using `hypothesis` locally. For example, on the feature branch from the above PR, with a test like this one (probably could be much more thorough, but even this simple probing already finds bad inputs):

```diff
--- a/tests/explainers/test_tree.py
+++ b/tests/explainers/test_tree.py
@@ -11,6 +11,8 @@ import pytest
 import sklearn
 import sklearn.pipeline
 from sklearn.utils import check_array
+from hypothesis import given, strategies as st
+import hypothesis.extra.numpy as npst
 
 import shap
 from shap.explainers._tree import SingleTree
@@ -1911,3 +1913,18 @@ def test_gh_3948(n_rows, n_estimators):
     clf.predict_proba(X)
     exp = shap.TreeExplainer(clf, X)
     exp.shap_values(X)
+
+
+@given(npst.arrays(
+    dtype=np.float32,
+    shape=(10, 3),
+    elements=st.floats(min_value=0, max_value=1)
+))
+def test_tight_sensitivity_extra(X):
+    rng = np.random.default_rng(0)
+    y = rng.integers(low=0, high=2, size=10)
+    clf = sklearn.ensemble.ExtraTreesClassifier(n_estimators=100, random_state=0)
+    clf.fit(X, y)
+    clf.predict_proba(X)
+    exp = shap.TreeExplainer(clf, X)
+    exp.shap_values(X)
```

`pytest --import-mode=append -k "test_tight_sensitivity_extra"`

```
E           shap.utils._exceptions.ExplainerError: Additivity check failed in TreeExplainer! Please ensure the data matrix you passed to the explainer is the same shape that the model was trained on. If your data shape is correct then please report this on GitHub. This check failed because for one of the samples the sum of the SHAP values was 0.006000, while the model output was 0.000000. If this difference is acceptable you can set check_additivity=False to disable this check.
E           Falsifying example: test_tight_sensitivity_extra(
E               X=array([[1.     , 1.     , 0.99999],
E                      [1.     , 1.     , 1.     ],
E                      [1.     , 1.     , 1.     ],
E                      [1.     , 1.     , 1.     ],
E                      [1.     , 1.     , 1.     ],
E                      [1.     , 1.     , 1.     ],
E                      [1.     , 1.     , 1.     ],
E                      [1.     , 1.     , 1.     ],
E                      [1.     , 1.     , 1.     ],
E                      [1.     , 1.     , 1.     ]], dtype=float32),
E           )
E           Explanation:
E               These lines were always and only run by failing examples:
E                   /Users/treddy/python_venvs/py_311_ldrd_host_virus/lib/python3.11/site-packages/shap/explainers/_tree.py:798
E                   /Users/treddy/python_venvs/py_311_ldrd_host_virus/lib/python3.11/site-packages/shap/explainers/_tree.py:800

../../python_venvs/py_311_ldrd_host_virus/lib/python3.11/site-packages/shap/explainers/_tree.py:811: ExplainerError
```

adamwitmer:
* Closes #3992

* modified the `check_sum` function to use `np.allclose` for comparing `sum_val` and `model_outputs` because the relative error function that was being used was sensitive to very small differences (values very close to zero) between the model outputs and calculated SHAP sums

* added two variables `abs_tol` and `rel_tol` to allow for customization of the absolute and relative tolerance values when calling `explainer.shap_values`

* added a portable regression test that reproduces the error locally

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,7 +88,6 @@ test = [
   # Constraint to prevent the combination of tf<2.15 and numpy>=2.0.
   # See GH #3707, #3768, 3922
   "numpy<2.0",
-  "hypothesis",
 ]
 
 test_notebooks = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,6 +88,7 @@ test = [
   # Constraint to prevent the combination of tf<2.15 and numpy>=2.0.
   # See GH #3707, #3768, 3922
   "numpy<2.0",
+  "hypothesis",
 ]
 
 test_notebooks = [
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -457,6 +457,8 @@ class TreeExplainer(Explainer):
         approximate: bool = False,
         check_additivity: bool = True,
         from_call: bool = False,
+        abs_tol: float = 1e-2,
+        rel_tol: float = 1e-2,
     ):
         """Estimate the SHAP values for a set of samples.
 
@@ -483,6 +485,13 @@ class TreeExplainer(Explainer):
             Run a validation check that the sum of the SHAP values equals the output of the model. This
             check takes only a small amount of time, and will catch potential unforeseen errors.
             Note that this check only runs right now when explaining the margin of the model.
+
+        abs_tol : float
+            absolute tolerance value for checking the difference between shap calculations and model
+            output when calling `check_additivity=True`
+
+        rel_tol : float
+            relative tolerance value for handling very small numbers when calling `check_sum`
 
         Returns
         -------
@@ -573,7 +582,7 @@ class TreeExplainer(Explainer):
                     out = phi[:, :-1]
 
                 if check_additivity and model_output_vals is not None:
-                    self.assert_additivity(out, model_output_vals)
+                    self.assert_additivity(out, model_output_vals, abs_tol, rel_tol)
                 if isinstance(out, list):
                     out = np.stack(out, axis=-1)
                 return out
@@ -630,7 +639,7 @@ class TreeExplainer(Explainer):
 
         out = self._get_shap_output(phi, flat_output)
         if check_additivity and self.model.model_output == "raw":
-            self.assert_additivity(out, self.model.predict(X))
+            self.assert_additivity(out, self.model.predict(X), abs_tol, rel_tol)
 
         # This statements handles the case of multiple outputs
         # e.g. a multi-class classification problem, multi-target regression problem
@@ -791,10 +800,10 @@ class TreeExplainer(Explainer):
                 out = np.stack([phi[:, :-1, :-1, i] for i in range(self.model.num_outputs)], axis=-1)
         return out
 
-    def assert_additivity(self, phi, model_output):
-        def check_sum(sum_val, model_output):
+    def assert_additivity(self, phi, model_output, abs_tol, rel_tol):
+        def check_sum(sum_val, model_output, abs_tol, rel_tol):
             diff = np.abs(sum_val - model_output)
-            if np.max(diff / (np.abs(sum_val) + 1e-2)) > 1e-2:
+            if not np.allclose(sum_val, model_output, atol=abs_tol, rtol=rel_tol):
                 ind = np.argmax(diff)
                 err_msg = (
                     "Additivity check failed in TreeExplainer! Please ensure the data matrix you passed to the "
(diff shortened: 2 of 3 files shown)


shap/explainers/_tree.py
```diff
@@ -457,6 +457,8 @@ def shap_values(
         approximate: bool = False,
         check_additivity: bool = True,
         from_call: bool = False,
+        abs_tol: float = 1e-2,
```
CloseChoice: Please let's not fix a problem AND extend the interface in one PR.

On a general I think if we change the interface, we'll do it on all explainers simultaneously to keep consistency. Furthermore, we would not only need to add it to the `shap_values` method (actually that is kind of a legacy thing and at best shouldn't be touched at all even though your change is backwards compatible) but also to the `__call__` methods.

So may I ask you to remove abs_tol and rel_tol, please?

tests/explainers/test_tree.py
```diff
@@ -5,11 +5,14 @@
 import pickle
 import sys
 
+import hypothesis.extra.numpy as npst
```
CloseChoice: I think adding hypothesis in general is a good thing, especially for flaky tests, which fortunately we don't have a lot. Could you please add it to the [test requirements](https://github.com/shap/shap/blob/master/pyproject.toml#L67-L90) so that the tests pass?

tests/explainers/test_tree.py
```diff
@@ -1898,3 +1901,24 @@ def test_consistency_approximate(expected_result, approximate):
     explanations_shap_values_approx = exp.shap_values(arr, approximate=approximate)
     np.testing.assert_allclose(explanations_call_approx.values, explanations_shap_values_approx)
     np.testing.assert_allclose(explanations_call_approx.values, expected_result)
+
+
+@pytest.mark.slow
```
CloseChoice: Thanks for the test. In here we basically just check the `assert_additivity` method, right? Since our pipelines run quite a while already and for the sake of keeping things simple I think it is better if we create the model and explainer once and then just pass the `phi` and `model_output` arrays directly to `assert_additivity` so that we generate directly what we test here. One example input should be:

```python

(Pdb++) phi
[array([[ 0.        ,  0.        , -0.24750001],
       [ 0.        ,  0.        ,  0.0825    ],
       [ 0.        ,  0.        ,  0.0825    ],
       [ 0.        ,  0.        ,  0.0825    ]]), array([[ 0.        ,  0.        ,  0.24749997],
       [ 0.        ,  0.        , -0.08249999],
       [ 0.        ,  0.        , -0.08249999],
       [ 0.        ,  0.        , -0.08249999]])]
(Pdb++) model_output
array([[0.        , 1.        ],
       [0.33333333, 0.66666667],
       [0.33333333, 0.66666667],
       [0.33333333, 0.66666667]])
```
Since these inputs crash current master

CloseChoice (changes requested):
Thanks a lot for the PR, it's deeply appreciated. This looks like a reasonable change, since the previous comparison method has its flaws

tylerjereddy:
@adamwitmer it would be good to ask the `shap` team questions if you have them--we discussed earlier this week that the testing situation still looks a bit weird here for example, so if you need clarification there I'm sure they'd be happy to answer if you are clear about what you don't understand.

tests/explainers/test_tree.py
adamwitmer: Would the new design still necessitate the use of `hypothesis` in the test? Currently, I have removed that and replaced it with a test that is parametrized on `phi` and `model_output`, however, as @tylerjereddy mentioned the implementation is a bit awkward. If you think it's best to keep `hypothesis`, I can revise my test accordingly, otherwise I can fix the current test to make it more streamlined.

shap/explainers/_tree.py
adamwitmer: The current tolerances used for the `check_sum` function seem quite large, is there a reason for this, and do they still seem reasonable based on the change in comparison method?

pyproject.toml
```diff
@@ -87,6 +87,7 @@ test = [
   # Constraint to prevent the combination of tf<2.15 and numpy>=2.0.
   # See GH #3707, #3768, 3922
   "numpy<2.0",
+  "hypothesis",
```
CloseChoice: we don't need hypothesis anymore, do we?

CloseChoice (changes requested):
This looks good. Please remove the dependency, then we're ready to merge

CloseChoice (approved):
LGTM, thanks. Nice work

CloseChoice:
Thanks again @adamwitmer, looking forward to see more of you ;)

PR #4006, opened 2025-02-28
BUG: fix additivity check failure uint32 overflow

arhall0:
* related to shap#4002

* addresses issue for 'tree_path_dependent'

* adds regression test that reproduces error before fix

## Overview

This PR attempts to correct part of shap#4002 for `feature_perturbation='tree_path_dependent'`. The test for this is quite slow; around ~120s on my local machine.

## Checklist

### DIFF
--- a/tests/explainers/test_tree.py
+++ b/tests/explainers/test_tree.py
@@ -1979,6 +1979,7 @@ def test_sklearn_tree_explainer_with_missing_values(X, y, expected_shap_values):
     np.testing.assert_allclose(shap_values, expected_shap_values)
 
 
+@pytest.mark.xslow
 def test_overflow_tree_path_dependent():
     """GH #4002
     Test SHAP values computation for `feature_perturbation='tree_path_dependent'` with large number of features."""
--- a/tests/explainers/test_tree.py
+++ b/tests/explainers/test_tree.py
@@ -1977,3 +1977,19 @@ def test_sklearn_tree_explainer_with_missing_values(X, y, expected_shap_values):
 
     # Verify SHAP values match expected values
     np.testing.assert_allclose(shap_values, expected_shap_values)
+
+
+def test_overflow_tree_path_dependent():
+    """GH #4002
+    Test SHAP values computation for `feature_perturbation='tree_path_dependent'` with large number of features."""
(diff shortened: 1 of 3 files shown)


tests/explainers/test_tree.py
```diff
@@ -1911,3 +1911,19 @@ def test_gh_3948(n_rows, n_estimators):
...
+    n_rows = 2_000
+    rng = np.random.default_rng(seed)
+    X = rng.integers(low=0, high=2, size=(n_rows, 1_100_000)).astype(np.float64)
+    y = rng.integers(low=0, high=2, size=n_rows)
+
+    clf = sklearn.ensemble.RandomForestClassifier(random_state=seed)
+    clf.fit(X, y)
+    clf.predict_proba(X)
+    exp = shap.Explainer(clf, algorithm="tree", feature_perturbation="tree_path_dependent")
+    exp(X)
```
tylerjereddy, line 1996: I confirmed fail before/pass after the patch locally on ARM Mac.

For the issue of the test causing the CI to barf because of its excessive memory usage, the typical solution would be to use a `pytest.mark.xslow` or similar so that the test is only run periodically/locally by a maintainer, given current GHA memory limits: https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners

That said, I'd wait for the `shap` team to comment before making adjustments. It isn't an ideal solution, since things could slip through easily that way, but it at least provides a method to `git bisect` on a high-memory local machine if something comes up, and also a more rigorous way to test locally if someone touches this file in the future.

CloseChoice: I agree, not ideal but the bug is only reproducible in this way, so I totally support adding `pytest.mark.xslow` here and we can run this locally from time to time.

arhall0: I did not see support for `pytest.mark.xslow` so I added it. Let me know if you have a preferred method/mark to use instead.

arhall0:
@CloseChoice Could you give this another look? I think it may be ready.

CloseChoice (approved):
LGTM, thanks for the work. Really appreciate your contributions.

PR #4014, opened 2025-03-08
All dependabot changes

CloseChoice:
## Overview

Replaces #4008 #3942 and #3942  

## Checklist


javascript/tests/__snapshots__/render.spec.jsx.snap
```diff
@@ -1,20 +1,12 @@
...
-      "sansSerif": true,
-      "userSelect": "none",
-    }
-  }
->
-  <style
-    dangerouslySetInnerHTML={
-      {
-        "__html": "
+exports[`AdditiveForceVisualizer renders correctly 1`] = `
```
CloseChoice, line 3: I updated this, since https://github.com/shap/shap/pull/3941 breaks this test. The true tests IMO are the mpl comparisons in tests/plots

javascript/Makefile
```diff
@@ -0,0 +1,3 @@
+buildjs:
```
CloseChoice, line 1: General note:
we have bundle.js at two places: first we create it using `npm run build` in `javascript/build/bundle.js`. This file needs to end up in `shap/plots/resources/bundle.js`.

This is the first step to automatically create and copy the bundle.js to the right place.
I also thought about creating the bundle.js as a softlink but this might be an issue on different OS and/or when building wheels (idk whether it really causes problems there, I just remember that linked files cannot be copied easily into docker containers and think we might run into similar issues here). 

Maybe in the future we'll have one github workflow that copies the js file over and then runs the tests in `tests/plots` to make sure that even with the new bundle everything is as expected, this works well with the dependabot updates. Or we'll create a pre-commit flow that compares the two bundle.js and enforces that they're equal. Just some thoughts, curious for your input here

PR #4017, opened 2025-03-12
Fix/check for attribute to extract colors

fabianliebig:
## Overview
Adds further checks to avoid AttributeError.

The `beeswarm` function checks the if the given `color` is of type `matplotlib.colors.Colormap` in one else path of the function and access the attribute `colors` if that check is true (it is in line [454](https://github.com/shap/shap/blob/cadb4d0300eb2c9104a48a604910d114ffd0a84c/shap/plots/_beeswarm.py#L454)). However, not all subtypes of `matplotlib.colors.Colormap` have that attribute (see this [discussion])(https://stackoverflow.com/questions/59395038/why-the-structure-of-matplotlibs-color-maps-differ). Therefore this PR just add a further attribute check as well as a test, assuming that the type check should persist.
## Checklist

### DIFF
--- a/shap/plots/_beeswarm.py
+++ b/shap/plots/_beeswarm.py
@@ -451,7 +451,7 @@ def beeswarm(
                 rasterized=len(shaps) > 500,
             )
         else:
-            if safe_isinstance(color, "matplotlib.colors.Colormap"):
+            if safe_isinstance(color, "matplotlib.colors.Colormap") and hasattr(color, "colors"):
                 color = color.colors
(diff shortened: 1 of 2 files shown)


CloseChoice:
@fabianliebig Thanks, this looks great. If the last test passes, I'll merge this. Don't worry about the macos test, this has nothing to do with your changes.

CloseChoice:
@fabianliebig Thanks a lot for your efforts. Happy to see you contribute to shap

PR #4019, opened 2025-03-14
Fix huggingface issue

CloseChoice:
## Overview

Fix the flaky tests in CI that download models and tokenizers from huggingface.
This PR adds a retry in case of failure to download

## Checklist


CloseChoice:
@connortann would be great if you could have a timely look at this to get our tests reliable again.

PR #4020, opened 2025-03-15
Ci warning matplotlib vert bool future deprecation

### LINKED ISSUE #3975: CI warning: matplotlib: vert: bool will be deprecated in a future version
```
tests/plots/test_violin.py::test_violin
  /opt/hostedtoolcache/Python/3.12.8/x64/lib/python3.12/site-packages/shap/plots/_violin.py:248: PendingDeprecationWarning: vert: bool will be deprecated in a future version. Use orientation: {'vertical', 'horizontal'} instead.
    parts = pl.violinplot(
```

The parameter `vert` is used in `_beeswarm.py` and `_violin.py`.

It looks like the parameter `orientation` was introduced in matplotlib `3.10`. According to [SPEC 0](https://scientific-python.org/specs/spec-0000/) , we need to support matplotlib versions back to at least `3.7`, so for now we need to handle both the new and the old arguments.

So, we need to ensure our code handles both cases, e.g.

```python
from packaging import version
if version.parse(matplotlib.__version__) >= version.parse("3.10"):
    orientation_kwarg = dict(orientation="horizontal")
else:
    # TODO: simplify this when we drop support for matplotlib 3.9
    orientation_kwarg = dict(vert=False)

pl.violinplot(..., **orientation_kwarg)

Fredheda:
## Overview

Closes #3975 

MAINT: 

Housekeeping PR to ensure CI warnings related to a deprecation change matplotlib are addressed by ensuring that the right kwargs are used in the violinplot function, depending on which version of matplotlib is installed.

If matplotlib version 3.10 or higher is installed: orientation="horizontal" is used. If the package version is lower then vert=False is used. Solution follows guidance outlined [here](https://github.com/shap/shap/issues/3975#issue-2806773008)

## Checklist

### DIFF
--- a/shap/plots/_beeswarm.py
+++ b/shap/plots/_beeswarm.py
@@ -29,6 +29,7 @@ from ._utils import (
     sort_inds,
 )
 
+# TODO: simplify this when we drop support for matplotlib 3.9
 if version.parse(matplotlib.__version__) >= version.parse("3.10"):
     ORIENTATION_KWARG = {"orientation": "horizontal"}
 else:
--- a/shap/plots/_beeswarm.py
+++ b/shap/plots/_beeswarm.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import warnings
 from typing import Literal
 
+import matplotlib
 import matplotlib.pyplot as pl
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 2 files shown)


CloseChoice (approved):
looks great, thanks a lot

PR #4041, opened 2025-03-23
Fix/single unique value jitter

### LINKED ISSUE #4037: BUG: To small or all equal values in data produce a ValueError in `_suggest_x_jitter`
### Issue Description

I recently encountered a ValueError when creating a scatter plot with data that contains only one unique value or where the differences between the unique values are smaller than 1e-8. This issue is related to the function  [`_suggest_x_jitter`](https://github.com/shap/shap/blob/d47eb01f8a3a82d0cbe9bf3f340c1aa1fae34cea/shap/plots/_scatter.py#L443) which is called when [`x_jitter` is set to "auto"](https://github.com/shap/shap/blob/d47eb01f8a3a82d0cbe9bf3f340c1aa1fae34cea/shap/plots/_scatter.py#L235).  In this function, the distances between unique values are filtered, resulting in an empty list for cases such as the synthetic minimal example. Consequently, [`np.min`  throws a ValueError.](https://github.com/shap/shap/blob/0d1ae5a62c9e0cafc2368fddd1b76e670cf44cb5/shap/plots/_scatter.py#L449)

I'm not entirely sure if this is a bug or an invalid input but it already was mentioned in #2632 once. This issue however, were closed automatically. 

```python
import shap
import numpy as np

explanations = shap.Explanation(
    np.array([[-0.20760354], [-0.21398779]]),
    data=np.array([[1], [1]]),
    feature_names=["feature1"],
)

shap.plots.scatter(explanations)
```

### Traceback

```shell
Cell In[24], line 10
      2 import numpy as np
      4 explanations = shap.Explanation(
      5     np.array([[-0.20760354], [-0.21398779]]),
      6     data=np.array([[1], [1]]),
      7     feature_names=["feature1"],
      8 )
---> 10 shap.plots.scatter(explanations)

File c:\Users\fabi\Repos\shap\shap\plots\_scatter.py:146, in scatter(shap_values, color, hist, axis_color, cmap, dot_size, x_jitter, alpha, title, xmin, xmax, ymin, ymax, overlay, ax, ylabel, show)
    144 for i in inds:
    145     ax = plt.subplot(1, len(inds), i + 1)
--> 146     scatter(shap_values[:, i], color=color, show=False, ax=ax, ymin=ymin, ymax=ymax)
    147     if overlay is not None:
    148         line_styles = ["solid", "dotted", "dashed"]

File c:\Users\fabi\Repos\shap\shap\plots\_scatter.py:235, in scatter(shap_values, color, hist, axis_color, cmap, dot_size, x_jitter, alpha, title, xmin, xmax, ymin, ymax, overlay, ax, ylabel, show)
    233 # pick jitter for categorical features
    234 if x_jitter == "auto":
--> 235     x_jitter = _suggest_x_jitter(features[:, ind])
    237 # guess what other feature as the stongest interaction with the plotted feature
    238 if interaction_index == "auto":

File c:\Users\fabi\Repos\shap\shap\plots\_scatter.py:449, in _suggest_x_jitter(values)
    446 try:
    447     # Identify the smallest difference between unique values
    448     diffs = np.diff(unique_vals)
--> 449     min_dist = np.min(diffs[diffs > 1e-8])
    450 except TypeError:
    451     # If unique_vals contains non-numeric values, set arbitrarily at 1
    452     min_dist = 1

File c:\Users\fabi\Repos\shap\.venv\lib\site-packages\numpy\core\fromnumeric.py:2953, in min(a, axis, out, keepdims, initial, where)
   2836 @array_function_dispatch(_min_dispatcher)
   2837 def min(a, axis=None, out=None, keepdims=np._NoValue, initial=np._NoValue,
   2838         where=np._NoValue):
   2839     """
   2840     Return the minimum of an array or minimum along an axis.
   2841 
   (...)
   2951     6
   2952     """
-> 2953     return _wrapreduction(a, np.minimum, 'min', axis, None, out,
   2954                           keepdims=keepdims, initial=initial, where=where)

File c:\Users\fabi\Repos\shap\.venv\lib\site-packages\numpy\core\fromnumeric.py:88, in _wrapreduction(obj, ufunc, method, axis, dtype, out, **kwargs)
     85         else:
     86             return reduction(axis=axis, out=out, **passkwargs)
---> 88 return ufunc.reduce(obj, axis, dtype, out, **passkwargs)

ValueError: zero-size array to reduction operation minimum which has no identity
```

### Expected Behavior

I would eighter expect the `_suggest_x_jitter` to catch the exception and set `min_dist` to 1 as with `TypeError` exceptions or return a validation error if such an input is unexpected.

### Bug report checklist

### Installed Versions

0.47.0

fabianliebig:
## Overview

Closes #4037

This pull request incorporates the recommendations from issue #4037 and introduces basic tests related to generating jitter for the scatter plot. The implementation addresses scenarios where all inputs for the [`x_jitter`](https://github.com/shap/shap/blob/d47eb01f8a3a82d0cbe9bf3f340c1aa1fae34cea/shap/plots/_scatter.py#L235) have identical values or where the value-differences between unique values are less than 1e-8.

## Checklist


tests/plots/test_scatter.py
```diff
@@ -96,3 +97,24 @@ def test_scatter_categorical(categorical_explanation):
     shap.plots.scatter(categorical_explanation[:, "sex"], ax=ax, show=False)
     plt.tight_layout()
     return fig
+
+
+@pytest.mark.parametrize(
```
CloseChoice: could you pls just test the example from the issue? We don't test private methods since they can be subject to change.

CloseChoice (changes requested):

tests/plots/test_scatter.py
fabianliebig: Sure, I've pushed an updated version :). Is that more appropriate?

tests/plots/test_scatter.py
```diff
@@ -96,3 +97,15 @@ def test_scatter_categorical(categorical_explanation):
     shap.plots.scatter(categorical_explanation[:, "sex"], ax=ax, show=False)
     plt.tight_layout()
     return fig
+
+
+@pytest.mark.parametrize("input", [np.array([[1], [1]]), np.array([[1e-10], [1e-9]]), np.array([[1]])])
+def test_scatter_plot_value_input(input):
+    """Test scatter plot with different input values."""
```
CloseChoice: looks good. Two final things: would you be so kind and refer to the issue as well:

GH #4037

So that we can easily check where this is originating from.

Secondly, could you pls return the figure, like it's done in other tests and add the 
`@pytest.mark.mpl_image_compare` decorator and run

```bash
pytest test/plots/test_scatter.py::test_scatter_plot_value_input --mpl-generate-path=tests/plots/baseline
```
this will then add a file with the name of the test to `tests/plots/baseline` and then add this to git as well? This will help us keep the plots consistent across releases.

CloseChoice (changes requested):

tests/plots/test_scatter.py
fabianliebig: Okay, done :) Thanks for your feedback.

CloseChoice (approved):
looks good to me, thanks for the effort

CloseChoice (approved):
LGTM. Thanks for the effort.

PR #4049, opened 2025-04-02
MAINT: Changed alias

### LINKED ISSUE #4048: STYLE: Inconsistent aliasing when importing matplotlib.pyplot
Hi all,

I was reading through the code in `shap/plots`, and I noticed that in some files (including but not limited to `_bar.py` and `_embedding.py`) pyplot is imported using the alias `import matplotlib.pyplot as pl`.

In other files, however, pyplot is imported using the alias `import matplotlib.pyplot as plt`. Some examples of this can be found in `_benchmark.py` and `_waterfall.py`.

Should these be changed to use a consistent alias across all files? If so, which is preferred? `pl` or `plt`?

Ja-Tink:
## Overview

Closes #4048  

I replaced all instances of the alias `import matplotlib.pyplot as pl` with `import matplotlib.pyplot as plt` in `shap/plots`.

If `pl` is the preferred alias, I can edit this PR to switch to that, instead.

## Checklist


CloseChoice (approved):
Looks great, thanks for the PR.

Issue #3819, opened 2024-08-20
ENH: Remove the `show` argument in shap.plots


connortann:
### Problem Description

At present the plotting functions accept a parameter `show` which defaults to `True`, which calls `plt.show()`. I suggest changing the default behaviour to not call `plt.show`, and deprecating the `show` parameter.

### Rationale

Overall I think this parameter creates needless complexity for no added value.

- It's not needed. In typical notebook environments (e.g. VS Code, jupyter, jupyterlab), matplotlib plots are rendered automatically without needing to call `plt.show()`.
- It can be frustrating, as plots can displaying twice (once by plt.show, and again by the notebook environment).
- It makes the API more complex. Other common plotting libraries (e.g. seaborn) do not have this parameter, as far as I know. Our plotting modules are already complicated enough as is!

### API complexity

We generally want to return `ax` to fit with matplotlib's [recommended signature](https://matplotlib.org/stable/users/explain/quick_start.html#making-a-helper-functions), but this is only workable _before_ `plt.show()` is called. At present we have to return the `ax` only if `show=False`, and otherwise return `None`.

This means the type signatures need to be quite complex:

```python
@overload
def plot(..., show: Literal[True]) -> None: ...
@overload
def plot(..., show: Literal[False]) -> plt.Axes: ...
def plot(..., show: bool) -> plt.Axes | None: ...
```

The simpler API could just be:

```python
def plot(...) -> plt.Axes
```

### Suggested implementation plan

- [ ] Change the default behaviour to not call `plt.show()`, and raise a deprecation warning if `show` is passed (whilst still being backwards-compatible).
- [ ] In a several releases' time, remove the `show` argument. To be tracked on #3507

Anushreebasics:
Hi @connortann ,I would like to work on this issue

PR #4056, opened 2025-04-08
MAINT: Added deprecation warning to show argument for waterfall plot

Ja-Tink:
## Overview

Implements #3819 (to be tracked on #3507)  

I added a deprecation warning to the "show" argument in `_waterfall.py`. If this looks good, I am happy to add a similar warning to all the other plotting functions as outlined in #3819.

Screenshot 2025-04-07 at 10 48 22 PM

## Checklist


Ja-Tink:
Also, should I update the docs to reflect this change? Or should that wait until all the plot functions have been changed?

shap/plots/_waterfall.py
```diff
@@ -43,6 +45,14 @@ def waterfall(shap_values, max_display=10, show=True):
     See `waterfall plot examples <https://shap.readthedocs.io/en/latest/example_notebooks/api_examples/plots/waterfall.html>`_.
 
     """
+    # Deprecation warning
+    if show is not None:
+        warnings.warn(
+            "WARNING: `show` argument is deprecated and will be removed in a future version. Please omit this argument and call plt.show() manually.",
+            category=UserWarning,
```
jack-mcivor: Should this be `DeprecationWarning` ?

shap/plots/_waterfall.py
```diff
@@ -43,6 +45,14 @@ def waterfall(shap_values, max_display=10, show=True):
     See `waterfall plot examples <https://shap.readthedocs.io/en/latest/example_notebooks/api_examples/plots/waterfall.html>`_.
 
     """
+    # Deprecation warning
+    if show is not None:
+        warnings.warn(
+            "WARNING: `show` argument is deprecated and will be removed in a future version. Please omit this argument and call plt.show() manually.",
```
jack-mcivor: The `WARNING:` prefix is not necessary

shap/plots/_waterfall.py
Ja-Tink: Thanks for the catch, fixed

shap/plots/_waterfall.py
Ja-Tink: Fixed

shap/plots/_waterfall.py
```diff
@@ -13,7 +15,7 @@
 # plot that is associated with that feature get overlaid on the plot...it would quickly allow users to answer
 # why a feature is pushing down or up. Perhaps the best way to do this would be with an ICE plot hanging off
 # of the bar...
-def waterfall(shap_values, max_display=10, show=True):
+def waterfall(shap_values, max_display=10, show=None):
```
CloseChoice, line 18: Please let's not change the default here since this is a breaking change.

CloseChoice (changes requested):

shap/plots/_waterfall.py
Ja-Tink: Gotcha, I'll change that.

In that case, when should the deprecation warning trigger? Is there a way to detect whether `show=True` because it is the default argument versus it being explicitly passed by the user? If not, should the warning trigger whenever the function is called, regardless of the arguments used?

CloseChoice: Hmm, I am a bit reluctant to add this, since we use the `show` parameter in a lot of different plots and I would like to deprecate all of them at once.

CloseChoice:
Note: we should use this PR and enhance it to deprecate all show arguments for the upcoming release

PR #4058, opened 2025-04-09
fix: fix neutral language in SHAP value description

Hrafz:
Replace "improves" with "increases" when describing how median income affects house prices, as this is more neutral and accurate terminology. Higher median income increases predicted house prices without implying this is necessarily better or worse.

## Overview

- Changed wording in documentation from "improves" to "increases" when describing the relationship between median income and predicted house prices
- This provides more neutral and precise terminology that avoids implying a value judgment about whether higher prices are better

## Checklist
- [ ] Unit tests added (not applicable for documentation change)

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ shap.plots.scatter(shap_values[:, "Latitude"], color=shap_values)
 </p>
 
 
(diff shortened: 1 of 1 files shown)


CloseChoice (approved):
LGTM, thanks for the PR.

PR #4091, opened 2025-05-21
Propagate _xgb_dmatrix_props to SHAP value calculation for DMatrices

cedricdonie:
Calculating interaction values and dependence plots did not work, because the constructed DMatrix did not have category enabled, and encoding category values was broken when the categorical value also contained missing values.

## Overview

- Falls back to the builtin `set`  for unique value calculation if `np.unique` fails. This will be the case if the array to be encoded contains NaN (`np.float64`) and strings/objects. See also https://github.com/numpy/numpy/issues/18147, https://github.com/numpy/numpy/issues/28364.
- Checks for the existence of DMatrix settings and uses these in `shap_interaction_values`, as is already done in. A similar fix was performed in #3314 for `shap_values`.

## Checklist

- [X] Unit tests added (if fixing a bug or adding a new feature


shap/utils/_general.py
```diff
@@ -161,7 +161,11 @@ def encode_array_if_needed(arr, dtype=np.float64):
     try:
         return arr.astype(dtype)
     except ValueError:
-        unique_values = np.unique(arr)
+        # np.unique() sorts the array, which will not work when arr contains both strings and NaN.
+        if arr.dtype == np.object_:
```
cedricdonie: This could also be solved by a `try` ... `except TypeError`. Apart from this, I am not sure whether it makes sense to do the encoding on an individual numpy array level. If SHAP were to use the `pd.Series.cat.codes` after receiving a DataFrame input, the encoding could be more easily reversed by the library user for subsequent analysis.

We need to use a hash-based method such as `set` because not all object instances are comparable. However the set order may be non-determinstic or differ between interpretations (https://stackoverflow.com/a/51578541). This may be an issue because encoding can then differ, again suggesting that `.cat.codes` may be better. Instead of `.cat.codes`, it might be possible to define a custom unique-values scheme such as:
* Convert everything to string and then sort (risk of collapsing `1.0` and `"1.0"` into the same class erroneously).
* Split arrays by dtype and sort the single-dtype sub-arrays separately, merging again afterwards
* Using an existing encoder from a package
* Using a custom hash function

tests/explainers/test_tree.py
```diff
@@ -1171,6 +1171,10 @@ def test_xgboost_dmatrix_propagation(self, Clf):
         shap_values = explainer.shap_values(X_nan)
         # check that SHAP values sum to model output
         assert np.allclose(margin, explainer.expected_value + shap_values.sum(axis=1))
+        interaction_values = explainer.shap_interaction_values(X_nan)
```
CloseChoice, line 1339: are your code changes tested by this? I set a breakpoint and this test doesn't run into `encode_array_if_needed`

shap/utils/_general.py
CloseChoice: Thanks for your investigation and your thoughts on that. I agree that using proper categories would be beneficial, not just here but in general. Unfortunately we don't tackle that consistently across shap. So I think we would first figure out all the different ways how categorical values are handled, think of a unifying way and then implement it across the code base.
As far as I see this is only called in `potential_interactions` which is not used at anywhere and just exported and in the scatter/dependence_legacy plot. Therefore we won't ever hand over pd.DataFrames here. While I agree that we could improve that, I would first test these lines properly by adding a scatter plot test with mixed types in the feature names. 

I think we would need to somehow handle categorical values correctly in the explanation class and then use instances of that in the plots.

CloseChoice (changes requested):
Thanks for the PR. 
I think your changes are an improvement here, but am unsure whether these lines are covered by tests. I approved the CI to run, so let's see what the coverage check says, if they are covered, pls ping me again, otherwise pls implement and ping.

shap/utils/_general.py
cedricdonie: Thanks for the detailed reply and for maintaining this awesome package!

> I agree that using proper categories would be beneficial, not just here but in general. Unfortunately we don't tackle that consistently across shap. So I think we would first figure out all the different ways how categorical values are handled, think of a unifying way and then implement it across the code base.
> [...]
> I think we would need to somehow handle categorical values correctly in the explanation class and then use instances of that in the plots.

I agree that consistently handling categories would be great and also make SHAP even more user-friendly. Unfortunately, for now I won't have time for such a far-reaching change. Thus, I reduced the scope of this pull request to only fix the interaction calculation for mixed string/float values.

The revised PR is now fully tested to the best of my knowledge.
```
pytest -vvv tests/explainers/test_tree.py::TestExplainerXGBoost::test_xgboost_dmatrix_propagation # passes, as expected
git checkout 97e3a8eba42bdfc885a4259d8c4ef6b3f7c6d2e7 -- shap/explainers/_tree.py # merge-base with master
pytest -vvv tests/explainers/test_tree.py::TestExplainerXGBoost::test_xgboost_dmatrix_propagation # fails, as expected

File: tests/explainers/test_tree.py (hunk 2 above)
Comment (cedricdonie) [reply tests/explainers/test_tree.py:1339]: Hi @CloseChoice, as described in my comment above, I have reduced the scope of this PR. However, I now believe that it is fully tested.

Some tests are still failing on my machine, but they also fail with the latest master. I use an RTX5090, which might be buggy; I had to add `--extra-index-url https://download.pytorch.org/whl/cu128` to `pip install` and needed to build the latest tensorflow master from source to avoid CUDA errors.
```
FAILED tests/explainers/test_gpu_tree.py::test_gpu_tree_explainer_shap[interventional-xgboost.sklearn.XGBClassifier1] - ValueError: operands could not be broadcast together with shapes (100,4,5) (5,100,4)
FAILED tests/explainers/test_gpu_tree.py::test_gpu_tree_explainer_shap[interventional-lightgbm.sklearn.LGBMClassifier1] - ValueError: operands could not be broadcast together with shapes (100,4,5) (5,100,4)
FAILED tests/explainers/test_gpu_tree.py::test_gpu_tree_explainer_shap[interventional-sklearn.ensemble._forest.RandomForestClassifier0] - ValueError: operands could not be broadcast together with shapes (100,4,2) (2,100,4)
FAILED tests/explainers/test_gpu_tree.py::test_gpu_tree_explainer_shap[interventional-sklearn.ensemble._forest.RandomForestClassifier1] - ValueError: operands could not be broadcast together with shapes (100,4,5) (5,100,4)
FAILED tests/explainers/test_gpu_tree.py::test_gpu_tree_explainer_shap[tree_path_dependent-xgboost.sklearn.XGBClassifier1] - ValueError: operands could not be broadcast together with shapes (100,4,5) (5,100,4)
FAILED tests/explainers/test_gpu_tree.py::test_gpu_tree_explainer_shap[tree_path_dependent-lightgbm.sklearn.LGBMClassifier1] - ValueError: operands could not be broadcast together with shapes (100,4,5) (5,100,4)
FAILED tests/explainers/test_gpu_tree.py::test_gpu_tree_explainer_shap[tree_path_dependent-sklearn.ensemble._forest.RandomForestClassifier0] - ValueError: operands could not be broadcast together with shapes (100,4,2) (2,100,4)
FAILED tests/explainers/test_gpu_tree.py::test_gpu_tree_explainer_shap[tree_path_dependent-sklearn.ensemble._forest.RandomForestClassifier1] - ValueError: operands could not be broadcast together with shapes (100,4,5) (5,100,4)
FAILED tests/explainers/test_gpu_tree.py::test_gpu_tree_explainer_shap_interactions[tree_path_dependent-xgboost.sklearn.XGBClassifier1] - ValueError: operands could not be broadcast together with shapes (4,100) (5,)
FAILED tests/explainers/test_gpu_tree.py::test_gpu_tree_explainer_shap_interactions[tree_path_dependent-lightgbm.sklearn.LGBMClassifier1] - ValueError: operands could not be broadcast together with shapes (4,100) (5,)
FAILED tests/explainers/test_gpu_tree.py::test_gpu_tree_explainer_shap_interactions[tree_path_dependent-sklearn.ensemble._forest.RandomForestClassifier0] - ValueError: operands could not be broadcast together with shapes (4,100) (2,)
FAILED tests/explainers/test_gpu_tree.py::test_gpu_tree_explainer_shap_interactions[tree_path_dependent-sklearn.ensemble._forest.RandomForestClassifier1] - ValueError: operands could not be broadcast together with shapes (4,100) (5,)
FAILED tests/plots/test_force_js.py::test_force_js_visual - AssertionError: Screenshot dimensions don't match baseline
FAILED tests/plots/test_force_js.py::test_force_array_js_visual - AssertionError: Screenshot dimensions don't match baseline
```

cedricdonie:
Hello @CloseChoice, could you please take another look at the recent updates? Thanks!

CloseChoice (approved):
Thanks for the fix, we'll merge once the pipelines are green (except for macos which will fail due to http errors)

CloseChoice (changes requested):
I don't quite see how this change could affect any xgboost tests, but some tests in the pipelines seem to fail here, could be hard to reproduce due to the very same tests are passing on different python versions/OSs. We need to take another look here.

tudstudent:
Hi @cedricdonie, thanks for this fix!

Our PR #4263 already includes the same `_xgb_dmatrix_props` propagation fix for `shap_interaction_values()`, along with a broader set of changes (interventional interaction values algorithm, categorical split support, OpenMP parallelization, and several other bug fixes).

We've also adopted your `test_xgboost_mixed_category_and_nan` test (adapted for XGBoost 2.x+ compatibility with integer category dtypes) and added interaction value checks to the existing `test_xgboost_dmatrix_propagation` test, giving credit to your PR in the test docstring.

This PR may be superseded by #4263 if it's merged first.

PR #4092, opened 2025-05-21
Update _bar.py

kush2803:
The original code was reading the first feature name from the list, and appending ordinal string to it.  The revised one appends ordinal string only if feature_name contains 1 element. Otherwise, it assigns the feature_name from the list.

## Overview

Closes #XXXX  

## Checklist


shap/plots/_bar.py
```diff
@@ -136,8 +136,12 @@ def bar(
         show_data = len(transforms) == 0
 
     # TODO: Rather than just show the "1st token", "2nd token", etc. it would be better to show the "Instance 0's 1st but", etc
-    if issubclass(type(feature_names), str):
+    if len(feature_names) == 1:
```
CloseChoice, line 139: this wouldn't match since if we hand over `feature_names="dummy"` this does not have length 1

shap/plots/_bar.py
```diff
@@ -136,8 +136,12 @@ def bar(
...
         feature_names = [ordinal_str(i) + " " + feature_names for i in range(len(values[0]))]
+    elif len(feature_names) != len(values[0]):
```
CloseChoice, line 141: I agree that this would make sense if the first check passes, but I would consider it more naturally to first check all the things that should pass and raise the error in the final `else` clause.

CloseChoice (changes requested):

CloseChoice:
pls fix the tests, too

CloseChoice:
Thanks for your efforts and this PR, we appreciate it. Due to lack of activity I'll close this for now. Feel free to come back, reopen and finish this whenever it suits you

PR #4111, opened 2025-06-12
WIP: basic lazy loading without stubs

CloseChoice:
## Overview

Implements lazy loading.
As of now, no stubs created, so package resolution does not work properly.

Closes #XXXX  

Open todos:
 - [ ] check if behaviour without matplotlib is still the same (so no import errors, only a warning if one tries to use plots)
 - [x] check if legacy plots are imported correctly
 - [x] fix ImportError when using gpu_tree
 - [ ] check how to distribute the stubs

## Checklist


CloseChoice:
closing for now. This is part of the scientific python spec, but didnt show performance enhancements and makes the code more complicated and type checking harder.

PR #4113, opened 2025-06-14
FIX: plots.image show labels for every row if provided

### LINKED ISSUE #4112: BUG: shap.plots.image no longer displays labels for rows other than the first one
### Issue Description

PR #3636 broke the display of labels on the second and subsequent rows in case:
- labels for each row are provided,
- no labels are provided but inferred from the Explainer object

Running the [Explain ResNet50 using the Partition explainer](https://shap.readthedocs.io/en/latest/example_notebooks/image_examples/image_classification/Explain%20ResNet50%20using%20the%20Partition%20explainer.html) example now no longer displays the labels on the second and subsequent rows.

```python
# Almost straight from the example, unless marked otherwise

import json
from tensorflow.keras.applications.resnet50 import ResNet50, preprocess_input
import shap

model = ResNet50(weights="imagenet")
X, y = shap.datasets.imagenet50() 
X = X.astype(int) # Different from example to fix that actual images not showing correctly (unrelated issue)

url = "https://s3.amazonaws.com/deep-learning-models/image-models/imagenet_class_index.json"
with open(shap.datasets.cache(url)) as file:
    class_names = [v[1] for v in json.load(file).values()]

def f(x):
    tmp = x.copy()
    preprocess_input(tmp)
    return model(tmp)

masker = shap.maskers.Image("inpaint_telea", X[0].shape)
explainer = shap.Explainer(f, masker, output_names=class_names)

# Reduced number of evals for speed
# Reduced number of outputs shown to 3 (to have 2x3 instead of 2x4 so that one dimension is not a multiple of the other dimension)
shap_values = explainer(X[1:3], max_evals=20, batch_size=50, outputs=shap.Explanation.argsort.flip[:3])

# No labels on second row when running this:
shap.image_plot(shap_values)

# Same result when labels are provided
shap.image_plot(shap_values, labels=['a','b','c','d','e','f'])
shap.image_plot(shap_values, labels=[['a','b','c'],['d','e','f']])
```

### Traceback

```shell

```

### Expected Behavior

Labels should be displayed on the second row (and subsequent rows if more examples).

### Bug report checklist

### Installed Versions

```
shap==0.48.0
shap-0.48.1.dev2 (master)
```

julesvanrie:
## Overview

Closes #4112

`plots.image()` currently only displays labels for the first row, even if labels for every row (every image in the sample) have been provided. This breaks the use case in the [Explain ResNet50 using the Partition explainer](https://shap.readthedocs.io/en/latest/example_notebooks/image_examples/image_classification/Explain%20ResNet50%20using%20the%20Partition%20explainer.html) example. The use case is where the classes shown in the plot are not the same for every row (e.g. when one wants to only display the `top_k` classes, which would be different for every row).

This PR fixes this problem so that `plots.image()` displays:
- Labels for every row if they are provided.
- Only labels for the first row if only set of labels is provided.

This PR also adds a test case checking that `plots.image()` does not error when labels are provided in different formats that should be allowed to work.

## Checklist


tests/plots/test_image.py
```diff
@@ -89,3 +89,49 @@ def __init__(self, data, values, output_names):
 
     shap_values_test = MockImageExplanation(test_data, test_values, test_output_names)
     shap.plots.image_to_text(shap_values_test)
+
+
+def test_image_plot_with_labels():
```
CloseChoice: Thanks for the test, this looks great. Would it be possible to set the seed for numpy and then use pytest-mpl to compare against the image?

Here is a blueprint: [test](https://github.com/shap/shap/blob/master/tests/plots/test_force.py#L94) and [baseline image](https://github.com/shap/shap/blob/master/tests/plots/baseline/test_force_plot_negative_sign.png).

CloseChoice (changes requested):
Thanks for the fix, that looks great. Would be great if you could update the test, then we're good to merge.

CloseChoice:
nevermind about the test failures, those are just connection issues when we try to download the dataset

tests/plots/test_image.py
julesvanrie: Thanks for the review and the suggestion.

I have actually rewritten the tests to largely reuse the existing logic and example explanation of `test_image_multi` (by moving it into a fixture) and compare against baselines images that way.

CloseChoice (approved):
looks great, ready to merge when pipelines succeed! Thanks a lot, awesome work

Issue #4108, opened 2025-06-11
ENH: CoalitionExplainer efficiency


CousinThrockmorton:
### Problem Description

Hi,

The CoalitionExplainer is now in the SHAP library, yay. But, it is not very efficient, that is, it is terribly inefficient compared to the theoretical promise. I plan to implement algorithm 5.4 from ["Values for level structures with polynomial-time algorithms,
relevant coalition functions, and general considerations" Besner (2022)](https://www.sciencedirect.com/science/article/abs/pii/S0166218X21004637?fr=RR-1&ref=cra_js_challenge) as a more efficient top down computation. 

Additionally, this would theoretically make it possible to provide temporary results of the larger coalitions attribution for visualisation that will get broken down to individual feature attributions as the code runs. 

### Alternative Solutions

Leave the existing functionality as is until requested. 

### Feature request checklist

- [x] I have checked the issue tracker for duplicate issues.
- [x] I'd be interested in making a PR to implement this feature

CloseChoice:
Thanks for looking into this. I also encountered that the coalition test is one of the longest running tests we have. Any upgrade here is highly appreciated.

IMW112:
@CousinThrockmorton, I am very impressed with the new CoalitionExplainer. However, I am running into out-of-memory errors when trying to explain a model with ~150 features in total, organised into a three-level deep a-priori hierarchy. Would any of the potential optimisations you have in mind reduce memory usage? Or is this just an unavoidable complication due to a vast number of possible permutations when working with many features and a relatively complex coalition structure?

PR #4116, opened 2025-06-21
ENH: Improve Coalition Explainer User Guidance and Fix Tree Building

EnzoFanAccount:
## Overview

This PR proposes making `CoalitionExplainer` as user-friendly as `PartitionExplainer` by automatically generating the `partition_tree` from the masker's clustering information if one isn't provided by the user. Users familiar with `PartitionExplainer` will find `CoalitionExplainer` works as they expect. This also fixes the `create_partition_hierarchy` helper function, replacing it with a correct implementation that produces a `partition_tree` dictionary compatible with the explainer's `_build_tree` method.

## Checklist


CloseChoice:
@CousinThrockmorton: may I ask for your opinion here?

CousinThrockmorton:
Hi @EnzoFanAccount,

Thank you for the contribution and checking out my code. Would love to hear your thoughts. 

There are two reasons why automatic partition generation is not supported in the coalition explainer:
1. First, conceptually. The `CoalitionExplainer` is built to add support for custom and non-binary partitions to SHAP, which are the mathematical grounding of the repeated Owen/Winter value based extension/approximation of the Shapley values. That is, it is made for the case where users would like to explicitly specify a partition of the feature space based on domain knowledge. Hence, the automatic creation of the partition is not the goal. It would be great to add at some point support to specify the top coalitions via domain knowledge and use scipy clustering to automatically create a binary partition within the sub-coalitions. 

2. Second, computationally. The `PartitionExplainer` already provides a great implementation calculating the Winter values using partitions created via scipy's binary correlation results efficiently. The current `CoalitionExplainer` implementation is not as efficient as the `PartitionExplainer`. First, cause it needs more work optimising the implementation see [Issue 4108](https://github.com/shap/shap/issues/4108), second, the binary partition creates simplification to the computation, that can't be exploited for non-binary partitions in the `CoalitionExplainer`.

Maybe the confusion is caused by using `partition_tree` variable naming in the `CoalitionExplainer` as well, since it is mathematically the same concept?? When it is a different attribute between the `CoalitionExplainer` (custom dictionary) and `PartitionExplainer` (scipy clustering).

Not sure about the `create_partition_hierarchy` helper function, it worked fine for the tests comparing with the `PartitionExplainer` results in the tests. [@CloseChoice Comparing with the `ExactExplainer` and `PartitionExplainer` values also might be the reason for the slow tests.] That is the main purpose of the `create_partition_hierarchy` in my mind. I suppose for large clustering memoization might be more efficient. But, again, it is not a great idea to pass an auto generated binary partition to the `CoalitionExplainer`, when the `PartitionExplainer` provides a tailored solution for calculating the same result.

That is my two cents on this PR. Thank you for the work and as always would love to hear your thoughts on the `CoalitionExplainer`.

Ayanokojinono:
Exq

EnzoFanAccount:
> Hi @EnzoFanAccount,
> 
> Thank you for the contribution and checking out my code. Would love to hear your thoughts.
> 
> There are two reasons why automatic partition generation is not supported in the coalition explainer:
> 
> 1. First, conceptually. The `CoalitionExplainer` is built to add support for custom and non-binary partitions to SHAP, which are the mathematical grounding of the repeated Owen/Winter value based extension/approximation of the Shapley values. That is, it is made for the case where users would like to explicitly specify a partition of the feature space based on domain knowledge. Hence, the automatic creation of the partition is not the goal. It would be great to add at some point support to specify the top coalitions via domain knowledge and use scipy clustering to automatically create a binary partition within the sub-coalitions.
> 2. Second, computationally. The `PartitionExplainer` already provides a great implementation calculating the Winter values using partitions created via scipy's binary correlation results efficiently. The current `CoalitionExplainer` implementation is not as efficient as the `PartitionExplainer`. First, cause it needs more work optimising the implementation see [Issue 4108](https://github.com/shap/shap/issues/4108), second, the binary partition creates simplification to the computation, that can't be exploited for non-binary partitions in the `CoalitionExplainer`.
> 
> Maybe the confusion is caused by using `partition_tree` variable naming in the `CoalitionExplainer` as well, since it is mathematically the same concept?? When it is a different attribute between the `CoalitionExplainer` (custom dictionary) and `PartitionExplainer` (scipy clustering).
> 
> Not sure about the `create_partition_hierarchy` helper function, it worked fine for the tests comparing with the `PartitionExplainer` results in the tests. [@CloseChoice Comparing with the `ExactExplainer` and `PartitionExplainer` values also might be the reason for the slow tests.] That is the main purpose of the `create_partition_hierarchy` in my mind. I suppose for large clustering memoization might be more efficient. But, again, it is not a great idea to pass an auto generated binary partition to the `CoalitionExplainer`, when the `PartitionExplainer` provides a tailored solution for calculating the same result.
> 
> That is my two cents on this PR. Thank you for the work and as always would love to hear your thoughts on the `CoalitionExplainer`.

Hi @CousinThrockmorton,

Thank you for taking your time to check out my PR. The context you provided is incredibly helpful, and I appreciate you taking the time to explain the design philosophy behind the two explainers. Specially, thank you for shining a light on this: `PartitionExplainer` is the optimized path for auto-generated binary hierarchies from clustering, while `CoalitionExplainer` is intentionally designed for custom, user-defined partitions (which can be non-binary).

I really didn't know that was the intentional design! My initial goal was to reduce friction for the user, but, looking at it now, I see that my proposed change kinda blurs this important line and could end up guiding users towards a computationally less efficient path.

If you're down I could pivot this PR to focus on clarifying this distinction in the API to prevent user confusion and guide them to the right tool. This could involve making the `partition_tree` argument strictly required, where a `ValueError` pointing that `CoalitionExplainer` is for custom hierarchies and that `shap.PartitionExplainer` should be used for automatic clustering would be raised in the case `partition_tree` is `None`.

I think I'm still down to keep the fix for `create_partition_hierarchy`. While I understand its main purpose is for the test suite, the recursive implementation in this PR is more robust against potential stack overflows on very deep clusterings and correctly handles the dictionary structure required by `_build_tree`, making the tests that rely on it more stable.

Of course, I don't think this change is strictly necessary, I just think it improves user experience by a bit. Let me know what you think! In any case, thank you for the feedback and for your work on SHAP, it's an amazing library.

CousinThrockmorton:
Sounds dope! It would be very nice to clarify the distinction and it is clearly needed. I agree with the proposed changes, you could probably implement it as an error message to point people to the `PartitionExplainer`. 

I am fine with keeping the fix on the `create_partition_hierarchy`. Just make sure the current test for `test_tabular_coalition_partition_match` passes.

Thank you for the PR and pointing out the potential confusion for users and feedback! And for being proactive in fixing it! :-)

CloseChoice:
Thanks for the helpful discussion, I am down to whatever @CousinThrockmorton suggests here. So @EnzoFanAccount please go ahead and as soon as you get an approval and the tests are green we'll merge your changes.

EnzoFanAccount:
@CousinThrockmorton I've just updated the PR to reflect the approach! All tests are passing all right, no issues there. Check it out and let me know what you think. Also, I think the `CoalitionExplainer` is great, only issue being it still needs some optimization, as you have correctly pointed out. Starting next week, I should have more time and could try helping with that if you don't mind! As always, thanks a bunch for the constructive feedback and for your work in this explainer!

shap/explainers/_coalition.py
```diff
@@ -302,28 +310,31 @@ def _build_tree(d, root):
             _build_tree(value, node)
     elif isinstance(d, list):
         for item in d:
-            node = Node(item)
-            root.child.append(node)
+            if isinstance(item, dict):
+                _build_tree(item, root)
+            else:
+                node = Node(item)
```
CloseChoice, line 316: with your changes, is this still covered by tests? Is this even reachable?

CloseChoice (changes requested):

shap/explainers/_coalition.py
CloseChoice: yes it is, resolved

PR #4119, opened 2025-06-29
Add options to customize the color bar of the violin plot

FanwangM:
## Overview

The current violin plot does not support change the label font sizes, which can be useful to have. In addition, the docstring is outdated and has been fixed as well.

## Checklist


shap/plots/_violin.py
```diff
@@ -37,48 +37,64 @@ def violin(
     color_bar=True,
     plot_size="auto",
     layered_violin_max_num_bins=20,
-    class_names=None,
```
CloseChoice, line 40: why did you remove these? That would make this a breaking change which we try to avoid

CloseChoice (changes requested):
Thanks for the PR, I think adding possibility to customize the bar is totally fine.

shap/plots/_violin.py
```diff
@@ -37,48 +37,64 @@ def violin(
...
-    class_inds=None,
     color_bar_label=labels["FEATURE_VALUE"],
     cmap=colors.red_blue,
+    color_bar_label_size=12,
+    color_bar_tick_size=12,
```
CloseChoice: Can you pls set this to 11, so that we can recover the previous behaviour?

```suggestion
    color_bar_tick_size=11,
```

CloseChoice (changes requested):
Also thanks for documenting missing parameters.

I would like to keep this as backwards compatible as possible, so pls reintroduce class_names and class_inds, if there is no good reason to drop those.

shap/plots/_violin.py
FanwangM: Hi. Thanks for the code review. This argument never gets used in https://github.com/shap/shap/blob/master/shap/plots/_violin.py and I think it's safe to remove it.  @CloseChoice

FanwangM: The same is for ` class_inds`.

shap/plots/_violin.py
FanwangM: Fixed in f42ad52.

shap/plots/_violin.py
CloseChoice: Technically this is still a breaking change, since code breaks if people added something in here, though it's unlikely. We committed to communicate breaking chance 2 releases prior, so pls reintroduce this argument.

If you want, you can add a deprecation warning and warn the user that this argument is removed if it is actually used.

shap/plots/_violin.py
```diff
@@ -37,48 +37,64 @@ def violin(
     color_bar=True,
     plot_size="auto",
     layered_violin_max_num_bins=20,
-    class_names=None,
-    class_inds=None,
     color_bar_label=labels["FEATURE_VALUE"],
     cmap=colors.red_blue,
+    color_bar_label_size=11,
```
CloseChoice:
```suggestion
    color_bar_label_size=12,
```
to recover previous behaviour

CloseChoice (changes requested):

shap/plots/_violin.py
FanwangM: Fixed in [039a61e](https://github.com/shap/shap/pull/4119/commits/039a61e09f42745dfa8f5e2cef9fd4b10fd5c3e9).

FanwangM:
Hi @CloseChoice . Are there any other comments you have for this pull request? Thanks.

CloseChoice:
Hi @FanwangM, sorry for the late reply. I am back now then and willing to push this forward. Pls ping on green.

CloseChoice:
@FanwangM, thanks again for the PR, your work and the patience. The macos tests are unrelated to this PR, we're good to merge

FanwangM:
Thanks! Happy to contribute to the community.

PR #4140, opened 2025-07-24
Fixed custom axis bug for partial dependence and added tests

### LINKED ISSUE #3206: Partial Dependence Plots doesn’t work with custom ax
### Issue Description

Hi! This issue is the same as [this](https://stackoverflow.com/questions/74381596/shap-partial-dependence-subplots-issue) stack Overflow question.

I want to plot different feature pdp in different axes but doing like this it doesn’t work as expected. Only the first feature is plotted but in the last ax. The rest are like overlapping the first.

```python
import shap
import lightgbm as lgbm
import pandas as pd

# Load the "adult" dataset from shap
X, y = shap.datasets.adult()
features = X.columns

# Split into training and testing sets
df = pd.concat([X, pd.Series(y, name='y_test')], axis=1)

# Your code snippet
X = df.loc[:, features]
X = X.fillna(X.median())
y_test = df.loc[:, 'y_test'].values
model = lgbm.LGBMRegressor()
model.fit(X, y_test)

fig, axes = plt.subplots(nrows=3, ncols=3, figsize=(16, 16))
axes_dic = {
    0: axes[0, 0],
    1: axes[0, 1],
    2: axes[0, 2],
    3: axes[1, 0],
    4: axes[1, 1],
    5: axes[1, 2],
    6: axes[2, 0],
    7: axes[2, 1],
    8: axes[2, 2]
}
fig, axes = plt.subplots(nrows=3, ncols=3, figsize=(16, 16))
fig.suptitle('Partial Dependence SHAP plots', fontsize=20)
for i in range(min(len(axes_dic), len(features))):
   shap.partial_dependence_plot(
       features[i], model.predict, X, model_expected_value=True,
       feature_expected_value=True, ice=False, show=False, ax=axes_dic[i]
   )
plt.tight_layout()
plt.show()
```

### Traceback

_No response_

### Expected Behavior

_No response_

### Bug report checklist

### Installed Versions

0.42.0

SamuelePunzo:
## Overview

Closes #3206 

This pull request implements support for passing custom axes to the partial_dependence function, enabling proper visualization when using subplots.

Additionally, a new test file has been created with three tests to ensure:
- The partial_dependence function runs without crashing.
- Setting show=False correctly prevents plt.show() from being called.
- Passing custom axes to the function works without errors.

These changes address an open issue regarding subplot support and improve the plotting functionality.

## Checklist

- [✓] All [pre-commit checks](https://pre-commit.com/#install) pass.
- [✓] Unit tests added (if fixing a bug or adding a new feature)


CloseChoice (changes requested):
I don't quite understand what the PR does and the link to the SO question is no longer working. Could you pls elaborate a bit more. Also a test using pytest-mpl could help, like e.g. here: https://github.com/shap/shap/blob/master/tests/plots/test_beeswarm.py#L80?

enzozon:
@CloseChoice happy to help clarify what this PR fixes. I ran into this while using SHAP for churn model explainability, so I dug into the mechanism — and while doing so I found two things worth addressing before merge.

## What the bug actually is

`partial_dependence` accepts an `ax` kwarg but never reads it on the 1D path:

```python
if ax is None:
    fig = plt.figure()
    ax1 = plt.gca()
else:
    fig = plt.gcf()
    ax1 = plt.gca()   # `ax` is never used
```

Both branches are identical in effect. That alone explains "my `ax` is ignored", but not the specific symptom reported in #3206 — *"only the first feature is plotted, in the last ax, the rest are overlapping"*. That comes from a two-step cascade:

1. `plt.gca()` right after `plt.subplots(2, 2)` returns the **last** created subplot, which is why everything lands in the bottom-right one.
2. `ax2 = ax1.twinx()` creates a new axes **and makes it current**. So the next call's `plt.gca()` returns the previous call's twin, the call after that returns the twin of the twin, and so on.

Instrumented run on shap 0.52.0 / matplotlib 3.11.1, passing four distinct subplots:

```
plt.gca() before = subplot[3]

call 0 -> passed ax=subplot[0]  | drew on a NEW twin axes | axes in figure: 5
call 1 -> passed ax=subplot[1]  | drew on a NEW twin axes | axes in figure: 6
call 2 -> passed ax=subplot[2]  | drew on a NEW twin axes | axes in figure: 7
call 3 -> passed ax=subplot[3]  | drew on a NEW twin axes | axes in figure: 8

subplot[0]: 0 lines
subplot[1]: 0 lines
subplot[2]: 0 lines
subplot[3]: 1 line
```

The figure ends up with 8 axes where it should have 4, and 3 of the 4 requested subplots stay empty.

## The regression test doesn't catch the regression

`test_partial_dependence_subplots` asserts only:

```python
assert len(plt.get_fignums()) == 1
```

But the bug never creates an extra *figure* — it stacks extra *axes* inside the existing one. Running that test verbatim against unpatched `master`:

```
PR assertion: PASSED   <-- on buggy code
reality: 2 of 3 subplots are EMPTY
figures open: 1 | axes in figure: 6
```

So the suite would stay green if the fix were reverted. Something like this fails before and passes after:

```python
def test_partial_dependence_respects_ax():
    X = np.random.randn(30, 3)
    fig, axes = plt.subplots(1, 3)
    for i, ax in enumerate(axes):
        shap.partial_dependence_plot(i, model, X, ax=ax, show=False)
    for ax in axes:
        assert ax.lines, "each supplied ax should receive the PD curve"
```

Happy to add a `pytest-mpl` baseline on top as you suggested — though I'd keep an assertion-based test alongside it, since an image baseline won't tell a future reader *why* it broke.

## The 2D branch fix raises UnboundLocalError

```python
if ax is None:
    fig = plt.figure()
    ax = fig.add_subplot(111, projection="3d")
```

When `ax` **is** supplied, this block is skipped and `fig` is never bound — but the function ends in `return fig, ax`. With the PR applied:

```python
fig = plt.figure()
ax3d = fig.add_subplot(111, projection="3d")
shap.partial_dependence_plot((0, 1), model, X, ax=ax3d, show=False, npoints=5)
# UnboundLocalError: cannot access local variable 'fig' where it is not associated with a value
```

That turns "silently ignores your ax" into a crash on the 2D path. It needs:

```python
else:
    fig = ax.get_figure()
```

---

@SamuelePunzo are you still planning to finish this? I'm happy to open a follow-up PR with the missing `else` branch and the assertion-based tests if you'd rather not — your call, it's your patch.

PR #4141, opened 2025-07-24
docs(shap_nips.bib): update reference

Helias:
## Overview
Not sure why with the booktitle, editor and pages the bibtex is interpreted wrongly as:

image

---

**After** removing those parts it's "fixed", I do not know if there are others way of fixing it.

image

Note: this PR is also for reporting the issue, I am not sure about my fix

## Checklist


CloseChoice (approved):
Thanks for the PR and the fix.

PR #4170, opened 2025-09-29
Add VERSION type stub for version management

shinjiaaa:
- Added a `VERSION: str` variable to the `_exp_val` function file.
- This variable is intended to store versioning or metadata information for easier tracking.

---

## Checklist


CloseChoice:
What's the motivation here? I don't see why we would just add another variable that does nothing?

CloseChoice:
I'll close this for now as I don't see a concrete use case. Feel free to come back and provide more details anytime.

Issue #4184, opened 2025-10-21
BUG: `TreeExplainer` throws `ValueError: could not convert string to float: '[2.0719469E0]'` with xgboost 3.1.0


harupy:
### Issue Description

```
Traceback (most recent call last):
  File "/Users/harutaka.kawamura/Desktop/repositories/mlflow/outputs/xgboost_shap_example.py", line 28, in <module>
    explainer = shap.TreeExplainer(model)
  File "/Users/harutaka.kawamura/.cache/uv/environments-v2/xgboost-shap-example-bfc501b85b0166b9/lib/python3.10/site-packages/shap/explainers/_tree.py", line 278, in __init__
    self.model = TreeEnsemble(model, self.data, self.data_missing, model_output)
  File "/Users/harutaka.kawamura/.cache/uv/environments-v2/xgboost-shap-example-bfc501b85b0166b9/lib/python3.10/site-packages/shap/explainers/_tree.py", line 1279, in __init__
    self._set_xgboost_model_attributes(
  File "/Users/harutaka.kawamura/.cache/uv/environments-v2/xgboost-shap-example-bfc501b85b0166b9/lib/python3.10/site-packages/shap/explainers/_tree.py", line 1506, in _set_xgboost_model_attributes
    loader = XGBTreeModelLoader(self.original_model)
  File "/Users/harutaka.kawamura/.cache/uv/environments-v2/xgboost-shap-example-bfc501b85b0166b9/lib/python3.10/site-packages/shap/explainers/_tree.py", line 2104, in __init__
    self.base_score = float(learner_model_param["base_score"])
ValueError: could not convert string to float: '[2.0719469E0]'
```

### Minimal Reproducible Example

```python
# /// script
# requires-python = ">=3.10"
# dependencies = [
#     "scikit-learn",
#     "shap",
#     "xgboost",
# ]
# ///

import shap
import xgboost as xgb
from sklearn.datasets import fetch_california_housing
from sklearn.model_selection import train_test_split

# Load data
X, y = fetch_california_housing(return_X_y=True)
X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.2, random_state=42)

# Train model
model = xgb.XGBRegressor(n_estimators=100, max_depth=5, random_state=42)
model.fit(X_train, y_train)

print(f"XGBoost: {xgb.__version__}, SHAP: {shap.__version__}")
print(f"Model R²: {model.score(X_test, y_test):.4f}")

# Compute SHAP values
[...]

PR #4187, opened 2025-10-26
Improve base_score assignment

lsdxp:
Refactor base_score assignment to handle string and collection types. in xgboost 310, the base_score return an array instead of a simple float

## Overview

BUG is reported in the issue #4184  due to changes in xgboost base_score return data type.

Diff of the file in xgboost is here: [Xgboost base_score](https://github.com/dmlc/xgboost/pull/11651/files#diff-92691cadfd8852da44fa622dceb9ada7c352a0b8a62e158ee2e87a95feeab6baR85-R87
)
from:

 Line 86 float base_score{ObjFunction::DefaultBaseScore()};

to

Line 86  common::ParamArray<float, true> base_score{"base_score", ObjFunction::DefaultBaseScore()};

## Checklist

### DIFF
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -2112,7 +2112,7 @@ class XGBTreeModelLoader:
             try:
                 base_score = ast.literal_eval(base_score)
                 if not isinstance(base_score, (list, float, int, tuple, np.ndarray)):
-                    raise ValueError   
+                    raise ValueError
             except ValueError as e:
                 emsg = f"Expected the base_score to contain a list or float, received {base_score}"
                 raise ValueError(emsg) from e
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import ast
 import io
 import json
 import os
@@ -7,7 +8,6 @@ import time
 import warnings
 from typing import Any
 
-import ast
 import numpy as np
 import pandas as pd
 import scipy.sparse
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -2111,8 +2111,11 @@ class XGBTreeModelLoader:
         if isinstance(base_score, str):
             try:
                 base_score = ast.literal_eval(base_score)
-            except Exception:
-                pass
+                if not isinstance(base_score, (list, float, int, tuple, np.ndarray)):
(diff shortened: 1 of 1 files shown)


shap/explainers/_tree.py
```diff
@@ -2101,13 +2102,21 @@ def __init__(self, xgb_model) -> None:
...
 
         self.name_obj = objective["name"]
         self.name_gbm = booster["name"]
         # handle the link function.
-        base_score = float(learner_model_param["base_score"])
+        base_score = learner_model_param["base_score"]
+        if isinstance(base_score, str):
+            try:
+                base_score = ast.literal_eval(base_score)
+            except Exception:
```
CloseChoice: Could you please catch the correct exception (I believe it's a value error) and then throw an exception that is infromative for the user?

Something like: "expected the base_score to contain a list or float, received {base_score}"?

CloseChoice (changes requested):
Thanks for the PR, this looks good so far. 

Not quite sure if we'll need a test, this depends on this pipeline on the master branch: https://github.com/shap/shap/actions/runs/18836154340

If this passes successfully, we'll need one, otherwise this PR fixes the current pipeline

shap/explainers/_tree.py
lsdxp: Added the exception message and restricted the base_score to be certain types to match the message

lsdxp:
> Thanks for the PR, this looks good so far.
> 
> Not quite sure if we'll need a test, this depends on this pipeline on the master branch: https://github.com/shap/shap/actions/runs/18836154340
> 
> If this passes successfully, we'll need one, otherwise this PR fixes the current pipeline

looks tests fail except with python3.9, assume current xgboost related tests already covering this part?

CloseChoice:
> > Thanks for the PR, this looks good so far.
> > Not quite sure if we'll need a test, this depends on this pipeline on the master branch: https://github.com/shap/shap/actions/runs/18836154340
> > If this passes successfully, we'll need one, otherwise this PR fixes the current pipeline
> 
> looks tests fail except with python3.9, assume current xgboost related tests already covering this part?
the latest xgboost release probably does not support python 3.9, therefore they pass. I let the CI run, let's see if this fixes the currently failing pipelines

CloseChoice (approved):
LGTM, thanks a lot for the quick fix, great work!

PR #4192, opened 2025-10-31
Bugfix: always force transformers label2id ids to integers

evamaxfield:
## Overview

In the process of using `shap` I found that some transformer models had `label2id` dictionaries with ids as strings. I.e.:

```python
{
    "label_one": "1",
    "label_two": "2",
    "label_three": "3",
}
```

It doesn't seem like there is a requirement/enforcement on the `transformers` side for integer ids and as such, a simple cast to integer can fix the following `max() + 1` call as the addition results in the following error: `TypeError: can only concatenate str (not "int") to str`.

## Checklist

### DIFF
--- a/shap/models/_transformers_pipeline.py
+++ b/shap/models/_transformers_pipeline.py
@@ -19,6 +19,7 @@ class TransformersPipeline(Model):
 
         # self.tokenizer = self.inner_model.model.tokenizer
         self.label2id = self.inner_model.model.config.label2id
+        self.label2id = {k: int(v) for k, v in self.label2id.items()}
         self.id2label = self.inner_model.model.config.id2label
         self.output_shape = (max(self.label2id.values()) + 1,)
         if len(self.output_shape) == 1:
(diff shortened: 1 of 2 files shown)


CloseChoice:
@evamaxfield thanks for the fix. Would you be so kind a provide a test for this? We are currently using a couple of tiny models from huggingface, see e.g. [here](https://github.com/shap/shap/blob/f501dca7f5c04409c3b5a656819a4042c9f78682/tests/explainers/test_explainer.py#L37)

evamaxfield:
Hey @CloseChoice I couldn't find a text-classification or sentiment-analysis model in the hf-internal-testing org model list. Instead I found a real text-classification model that is pretty small (67M params) for testing. Hopefully the test makes sense but please let me know if there is something more I can do!

CloseChoice (approved):
Perfect, thanks for the fix and adding the test. I verified that this failed on master and this looks good and clean. Ready to merge.

CloseChoice:
> Hey @CloseChoice I couldn't find a text-classification or sentiment-analysis model in the hf-internal-testing org model list. Instead I found a real text-classification model that is pretty small (67M params) for testing. Hopefully the test makes sense but please let me know if there is something more I can do!

Great work, thanks for the effort and getting this through so quickly. Looking forward to seeing more contributions from you ;)

Issue #4146, opened 2025-08-01
BUG: shap plot text not color in highlightings in Colab


NataliaSalaberry:
### Issue Description

I'm trying to run this code on google colab from [here](https://github.com/shap/shap/blob/master/notebooks/text_examples/translation/Machine%20Translation%20Explanations.ipynb)

from shap.plots.text(shap_values) is highlightings colors, but here is how mine looks like

<img width="878" height="181" alt="Image" src="https://github.com/user-attachments/assets/68b0c947-86a6-469e-aad5-8a6d6829c7dc" />

Any idea why?

### Minimal Reproducible Example

```python
from transformers import AutoModelForSeq2SeqLM, AutoTokenizer

import shap

tokenizer = AutoTokenizer.from_pretrained("Helsinki-NLP/opus-mt-en-es")
model = AutoModelForSeq2SeqLM.from_pretrained("Helsinki-NLP/opus-mt-en-es").cuda()

data = [
    "Transformers have rapidly become the model of choice for NLP problems, replacing older recurrent neural network models"
]
explainer = shap.Explainer(model, tokenizer)

shap_values = explainer(data, fixed_context=1)
```

### Traceback

```shell

```

### Expected Behavior

<img width="611" height="173" alt="Image" src="https://github.com/user-attachments/assets/2a829c27-1085-4160-a07c-00042499053c" />

### Bug report checklist

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest release](https://github.com/shap/shap/releases) of shap.
- [x] I have confirmed this bug exists on the [master branch](https://github.com/shap/shap/blob/master/CONTRIBUTING.md#installing-from-the-master-branch) of shap.
- [x] I'd be interested in making a PR to fix this bug

### Installed Versions

shap==0.48.0

NataliaSalaberry:
Actually, the solution for me in Google Colab was:

1- change to GPUT4 (because Nvidia requeried for cuda())

2- !pip install shap==0.48.0

3- !pip install numpy==1.26.4

Finally, I got the same plot (with colors) of original notebook.

Possible: I think the problem is the last version 2.3.2 (Really, with alll versions >=2....) of numpy with shap 0.48.0

fomightez:
[...]

PR #4204, opened 2025-11-11
Make shap.plots.text compatible with numpy2.

ljw20180420:
## Overview

According to https://github.com/shap/shap/issues/4146, `shap.plots.text` has not color when used with `numpy 2.0.0+`.
I find that this is because directly convert `np.ndarray` to `tuple` by `tuple(np.ndarray)` will results in something like
```python
(np.float(xxx), np.float(xxx), np.float(xxx))`
```
I fix this by `tuple(np.ndarray.tolist())`.

## Checklist

### DIFF
--- a/shap/plots/_text.py
+++ b/shap/plots/_text.py
@@ -140,11 +140,7 @@ def text(
         for i in range(shap_values.shape[-1]):
             values, clustering = unpack_shap_explanation_contents(shap_values[:, i])
             tokens, values, group_sizes = process_shap_values(
-                shap_values[:, i].data,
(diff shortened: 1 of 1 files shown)


maciejskorski:
not fully working 🚫 

fixes the feature diagram rendering  (+), but not  in-line text rendering (-)

image

this is the expected behaviour

image

https://shap.readthedocs.io/en/latest/example_notebooks/text_examples/sentiment_analysis/Positive%20vs.%20Negative%20Sentiment%20Classification.html

PR #4249, opened 2026-01-26
Fix waterfall plot label cutoff when saving (closes #3924)

### LINKED ISSUE #3924: BUG: Labels are cut off if the plot is saved (waterfall plot for a single point)
### Issue Description

A error that could be fixed was also mentioned in #3213 and #308 . I tried to save a bunch of images with the code below. First I did not include the plt.tight_layout(). Without that line all labels on the left of the figures are cut off and I could not read the labels. 

Hopefully you could integrate that line of code to prevent such behavior. The sad thing is, that the figures are displayed correctly if one use .show() . Therefore I had to generate all figures again and that took a long time. Perhaps a run in parallel when calculation the values would be great also for the possibility of plottong all the data. 

```python
shap.plots.waterfall(shap_values[0], show=False)
    plt.tight_layout()
    plt.savefig(namepath + '_single.png')
```

### Traceback

_No response_

### Expected Behavior

_No response_

### Bug report checklist

### Installed Versions

newest one from today 2024/12/05

VARUN3WARE:
## Overview

Closes #3924 

- Add automatic tight_layout() call in waterfall() and waterfall_legacy()
- Prevents feature labels from being cut off in saved figures
- Fixes issue where plt.savefig() would just crop y-axis labels
- 8/8 waterfall plot tests are passed

## Checklist

> unit tests related to issue i was working on passed but some other failed tests occured in else code sections..


CloseChoice (changes requested):
Pls add a test that fails on master and is fixed through this PR, make sure to commit an image for this test as well. See our other image based tests for reference.

VARUN3WARE:
added the test which failed without my implied changes on the master branch 
and also commited the image in baseline as done in other plot related tests

**before:**
image

**after:**
image

tests/plots/baseline/test_waterfall_with_long_feature_names_savefig.png
CloseChoice, line 1: So I ran your test on master and it looks like this:

Image

tbh. I think this is the better choice currently. I think for long feature names we would need some resizing of the names and automatic linebreaks.

CloseChoice (changes requested):
Thanks for the PR again, and thanks for the tests, I really appreciate that.
 As mentioned I don't see this as a clear improvement yet. I believe this needs some more work and some clever strategies to handle feature names of varying sizes (or simply live with cutting off too long names, but with too long I mean way longer than the current ones).

I also think this would need more tests, since I would have probably approved hadn't I seen the messed up x-labels in your plot. But we can talk about specific tests in the next review round ;)

PR #4252, opened 2026-01-28
Fix LinearExplainer ignoring link parameter (issue #3767)

### LINKED ISSUE #3767: BUG: LinearExplainer ignores "link" argument
### Issue Description

The Shapley values returned for a logistic regression by `LinearExplainer` with or without `link=shap.links.logit` are the same, and they are at the scale of the log(odds), so as if the logit function was there. 

When the link is the identity, I'd expect the shap values to explain the probability, not the log(odds). That's what happens with other explainers, like Exact or Permutation.

```python
import shap
from sklearn.model_selection import train_test_split
from sklearn.linear_model import LogisticRegression

X, y = shap.datasets.adult()
X_train, X_test, y_train, y_test = train_test_split(
X, y, test_size=0.2, random_state=1
)

model = LogisticRegression(max_iter=100)
model.fit(X_train, y_train)
X_sub = shap.sample(X_train, 100)

# this:
ex = shap.LinearExplainer(model, X_sub)
shap_values = ex(X_test.iloc[0:5])
shap_values[0]

# gives the same result as this:
ex = shap.LinearExplainer(model, X_sub, shap.links.logit, linearize_link=False)
shap_values = ex(X_test.iloc[0:5])
shap_values[0]
```

### Traceback

_No response_

### Expected Behavior

I expected the first set up, with the identity link, which is the default, to explain the probabilities.

### Bug report checklist

### Installed Versions

'0.45.0'

VARUN3WARE:
## Overview

Closes #3767  

- Auto-detect classifiers and default to logit link
- Warn users when using identity link with classifiers

## Checklist

### DIFF
--- a/shap/explainers/_linear.py
+++ b/shap/explainers/_linear.py
@@ -106,7 +106,7 @@ class LinearExplainer(Explainer):
         self,
         model: Any,
         masker: Any,
-        link: Callable[[Any], Any] = links.identity,
+        link: Callable[[Any], Any] | Literal["auto"] = "auto",
         nsamples: int = 1000,
         feature_perturbation: None | Literal["interventional", "correlation_dependent"] = None,
         **kwargs: Any,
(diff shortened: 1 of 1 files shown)


CloseChoice (changes requested):
Can you pls explain why this actually does not use the identity function in the first place (or how it somehow does not make a difference if we hand over the logit link or the identity link function?) 

Pls also add a test that fails on master and is fixed with this PR.

PR #4271, opened 2026-02-17
Add AdaBoostClassifier support to TreeExplainer

### LINKED ISSUE #335: AdaBoost Support

tudstudent:
## Summary

- Adds `AdaBoostClassifier` (SAMME) support to `TreeExplainer`, a feature requested for 5+ years
- Implements correct mathematical decomposition: weighted hard votes → decision function → softmax probabilities
- New C++ `softmax` vector transform with numerically stable implementation (max-subtraction, `1/(K-1)` scaling per Zhu et al. 2009)
- Binary case (K=2) reduces to existing logistic transform; multiclass (K>2) uses new softmax path
- Supports both `model_output="raw"` (decision function) and `model_output="predict_proba"` (probabilities)

### Why a new transform?

Unlike gradient boosting models where each tree's output is an additive margin score that can be transformed element-wise (logistic, squared loss), AdaBoost's `predict_proba` applies **softmax across all K classes simultaneously**. The existing per-element transform infrastructure cannot handle this, so we add a vector-valued softmax path in `dense_independent()` that precomputes full margin/softmax vectors and uses per-class rescaling for exact SHAP additivity.

### Changes

| File | Change |
|------|--------|
| `shap/cext/tree_shap.h` | `softmax_vector()`, `MODEL_TRANSFORM::softmax`, vector softmax path in `dense_independent()` and `dense_tree_predict()` |
| `shap/explainers/_tree.py` | `AdaBoostClassifier` detection block, `"softmax"` transform code, `get_transform()` update |
| `tests/explainers/test_tree.py` | 4 new tests: binary/multiclass × raw/probability — all verify SHAP additivity |

## Test plan

- [x] `test_adaboost_binary_raw` — SHAP values sum to `decision_function()` (tol 1e-4)
- [x] `test_adaboost_binary_proba` — SHAP values sum to `predict_proba()` per class (tol 1e-3)
- [x] `test_adaboost_multiclass_raw` — 3-class iris, SHAP values sum to `decision_function()` per class (tol 1e-4)
- [x] `test_adaboost_multiclass_proba` — 3-class iris, SHAP values sum to `predict_proba()` per class (tol 1e-3)
- [x] Full test suite: 115 passed, 1 skipped, 0 regressions (only pre-existing `test_overflow_tree_path_dependent` fails due to 16GB RAM requirement)

Closes #335
Closes #1219
Supersedes #1546

### DIFF
--- a/shap/cext/tree_shap.h
+++ b/shap/cext/tree_shap.h
@@ -152,11 +152,26 @@ inline tfloat squared_loss_transform(const tfloat margin, const tfloat y) {
     return (margin - y) * (margin - y);
 }
 
+inline void softmax_vector(const tfloat *d, tfloat *out, unsigned num_outputs) {
+    const tfloat scale = 1.0 / (num_outputs - 1);
+    tfloat max_val = d[0] * scale;
+    for (unsigned k = 1; k < num_outputs; ++k) {
+        if (d[k] * scale > max_val) max_val = d[k] * scale;
+    }
+    tfloat sum_exp = 0;
+    for (unsigned k = 0; k < num_outputs; ++k) {
+        out[k] = exp(d[k] * scale - max_val);
+        sum_exp += out[k];
+    }
+    for (unsigned k = 0; k < num_outputs; ++k) out[k] /= sum_exp;
+}
+
 namespace MODEL_TRANSFORM {
     const unsigned identity = 0;
     const unsigned logistic = 1;
     const unsigned logistic_nlogloss = 2;
     const unsigned squared_loss = 3;
+    const unsigned softmax = 4;
 }
 
 inline transform_f get_transform(unsigned model_transform) {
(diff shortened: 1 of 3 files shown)


CloseChoice (commented):
Hmm, so Adaboost has DecisionTrees as default weak learners, but it could be used with others as well. You capture that in the Python `_tree.py` changes, but this looks like a lot of changes with just the benefit of supporting a special (though the default) case for adaboost. I am a bit unsure and tend to simply not supporting this.

tudstudent:
Thanks for the review. A few points to consider:

- AdaBoost with non-tree base learners can't use TreeExplainer by design — same as every other model in `_tree.py` (GradientBoosting, XGBoost, LightGBM all require tree learners too). The Python `assert` already guards against non-tree estimators, so this isn't a special case — it's the standard pattern.

- The C++ changes are actually quite small: `softmax_vector()` is ~15 lines, `MODEL_TRANSFORM::softmax` is 1 line, and the rest is plumbing it through `dense_independent()` and `dense_tree_predict()`. The softmax transform is generic and could serve any future model that needs a vector-valued output transform rather than per-element scalar transforms.

- The Python detection block follows the exact same dict-based `SingleTree` pattern as ngboost (lines 1496-1510) — nothing novel in the approach.

- This has been requested for 5+ years (#335, #1219) with a stale PR (#1546) that failed because it used the wrong mathematical decomposition. The demand is clearly there.

Happy to simplify anything specific if you see unnecessary complexity.

tudstudent:
@CloseChoice Wanted to send a gentle reminder how we can proceed with this PR. Would you be so kind to give direction if changes need to be made or what needs to happen to get this moved forward (hopefully into a merge?)
I will rebase when we can proceed, I am still following up, but not seen progress on this PR.

PR #4272, opened 2026-02-17
Fix path-dependent SHAP NaN with small background datasets

tudstudent:
## Summary

- Path-dependent TreeSHAP produces NaN when any node has zero background coverage, because `unwind_path()` divides by `zero_fraction`
- This blocks LightGBM multiclass models with small backgrounds and also affects sklearn trees where floating-point threshold mismatches in `tree_update_weights` cause some leaves to get zero coverage even with full training data
- **Fix**: after background weight recomputation, replace zero weights with epsilon (1e-6), giving uncovered nodes negligible probability and preventing division by zero
- Additivity holds to <1e-7 precision and values converge as background size increases

Addresses #3574

## Test plan

- [x] `test_path_dependent_small_background` — LightGBM multiclass with 50-sample background: values finite, symmetric, row-sum consistent, additive, and converging toward full-background results

### DIFF
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -2106,6 +2106,27 @@ class SingleTree:
                 data_missing,
             )
 
+            # Fix for zero-weight nodes causing NaN in path-dependent SHAP.
+            #
+            # Original issue: when a background dataset doesn't cover all
+            # leaves, the fully_defined_weighting check raises ExplainerError,
+            # blocking path-dependent mode entirely (#3574).
+            #
+            # Root cause: tree_shap_recursive's unwind_path() divides by
+            # zero_fraction. When a leaf has zero background coverage (w=0),
+            # samples routing through that subtree encounter zero_fraction=0
+            # on cold paths, producing NaN via 0/0. This also occurs with
+            # full training data as background due to floating-point
+            # threshold comparison mismatches in tree_update_weights vs the
+            # original model training.
+            #
+            # Fix: replace zero weights with epsilon (1e-6) so uncovered
+            # nodes have negligible but non-zero probability. Additivity
+            # holds to <1e-7 and values converge as background size grows.
+            zero_mask = self.node_sample_weight == 0.0
+            if zero_mask.any():
+                self.node_sample_weight[zero_mask] = 1e-6
+
         # we compute the expectations to make sure they follow the SHAP logic
         self.max_depth = _cext.compute_expectations(
             self.children_left, self.children_right, self.node_sample_weight, self.values
--- a/tests/explainers/test_tree.py
+++ b/tests/explainers/test_tree.py
@@ -2937,3 +2937,32 @@ def test_tree_explainer_random_forest_regressor():
         else:
             shap_sum = shap_values.sum(1) + explainer.expected_value
         assert np.abs(shap_sum - predictions).max() < 1e-4
+
+def test_path_dependent_small_background():
+    """Path-dependent SHAP with small background that has uncovered leaves.
+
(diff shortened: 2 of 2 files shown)


tests/explainers/test_tree.py
```diff
@@ -2937,3 +2937,46 @@ def test_tree_explainer_random_forest_regressor():
         else:
             shap_sum = shap_values.sum(1) + explainer.expected_value
         assert np.abs(shap_sum - predictions).max() < 1e-4
+
+
+def test_path_dependent_small_background():
```
CloseChoice, line 2942: To me, this does not look like it's deterministic. So I assume we cannot guarantee that the problem is fixed. Pls produce an example with deterministic training (so fix a seed) and make sure that this contains uncovered leafs. Usually hypothesis is good for this, so that you can iterate over seeds quickly and note the seeds that result in uncovered leafs

tests/explainers/test_tree.py
```diff
@@ -2937,3 +2937,46 @@ def test_tree_explainer_random_forest_regressor():
...
+    np.testing.assert_allclose(iv.sum(axis=2), sv, atol=1e-6)
+
+    # Additivity: SHAP values sum to prediction - expected_value
+    pred = explainer.model.predict(X[:5])
+    for c in range(3):
+        shap_sum = explainer.expected_value[c] + sv[:, :, c].sum(axis=1)
+        np.testing.assert_allclose(shap_sum, pred[:, c], atol=1e-6)
+
+    # Values should converge toward full-background results
+    explainer_full = shap.TreeExplainer(model, data=X, feature_perturbation="tree_path_dependent")
```
CloseChoice: also can we trim down the test to the most essential parts only pls?

shap/explainers/_tree.py
```diff
@@ -2106,6 +2106,27 @@ def extract_data(node, tree):
...
+            # full training data as background due to floating-point
+            # threshold comparison mismatches in tree_update_weights vs the
+            # original model training.
+            #
+            # Fix: replace zero weights with epsilon (1e-6) so uncovered
+            # nodes have negligible but non-zero probability. Additivity
+            # holds to <1e-7 and values converge as background size grows.
+            zero_mask = self.node_sample_weight == 0.0
+            if zero_mask.any():
+                self.node_sample_weight[zero_mask] = 1e-6
```
CloseChoice, line 2128: This looks like the right idea to fix this!

CloseChoice (changes requested):

tests/explainers/test_tree.py
tudstudent: Rewritten with deterministic training (`random_state=0`). The test now explicitly asserts that epsilon-replaced zero-weight nodes exist before checking SHAP values are finite and additive. Trimmed to essentials — dropped convergence and interaction checks.

CloseChoice: When run with `random_state=0` are there uncovered leaves?

tudstudent: Tested this morning. But if you doubt test yourself.

tests/explainers/test_tree.py
tudstudent: Yes you can if you feel its needed, feel free to make adjustments.

tests/explainers/test_tree.py
CloseChoice: Yep, this works, the test fails on master! Nice!

CloseChoice (approved):
Nice work, thanks for the fix!

CloseChoice:
Very nice @tudstudent, thanks a lot for fixing this longstanding issue!

PR #4273, opened 2026-02-17
Add categorical split support to GPU TreeSHAP

tudstudent:
## Summary

- GPU TreeSHAP (`_cext_gpu.cu`) ignores categorical splits entirely — all splits are evaluated as numerical `x <= threshold`, producing wrong SHAP values for models with categorical features
- **Fix**: extend `ShapSplitCondition` with `is_categorical` + `category_mask` fields, and add categorical branch to `EvaluateSplit()` using bitmask encoding matching the CPU `category_in_threshold()`
- Python side (`_gpu_tree.py`): remove `_xgboost_cat_unsupported()` guard, populate `threshold_types` array from tree data

**Note**: We understand there are plans to migrate the GPU backend to RAPIDS `gputreeshap`. This PR fixes bugs in the current in-tree GPU code that users depend on today. If the migration happens, this code would naturally be superseded.

## Test plan

- [x] GPU categorical test in `test_gpu_tree.py` — XGBoost model with categorical features, SHAP values match CPU TreeExplainer

### DIFF
--- a/shap/cext/_cext_gpu.cu
+++ b/shap/cext/_cext_gpu.cu
@@ -7,19 +7,35 @@ const float inf = std::numeric_limits<tfloat>::infinity();
 
 struct ShapSplitCondition {
   ShapSplitCondition() = default;
+
+  // Constructor for numeric splits
   ShapSplitCondition(tfloat feature_lower_bound, tfloat feature_upper_bound,
                      bool is_missing_branch)
       : feature_lower_bound(feature_lower_bound),
         feature_upper_bound(feature_upper_bound),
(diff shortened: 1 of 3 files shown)


CloseChoice (changes requested):
As said, this code will change soon. We'll move to the latest version of https://github.com/rapidsai/gputreeshap. So any changes on this part of the code is currently blocked!

tudstudent:
@CloseChoice 
Do you have any timeline, days, weeks, months?
In the latter why not implement, as we then at least can test and use it, as well have a "benchmark" when moving to gpytreeshap?

PR #4274, opened 2026-02-17
Implement interventional SHAP interaction values and fix categorical split handling

tudstudent:
## Summary

Implements interventional SHAP interaction values for tree models, fixes categorical split handling across CPU paths, and corrects the GPU categorical encoding for CPU/GPU parity.

### New feature: Interventional SHAP interaction values

Implements the algorithm from [Zern et al. (AAAI 2023)](https://ojs.aaai.org/index.php/AAAI/article/view/26234). Previously, `shap_interaction_values()` with `feature_perturbation="interventional"` returned all zeros silently (#1824). Now produces correct interaction values verified against brute-force Shapley computation for 3 through 6 features and tested across sklearn, LightGBM, XGBoost, and CatBoost.

### Categorical splits: uint32 bitset representation

Replaces the packed integer mask categorical representation (`1 << cat` stored in a float threshold) with a uint32 bitset buffer design. The bitset approach removes the ~31 category limit, provides explicit bounds checking for malformed inputs, and aligns with how LightGBM stores categorical splits internally.

The bitset design is adopted from PR #5079 by @stevenellis-od, which solves the same representation problem for the path-dependent code path. We apply it coherently across all CPU entry points that handle categorical thresholds: interventional SHAP values and interactions, exact path-dependent Tree SHAP, prediction, background weight updates, and approximate (Saabas) explanations. Global path-dependent mode explicitly rejects categorical splits rather than silently misrouting offset values as numeric thresholds.

Legacy packed-mask custom trees are not migrated. Categorical nodes must provide a proper `cat_bitsets` buffer or they raise a clear error.

### Bug fixes

- **`category_in_threshold` 0-based encoding**: was `1 << (cat-1)` (undefined behavior for category 0), now uses 0-based bitset lookup. GPU `CategoryInMask` in `_cext_gpu.cu` updated to match for CPU/GPU parity.
- **XGBoost `dmatrix_props` not propagated** for interaction values (#3510).
- **`tree_saabas()` ignored `threshold_types`**: categorical nodes were routed as numeric thresholds. Now bitset-aware.
- **`build_merged_tree_recursive()` dropped `threshold_types`**: merged trees lost categorical type info. Now preserved.

### Key changes

| File | Change |
|------|--------|
[...]

PR #4276, opened 2026-02-18
Add OpenMP parallelization for interventional TreeSHAP

tudstudent:
## Summary

- Parallelizes the sample loop in `dense_independent()` and `dense_independent_interactions()` with OpenMP for near-linear speedup
- Extracted `from_flag` from the `Node` struct into a per-thread `char[]` array (~1 KB/thread), keeping `node_trees` fully shared and read-only — avoids deep-copying the entire tree array (~16 MB/thread)
- Per-thread workspace vectors allocated once per `#pragma omp parallel` region; `schedule(dynamic, 1)` for load balancing
- Removed `print_progress_bar()` calls from parallel regions (requires GIL via `PySys_WriteStderr`)
- Build: `-fopenmp` on Linux/macOS, `/openmp` on Windows

**Depends on #4274** — this modifies `dense_independent_interactions()` and the `Node` struct introduced there.

## Test plan

- [x] `test_openmp_interventional_shap_values` — 1-thread vs 4-thread results match within `atol=1e-10`
- [x] `test_openmp_interventional_interaction_values` — 1-thread vs 4-thread match + symmetry check
- [x] `test_openmp_interventional_with_transform` — probability transform output matches across threads

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,6 +99,12 @@ execute_process(
 
 target_include_directories(_cext PUBLIC ${NUMPY_INCLUDE_DIR})
 
+# OpenMP for parallelized interventional TreeSHAP
+find_package(OpenMP)
+if(OpenMP_CXX_FOUND)
+  target_link_libraries(_cext PRIVATE OpenMP::OpenMP_CXX)
+endif()
+
 # Install directive for scikit-build-core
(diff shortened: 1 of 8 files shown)


tudstudent:
@CloseChoice Wanted to send a gentle reminder how we can proceed with this PR. Would you be so kind to give direction if changes need to be made or what needs to happen to get this moved forward (hopefully into a merge?)
I will rebase when we can proceed, I am still following up, but not seen progress on this PR.

PR #4279, opened 2026-02-20
add ai usage policy at appropriate place in contributing, mention contributing in readme

CloseChoice:
cc @tylerjereddy, maybe this is of interest to you. Happy to hear your thoughts on this.

Will still merge this in right away, feel free to leave comments on the closed PR

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,6 +12,7 @@
 - [Pull Requests (PRs)](#pull-requests-prs)
   - [Etiquette for creating PRs](#etiquette-for-creating-prs)
   - [Checklist for publishing PRs](#checklist-for-publishing-prs)
+  - [AI Usage Policy](#ai-usage-policy)
 - [Documentation](#documentation)
   - [Previewing changes on Pull Requests](#previewing-changes-on-pull-requests)
   - [Building the docs locally](#building-the-docs-locally)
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -202,6 +203,17 @@ please ensure:
 - All [pre-commit hooks](#code-checks-with-precommit) pass, and
 - Unit tests have been added (if your PR adds any new features or fixes a bug).
 
+### AI Usage Policy
+
+This repository accepts PRs written by LLMs in any capacity. When filing a PR, please keep the following in mind:
```
tylerjereddy, line 208: @CloseChoice interesting--the policy reads quite favorably on LLM usage up front, and then adds the copyright bullet below. The need for disclosure of AI usage seems pretty universal, and your policy has that too.

Although you don't have an item that explicitly prohibits PRs opened and communicated entirely by "agents" without a human driver, the point about needing to understand the code does implicitly exclude that I suppose.

Overall, you seem a bit more on the "pro LLM" side, which may be ok, as long the licensing issue remains under control. If I could pick one more rule to add, it would probably be to (reserve the right to) explicitly say "no" to PRs where it is obviously just the AI and not a person working. But maybe that's subjective--some of the communities I work with tend to find those interactions less rewarding, and less likely to lead to a long-term maintainer that might be onboarded to help deal with the firehose of contributions.

@InessaPawson has read even more of these than I have I think, maybe she has seen a few that have an angle similar to yours.

CloseChoice: I think that summarizes it well. I am more on the "pro LLM" side, I think used in the "right" way, they can really help productivity. I thought about adding a paragraph that when someone uses AI heavily they must make sure that there is an issue with maintainer comments that this is ready to be implemented/fixed. To limit the AI slop we get where people just think something might be good to have but is not on our roadmap.

PR #4294, opened 2026-03-13
ENH: add logging.info when background data is sub-sampled (closes #3461)

### LINKED ISSUE #3461: ENH: increase transparency of background dataset sub-sampling

Ebube22:
Closes #3461

Adds a logging.info() message when the background dataset 
is larger than max_samples and sub-sampling occurs. This 
improves transparency by informing users that their SHAP 
values are computed on a subset of the provided background 
data, and suggests how to avoid sub-sampling if desired.

Note: logging infrastructure (import logging, log = 
logging.getLogger("shap")) was already present in the file.

### DIFF
--- a/shap/maskers/_tabular.py
+++ b/shap/maskers/_tabular.py
@@ -54,7 +54,7 @@ class Tabular(Masker):
             data = np.expand_dims(data["mean"], 0)
 
         if hasattr(data, "shape") and data.shape[0] > max_samples:
-            log.info(
+            log.warning(
                 "Background dataset has %d samples but max_samples=%d. "
                 "Subsampling to %d samples for SHAP value computation. "
                 "To use all samples, set max_samples=%d when initializing the masker.",
--- a/shap/maskers/_tabular.py
+++ b/shap/maskers/_tabular.py
@@ -54,6 +54,15 @@ class Tabular(Masker):
             data = np.expand_dims(data["mean"], 0)
 
         if hasattr(data, "shape") and data.shape[0] > max_samples:
+            log.info(
+                "Background dataset has %d samples but max_samples=%d. "
+                "Subsampling to %d samples for SHAP value computation. "
+                "To use all samples, set max_samples=%d when initializing the masker.",
+                data.shape[0],
+                max_samples,
+                max_samples,
+                data.shape[0],
+            )
             data = utils.sample(data, max_samples)
 
         self.data = data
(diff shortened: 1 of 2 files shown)


CloseChoice (approved):
Thanks for this, we had this a long time coming. Finally it's there. I made some slight changes though.

Ebube22:
That's fine, very much welcome.

PR #4298, opened 2026-03-17
Fix TreeExplainer crash with pandas nullable dtypes

### LINKED ISSUE #4011: BUG: Incompatibility with pandas Int64 dtype (int64 with NAs)
### Issue Description

While using pandas dtype `Int64` which supports `NA`s (unlike numpy `int64`) I get an error in `TreeExplainer` with background datasets. The type of error displayed seems to depend on the model and the data.

In the example provided commenting line `X = X.convert_dtypes()` which converts dtypes to pandas dtypes fixes the issue. However, this does not allow for passing features that are int but also have NAs.

```python
import numpy as np
import pandas as pd
from catboost import CatBoostClassifier
from xgboost import XGBClassifier
import shap

x1 = rng.standard_normal(100_000)
x2 = rng.integers(-100,100, 100_000)
X = pd.DataFrame({"x1": x1, "x2": x2})
X = X.convert_dtypes()
# X.loc[X["x2"]>50,"x2"] = None
X.info()

y = (X.sum(axis=1)>0)*1

# model = CatBoostClassifier(verbose=False)
# model.fit(X,y)

model = XGBClassifier(verbose=False)
model.fit(X,y)

explainer = shap.TreeExplainer(model, X, feature_perturbation="interventional")
```

### Traceback

```shell
1. with `NA`s (uncommented line `X.loc[X["x2"]>50,"x2"] = None`) I get:

TypeError: Cannot cast array data from dtype('O') to dtype('float64') according to the rule 'safe'

2. for `XGBClassifier` I get:
TypeError: Cannot cast array data from dtype('O') to dtype('float64') according to the rule 'safe'

3. for `CatBoostClassifier` I get:

```

### Expected Behavior

_No response_

### Bug report checklist

### Installed Versions

0.47.0

tudstudent:
## Summary

- `TreeExplainer.shap_values()` crashes with `TypeError: float() argument must be a string or a real number, not 'NAType'` when input DataFrame uses pandas nullable dtypes (`Int64`, `Float64`) containing `pd.NA`
- Root cause: `DataFrame.values` on nullable dtypes produces object arrays, and `astype(np.float32)` can't convert `pd.NA` to float
- Fix: replace `.values` with `.to_numpy(dtype=..., na_value=np.nan)` which properly handles nullable NA during cast
- Applied in both `_validate_inputs()` and `TreeEnsemble.predict()`

Fixes #4011

## Test plan

- [x] `test_nullable_pandas_dtype` — creates DataFrame with `Int64` + `Float64` columns containing `pd.NA`, asserts the dtype precondition exists, computes SHAP values successfully
- [x] Verified old code path (`DataFrame.values.astype(np.float32)`) raises `TypeError` on the same data


CloseChoice (approved):
LGTM, thanks a lot for the PR. I checked this locally, the error appears on master and the fixes work. I run the pipelines once again, if they pass, we'll merge!

Thanks, this is a nice fix!

PR #4302, opened 2026-03-23
Fix deprecated NumPy dtype aliases in test_kernel.py

peteroyce:
## Summary
- Replaced deprecated `np.bool_` and `np.object_` aliases with built-in `bool` and `object` in `tests/explainers/test_kernel.py`
- These aliases have been deprecated since NumPy 1.20 and will be removed in a future version

## Test plan
- [ ] Run `pytest tests/explainers/test_kernel.py` to verify tests still pass
- [ ] Confirm no NumPy deprecation warnings for dtype aliases


CloseChoice (approved):
thanks, LGTM

PR #4314, opened 2026-04-02
refactor: extract TreeEnsemble handling into strategy pattern (#4212)

### LINKED ISSUE #4212: Open Discussion: Dropping support for smaller tree model libraries?
It just happened again: our pipelines crashed due to internal changes in xgboost (see [the fix here](https://github.com/shap/shap/pull/4187)) and just now once again due to changes in gpboost (see [here](https://github.com/fabsig/GPBoost/issues/174)). Things like these happen frequently and are a huge burden for maintainers. This is blocking us from releasing new versions (since our pipelines need to be fixed before we can do that) and is eating up time that is lacking to implement new features. I would love to be able to support all these models, but maybe we have come to a point where we need to discuss other measures. 

Our pipelines usually don't use pinned versions of upstream libraries. We do this intentionally to always use the latest version available, but if this version breaks our code, we first need to fix that in order to continue working on SHAP. I want to brainstorm ways forward with the community on this matter. Here are a couple potential solutions:
 1. simply drop support for models some models and provide a generic interface that calculates SHAP values for tree models if the tree structure and base values, etc. are provided in a standard format. We could introduce documentation how this can be done for various models BUT this is not good UX for users. I would also not consider dropping support for sklearn, xgboost, lightgbm, catboost but for smaller libraries llike gpboost, ngboost, etc. And this would also mean that we won't add support for any new models in the near future.
 2. we'll just pin the versions of these libraries and explicitly support only given versions of the library. This would mean that our pipelines don't crash suddenly if a new version appears, but might be more work, since we would need to update the pins regularly on user requests (e.g. for new versions of lightgbm, xgboost, catboost, sklearn, etc.)
 3. continue as we do now, potentially finding people willing to own a specific implementation.

As mentioned I would like to get your feedback on this. Which libraries do you actually use? How would you feel about these potential changes? 
And would you even be willing to help improve the library with some ownership for a given implementation?

Pls also see the discussion here and pls vote: #4213

Abhishek9639:
## Overview

Closes #4212

Refactors the monolithic `TreeEnsemble.__init__` by extracting library-specific model handling into a dedicated `_tree_handlers` package using the strategy pattern, as suggested by @fkiraly.

### What changed

- **`shap/explainers/_tree.py`**: Removed the ~535-line `if/elif` chain from `TreeEnsemble.__init__` and replaced it with a single call to `_dispatch(model, self, data, data_missing)`.

- **`shap/explainers/_tree_handlers/`** (new package):
  - `_base.py` - Abstract base class `TreeModelHandler` with `can_handle()` and `handle()` interface
  - `_sklearn.py` - scikit-learn, econml, causalml, skopt
  - `_xgboost.py` - XGBoost (Booster, XGBClassifier, etc.)
  - `_lightgbm.py` - LightGBM
  - `_catboost.py` - CatBoost
  - `_gpboost.py` - GPBoost
  - `_pyspark.py` - PySpark ML
  - `_ngboost.py` - NGBoost
  - `_other.py` - dict/list models, imblearn, pyod
  - `__init__.py` - Handler registry and `_dispatch()` entry point

### Why

The previous code had all model-specific logic in a single 500+ line if/elif block inside `TreeEnsemble.__init__`. This made it hard to maintain, test individually, or extend with new libraries. The strategy pattern isolates each library into its own module.

## Checklist

- [x] Unit tests pass (78 passed, 0 failed).


Abhishek9639:
@fkiraly or @CloseChoice, 
This PR implements the strategy pattern refactoring you suggested in #4212.

Quick summary of what's done:

- The ~535-line `if/elif` chain in `TreeEnsemble.__init__` is now replaced with a single `_dispatch(model, self, data, data_missing)` call.
- Each library gets its own handler class (e.g. `SklearnHandler`, `XGBoostHandler`) in a new `_tree_handlers/` package.
- All handlers inherit from `TreeModelHandler` which defines the `can_handle()` / `handle()` interface.
- `get_handler(model)` picks the right one from a registry no mega if/else.

All existing tests pass (78 passed, 0 failed) and pre-commit checks are clean.

Happy to make changes if anything needs adjusting.
Thanks

CloseChoice (commented):
Thanks for the PR and your effort, but I have to close this. 
A few notes:
1. Pls disclose your AI usage if any.
2. Before picking up such large issues, please reach out first, explain how you want to approach the problem and what the rationale behind that is.

If you had done 2. I could have told you what I imagine here. If we touch this code, I would a try to build this right from the ground up. I have written more about the problem [here](https://github.com/shap/shap/wiki/ESoC-2026-Project-Ideas#unify-shap-value-calculation-from-tree-based-models). But this is very much a problem to iterate on.

If you want to work on SHAP feel very welcome to do so, we have a couple good issues open, which are easy and quick to fix (e.g. start [here](https://github.com/shap/shap/issues/3036)). Once done, we can find harder problems to work on for you.

Abhishek9639:
@CloseChoice. 
Thanks for the honest feedback, I appreciate it.

1. Yes, I used AI assistance for parts of this PR. I should have disclosed that upfront, my mistake.
2. You're right I should have reached out first to discuss the approach before jumping into the implementation. I'll make sure to do that going forward.

I'll start with the smaller issues you suggested and build up from there. Thanks for keeping the door open.

CloseChoice:
@Abhishek9639 you are more than welcome here and I am also happy if you use AI, am happy to see you contributing further, thanks already. I just think large issues need a bit more communication, so that we are aligned and you can spend your time well. Looking forward to seeing more from you here.

Abhishek9639:
Thanks @CloseChoice, 
Really appreciate the encouragement. I'll make sure to discuss the approach first on larger issues. Working on a couple of smaller ones now will keep the PRs coming.
Thanks

PR #4315, opened 2026-04-02
fix: make force plot tests deterministic to fix MacOS flakiness (#4102)

### LINKED ISSUE #4102: BUG: Flaky test on MacOS
### Issue Description

The test test_random_force_plot_mpl_text_rotation_with_data is flaky on MacOS. Have a look at the pipelines: 
- https://github.com/shap/shap/actions/runs/15451787983/job/43495454593#step:5:1498
- https://github.com/shap/shap/actions/runs/15427914110/job/43419349407#step:5:1465

```python
see description
```

### Traceback

```shell

```

### Expected Behavior

_No response_

### Bug report checklist

### Installed Versions

master

Abhishek9639:
## Overview

Closes #4102

The flaky tests were caused by `RandomForestRegressor` being created without a `random_state`, leading to non-deterministic SHAP values on each run. On MacOS, this combined with different font rendering (Core Text vs FreeType) made the matplotlib force plot label positioning inconsistent.

## Changes

- Added `random_state=0` to `RandomForestRegressor` in the test fixture.
- Removed `@pytest.mark.skipif(sys.platform == "darwin")` from both affected tests.
- Removed unused `import sys`.
- Added `plt.close("all")` for proper matplotlib cleanup.

## Checklist

### DIFF
--- a/tests/plots/test_force.py
+++ b/tests/plots/test_force.py
@@ -1,4 +1,3 @@
-import sys
 from contextlib import nullcontext as does_not_raise
 
 import matplotlib.pyplot as plt
@@ -15,7 +14,7 @@ def data_explainer_shap_values():
 
     # train model
     X, y = shap.datasets.california(n_points=500)
-    model = RandomForestRegressor(n_estimators=100)
+    model = RandomForestRegressor(n_estimators=100, random_state=0)
     model.fit(X, y)
 
     # explain the model's predictions using SHAP values
@@ -64,9 +63,6 @@ def test_verify_valid_cmap(cmap, exp_ctx):
         verify_valid_cmap(cmap)
 
 
-@pytest.mark.skipif(
-    sys.platform == "darwin", reason="Since this test is flaky on MacOS, we skip it for now. See GH #4102."
-)
 def test_random_force_plot_mpl_with_data(data_explainer_shap_values):
     """Test if force plot with matplotlib works."""
     X, explainer, shap_values = data_explainer_shap_values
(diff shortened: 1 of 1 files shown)


Abhishek9639:
Greetings @CloseChoice,
This should fix the flaky MacOS tests from #4102.

The root cause was `RandomForestRegressor` being created without `random_state` in the test fixture. Without a seed, each run produced different SHAP values, and on MacOS the different font backend (Core Text) made the label layout in `draw_labels()` inconsistent.

The fix:
- `RandomForestRegressor(n_estimators=100, random_state=0)` - deterministic model.
- Removed both `skipif(sys.platform == "darwin")` - tests now run on MacOS again.
- Added `plt.close("all")` - cleanup between tests.

All pre-commit hooks pass (ruff + ruff format). 
Happy to adjust if needed.
Thanks

CloseChoice:
Hmm, I understand that fixing the seed is desirable here, but I don't understand why this would fix a MacOS-exclusive problem. I would have expected that this issue appears on all OSs. Any insights here?

Abhishek9639:
@CloseChoice,

The missing seed makes the test non-deterministic on all platforms, but it only shows up as flaky on MacOS because of how matplotlib measures text bounding boxes in `draw_labels()`:

On MacOS, font metrics (text widths/heights) are slightly different even with the Agg backend, system-level font libraries and default fonts differ from Linux. So the label positioning hits edge cases more often when SHAP values vary between runs.

With a fixed seed, the SHAP values are always the same, so the layout stays consistent regardless of small font metric differences across platforms.

That said, you're right that the flakiness could technically appear on any OS MacOS just happens to be more sensitive to it.

CloseChoice (approved):
LGTM, thanks a lot for picking this up. Let's get this in!

CloseChoice:
@Abhishek9639, thanks a lot for working on this, your effort and the additional explanation. Really appreciate it.

PR #4319, opened 2026-04-02
DOCS: Fix Python installation version from 3.11 to 3.12

kronzter:
Updated Python version in conda environment setup docs

``` pip install --editable '.[test,plots,docs]'```
Build will fail if using 3.11.
Because its a trivial fix, didn't open Issue

## Checklist

- [✓ ] All [pre-commit checks](https://pre-commit.com/#install) pass.


CloseChoice (approved):
Thanks, I missed that when upgrading

PR #4322, opened 2026-04-02
Fix NameError when with_binary=False and with_cuda=True by initializing compile_args

### LINKED ISSUE #4321: BUG: NameError when building CUDA extension with with_binary=False due to uninitialized compile_args
### Issue Description

The variable `compile_args` is defined only inside the `if with_binary:` block in `setup.py`. However, it is later referenced inside the `if with_cuda:` block regardless of whether `with_binary` is True or False.

If `with_binary=False` and `with_cuda=True`, the code attempts to execute:

compile_args.append("-fPIC")

but `compile_args` has not been defined in this execution path, resulting in a `NameError`.

This happens because `compile_args` is initialized conditionally instead of being initialized at the start of the function.

To prevent this issue, `compile_args` should be initialized before the conditional blocks so it is always defined regardless of configuration.

```python
# Simplified version of the problematic pattern

def run_setup(with_binary, with_cuda):
    if with_binary:
        compile_args = []

    if with_cuda:
        compile_args.append("-fPIC")  # NameError if with_binary=False

run_setup(with_binary=False, with_cuda=True)
```

### Traceback

```shell
  File "setup.py", line XX, in run_setup
    compile_args.append("-fPIC")
NameError: name 'compile_args' is not defined
```

### Expected Behavior

compile_args should always be initialized before being used.

Suggested fix:

def run_setup(*, with_binary, with_cuda):
    ext_modules = []
    compile_args = []  # initialize here

    if with_binary:
        if sys.platform == "zos":
            compile_args.append("-qlonglong")
        if sys.platform == "win32":
            compile_args.append("/MD")

    if with_cuda:
        compile_args.append("-fPIC")

### Bug report checklist

### Installed Versions

shap: latest version (tested on master branch)
Python: 3.11
OS:  Windows

mohityadav8:
…ng compile_args

## Overview
Fixes a bug in setup.py where `compile_args` was only initialized inside the
`if with_binary:` block but later used inside the `if with_cuda:` block.

When `with_binary=False` and `with_cuda=True`, this caused:
NameError: name 'compile_args' is not defined.

This PR initializes `compile_args` at the beginning of the function so it is
always defined regardless of configuration.

Closes #4321 
## Overview
Fixes a bug in setup.py where `compile_args` was only initialized inside the
`if with_binary:` block but later used inside the `if with_cuda:` block.

When `with_binary=False` and `with_cuda=True`, this caused:
NameError: name 'compile_args' is not defined.

This PR initializes `compile_args` at the beginning of the function so it is
always defined regardless of configuration.

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -98,8 +98,9 @@ def compile_cuda_module(host_args):
 
 def run_setup(*, with_binary, with_cuda):
     ext_modules = []
+    compile_args = []   # initialize here so it always exists
+
     if with_binary:
-        compile_args = []
         if sys.platform == "zos":
             compile_args.append("-qlonglong")
         if sys.platform == "win32":
@@ -113,6 +114,7 @@ def run_setup(*, with_binary, with_cuda):
                 extra_compile_args=compile_args,
             )
         )
+
     if with_cuda:
         try:
             cuda_home, _ = get_cuda_path()
@@ -137,7 +139,6 @@ def run_setup(*, with_binary, with_cuda):
             )
         except Exception as e:
             raise Exception("Error building cuda module: " + repr(e)) from e
-
     ext_modules.append(
         Extension("_kernel_lib", sources=["shap/explainers/_kernel_lib.pyx"], include_dirs=[np.get_include()])
     )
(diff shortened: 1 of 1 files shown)


mohityadav8:
@CloseChoice review this please

CloseChoice (approved):
Thanks good catch.

PR #4324, opened 2026-04-03
DOC: Update Iris classification notebook to use shap.plots.force API

bhrigushi:
Fixes part of #3036

Updated the Iris classification with scikit-learn notebook to use the 
current shap.plots.force() API instead of the deprecated shap.force_plot().

Changes:
- Replaced all instances of shap.force_plot() with shap.plots.force()
- Notebook runs top to bottom without errors


CloseChoice (changes requested):
We recently merged a PR regarding this notebook. Can you please resolve the merge conflicts and check if this PR adds something beyond the recent changes?

Issue #4327, opened 2026-04-03
ENH: Migrate numba functions to C++


daidahao:
### Problem Description

I noticed shap dependency on numba when developing a SHAP-based explainer in Darts https://github.com/unit8co/darts/pull/3049. Since Darts depend on shap which depends on numba, removing numba from shap could help eventually remove it from Darts as well and keep up with new Python releases.

At the same time, I noticed @CloseChoice [proposal](https://github.com/shap/shap/wiki/ESoC-2026-Project-Ideas#move-performance-critical-functions-from-numba-to-c) for moving away from numba which I am highly in favour of. I am creating this issue here as a starting point of removing numba.

EDIT (from @CloseChoice): here a bit more rationale behind this:
numba comes with llvmlite and these two packages are known to take a while until they support new Python versions, so in order to speed up our support for newer versions moving to nanobind seems like a good option. Furthermore we get rid of JIT-compile time overhead at the cost of writing in a low-level language with all its downsides. Since we are using C++ already for the TreeExplainer, staying in that ecosystem seems natural, but doing this in a more modern way facilitating nanobind seems like a good and easy option to me. This will allow us to include more dependencies easily via `CMakeLists.txt`, like https://github.com/rapidsai/gputreeshap and https://github.com/ROCm/rocgputreeshap and provides some ergonomics for numpy array handling in C++ without needs for copies.

### Concrete steps (in order of priority)

Updated on 18 Apr, 2026 for a complete TODO list:

#### Stage 1:
- [x] Stage 1: PoC and summary PR where all sub-PRs should merge into #4366
- [x] move Cython function to nanobind
- [x] Establish scikit-buiild-core as sole backend #4452
- [x] Enable tests on macOS-arm64 #4545 
- [x] Add full argument names to `cutils.cpp`.
- [x] Exclude `docs*` and `notebooks*` from sdist.
- [x] check that CI runs with these changes and wheels are built correctly
- [x] Tree logic is now built with scikit-buiild-core, see https://github.com/shap/shap/issues/4327#issuecomment-4194932514 
- [x] Add stub generation #4539 
- [x] Fix cutils import #4541 
- [x] Verify binary builds locally: https://test.pypi.org/project/shap/0.51.1.dev106/
- [ ] Update conda-forge recipe https://github.com/conda-forge/shap-feedstock/issues/108

#### Related but not in scope:
- [x] Move optional dependencies into `dependency-groups` #4577 (superseded by #4589 and moved out of nanobind scope).
[...]

PR #4328, opened 2026-04-03
ENH: Migrate `lower_credit` to C++

daidahao:
## Overview

Supports #4327.

**THIS IS ONLY A PROTOTYPE.**

I am not familiar with SHAP build pipeline, so would need your guidance on merging it with nanobind's preferred scikit-build-core backend. @CloseChoice 

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,8 +55,8 @@ nanobind_add_module(
 
   # Source code goes here
   src/bind.cpp
-  # src/partition.cpp
+  src/partition.cpp
 )
 
 # Install directive for scikit-build-core
-install(TARGETS _ext LIBRARY DESTINATION .)
+install(TARGETS _ext LIBRARY DESTINATION ${SKBUILD_PROJECT_NAME})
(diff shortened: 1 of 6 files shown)


CloseChoice:
@daidahao thanks for this work. I am currently working on an implementation for another function but am including introducing (scikit-build-core)[https://scikit-build-core.readthedocs.io/en/stable/] as additional build system. But to build and distribute this properly, we would actually need to move the Cython function + cext over to nanobind, numba could stay as is for now. So thanks again for this work, I would suggest to wait with further progress until I figured all the details out, and then merge this branch into what I am currently working on.

daidahao:
@CloseChoice Thanks for the update. Using nanobind would almost certainly require cmake and it seems scikit build backend is the preferred way to do so.

Correct me if I am wrong: My understanding is that existing cython and cext can still be built with the new scikit backend, so perhaps reimplementing them is not needed?

Happy to take a pause here. But let me know if there is anything I could help with. Thank you for maintaining such a great library!

daidahao:
Closed and superseded by #5081.

PR #4330, opened 2026-04-03
Issue #3954: used default matplotlib style for tests

### LINKED ISSUE #3954: MAINT: use "default" matplotlib style for tests

Anushreebasics:
Closes #3954

This pull request updates plot test behavior to use Matplotlib’s modern default style with pytest-mpl, so generated test images better reflect what users see in current Matplotlib versions.

Changes included:

Added mpl-default-style = "default" under pytest options in [pyproject.toml:215]
Regenerated plot baseline images in [baseline] to match the new style output.
Verified the plots test suite passes with the regenerated baselines.


Anushreebasics:
@connortann @CloseChoice

CloseChoice (approved):
Thanks for the PR, this looks great. Let's get it in once the pipelines are green, pls ping then.

CloseChoice:
I think some more image test files need to be updated

Anushreebasics:
@CloseChoice please review the changes

PR #4331, opened 2026-04-03
Draft PR for abstract summarization notebook

kronzter:
## Overview

Aims to fix ``` Abstract Summarization Explanation Demo.ipynb ``` of #3036 

Fixed 2 typos.

Below are some questions and my research into how I think they can be fixed.
**Questions** 
1. This is a gpu notebook, is there a requirement of adding a conditional for switching to cpu if no cuda found? within ```tokenizer = AutoTokenizer.from_pretrained("sshleifer/distilbart-xsum-12-6")```
```model = AutoModelForSeq2SeqLM.from_pretrained("sshleifer/distilbart-xsum-12-6").cuda()```
I think maybe not? Probably self - explanatory and highly likely that everyone running it already has a gpu or is on a cloud nb.

2. Again because this is a gpu notebook, if removed from the exclude list as mentioned in #3036, this will cause CI to fail(Timeout due to no gpu) ? How should the pr for this be created in such a case? 

3. Although I am still investigating this, it seems that the code is generating illegal style values in the html, this can potentially break the render on some browsers. and ultimately lead to incorrect renders potentially. as mentioned in #2786 
Sample output: ``` background: rgba(np.float64(30.0), np.float64(136.0), np.float64(229.0), np.float64(0.46397306397306387))```
*can be viewed by checking json source of the notebook
It appears that the entire numpy tuple string is being printed instead of the values?
the builder code appears to be within (text function) ```shap\shap\plots\_text.py```

4. One warning that appears, due to upstream library ``` WARNING:tensorflow:From [c:\Users\harsh\anaconda3\envs\shap\Lib\site-packages\tf_keras\src\losses.py:2976](file:///C:/Users/harsh/anaconda3/envs/shap/Lib/site-packages/tf_keras/src/losses.py:2976): The name tf.losses.sparse_softmax_cross_entropy is deprecated. Please use tf.compat.v1.losses.sparse_softmax_cross_entropy instead.```
Maybe this can be left as is? Without suppressing?

Besides the above, notebook runs without issues or deprecations

## Checklist

### CODE REVIEW

**Comment by kronzter** [comment]
Ping

File: scripts/run_notebooks_timeouts.py [hunk 1]
```diff
@@ -62,7 +62,6 @@
     Path("text_examples/sentiment_analysis/Keras LSTM for IMDB Sentiment Classification.ipynb"),
     Path("text_examples/sentiment_analysis/Positive vs. Negative Sentiment Classification.ipynb"),
     Path("text_examples/sentiment_analysis/Using custom functions and tokenizers.ipynb"),
```
CloseChoice, line 64: since this is gpu notebook, I would simply ignore it here. Pls comment it out and add a comment mentioning the gpu problem.

CloseChoice (changes requested):
Thanks for the PR. I think sticking to GPU here is fine, though if you want to, pls feel free to try if a cpu version of this is easily implementable (just a few lines of code, we don't want code bloat in the notebooks for these technical things) and if that is still fast enough for the CI to run. Otherwise I left a comment with a workaround.

PR #4333, opened 2026-04-03
MAINT: use "default" matplotlib style for tests

### LINKED ISSUE #3954: MAINT: use "default" matplotlib style for tests

Abhishek9639:
## Overview

Closes #3954

By default, pytest-mpl uses the "classic" matplotlib style, but modern matplotlib versions use the "default" style. This PR updates `pyproject.toml` to explicitly use "default" so that test plots better reflect what users actually see.

### Changes
- Added `mpl-default-style = "default"` in `pyproject.toml`
- Regenerated all baseline test images using matplotlib 3.10

## Checklist


Abhishek9639:
Greetings @connortann or @CloseChoice, 
Could you take a look when you get a chance? I picked this up since #3982 has been inactive for a while. Let me know if anything needs to be updated. 
All CI checks are passing now.
Thanks

CloseChoice (commented):
Thanks a lot for working on this. This looks good. We've got an approved PR for this problem already, see https://github.com/shap/shap/pull/4330, which came in first. I approved that, so if pipelines succeed I'll merge that and close this PR.

Abhishek9639:
ok

CloseChoice:
thanks for the PR. As mentioned an earlier PR addressing the same issue was merged.

PR #4336, opened 2026-04-03
Fix/remove debug print and typo in serializable

### LINKED ISSUE #4335: BUG: Debug print statement left in Deserializer.load() + typo in error message
There's a print on line 177 of shap/_serializable.py that looks like it was left in from debugging.

 It prints to stdout every time you load a saved explainer, which clutters the output.
The log.debug() call right above it already does the same thing properly, so the print can just be removed.

Also spotted a small typo on line 170 — "wsa" instead of "was" in an error message.

dhiraj-143r:
fix: #4335

Removed a leftover print() that was cluttering stdout on every model load. Also fixed a typo ("wsa" → "was") in a nearby error message.

### DIFF
--- a/shap/_serializable.py
+++ b/shap/_serializable.py
@@ -163,14 +163,14 @@ class Deserializer:
             if end_token == "END_BLOCK___":
                 return
             self._load_data_value()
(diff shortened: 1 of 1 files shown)


CloseChoice (approved):
LGTM, thanks!

PR #4337, opened 2026-04-03
Add tree path dependent approach to CatBoost SHAP value calculation

### LINKED ISSUE #4165: Request: Add tree path dependent approach to catboost shap value calculation
https://github.com/shap/shap/blob/d9b10f69baffeb55d9d1b34126bd7e189cae5116/shap/explainers/_tree.py#L603

The current implementation just uses Catboost's implementation, which always relies on reference data (either explicitly or implicitly). Requesting the ability to generate shap values using tree path (similar to implementation for xgboost).

PredictiveManish:
## Overview

Closes #4165   

Adds tree path dependent approach to CatBoost SHAP value calculation in TreeExplainer.

## Changes:
- Modified `shap/explainers/_tree.py` to pass `reference_data=None` to `CatBoost's get_feature_importance()` method, enabling tree path dependent SHAP values (no reference/background dataset required)

- Added additivity verification support using `predict()` with `prediction_type="RawFormulaVal"` to verify SHAP values sum to model output

## Checklist


shap/explainers/_tree.py
```diff
@@ -630,14 +630,18 @@ def shap_values(
...
-            elif self.model.model_type == "catboost":  # thanks to the CatBoost team for implementing this...
+            elif self.model.model_type == "catboost":
                 assert not approximate, "approximate=True is not supported for CatBoost models!"
                 assert tree_limit == -1, "tree_limit is not yet supported for CatBoost models!"
                 import catboost
 
                 if not isinstance(X, catboost.Pool):
                     X = catboost.Pool(X, cat_features=self.model.cat_feature_indices)
-                phi = self.model.original_model.get_feature_importance(data=X, fstr_type="ShapValues")
+                phi = self.model.original_model.get_feature_importance(
```
CloseChoice, line 640: is this setting the tree dependent approach as the only usable one? To me it looks like that.

CloseChoice (changes requested):
Thanks for working on this.

Pls address my question and we would need at least 2 tests making sure that this works correctly.

shap/explainers/_tree.py
```diff
@@ -630,14 +630,18 @@ def shap_values(
...
+                    data=X, fstr_type="ShapValues", reference_data=None
+                )
+                if check_additivity and self.model.model_output == "raw":
+                    model_output_vals = self.model.original_model.predict(X, prediction_type="RawFormulaVal")
```
CloseChoice, line 644: what is this for and why did you include it?

CloseChoice (commented):
The gold standard to test this would be:
 - train a CatBoost tree
 - extract its tree structure (nodes, leafs, weights, coverage, etc.)
 - create a DecisionTree with the exact same structure
 - calculate the shap values once via CatBoost internal shap calculation and once via our cext extension (just feeding both trees with the same data into `shap.TreeExplainer` does the trick) and make sure that the outputs are identical.

shap/explainers/_tree.py
PredictiveManish: There was no model_output_vals set for CatBoost, meaning additivity was never checked in the original shortcut path.

I incorrectly assumed it needed to match XGBoost/LightGBM pattern, but:

1. The general code below (lines 646-656) extracts expected_value from CatBoost's output and creates out (SHAP values)
2. The additivity check at 655-656 only runs if model_output_vals is not None
3. Since original didn't set it, additivity wasn't checked in tree_path_dependent path for CatBoost


The valid reason for my addition: I thought I was adding missing additivity check for consistency.

## Reason for irrelevancy: 
The original CatBoost shortcut also didn't have additivity checking - my code just made it consistent but also different from original behavior. Should probably remove it to keep backward compatible, or keep if we want additivity checking.

shap/explainers/_tree.py
PredictiveManish: No. The shortcut path (lines 584-588) only runs when ALL three conditions are true:

1. `self.feature_perturbation == "tree_path_dependent"`
2. `self.model.model_type != "internal"`
3. `self.data` is None (no background data provided)

If user provides background data (self.data is not None), the code falls through to the C extension path (line 660+) which uses reference data.

So:

- With `data=None:` uses tree_path_dependent (tree path dependent SHAP)
- With `data=background_set:` uses interventional approach with reference data

shap/explainers/_tree.py
PredictiveManish: Will follow what you'll say!

PredictiveManish:
> The gold standard to test this would be:
> 
> * train a CatBoost tree
> * extract its tree structure (nodes, leafs, weights, coverage, etc.)
> * create a DecisionTree with the exact same structure
> * calculate the shap values once via CatBoost internal shap calculation and once via our cext extension (just feeding both trees with the same data into `shap.TreeExplainer` does the trick) and make sure that the outputs are identical.

@CloseChoice I understand the value of gold standard test, but my change is simply adding `reference_data=None`. Extracting and reconstructing full tree is significant work. Can we merge this simple fix now and I'll open a follow-up PR with the comprehensive test? Or Will do in this itself if you suggest this.

PR #4341, opened 2026-04-03
Fix NameError by initializing compile_args before conditional blocks

### LINKED ISSUE #4237: Reenable CausalML tests
Due to the error described in https://github.com/uber/causalml/issues/859, we disabled the only causalml test. Getting this back running is important, to reliably make sure that we can explain causalml correctly.

SAJLENDRAPANDEY:
## Overview

Closes #4237

This PR fixes a NameError issue where `compile_args` was only initialized inside the `if with_binary` block but later used in the `if with_cuda` block.

When `with_binary=False` and `with_cuda=True`, this caused a crash because `compile_args` was not defined.

## Changes

- Initialized `compile_args = []` at the beginning of the `run_setup` function
- Ensured `compile_args` is always defined before use
- Prevented NameError in CUDA execution path

## Checklist

- [ ] Unit tests added (not required for this fix)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -98,8 +98,12 @@ def compile_cuda_module(host_args):
 
 def run_setup(*, with_binary, with_cuda):
     ext_modules = []
+    
+    # ✅ FIX: initialize here (IMPORTANT)
+    compile_args = []
+
+    # -------- Binary extension --------
     if with_binary:
-        compile_args = []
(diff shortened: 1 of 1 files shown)


CloseChoice:
thanks but we only had a PR addressing this, sorry. We really appreciate the effort, and I work through the PRs in order of creation.

PR #4343, opened 2026-04-04
fix: assign result of np.flipud to fix no-op in AdditiveForceArrayVisualizer

### LINKED ISSUE #4342: Bug: np.flipud(clustOrder) in AdditiveForceArrayVisualizer is a no-op
### Bug Description
In `shap/plots/_force.py`, on line 544, there is a call to `np.flipud(clustOrder)` that is not assigned back to the variable. Since `np.flipud()` returns a new array (or a view) and does not modify the original array in place, this operation is effectively a no-op.

### Code location
`shap/plots/_force.py`
Line: 544

```python
# make sure that we put the higher predictions first...just for consistency
    np.flipud(clustOrder)  # reverse

### PR DESCRIPTION
**Pull request by Mahaveerjain-18**
## Summary

Fixes a silent no-op bug in `AdditiveForceArrayVisualizer.__init__()` where `np.flipud(clustOrder)` was called but its return value was discarded.

Fixes #4342

## Problem

In `shap/plots/_force.py`, line 544:

```python
# make sure that we put the higher predictions first...just for consistency
    np.flipud(clustOrder)  # reverse  <-- BUG: result is discarded
```

`np.flipud()` returns a new array (or view)  - it does **not** modify the array in-place. Since the return value was never assigned back, `clustOrder` was never actually reversed. The stacked force plot ordering was therefore never corrected, silently ignoring the intent described in the comment.

## Fix

```python
clustOrder = np.flipud(clustOrder)  # reverse
```

One character change: assign the result back to `clustOrder`.

## Testing

The fix is a one-line assignment correction. The logic now matches the comment's stated intent: stacked force plots with higher predictions will be displayed first.


Mahaveerjain-18:
@CloseChoice This PR is ready for review

CloseChoice (changes requested):
Thanks for the PR. Could you pls add a test that shows fails on master and goes through the changed code path?

Mahaveerjain-18:
@CloseChoice  I had added the test as requested and the PR is ready for review

CloseChoice (approved):
Thanks, I verified that this fails on master. Pls ping on green pipelines.

Mahaveerjain-18:
@CloseChoice I have pushed a fix for the failing browser tests, The np.flipud correct fix inverted the visual plotting sequence ,so I just needed to regenerate and update the baseline screenshot for the selenium plot test. Could you please approve the workflow runs so we can get green pipelines?"

CloseChoice (approved):
looks good. Let's get this in

PR #4349, opened 2026-04-04
test: increase coverage for shap/_explanation.py

tarun-227:
## Summary
- Add 29 new test cases covering previously uncovered code paths in `shap/_explanation.py`
- Tests cover: Explanation init from another Explanation, display_data setter with DataFrame, binary operators between Explanations and scalars, argsort/flip/min/max properties, sum with feature grouping, percentile/sample methods, hclust method, cohorts with array labels, MetaExplanation class properties, _compute_shape edge cases, is_1d/list_wrap utilities, and property setters
- Coverage improved from **68% to 87%**

## Test plan
- [x] All 43 tests pass locally (14 existing + 29 new)
- [x] No existing tests broken
- [x] Coverage improved from 68% to 87%

Ref: #3690

**Note:** Tests were scaffolded with AI assistance and reviewed/validated locally.

### DIFF
--- a/tests/test_explanation.py
+++ b/tests/test_explanation.py
@@ -245,3 +245,407 @@ def test_cohorts_generation_with_one_feature():
     cohorts = exp.cohorts(3)
     assert isinstance(cohorts, shap.Cohorts)
     assert len(cohorts.cohorts) == 3
+
+
+def test_cohorts_with_array():
+    """Test cohorts created from an array of cohort labels."""
+    exp = shap.Explanation(
+        values=np.random.RandomState(0).randn(10, 3),
+        data=np.random.RandomState(0).randn(10, 3),
+        feature_names=list("abc"),
+    )
+    labels = np.array(["a", "a", "a", "b", "b", "b", "c", "c", "c", "c"])
+    ch = exp.cohorts(labels)
+    assert isinstance(ch, shap.Cohorts)
+    assert len(ch.cohorts) == 3
+
+
+def test_cohorts_repr():
+    """Test the __repr__ method of Cohorts."""
+    exp = shap.Explanation(
+        values=np.random.RandomState(0).randn(10, 3),
+        data=np.random.RandomState(0).randn(10, 3),
+        feature_names=list("abc"),
+    )
+    ch = exp.cohorts(np.array(["x", "x", "x", "x", "x", "y", "y", "y", "y", "y"]))
+    repr_str = repr(ch)
+    assert "Cohorts" in repr_str
+    assert "2 cohorts" in repr_str
+
+
+def test_cohorts_invalid_type():
+    """Test that cohorts raises TypeError for invalid input types."""
+    exp = shap.Explanation(
+        values=np.random.RandomState(0).randn(10, 3),
+        data=np.random.RandomState(0).randn(10, 3),
+        feature_names=list("abc"),
+    )
+    with pytest.raises(TypeError, match="not recognized"):
+        exp.cohorts("invalid")
+
+
+def test_explanation_init_from_explanation():
+    """Test initializing an Explanation from another Explanation."""
+    exp1 = shap.Explanation(
+        values=np.array([1.0, 2.0, 3.0]),
+        base_values=0.5,
+        data=np.array([10.0, 20.0, 30.0]),
+    )
+    exp2 = shap.Explanation(exp1)
+    np.testing.assert_array_equal(exp2.values, exp1.values)
+    assert exp2.base_values == exp1.base_values
+    np.testing.assert_array_equal(exp2.data, exp1.data)
+
+
+def test_explanation_display_data_setter_with_dataframe():
+    """Test that setting display_data with a DataFrame converts to values."""
+    exp = shap.Explanation(values=np.array([[1.0, 2.0], [3.0, 4.0]]))
+    df = pd.DataFrame({"a": [10.0, 20.0], "b": [30.0, 40.0]})
+    exp.display_data = df
+    assert isinstance(exp.display_data, np.ndarray)
+    np.testing.assert_array_equal(exp.display_data, df.values)
+
+
+def test_explanation_binary_operators():
+    """Test arithmetic operations between Explanation objects."""
+    vals = np.array([1.0, 2.0, 3.0])
+    data = np.array([10.0, 20.0, 30.0])
+    exp1 = shap.Explanation(values=vals.copy(), base_values=1.0, data=data.copy())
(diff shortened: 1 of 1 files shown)


naxatra2:
Are these tests AI generated? Some testcases feel redundant and also you have created like 45+ testcase across different PRs in like 2 hour time difference.

tarun-227:
Yes, I used AI tools to help scaffold some of these tests, but I reviewed and validated each one locally — all tests pass and coverage numbers are verified. Happy to trim any redundant cases if you can point out which ones. The goal was to address #3690 which asks for increased test coverage.

naxatra2:
I get your point that you have locally validated them, but take for example `test_explanation_sum_grouping_rank1` & `test_explanation_sum_with_grouping` they are very similar. Only the input type is different and that can be optimised imo. If we take more time to create these testcases then it would be better in the long run. There are also other testcases like `test_explanation_binary_operators` and `test_explanation_binary_operators_with_scalar`. 

We can use some pytest features like `pytest.mark.parameterise` to fix these issues or something similar. Otherwise its very hard to read these because of the amount of new testcases with barely different purposes. It feels like AI slop. I think it will cause issues in the longer run.

tests/test_explanation.py
```diff
@@ -245,3 +245,407 @@ def test_cohorts_generation_with_one_feature():
...
+    result = exp.mean(axis=0)
+    assert result.clustering is not None
+    assert result.clustering.shape == (1, 4)
+
+
+def test_metaclass_properties():
+    """Test that MetaExplanation class properties return OpChain objects."""
+    from shap.utils._general import OpChain
+
+    assert isinstance(shap.Explanation.abs, OpChain)
```
CloseChoice: we don't need this test case!

tests/test_explanation.py
```diff
@@ -245,3 +245,407 @@ def test_cohorts_generation_with_one_feature():
...
+    """Test that MetaExplanation __getitem__ returns OpChain."""
+    from shap.utils._general import OpChain
+
+    result = shap.Explanation[0]
+    assert isinstance(result, OpChain)
+
+
+def test_compute_shape_edge_cases():
+    """Test _compute_shape with various edge cases."""
+    from shap._explanation import _compute_shape
```
CloseChoice: pls don't test private functions directly!

tests/test_explanation.py
```diff
@@ -245,3 +245,407 @@ def test_cohorts_generation_with_one_feature():
...
+    """Test the is_1d utility."""
+    from shap._explanation import is_1d
+
+    assert is_1d(["a", "b", "c"]) is True
+    assert is_1d([np.array([1, 2]), np.array([3, 4])]) is False
+
+
+def test_list_wrap():
+    """Test list_wrap utility function."""
+    from shap._explanation import list_wrap
```
CloseChoice: I would prefer to go through the public interface for this

CloseChoice (changes requested):
@tarun-227 thanks for the PR. Could you pls trim deduplicate test cases? Furthermore I would like to have these test cases more in real world oriented fashion, so going through the `shap.<Custom>Explainer` interface. If this is not a good option in some cases, pls note them and explain why. Furthermore there needs to more done on this PR than simply addressing the specific comments, but rather thinking through which code paths are hit by which test case.

tarun-227:
Thanks for the review @CloseChoice! I've pushed an update addressing your feedback:

- **Removed `test_metaclass_properties`** — agreed, not needed
- **Removed direct tests for private functions** (`_compute_shape`, `list_wrap`, `is_1d`) — these are now exercised indirectly through the public interface
- **Added a real-world test using `TreeExplainer`** that trains a small model and runs operations (`.mean()`, `.sample()`, `.display_data` setter) on the resulting Explanation object
- **Consolidated redundant tests with `pytest.mark.parametrize`** — binary ops, reduction ops, sum grouping, argsort/flip, and sample are now parametrized instead of having separate near-identical test functions

Coverage is at 85% (down slightly from 87% due to removing private function tests, but the test quality is better). Let me know if there's anything else to adjust.

tests/test_explanation.py
```diff
@@ -245,3 +245,286 @@ def test_cohorts_generation_with_one_feature():
...
+    assert sampled.shape == (5, 8)
+
+    # display_data setter with DataFrame
+    shap_values.display_data = X[:10]
+    assert isinstance(shap_values.display_data, np.ndarray)
+
+
+def test_explanation_property_setters():
+    """Test various property setters on Explanation."""
+    exp = shap.Explanation(
```
CloseChoice: I don't see any value in this test. We basically test that the setter works correctly here.

tests/test_explanation.py
```diff
@@ -245,3 +245,286 @@ def test_cohorts_generation_with_one_feature():
...
+def test_explanation_shape_len_and_copy():
+    """Test shape, len, and __copy__ behavior."""
+    import copy
+
+    exp = shap.Explanation(
+        values=np.zeros((10, 5)),
+        base_values=0.5,
+        data=np.ones((10, 5)),
+    )
+    assert exp.shape == (10, 5)
```
CloseChoice: this check is unnecessary

CloseChoice (changes requested):
Pls address the comments. Also pls take a critical look over all tests and just trim down to the must crucial ones. I haven't checked everything thoroughly but this still looks like quite a bit of unnecessary code and trivial checks. If you don't see if something **nontrivial** is checked in here (that's not checked previously) then just remove the test.

tarun-227:
Thanks for the follow-up review @CloseChoice. Pushed another round of trimming:

- Removed `test_explanation_property_setters` entirely (agreed, it only verified setters).
- Removed `test_explanation_shape_len_and_copy` — the shape/len checks on line 437 were trivial construction assertions, and `copy.copy` on its own isn't worth a dedicated test.
- Removed standalone `test_explanation_display_data_setter_with_dataframe` — the same DataFrame→ndarray conversion is already exercised in `test_explanation_from_tree_explainer_operations`.
- Folded `test_explanation_percentile` into the `test_explanation_reduction_ops` parametrize (it's the same reduction shape, just a different method).

Net change: -73 lines, +12 lines. All remaining tests pass locally (39 passed). Let me know if anything else still feels redundant.

tarun-227:
@CloseChoice — friendly deadline ping. The latest push (Apr 17) addressed all your review points: removed trivial tests, folded duplicates into parametrize, and kept only meaningful coverage additions. The ESoC contribution window closes Apr 30 (18:00 UTC). Would really appreciate a re-review when you have a moment. Thanks!

PR #4352, opened 2026-04-04
FIX: handle missing feature_names in scatter plots

### LINKED ISSUE #4351: BUG: shap.plots.scatter crashes when Explanation has no feature_names
### Issue Description

`shap.plots.scatter()` crashes when the passed `Explanation` has no `feature_names`.

The function currently evaluates `len(shap_values.feature_names)` before it normalizes unnamed inputs, so `Explanation` objects built from plain NumPy arrays fail immediately with `TypeError: object of type 'NoneType' has no len()`. This reproduces on the latest release (`0.51.0`) and on the current `master` commit `42249a3dff148d36f28121d794953027ec1226f2`.

```python
import numpy as np
import matplotlib
matplotlib.use("Agg")
import shap

explanations = shap.Explanation(
    values=np.array([[1.0], [3.0]]),
    base_values=np.array([0.0, 0.0]),
    data=np.array([[10.0], [30.0]]),
)

shap.plots.scatter(explanations, show=False)
```

### Traceback

```shell
TypeError: object of type 'NoneType' has no len()
```

### Expected Behavior

`shap.plots.scatter()` should generate default labels when `feature_names` is missing instead of crashing. The same unnamed `Explanation` should remain plottable for both single-feature and multi-feature scatter plots.

### Bug report checklist

### Installed Versions

- `shap`: `0.51.0` and `master` (`42249a3dff148d36f28121d794953027ec1226f2`)
- `python`: `3.12.3`
- `matplotlib`: `3.10.8`
- `numpy`: `2.4.4`

officialasishkumar:
## Overview

Closes #4351

- infer default feature names in `shap.plots.scatter()` when an `Explanation` does not provide them
- preserve the inferred labels for multi-column scatter plots and for the `color=Explanation` code path
- add regression tests covering unnamed single-column, multi-column, and color-by-explanation scatter plots

## Checklist


shap/plots/_scatter.py
```diff
@@ -131,8 +131,10 @@ def scatter(
     if not isinstance(shap_values, Explanation):
         raise TypeError("The shap_values parameter must be a shap.Explanation object!")
 
+    feature_names = _get_feature_names(shap_values)
```
CloseChoice: your fallback with `feature_names2` feels complicated. Couldn't we simply do something like

```python
  if shap_values.feature_names is not None:                                                                                                                                             
      feature_names = shap_values.feature_names                                                                                                                                         
  else:                                                                                                                               
      num_features = shap_values.shape[1] if len(shap_values.shape) > 1 else 1                                                                                                          
      feature_names = [f"{labels['FEATURE'] {str(i)}" for i in range(num_features)]   
``

File: shap/plots/_scatter.py [hunk 2]
```diff
@@ -143,7 +145,10 @@ def scatter(
         _ = plt.subplots(1, len(inds), figsize=(min(6 * len(inds), 15), 5))
         for i in inds:
             ax = plt.subplot(1, len(inds), i + 1)
-            scatter(shap_values[:, i], color=color, show=False, ax=ax, ymin=ymin, ymax=ymax)
+            column_shap_values = shap_values[:, i]
+            if column_shap_values.feature_names is None:
+                column_shap_values.feature_names = str(feature_names[i])
```
CloseChoice, line 157: this is a mutation of user input, can we pls avoid this somehow?

CloseChoice (changes requested):
thanks, this is a real issue, just verified that the tests fail on master. Pls address the comments below, I *believe* there is a simpler solution but could be wrong.

shap/plots/_scatter.py
officialasishkumar: Simplified the fallback feature-name handling along those lines.

shap/plots/_scatter.py
officialasishkumar: Avoided mutating the caller-provided `Explanation` by copying the sliced value before assigning a temporary name.

PR #4353, opened 2026-04-04
FIX: assign result of np.flipud in AdditiveForceArrayVisualizer

### LINKED ISSUE #4342: Bug: np.flipud(clustOrder) in AdditiveForceArrayVisualizer is a no-op

officialasishkumar:
## Overview

Closes #4342

`np.flipud()` returns a new array and does not modify the input in place. In `AdditiveForceArrayVisualizer.__init__()`, the return value was being discarded, making the call a silent no-op. The stacked force plot therefore never reordered its rows to put higher predictions first, contrary to the intent described in the accompanying comment.

This PR assigns the result back to `clustOrder` and adds a unit test that constructs explanations with linearly-spaced effects and verifies that the explanation with the highest total effects receives `simIndex == 1` (displayed first).

## Checklist


tests/plots/test_force.py
```diff
@@ -117,3 +117,51 @@ def test_force_plot_positive_sign():
...
     )
     return plt.gcf()
+
+
+def test_force_array_higher_predictions_first():
+    """Test that AdditiveForceArrayVisualizer puts higher predictions first (GH #4342).
+
+    np.flipud(clustOrder) must be assigned back; otherwise the reorder is a no-op.
+    """
+    from shap.plots._force import AdditiveExplanation, AdditiveForceArrayVisualizer, DenseData, Instance, Model
```
CloseChoice: could we pls use the public API? Internal classes are subject to change, so this test could break

tests/plots/test_force.py
```diff
@@ -117,3 +117,51 @@ def test_force_plot_positive_sign():
...
+    n_features = 3
+    feature_names = [f"f{i}" for i in range(n_features)]
+    model = Model(None, ["f(x)"])
+
+    # Build explanations whose total effects span negative to positive.
+    # Hierarchical clustering will produce a linear leaf ordering for these
+    # well-separated, linearly-spaced effects.
+    effect_sums = [-10.0, -5.0, 0.0, 5.0, 10.0]
+    explanations = []
+    for s in effect_sums:
```
CloseChoice: this test fails on master and reproduces the issue, but it feels overcomplicated. Could we trim it down?

CloseChoice (changes requested):
thanks for this, this is a real issue. Would be great if you could address the comments then we can get this one in.

tests/plots/test_force.py
officialasishkumar: Trimmed the test down to a small two-row case that still reproduces the ordering issue.

tests/plots/test_force.py
officialasishkumar: Updated the regression test to go through the public `shap.plots.force` API.

CloseChoice:
Sorry, saw too late that this is a duplicate of https://github.com/shap/shap/pull/4343. I appreciate the effort and the test you added but will close this nonetheless. What we'll accept is a visual test for this using pytest-mpl.

On a second thought, I leave this open, your test looks simpler than #4343. This could actually be an improvement to the test once #4343 is merged.

CloseChoice:
closing as https://github.com/shap/shap/pull/4343 is merged. Thanks again for your efforts

PR #4350, opened 2026-04-04
test: restore pytest discovery for shap.actions

JRV7903:
This PR restores test coverage (meta-issue #3690 ) for the `shap.actions` module by fixing pytest discovery.
The tests already existed under `tests/actions/`, but were named `_action.py` and `_optimizer.py`, so they were not collected by default pytest patterns.
- Renamed:
  - `tests/actions/_action.py` → `tests/actions/test_action.py`
  - `tests/actions/_optimizer.py` → `tests/actions/test_optimizer.py`
- Updated module docstrings to correctly reflect their purpose
- Minor fix in `test_action.py` to avoid mutating the original dataframe row during testing (uses a copy instead)
No production code was modified. This change simply enables existing tests to run as intended and contribute to coverage.
- AI Disclosure: LLM assistance was used during codebase exploration and to help draft this PR description. All changes were verified manually and validated through local test execution.

PR #4354, opened 2026-04-04
Added remaining tests for full covering _optimizer.py

PredictiveManish:
## Overview

Working towards #3690  

This PR adds 7 new test cases to improve test coverage for `shap/actions/_optimizer.py`, addressing the low coverage (15%) identified in the testing roadmap.

## Tests Added:

1. `test_warning_emitted` - Verifies that the alpha state warning is emitted when ActionOptimizer is initialized.
2. `test_no_valid_solution` - Tests the edge case when no valid solution exists and the queue empties without finding a solution.
3. `test_action_replacement_in_group` - Tests the core logic of replacing an action with a more expensive one in the same action group.
4. `test_multiple_action_groups_with_replacement` - Tests combination scenarios with multiple action groups where actions are replaced within the same group.
5. `test_immediate_satisfaction` - Tests the edge case when the model is satisfied on the first try without needing additional actions.
6. `test_single_action_group` - Tests the case when actions contains a single Action object (not wrapped in a list).
7. `test_action_list_sorting` - Verifies that actions provided in a list are properly sorted by cost.

## Coverage Impact

These tests significantly improve coverage of `shap/actions/_optimizer.py`, particularly around:

- Warning emission on initialization  
- Queue handling and action replacement logic  
- Edge cases for solution finding  
- Action sorting and grouping behavior  

Note: This file will be renamed to `test_optimzer.py` after #4350 is merged.

## Checklist


tests/actions/_optimizer.py
```diff
@@ -93,3 +93,115 @@ def test_run_out_of_group():
 def test_bad_action():
     with pytest.raises(InvalidAction):
         shap.ActionOptimizer(None, [None])  # type: ignore
+
+
+def test_warning_emitted():
```
CloseChoice, line 98: does this test add any value? I don't really see that.

tests/actions/_optimizer.py
```diff
@@ -93,3 +93,115 @@ def test_run_out_of_group():
...
+        [IncreaseFeature2(i) for i in range(1, 5)],
+    ]
+    optimizer = shap.ActionOptimizer(sum_equals_12, possible_actions)
+    actions = optimizer(X.iloc[0])
+    assert len(actions) == 2
+
+
+def test_immediate_satisfaction():
+    """Tests when model is satisfied on first try (no actions needed)."""
+    X, IncreaseFeature1, IncreaseFeature2, IncreaseFeature3, passed = create_basic_scenario()
```
CloseChoice, line 161: why not 

`X, IncreaseFeature1, _, _, _ = ...`?

CloseChoice (changes requested):
Pls resolve the merge conflict, so that these tests are actually run!

tests/actions/_optimizer.py
PredictiveManish: yes, i think too, added previously to ensure the warning isn't accidentally removed, but it does feel like testing implementation details rather than user-facing behavior.

tests/actions/_optimizer.py
PredictiveManish: yes, missed that.

PredictiveManish:
@CloseChoice Resolved the conflict and updated tests.

PR #4326, opened 2026-04-03
improve test coverage for clustering utilities

SiddhantGahankari:
## Overview

Relates #3690 

This PR adds unit tests for `shap/utils/_clustering.py` to improve coverage and ensure stability for edge cases.
The tests cover:
- Core functionality of internal leaf ordering (`hclust_ordering`)
- Handling of `pd.DataFrame` inputs in `hclust`
- Handling and imputation of NaN inputs in `hclust` and `partition_tree`
- Verification of protective warnings for constant features in XGBoost distance metric
- Verification of protective warnings when users incorrectly pass `y` targets to unsupervised metrics

## Checklist

- [x] All [pre-commit checks](https://pre-commit.com/#install) pass.
- [x] Unit tests added (if fixing a bug or adding a new feature)

PR #4360, opened 2026-04-04
chore: exclude numba jitted clustering functions from coverage report

SiddhantGahankari:
## Overview

This PR adds advanced exclusion rules to `pyproject.toml` to ignore specific Numba JIT-decorated functions in `shap/utils/_clustering.py` from the `pytest-cov` coverage report.

As discussed in #4326, testing these lines directly and having it register on `pytest-cov` is difficult because the coverage tracer cannot inspect inside Numba decorated functions without disabling the JIT compiler (`NUMBA_DISABLE_JIT=1`). Because disabling JIT significantly slows down the CI pipeline, the maintainers agreed that explicitly excluding these functions from the coverage report is the preferred solution.

**Functions excluded:**
- `partition_tree_shuffle`
- `delta_minimization_order`

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -298,6 +298,12 @@ omit = ["shap/benchmark/*"]
 [tool.coverage.paths]
 combine = ["shap", "*/site-packages/shap"]
 
+[tool.coverage.report]
+exclude_also = [
+    "def partition_tree_shuffle",
+    "def delta_minimization_order",
+]
+
 [tool.cibuildwheel]
 # Restrict the set of builds to mirror the wheels available in scipy. See #3028
 # skip *-musllinux_aarch64 since numpy doesn't provid those wheels
--- a/tests/utils/test_clustering.py
+++ b/tests/utils/test_clustering.py
@@ -76,12 +76,11 @@ def test_partition_tree_with_nan():
     assert isinstance(result, np.ndarray)
     assert result.shape == (2, 4)
 
+
 def test_hclust_accepts_dataframe():
-    #checks for whether hclust accepts dataframe
+    # checks for whether hclust accepts dataframe
     pytest.importorskip("xgboost")
-    X = pd.DataFrame(
-        np.column_stack((np.arange(1, 10), np.arange(100, 1000, step=100)))
-    )
+    X = pd.DataFrame(np.column_stack((np.arange(1, 10), np.arange(100, 1000, step=100))))
     y = np.where(X.iloc[:, 0] > 5, 1, 0)
     result = hclust(X, y, random_state=0)
     assert isinstance(result, np.ndarray)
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -298,6 +298,12 @@ omit = ["shap/benchmark/*"]
 [tool.coverage.paths]
 combine = ["shap", "*/site-packages/shap"]
 
+[tool.coverage.report]
+exclude_also = [
+    "def partition_tree_shuffle",
```
CloseChoice, line 303: why do you exclude only two njitted functions and not all?

tests/utils/test_clustering.py
```diff
@@ -40,3 +41,83 @@ def test_hclust_errors_on_unknown_linkages():
...
+    y = np.where(X.iloc[:, 0] > 5, 1, 0)
+    result = hclust(X, y, random_state=0)
+    assert isinstance(result, np.ndarray)
+    assert result.shape == (1, 4)
+
+
+def test_xgboost_distances_r2_constant_feature_warns():
+    # Constant features should trigger a warning.
+    pytest.importorskip("xgboost")
+    from shap.utils._clustering import xgboost_distances_r2
```
CloseChoice, line 93: pls don't use internel imports for testing. If this function is not exposed otherwise, it's meant to be "private" even if it's not underscored

tests/utils/test_clustering.py
```diff
@@ -40,3 +41,83 @@ def test_hclust_errors_on_unknown_linkages():
...
+        }
+    )
+    result = partition_tree(X)
+    assert isinstance(result, np.ndarray)
+    assert result.shape == (2, 4)
+
+
+def test_hclust_accepts_dataframe():
+    # checks for whether hclust accepts dataframe
+    pytest.importorskip("xgboost")
```
CloseChoice, line 82: why this import? That's not needed in here is it?

tests/utils/test_clustering.py
```diff
@@ -40,3 +41,83 @@ def test_hclust_errors_on_unknown_linkages():
...
+            [5.0, 500.0],
+            [6.0, 600.0],
+            [7.0, 700.0],
+            [8.0, 800.0],
+            [9.0, 900.0],
+        ]
+    )
+    result = hclust(X, random_state=0)
+    assert isinstance(result, np.ndarray)
+    assert result.shape == (1, 4)
```
CloseChoice, line 63: does this test add anything? I would have thought these basic cases are already covered.

CloseChoice (changes requested):

pyproject.toml
SiddhantGahankari: I will add all of them right away.

PR #4362, opened 2026-04-05
FIX: replace copy.deepcopy with copy.copy in Explanation.__getitem__ and fix missing return

### LINKED ISSUE #4052: BUG: Very slow indexing of explanation object
### Issue Description

Have recently upgraded to 0.47.1 and am seeing very slow performance when trying to generate plots for an Explanation object generated by the TreeExplainer.

Image

Image
Shape of the explanation object is 507,479 x 537

I specifically set the display data as I want to plot against my original data and not against transformations that take place in my pipeline. Accessing e.g. feature index 4 from the shap values or display data takes couple of milliseconds 

Image

When plotting e.g. scatter plots, you have to provide a specific feature you want to plot, e.g. if I index the explanation object directly looking for all examples for the 4th feature index, this takes 3 minutes 21 seconds!

Image

If I interrupt the kernel while it's running, it seems to be copying the whole explanation object in order to return just the bit that I need for the plot:

Image

```python
Unable to reproduce with a toy dataset
```

### Traceback

```shell

```

### Expected Behavior

_No response_

### Bug report checklist

### Installed Versions

0.47.1

sohamukute:
## Overview                                                                                                                                                                                
                                                                                                                                                                                             
  Closes #4052                                                                                                                                                                               
                                                                                                                                                                                          
  `__getitem__`, `percentile`, and `group_features` were all calling `copy.deepcopy` even though only a few attributes were being modified right after. Switched to `copy.copy` which shares array references instead of copying all the data. Also found a missing `return` in the 2D feature_names path that was silently dropping the result and causing an IndexError.              
                                                                                                                                                                                             
  ## Checklist                                                                                                                                                                               
                                                                                                                                                                                             

sohamukute:
Hey @CloseChoice, can you please review the fix.

CloseChoice (approved):
Thanks for this, this looks good. Did you run any tests to check if this improves test times/uses less memory?

Can you pls ping on green?

sohamukute:
> Thanks for this, this looks good. Did you run any tests to check if this improves test times/uses less memory?
> 
> Can you pls ping on green?

@CloseChoice Yes, ran the full suite locally 401 passed, no new failures. Memory improves since copy.copy skips duplicating the numpy arrays. The 45ms to 0.06ms number is from #4363, not this PR.

sohamukute:
@CloseChoice any update on this

PR #4363, opened 2026-04-05
ENH: add fast path in Explanation.__getitem__ for standard 2D numpy arrays

### LINKED ISSUE #4052: BUG: Very slow indexing of explanation object

sohamukute:
## Overview                                                                                                                                                                                
                                                                                                                                                                                             
  Closes #4052                                                                                                                                                                               
                                                                                                                                                                                             
  `Slicer.__getitem__` walks every attribute row by row in Python even for simple slices. On a 50000 row explanation `exp[:, 4]` was taking 45ms. Added `_fast_slice` which indexes the numpy arrays directly for the common case and skips Slicer entirely, bringing that down to 0.06ms. Falls back to Slicer for multi-output explanations and anything else not covered.            
                                                                                                                                                                                             
  ## Checklist                                                                                                                                                                               
                                                                                                                                                                                             
### DIFF
--- a/shap/_explanation.py
+++ b/shap/_explanation.py
@@ -418,6 +418,11 @@ class Explanation(metaclass=MetaExplanation):
         if len(item) == 0:
             return new_self  # type: ignore
         if new_self is None:
+            fast = _fast_slice(self, item)
+            if fast is not None:
+                fast.op_history = copy.copy(self.op_history)
+                fast.op_history.append(OpHistoryItem(name="__getitem__", args=(item,), prev_shape=self.shape))
+                return fast
             new_self = copy.copy(self)
         new_self._s = new_self._s.__getitem__(item)
         new_self.op_history.append(OpHistoryItem(name="__getitem__", args=(item,), prev_shape=self.shape))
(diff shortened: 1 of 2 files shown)


CloseChoice (changes requested):
Thanks for the PR. So you are doing 2 changes in one PR? Can you pls split this up? For issues like this I would also like to see a script with which I can reproduce the described problem.

sohamukute:
Hey @CloseChoice, this is the follow up to #4362, adds a fast path in __getitem__ that skips Slicer for the common 2D case. Would appreciate a review when you have time.

PR #4250, opened 2026-01-27
Fix logit function numerical stability (issue #4045)

VARUN3WARE:
## Overview

Closes #4045

Description of the changes proposed in this pull request:
- Add epsilon parameter (default 1e-10) to prevent inf/nan values
- Clip probabilities to [eps, 1-eps] before logit transformation
- Handles probabilities at or near 0 and 1 safely
- Add comprehensive unit test file 

## Checklist

- [X] All [pre-commit checks](https://pre-commit.com/#install) pass.
- [X] Unit tests added (if fixing a bug or adding a new feature)

PR #4368, opened 2026-04-05
Fix OverflowError in census notebook Occupation scaling

Soumya080:
### Summary
Fixes a runtime `OverflowError` in `Census income classification with scikit-learn.ipynb` caused by scaling the "Occupation" column when it is inferred as a low-bit integer dtype.

---

### What was changed

Replaced in-place scaling:

```python
X["Occupation"] *= 1000
X["Occupation"] = X["Occupation"].astype("int64") * 1000 
```

 ### Why this is needed

In newer pandas versions (≥ 2.x), dtype inference can assign the "Occupation" column as int8.
Multiplying by 1000 exceeds the valid range ([-128, 127]) and raises an OverflowError.

Older pandas versions often performed implicit upcasting, but newer versions enforce stricter dtype behavior, making this issue visible.

Casting to int64 ensures the scaling operation is safe and consistent across environments.
 #4250 
 ### Testing

Verified the notebook runs end-to-end without errors
Confirmed no overflow occurs
Model training and predictions remain unchanged
SHAP explanations execute correctly

### Notes
Minimal, single-line fix
No changes to model logic or outputs
Improves compatibility with newer pandas versions


CloseChoice (changes requested):
Can you pls remove this notebook from https://github.com/shap/shap/blob/master/scripts/run_notebooks_timeouts.py#L25?

Soumya080:
"Done! I've removed the notebook from the allow_to_timeout list as requested. 
Let me know if there's anything else needed!

PR #4369, opened 2026-04-06
TEST: Add comprehensive test coverage for AdditiveExplainer

divyam-jha123:
Add unit tests for shap/explainers/_additive.py to increase test coverage from 29% to 84%.

Tests added:
- Basic additivity verification
- Single sample and multiple sample scenarios
- Feature names preservation
- Main effects equality for additive models
- Edge cases: zero features, constant features, negative values
- Masker validation (requires Independent masker)
- supports_model_with_masker static method
- EBM classifier support (when interpret package available)
- explain_row method direct testing
- Large number of features
- Small background dataset
- Identical samples

The remaining uncovered lines (50-62, 116-118) are:
- EBM without masker code path (raises NotImplementedError)
- EBM with interactions error path

Closes part of #3690

## Overview

Closes #XXXX  

## Checklist


tests/explainers/test_additive.py
```diff
@@ -0,0 +1,370 @@
...
+
+    def model(X):
+        # Simple additive model: sum of features with different weights
+        X = np.atleast_2d(X)
+        return X[:, 0] * 2.0 + X[:, 1] * 3.0 + X[:, 2] * 1.5 + 1.0
+
+    return model
+
+
+def create_multi_output_additive_model():
```
CloseChoice: is this used anywhere?

tests/explainers/test_additive.py
```diff
@@ -0,0 +1,370 @@
+"""Unit tests for the AdditiveExplainer."""
+
+import numpy as np
+import pytest
+
+import shap
+
+
+def create_additive_model():
```
CloseChoice, line 11: I believe this should be a pytest.fixture

CloseChoice (changes requested):
Thanks for the PR. A couple things here: 
1. pls address the comments
2. can you somehow trim down the number of tests to only the necessary ones? For a first try I would suggest getting the coverage per line from coverage.py and eliminate tests that cover the exact same lines. If you want to be fancy you could also check if they kill the exact same mutants using e.g. mutmut or cosmic-ray

Furthermore, we have recently added https://github.com/shap/shap/blob/master/tests/conftest.py#L81, so that the tests actually return a file which stores the shap values so that we can be sure that the same values are produced across commits. I would like to see at least a couple tests in here following that pattern.

tests/explainers/test_additive.py
divyam-jha123: ya it is not used anywhere and only defined in this file

divyam-jha123:
@CloseChoice hey I addressed the suggested changes: converted create_additive_model to a proper
fixture and fixed usages, removed redundant additive tests based on per-test
line coverage overlap, removed the unused create_multi_output_additive_model,
and added baseline-backed SHAP regression coverage via 
compare_numpy_outputs_against_baseline
review https://github.com/shap/shap/pull/4369/changes/3938eb5ac55ae87e478600fa8a20604043445834

PR #4373, opened 2026-04-06
FIX: Use isinstance() instead of 'is' for type checks

### LINKED ISSUE #4372: BUG:  Type check uses `is` instead of `isinstance` in Impute  class
### Issue Description

 ```python

feature.

```python
import numpy as np

data_dict = {
    "mean": X_train.mean(axis=0),
    "cov":  np.cov(X_train.T),
}

```

### Expected Behavior

_No response_

### Bug report checklist

### Installed Versions

0.51.0

divyam-jha123:
## Overview

Closes #4372  

The expressions 'data is dict' and 'masker is dict' compare an instance
 to the dict type object, which always returns False. This caused the
 dict-based initialization paths to be unreachable.
 
 Fixed in:
 - shap/maskers/_tabular.py: Impute.__init__
 - shap/explainers/_explainer.py: Explainer.__init__

## Checklist

### DIFF
--- a/shap/explainers/_explainer.py
+++ b/shap/explainers/_explainer.py
@@ -116,7 +116,7 @@ class Explainer(Serializable):
                 self.masker = maskers.Text(masker)
         elif (masker is list or masker is tuple) and masker[0] is not str:
             self.masker = maskers.Composite(*masker)
-        elif (masker is dict) and ("mean" in masker):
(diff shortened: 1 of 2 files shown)


CloseChoice (approved):
LGTM, thanks. Pls ping on green pipelines

PR #4377, opened 2026-04-06
Fix optional dependency handling and add missing shap import in tree

Soumya080:
## Summary
Improves optional dependency handling in tree-based model notebooks and fixes a missing `shap` import in the CatBoost tutorial.

## What was changed
- Wrapped optional imports (CatBoost, LightGBM, XGBoost) in `try-except ImportError` blocks
- Added clear installation instructions when dependencies are missing
- Added missing `import shap` in the CatBoost notebook before `shap.initjs()`

## Why this is needed
These notebooks rely on optional dependencies that may not be installed in all environments. Without proper handling, they fail with unclear errors.

Additionally, the CatBoost notebook called `shap.initjs()` without importing `shap`, which results in a `NameError` when dependencies are available.

This change ensures:
- Early, clear failure if dependencies are missing
- No runtime errors due to missing imports
- Consistent behavior across environments

## Notes
- Changes are limited to import cells only
- No impact on model logic or outputs


Soumya080:
Hi, I’ve opened a PR fixing optional dependency handling and missing `shap` import in the CatBoost tutorial notebook.

Also ensured the notebook runs cleanly end-to-end and resolves execution order issues.

PR: #4377

Would appreciate feedback.

CloseChoice (changes requested):
pls don't wrap imports into try/except blocks. If that is not installed and an `ImportError` is thrown, that's enough. It does not get more explicit to users, no need to bloat the code. We can keep the missing shap import though

Soumya080:
Thanks for the feedback!

I've removed the try/except import blocks as suggested and kept only the required imports. The changes are limited to the affected notebooks.

Please let me know if any further adjustments are needed.

PR #4387, opened 2026-04-07
ENH: Prototype new plotting API for monitoring plot

### LINKED ISSUE #3803: Plot overhaul: monitoring

Related:
- #2877

TrishaG189:
### Overview  
Closes #3803  
Part of #3411  

### Description  
This PR updates the `shap.plots.monitoring` API to align with the explicit matplotlib interface requirements.

**Changes made:**
- Added an optional `ax` parameter and return it when `show=False`  
- Replaced implicit `plt.*` calls with explicit `ax.*` usage  
- Kept the original figure sizing behavior, using `plt.gcf()` only when `ax` isn’t provided  
- Passed the scatter plot object directly to the colorbar so it attaches to the correct axes  
- Added a `pytest-mpl` baseline test in `tests/plots/test_monitoring.py` to ensure visual consistency and improve coverage beyond 80%  

### Checklist  
- [x] Unit tests added


shap/plots/_monitoring.py
```diff
@@ -14,7 +14,7 @@ def truncate_text(text, max_len):
         return text
 
 
-def monitoring(ind, shap_values, features, feature_names=None, show=True):
+def monitoring(ind, shap_values, features, feature_names=None, ax=None, show=True):
```
CloseChoice:
```suggestion
def monitoring(ind, shap_values, features, feature_names=None, show=True*, ax=None):
```
The way you add `ax` as the second last argument is technically breaking, imagine if someone only uses positional arguments!

Therefore I suggest to add `ax` as a keyword only argument.

CloseChoice (changes requested):
looks good overall, but we cannot break users code.

shap/plots/_monitoring.py
TrishaG189: Great catch, I hadn’t thought about the positional argument issue.
I’ve updated the signature to **`show=True, *, ax=None` so `ax`**  adds as a keyword only arguement.
Just pushed the change—let me know if this looks good.

PR #4389, opened 2026-04-07
TST: Add test coverage for shap/plots/_embedding.py

shah-suhani:
## Overview

Related to #3690 

This PR adds test coverage for `shap/plots/_embedding.py`, which currently has 19% coverage.

It covers 6 tests:

- Integer index coloring
- `sum()` Index coloring
- Custom 2D numpy array as method
- Custom feature names
- Alpha parameter
- Unsupported method (issue #4394, fix in PR #4395)

## Note

I noticed a bug in `shap/plots/_embedding.py` - opened issue #4394 and submitted fix in PR #4395. The unsupported method test in this PR documents the buggy behaviour and will be updated after #4395 is merged.

## Checklist

- [x] All [pre-commit checks](https://pre-commit.com/#install) pass.
- [x] Unit tests added (if fixing a bug or adding a new feature)

PR #4395, opened 2026-04-07
FIX: Raise ValueError for unsupported method in embedding ()

### LINKED ISSUE #4394: BUG: UnboundLocalError in embedding () when passing unsupported method
### Issue Description

In `shap/plots/_embedding.py`, passing an unsupported `method` argument causes the function to print an error message but then immediately crash with an `UnboundLocalError`.

```python
import numpy as np
import shap

shap_values = np.random.randn(50, 5)
shap.plots.embedding(ind=0, shap_values=shap_values, method="unsupported", show=False)
```

### Traceback

```shell
Unsupported embedding method: unsupported
UnboundLocalError: cannot access local variable 'embedding_values' where it is not associated with a value
```

### Expected Behavior

Function prints the error message and returns gracefully without crashing.

### Bug report checklist

### Installed Versions

shap 0.1.dev2940, Python 3.12.13

shah-suhani:
## Overview

Fixes #4394 

When passing an unsupported `method` argument to `shap.plots.embedding()`, the function prints an error message but then immediately crashes with an `UnboundLocalError` because `embedding_values` is never assigned in the `else` branch.

Added `return` after the print statement in `shap/plots/_embedding.py` so the function exits after printing the error.

## Checklist

- [x] Unit tests added (test added in PR #4389)

### DIFF
--- a/shap/plots/_embedding.py
+++ b/shap/plots/_embedding.py
@@ -52,8 +52,8 @@ def embedding(ind, shap_values, feature_names=None, method="pca", alpha=1.0, sho
     elif hasattr(method, "shape") and method.shape[1] == 2:
         embedding_values = method
     else:
-        print("Unsupported embedding method:", method)
-        return
+        emsg = f"Unsupported embedding method: {method}. Expected 'pca' or a numpy array of shape (# samples x 2)."
+        raise ValueError(emsg)
 
     plt.scatter(embedding_values[:, 0], embedding_values[:, 1], c=cvals, cmap=colors.red_blue, alpha=alpha, linewidth=0)
     plt.axis("off")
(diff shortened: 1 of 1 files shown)


shap/plots/_embedding.py
```diff
@@ -53,6 +53,7 @@ def embedding(ind, shap_values, feature_names=None, method="pca", alpha=1.0, sho
         embedding_values = method
     else:
         print("Unsupported embedding method:", method)
+        return
```
amishhaa: How about we do something like https://github.com/shap/shap/blob/master/shap/plots/_force.py#L182? where we raise an error to handle it. Thoughts?

amishhaa (changes requested):

shap/plots/_embedding.py
shah-suhani: Makes sense! Raising a `ValueError` is cleaner than printing and returning. Will update it.
Thanks!

CloseChoice (changes requested):
Could we pls get a test, that checks that the error is correctly thrown? That will help us in the future to not break this behaviour again.

Issue #3808, opened 2024-08-12
Plot overhaul: waterfall


connortann:
This issue is part of the [Visualisation overhaul](https://github.com/orgs/shap/projects/1) project.

The things we wish to overhaul are detailed in #3411, and the current status of these TODO items for this plot is tracked in the project fields (see sidebar).

Anushreebasics:
@techaadii I have been working on this PR, kindly take up another issue

PR #4401, opened 2026-04-08
ENH: Plot overhaul: scatter and waterfall

techaadii:
## Summary
Part of the plotting API overhaul meta-issue #3411.
Covers scatter (#3804) and waterfall (#3808).

## AI Usage Disclosure
I used Claude (Anthropic) as an AI assistant throughout this PR.

- **Autonomous**: Claude identified all implicit `plt.xxx()` calls that needed 
  converting and generated the initial replacement code.
- **Assisted**: I reviewed each change, caught bugs (typos like `ig.` vs `fig.`, 
  `x_i` vs `ax_i`, `plt.axes` vs `ax_list`, `if plt.show` vs `if show`), 
  and fixed them manually.

I understand every line in this PR and can explain any of it on request.

## What I understood working on this

The core problem is that functions like `plt.plot()` operate on whichever axes 
matplotlib considers "current", which breaks when the caller has already set up 
their own figure layout. By accepting an `ax` parameter and defaulting to 
`plt.gca()`, we let the caller control exactly which axes we draw on, without 
breaking existing code that relies on `plt.gcf()` being respected.

For `waterfall`: resolved `ax` and `fig` early from `plt.gca()` and 
`ax.get_figure()`, then replaced every subsequent `plt.xxx()` call with 
`ax.xxx()` or `fig.xxx()`. Changed the return from `plt.gcf()` to `ax`.

For `scatter`: the multi-feature path was using `plt.subplot()` to create and 
select axes implicitly. Replaced with `plt.subplots()` returning an explicit 
list, used `np.atleast_1d()` to handle the single-subplot edge case, and 
ensured the function always returns the axes rather than conditionally on `show`.

### `shap/plots/_scatter.py`
- Multi-feature path: replaced `plt.subplot()` with `plt.subplots()`, 
  `plt.plot()` with `ax_i.plot()`, `plt.legend()` with `ax_list[-1].legend()`
- Always returns `ax` (single feature) or `ax_list` (multiple features), 
  regardless of `show` value
- Updated docstring Returns section accordingly

### `shap/plots/_waterfall.py`
- Added `ax: plt.Axes | None = None` parameter
- Resolves axes via `plt.gca()` when none provided, gets figure via 
  `ax.get_figure()`
- Replaced all `plt.barh`, `plt.arrow`, `plt.text`, `plt.errorbar`, 
  `plt.yticks`, `plt.axhline`, `plt.axvline`, `plt.gca()` calls with 
  `ax.xxx()` equivalents
- Always returns `ax`

### Tests
- `test_waterfall_returns_ax`: verifies `ax` is always returned
- `test_waterfall_accepts_ax`: verifies a passed `ax` is used and returned
- `test_scatter_always_returns_ax`: verifies `ax` returned regardless of `show`
- `test_scatter_multi_returns_axes_list`: verifies list of Axes for multi-feature

## Checklist
- [x] Uses explicit matplotlib interface
- [x] Accepts `ax` parameter (waterfall)
- [x] Returns `ax` always (both functions)
- [x] Unit tests added
- [x] Existing tests pass (21/21)
- [x] AI usage disclosed

@connortann I am happy to hear back from you.


techaadii:
@connortann  Please review this and guide if this solves the problem or not. Thank you

techaadii:
this is my first contribution to shap! This PR is part of the plotting API overhaul (#3411) and covers the scatter (#3804) and waterfall (#3808) plots. All 21 tests pass locally. Could a maintainer approve the CI workflows when you get a chance? Happy to address any review feedback. cc @connortann

naxatra2:
I think there is some return Type inconsistency in your scatter

In the updated `_scatter.py`, the function now returns a single `ax` for single features but an `ax_list` (list) for multiple features.

I think this is functionally correct, but shouldn't we follow the standard Matplotlib pattern of returning an ndarray of axes for subplots? or at least be extremely explicit in the type hints to avoid `AttributeError` for users expecting a single object.

tests/plots/baseline/test_scatter_multiple_cols_overlay.png
CloseChoice, line 1: I would prefer that we keep the old behaviour here. Stylewise and not breaking anything on users end.

tests/plots/test_waterfall.py
```diff
@@ -118,3 +118,19 @@ def f(x):
     explainer = shap.Explainer(f, med)
     shap_values = explainer(X)
     shap.plots.waterfall(shap_values[0], show=False)
+
+
+def test_waterfall_returns_ax(explainer):
```
CloseChoice, line 123: can we actually use pytest-mpl for acouple of the new tests?

CloseChoice (changes requested):

PR #4408, opened 2026-04-08
MAINT: tensorflow: Conversion of an array with ndim > 0 to a scalar is deprecated #3973

### LINKED ISSUE #3973: CI warning: tensorflow: Conversion of an array with ndim > 0 to a scalar is deprecated
```
tests/explainers/test_gradient.py::test_tf_input[numpy]
tests/explainers/test_gradient.py::test_tf_input[dataframe]
  /opt/hostedtoolcache/Python/3.12.8/x64/lib/python3.12/site-packages/tensorflow/python/framework/ops.py:315: DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
    return float(self._numpy())
```

Switcha57:
## Overview

Closes #3973 

Since the model used in the test does not contain any layer which behavior change between training and inference and the whole data fits in a batch we can use [the raw call to predict and not the wrapper to skip the keras preprocessing of the input](https://www.tensorflow.org/api_docs/python/tf/keras/Model#predict) which triggers the warning
## Checklist


CloseChoice (commented):
Pls ping once pipelines are green

CloseChoice (approved):
Thanks for the PR. Let's merge on green pipelines

Switcha57:
> Pls ping once pipelines are green

It looks like the CI workflows are currently paused and require a maintainer's approval to run. Could you please trigger them so we can get those green pipelines?

Switcha57:
@CloseChoice ?

PR #4409, opened 2026-04-08
Plot overhaul group difference

### LINKED ISSUE #3799: Plot overhaul: group_difference

techaadii:
## Summary
closes #3799 
Part of the plotting API overhaul meta-issue #3411.
Covers group_difference (#3799).

For this No AI has been used 

I understand every line and can explain it on request.

## What I understood
The original code used `if ax:` to detect a provided axes, then silently 
set `show = False` as a side effect. This is fragile — it overrides the 
caller's intent and still creates a new figure internally even when ax is 
provided. The fix uses `if ax is None` as the standard pattern, skips 
figure creation when the caller provides their own axes, and always returns 
ax so callers can chain further customisation.

### `shap/plots/_group_difference.py`
- Replace `if ax / else` block with `if ax is None: _, ax = plt.subplots(...)`
- Remove the `show = False` override side effect
- Always return `ax`
- Add type hints to all parameters and return type
- Complete docstring with all parameters and Returns section
- Add `from __future__ import annotations`

### `tests/plots/test_group_difference.py`
- `test_group_difference_returns_ax`: verifies ax is always returned
- `test_group_difference_accepts_ax`: verifies passed ax is used and returned

Closes #3799
Part of #3411

@connortann  would love to hear your feedback. The repository is very well maintained and its so fun to contribute to a legacy. Thanks


CloseChoice:
This is exactly the same as #4401. Once that is merged, I'll close this PR as well

PR #4411, opened 2026-04-08
FIX: wire ax parameter in partial_dependence_plot (#3206)

### LINKED ISSUE #3206: Partial Dependence Plots doesn’t work with custom ax

darraghmoran2025:
## Summary

Closes #3206

`partial_dependence_plot` accepted an `ax` parameter but both branches
of the `if ax is None` block called `plt.gca()` unconditionally, so any
externally created axes were silently ignored and the plot always drew
onto whatever axes matplotlib happened to have active.

**Fix:** use the provided `ax` directly as `ax1`, and derive the figure
via `ax.get_figure()` instead of `plt.gcf()`.

## Changes

- `shap/plots/_partial_dependence.py`: wire `ax` through correctly
- `tests/plots/test_dependence.py`: add regression test asserting the
  returned axes object is identical to the one passed in

## Test plan

- [x] `pytest tests/plots/test_dependence.py` — all 4 tests pass


tests/plots/test_dependence.py
```diff
@@ -17,6 +18,29 @@ def test_random_dependence_no_interaction():
     shap.dependence_plot(0, np.random.randn(20, 5), np.random.randn(20, 5), show=False, interaction_index=None)
 
 
+def test_partial_dependence_custom_ax():
```
CloseChoice, line 22: can you pls use pytest-mpl to generate an image from this test?

CloseChoice (changes requested):
loos good. We just need an image from the test.

tests/plots/test_dependence.py
darraghmoran2025: test_partial_dependence_custom_ax
does this work? i have also committed it to the PR. thank you!

PR #4413, opened 2026-04-09
Resolve: UserWarning: The structure of inputs doesn't match the expected structure.

### LINKED ISSUE #3066: MAINT: address all deprecation warnings in test suite

shaivimalik:
## Overview

Resolves warnings of this type across the tests: 
```python
/opt/hostedtoolcache/Python/3.12.8/x64/lib/python3.12/site-packages/keras/src/models/functional.py:237: UserWarning: The structure of `inputs` doesn't match the expected structure.
```

In `shap/explainers/_deep/deep_tf.py` and `shap/explainers/_gradient.py`:

- When a tuple of `(model.inputs, model.layers[-1].output)` is passed to the explainer, the `self.model` variable is created as:

```python
self.model = keras.Model(model[0], model[1])
```
where, `type(model[0])` is `<class 'list'>`, which leads to `self.model.input` being `[<KerasTensor shape=(None, 2), dtype=float32, sparse=False, ragged=False, name=keras_tensor>]`

- When the model is passed directly to the explainer, the `self.model` variable is created as:

```python
self.model = model
```

which results in `self.model.input` being `<KerasTensor shape=(None, 2), dtype=float32, sparse=False, ragged=False, name=keras_tensor>`

This creates an inconsistency: depending on how the user creates the explainer, the model expects a different input structure. This often results in the warnings raised by the test cases. 

Since most of the code in `shap/explainers/_deep/deep_tf.py` and `shap/explainers/_gradient.py` expects `self.model.input` to be a list of tensors (`[<KerasTensor shape=(None, 2), dtype=float32, sparse=False, ragged=False, name=keras_tensor>]`). I have modified the variable declaration so that it resembles an user passing a tuple to the explainer.

Interestingly, the tests in `tests/explainers/test_deep.py` do not cover the scenario where a user passes the model directly to the explainer for a single-input model. I have added a test to handle this case.

Supports #3066 

## Checklist


shaivimalik:
Hello @CloseChoice, could you please review it?

CloseChoice (commented):
Thanks, this looks good to me. Can you pls ping on green pipelines?

tests/explainers/test_deep.py
```diff
@@ -196,6 +196,45 @@ def test_tf_keras_linear():
...
 
 
+def test_tf_keras_linear_model_input():
+    """Test verifying that a linear model with linear data gives the correct result."""
+    # FIXME: this test should ideally pass with any random seed. See #2960
+    random_seed = 0
+
+    tf = pytest.importorskip("tensorflow")
+
+    # tf.compat.v1.disable_eager_execution()
```
CloseChoice: NIT: remove this line pls

shaivimalik:
@CloseChoice, quick question: why do we recreate the model on line 101 of `tests/explainers/test_gradient.py` (related to 1bb8d62beb9e22439b1a240ad3baae10230bffe8) ?

PR #4417, opened 2026-04-09
Improve test coverage in `tests/maskers/test_composite.py`

Jiya873:
Part of #3690

I have added tests in `tests/maskers/test_composite.py` to increase coverage for the Composite masker.

I have improved Composite masker test by targeting previously untested execution branches, exception handling (mismatched rows, joint clustering errors), and conditional flag logic.

### DIFF
--- a/tests/maskers/test_composite.py
+++ b/tests/maskers/test_composite.py
@@ -4,6 +4,7 @@ import numpy as np
 import pytest
 
 import shap
+from shap.utils._exceptions import InvalidMaskerError
 
 
 def test_composite_masker_init():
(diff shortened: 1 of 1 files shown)


tests/maskers/test_composite.py
```diff
@@ -237,3 +238,129 @@ def test_composite_masker_with_kwargs():
...
+        composite(mask, data1[0], data2[0])
+
+
+def test_composite_masker_joint_clustering_exception():
+    """Test that joining two non-trivial clusterings raises NotImplementedError."""
+
+    class MockClusteringMasker(shap.maskers.Masker):
+        def __init__(self):
+
+            self.clustering = np.array([[0, 1, 0.5, 2]])
```
CloseChoice: is this line needed here? If not, we could pull the class to the top of the file since more tests use it.

tests/maskers/test_composite.py
```diff
@@ -237,3 +238,129 @@ def test_composite_masker_with_kwargs():
...
+    composite = shap.maskers.Composite(masker1, masker2)
+
+    with pytest.raises(NotImplementedError, match="Joining two non-trivial clusterings is not yet implemented"):
+        composite.clustering("arg1", "arg2")
+
+
+def test_composite_masker_missing_clustering():
+    """Test that clustering attribute is absent if any submasker lacks it."""
+
+    class NoClusteringMasker(shap.maskers.Masker):
```
CloseChoice: see comment above.

CloseChoice (changes requested):
Let's clear these questions, then I believe this is good.

tests/maskers/test_composite.py
Jiya873: ohk. i will move these classes to top as these classes are not specific to the test.

tests/maskers/test_composite.py
Jiya873: Done

Jiya873:
I have applied the changes, plz have a look @CloseChoice

PR #4418, opened 2026-04-09
docs: clarify expected_value math formula in TreeExplainer #4414

DashratRajpurohit:
### What does this PR do?
This PR clarifies the documentation for `TreeExplainer.shap_values`, specifically the relationship between summed SHAP values and the model output.

### What problem does it solve?
As discussed in #4414, the current documentation can be confusing regarding how `expected_value + sum(shap_values)` relates to model predictions.

For models like XGBoost and LightGBM (with `model_output="raw"`), SHAP values are computed in the raw margin (log-odds) space, while users often compare them with probability outputs from `predict()` or `predict_proba()`.

This PR improves clarity by:
- Clarifying that the formula applies to the explainer's model output (raw margin)
- Adding a note explaining that users may need to apply a transformation (e.g., logistic function) to compare with probabilities

Fixes #4414

### Checklist
- [x] I have read the CONTRIBUTING.md
- [x] I have updated the documentation accordingly
- [x] Tests not required (documentation-only change)

Issue #4414, opened 2026-04-09
`shap_values` documentation issues and formula calculation gap.


khteh:
https://shap.readthedocs.io/en/latest/generated/shap.TreeExplainer.html

```
- **XGBoost and LightGBM** (with default ``model_output="raw"``) output
  a single value (raw margin/log-odds), resulting in SHAP values of shape
  ``(#num_samples, #num_features)`` and ``expected_value`` as a scalar.
```
I find this confusing. Where is that **single value** returned? Compared to sklearn, which has a saparate dimension for the 2 output values, XGBoost doesn't have that. So, where is that **single value** embedded somewhere? Could you please clarify on that?

About the assertion of the values:
(1) 
```
Single output: shap_values[i, :].sum() + expected_value = model_output[i]
```
I try this with XGBoost and this is what I observe:
```
prediction: (92, 2)
shap_values: (1, 20), [[ 0.0194523   0.03423912  0.7979535   0.5000503   0.13220659  0.36267897
   0.20176274  0.          0.528874    0.01454332  0.09491263 -0.00805646
  -0.00625417 -0.01392754  0.02101561  0.28726286  0.          0.02308488
   0.17511454  0.5701697 ]]
expected_value: (), 0.18604861199855804

assert shap_values[0, :].sum() + self._shap.expected_value == prediction[0]
```
There is a mismatach in the prediction shape expectation. It turns out to be 2-D array and therefore hit the `ValueError`:
```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
Then, I use `.all()` and the assertion fails:
```
    assert (shap_values[0, :].sum() + self._shap.expected_value == prediction[0]).all(), f"{shap_values[0, :].sum()} + {self._shap.expected_value} = {shap_values[0, :].sum() + self._shap.expected_value} != {prediction[0]}"
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
AssertionError: 3.7350831031799316 + 0.18604861199855804 = 3.9211316108703613 != [0.03359491 0.9664051 ]
```
(2)
```
Multiple outputs: shap_values[i, :, j].sum() + expected_value[j] = model_output[i, j]
```
I try it with RandomForest and this is what I observe:
```
prediction: (92, 2)
[...]

PR #4423, opened 2026-04-09
DOCS: Clarify sample alignment requirement for shap_values additivity…

BaibhavKundu2005:
## What does this PR do?
It adds a note to the `TreeExplainer.shap_values` docstring clarifying that 
the additivity formula requires SHAP values and model predictions to be 
computed on the same samples in the same order.

In issue #4414, users encounter assertion failures when computing 
SHAP values on a subset of samples and comparing against predictions from 
a different batch. The formula itself is correct - the root cause is 
sample index misalignment.

## Investigation
I reproduced the RandomForest case from #4414 with a controlled dummy 
dataset, testing both `feature_perturbation='tree_path_dependent'` and 
`feature_perturbation='interventional'`. All assertions passed in both 
modes when the same samples were used for both SHAP and predictions, 
confirming the root cause is sample misalignment and not a bug in SHAP.

## Related
- Partially addresses #4414 (output space confusion addressed separately in #4418)
- Related to #4418

## Checklist
- [x] I have read the CONTRIBUTING.md
- [x] Pre-commit hooks pass
- [x] Documentation builds cleanly
- [x] No new tests required (documentation-only change)

### DIFF
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -549,6 +549,12 @@ class TreeExplainer(Explainer):
 
             * Single output: ``shap_values[i, :].sum() + expected_value = model_output[i]``
             * Multiple outputs: ``shap_values[i, :, j].sum() + expected_value[j] = model_output[i, j]``
+            .. note::
+                This property holds only when ``shap_values`` and ``model_output``
+                are computed on the **same samples in the same order**. A common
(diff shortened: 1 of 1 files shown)


CloseChoice:
Thanks for the PR, unfortunately someone already worked on this. Duplicate of #4418

PR #4424, opened 2026-04-09
fix: replace identity type checks with isinstance in Explainer and Impute

### LINKED ISSUE #4372: BUG:  Type check uses `is` instead of `isinstance` in Impute  class

tarun-227:
## Summary

Closes #4372

Three type checks use `is` (identity) instead of `isinstance`, making their branches unreachable dead code:

| Location | Bug | Effect |
|---|---|---|
| `_explainer.py:117` | `masker is list or masker is tuple` | Composite masker path never executes |
| `_explainer.py:117` | `masker[0] is not str` | Same line, same pattern |
| `_explainer.py:119` | `masker is dict` | Dict-to-Independent masker path never executes |
| `_tabular.py:341` | `data is dict` | Impute dict initialization never executes |

`x is dict` compares an instance to the `dict` type object, which is always `False`. The correct pattern is `isinstance(x, dict)`.

## Changes
- Replace all four identity checks with `isinstance()` in `_explainer.py` and `_tabular.py`
- Add regression tests for each fixed code path (Impute dict init, Explainer dict masker, Explainer list masker)

## Test plan
- [x] Three new regression tests pass
- [x] All existing masker tests pass
- [x] Bug confirmed: `{'mean': ...} is dict` → `False`, `isinstance({'mean': ...}, dict)` → `True`

**Note:** This is a more complete fix than #4373, which addresses lines 119 and 341 but misses the `masker is list or masker is tuple` bug on line 117 and includes no tests.

**Note:** Tests were scaffolded with AI assistance and reviewed/validated locally.

### DIFF
--- a/shap/explainers/_explainer.py
+++ b/shap/explainers/_explainer.py
@@ -114,9 +114,9 @@ class Explainer(Serializable):
                 self.masker = maskers.Text(masker, mask_token="...", collapse_mask_token=True)
             else:
                 self.masker = maskers.Text(masker)
-        elif (masker is list or masker is tuple) and masker[0] is not str:
+        elif isinstance(masker, (list, tuple)) and not isinstance(masker[0], str):
             self.masker = maskers.Composite(*masker)
-        elif (masker is dict) and ("mean" in masker):
+        elif isinstance(masker, dict) and ("mean" in masker):
             self.masker = maskers.Independent(masker)
         elif masker is None and isinstance(self.model, models.TransformersPipeline):
             return self.__init__(  # type: ignore[misc]
--- a/shap/explainers/_explainer.py
+++ b/shap/explainers/_explainer.py
@@ -114,7 +114,12 @@ class Explainer(Serializable):
                 self.masker = maskers.Text(masker, mask_token="...", collapse_mask_token=True)
             else:
                 self.masker = maskers.Text(masker)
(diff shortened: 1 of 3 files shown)


tarun-227:
@CloseChoice gentle ping on this small refactor (replaces `type(x) is Y` identity checks with `isinstance` in `Explainer` and `Impute`). Happy to adjust if you'd prefer a different direction.

CloseChoice (approved):
Thanks this looks good. Pls ping on green pipelines.

EDIT: pls also fix the merge conflicts

tarun-227:
Thanks for the approval @CloseChoice! Spotted that the CI failures on `test_deep.py` were caused by the `isinstance` change itself: the previously-dead Composite branch (which used the identity check `masker is list or masker is tuple`, effectively always false) became live and started catching `[tensor1, tensor2]` style multi-input deep-explainer data — which then crashed inside `Composite.__init__` when it tried `masker.__call__.__code__`.

Pushed a follow-up that tightens the condition so Composite is only used when every element is a `shap.maskers.Masker` instance. The affected deep tests pass locally (18 passed) and `test_tabular.py` still passes.

PR #4429, opened 2026-04-10
Resolve: DeprecationWarning: __array_wrap__ must accept context and return_scalar arguments (positionally) in the future. (Deprecated NumPy 2.0)

shaivimalik:
## Overview

Resolves (part of) #3066 

```python
DeprecationWarning: __array_wrap__ must accept context and return_scalar arguments (positionally) in the future. (Deprecated NumPy 2.0)
```

This warning is raised when `explainer.expected_value[t]` and `output_phis[t].sum(axis=tuple(range(1, output_phis[t].ndim)))` are subtracted from`model_output_values[:, t]`. 

The warning is raised when a `numpy-ndarray` is subtracted from a `torch.Tensor` as illustrated in the toy example below (probably due to the implicit type conversion).

Screenshot 2026-04-10 at 7 32 45 PM

`model_output_values[:, t]` is either a `torch.Tensor` or a `tensorflow.python.framework.ops.EagerTensor`, depending on the framework being used. Both tensors support the `.numpy()` method, so in the updated code it is called to convert them to `numpy-ndarray`.

`output_phis[t].sum(axis=tuple(range(1, output_phis[t].ndim)))` is `numpy-ndarray`.

`explainer.expected_value[t]` is either a `tensor` or `numpy-ndarray`, or `np.float32`. Using `np.asarray` is the best method to convert to `numpy-ndarray` as it doesn't raise a warning when a numpy array or tensor is passed to it. Other ways to convert `explainer.expected_value[t]` to `numpy-ndarray` are to use `np.array` (which raises the DeprecationWarning) or do `explainer.expected_value[t].numpy() if hasattr(explainer.expected_value[t], 'numpy') else explainer.expected_value[t]` (less clean).

## Checklist

### DIFF
--- a/shap/explainers/_deep/deep_utils.py
+++ b/shap/explainers/_deep/deep_utils.py
@@ -11,12 +11,12 @@ def _check_additivity(explainer, model_output_values, output_phis):
     for t in range(len(explainer.expected_value)):
         if not explainer.multi_input:
             diffs = (
-                model_output_values[:, t]
-                - explainer.expected_value[t]
+                model_output_values[:, t].numpy()
+                - np.asarray(explainer.expected_value[t])
                 - output_phis[t].sum(axis=tuple(range(1, output_phis[t].ndim)))
             )
         else:
-            diffs = model_output_values[:, t] - explainer.expected_value[t]
+            diffs = model_output_values[:, t].numpy() - np.asarray(explainer.expected_value[t])
 
             for i in range(len(output_phis[t])):
                 diffs -= output_phis[t][i].sum(axis=tuple(range(1, output_phis[t][i].ndim)))


shap/explainers/_deep/deep_utils.py
```diff
@@ -11,12 +11,12 @@ def _check_additivity(explainer, model_output_values, output_phis):
     for t in range(len(explainer.expected_value)):
         if not explainer.multi_input:
             diffs = (
-                model_output_values[:, t]
-                - explainer.expected_value[t]
+                model_output_values[:, t].numpy()
```
CloseChoice, line 14: can we be sure that this is always a tensor? Can't this be a numpy array already?

shap/explainers/_deep/deep_utils.py
```diff
@@ -11,12 +11,12 @@ def _check_additivity(explainer, model_output_values, output_phis):
...
+                - np.asarray(explainer.expected_value[t])
                 - output_phis[t].sum(axis=tuple(range(1, output_phis[t].ndim)))
             )
         else:
-            diffs = model_output_values[:, t] - explainer.expected_value[t]
+            diffs = model_output_values[:, t].numpy() - np.asarray(explainer.expected_value[t])
```
CloseChoice, line 19: same here?

CloseChoice (changes requested):
Let's sort out the questions then take it from there.

shap/explainers/_deep/deep_utils.py
shaivimalik: I think so, I ran some type checking tests and they were always tensors. We can also trace from where `_check_additivity` is called ([here](https://github.com/shaivimalik/shap/blob/69e3ab699217f25caed9995b17ae0a47613d9a49/shap/explainers/_deep/deep_pytorch.py#L226) and [here](https://github.com/shaivimalik/shap/blob/69e3ab699217f25caed9995b17ae0a47613d9a49/shap/explainers/_deep/deep_tf.py#L347])). In both cases, `model_output_values` are the output predictions from either a tensorflow/keras or pytorch model, which are always going to be tensors.

PR #4432, opened 2026-04-10
fix: use isinstance() instead of identity check for dict in Impute

### LINKED ISSUE #4372: BUG:  Type check uses `is` instead of `isinstance` in Impute  class

SarthakB11:
Fixes #4372

**Problem**: `Impute.__init__` uses `data is dict` (identity check) instead of
`isinstance(data, dict)` to detect dict-typed input data. Since `data` is never the
`dict` class itself, the branch for extracting precomputed mean/covariance is silently
skipped for all dict inputs.

**Root cause**: `is` checks object identity, not type membership. The condition
`data is dict` is only `True` when `data` literally *is* the built-in `dict` class
object, not when `data` is a dict instance.

**Fix**: Replace `data is dict` with `isinstance(data, dict)` in
`shap/maskers/_tabular.py` line 341.

**Tests**: Added `tests/maskers/test_impute.py` with 4 regression tests:
- dict with mean + cov keys
- dict subclass (OrderedDict)
- dict with mean only (no cov)
- plain ndarray (unchanged path)

### AI Disclosure

AI tools (Claude) were used for initial investigation and test scaffolding. All code
was reviewed and verified manually.

### DIFF
--- a/shap/maskers/_tabular.py
+++ b/shap/maskers/_tabular.py
@@ -338,7 +338,7 @@ class Impute(Masker):  # we should inherit from Tabular once we add support for 
             The background dataset that is used for masking.
 
         """
-        if data is dict and "mean" in data:
+        if isinstance(data, dict) and "mean" in data:
             self.mean = data.get("mean", None)
             self.cov = data.get("cov", None)
             data = np.expand_dims(data["mean"], 0)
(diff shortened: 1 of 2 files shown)


SarthakB11:
Didn't notice the existing PRs, happy to close in favor of  #4373  / #4424

PR #4434, opened 2026-04-10
FIX: Correct max_evals calculation in PermutationExplainer.shap_values()

### LINKED ISSUE #4433: BUG: PermutationExplainer.shap_values() incorrectly calculates max_evals, resulting in fewer permutations
## Description
There is a logic error in `PermutationExplainer.shap_values()` when converting `npermutations` to `max_evals`. The method silently runs less than half of the user-requested permutations, leading to lower-quality SHAP value estimates without any warning. 

In `_permutation.py` on line 286, `max_evals` is computed as:
```python
explanation = self(X, max_evals=npermutations * X.shape[1], main_effects=main_effects)
```
However, inside explain_row() (line 165), the number of actual permutations is calculated dynamically based on max_evals:
```
python
npermutations = max_evals // (2 * len(inds) + 1)
```
Because one full permutation sequence evaluates the model forward, backward, plus a baseline 
(2 * num_features + 1), providing npermutations * X.shape[1] as max_evals restricts the explainer heavily and executes far fewer permutations than intended.

## Steps to Reproduce
Run the following minimal snippet using the legacy shap_values API and compare it against the __call__ API:
```
import numpy as np
import shap

# 1. Provide a minimal mock model and data
def model(X):
    return X.sum(axis=1)

X = np.random.randn(10, 5)  # 5 features
masker = shap.maskers.Independent(np.zeros((1, 5)))
explainer = shap.PermutationExplainer(model, masker)

# 2. Use the legacy API and request 10 permutations
sv_legacy = explainer.shap_values(X, npermutations=10)

# 3. Use the __call__ API manually configured for 10 permutations
sv_call = explainer(X, max_evals=10 * (2 * 5 + 1))

```
## Expected Behaviour
When passing npermutations=10 with a 5-feature dataset, the explainer should execute exactly 10 permutations (which requires an allocation of 10 * 11 = 110 maximum evaluations per row).

## Actual Behaviour
The explainer calculates max_evals = 10 * 5 = 50. From this budget, the explainer's loop computes 50 // 11 = 4 actual permutations. The user asks for 10 permutations but silently gets only 4.
## Proposed Fix
The max_evals calculation in shap_values() should account for the 2*M + 1 model evaluations required per permutation.
In shap/explainers/_permutation.py:
```
- explanation = self(X, max_evals=npermutations * X.shape[1], main_effects=main_effects)
+ explanation = self(X, max_evals=npermutations * (2 * X.shape[1] + 1), main_effects=main_effects)

```

## Bug report checklist

### Installed Versions

SHAP version: 0.46.0 (or current master)

AftAb-25:
## Description

Closes #4433.

I was looking into how `PermutationExplainer.shap_values()` works and noticed a bug where it ends up starving the explainer of evaluation budget. It computes `max_evals` using `npermutations * X.shape[1]`. However, since every individual permutation needs an evaluation for both the forward pass and the backward pass, plus one for the baseline, the actual requirement is `2 * n_features + 1` evaluations per permutation.

Because of this under-allocation, the underlying [explain_row()](cci:1://file:///c:/Users/aftab/Desktop/shap/shap/explainers/_permutation.py:127:4-250:9) method ends up running far fewer permutations than the user actually requested. 

## Changes:
- Updated the `max_evals` calculation in [shap_values()](cci:1://file:///c:/Users/aftab/Desktop/shap/shap/explainers/_permutation.py:252:4-286:62) to pass `npermutations * (2 * X.shape[1] + 1)`, ensuring the explainer can execute the correct number of permutations.

All tests in `test_permutation.py` are passing locally.

### DIFF
--- a/shap/explainers/_permutation.py
+++ b/shap/explainers/_permutation.py
@@ -283,7 +283,7 @@ class PermutationExplainer(Explainer):
             of such matrices, one for each output.
 
         """
-        explanation = self(X, max_evals=npermutations * X.shape[1], main_effects=main_effects)
+        explanation = self(X, max_evals=npermutations * (2 * X.shape[1] + 1), main_effects=main_effects)
         return explanation.values  # type: ignore[union-attr]
 
     def __str__(self) -> str:


shap/explainers/_permutation.py
```diff
@@ -283,7 +283,7 @@ def shap_values(
             of such matrices, one for each output.
 
         """
-        explanation = self(X, max_evals=npermutations * X.shape[1], main_effects=main_effects)
+        explanation = self(X, max_evals=npermutations * (2 * X.shape[1] + 1), main_effects=main_effects)
```
CloseChoice, line 286: Thanks, this is a great catch!

CloseChoice (approved):
Great work. Let's check if the pipelines run through or if we need to change anything, maybe the baselines files, let's see. Pls ping on green or if you get stuck. Want to get this in ASAP

AftAb-25:
Hey @CloseChoice, thanks for the review!

I've kept an eye on the CI pipelines and all the checks have finished running successfully and are completely green ✅. Tests are passing correctly and it looks like there were no issues with the baseline files either. 

Ready to merge whenever you are! Let me know if you need any other changes from me to get this wrapped up.

PR #4441, opened 2026-04-10
test: use default matplotlib style in test suite

### LINKED ISSUE #3954: MAINT: use "default" matplotlib style for tests

peteroyce:
Fixes #3954

Switches the test suite to use matplotlib's `default` style for consistent rendering across environments, matching what users actually see with a modern matplotlib installation.

**Changes:**
- Adds `mpl-default-style = "default"` to `[tool.pytest.ini_options]` in `pyproject.toml` so pytest-mpl uses the default style when generating/comparing baseline images
- Adds `matplotlib.style.use("default")` inside the `mpl_test_cleanup` autouse fixture in `tests/conftest.py`, scoped within `plt.rc_context()` so each test starts from a clean default style state

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -218,12 +218,19 @@ local_scheme = "no-local-version"
 [tool.pytest.ini_options]
 addopts = "--mpl -m 'not xslow'"
 testpaths = ["tests"]
+mpl-default-style = "default"
 filterwarnings = [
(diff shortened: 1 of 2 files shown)


CloseChoice:
Issue is closed already and an earlier PR was merged, see #4330. Therefore I close this.

PR #4445, opened 2026-04-10
TYP: add type hints for explainers.tf_utils

SarthakB11:
Part of #3730

**What:** Add type annotations to `shap/explainers/tf_utils.py` and enable
`check_untyped_defs` for this module.

**Changes:**
- Add return and parameter type hints to `_import_tf`, `_get_session`,
  `_get_graph`, `_get_model_inputs`, `_get_model_output`
- Type the module-level lazy `tf` global as `ModuleType | None`
- Add `type: ignore[union-attr]` comments where the optional `tf` module
  is accessed after the `_import_tf()` guard (mypy can't infer the narrowing
  across the global assignment)
- Remove `shap.explainers.tf_utils` from the mypy exclusion list in
  `pyproject.toml`

**Verification:** `mypy --check-untyped-defs` passes with zero errors on this
module. `ruff check` passes. Runtime imports verified. No TF-specific tests
affected (TF is an optional dependency).

### AI Disclosure

AI tools (Claude) were used for initial investigation and annotation scaffolding.
All code was reviewed and verified manually.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -150,7 +150,7 @@ module = [
   # "shap.explainers._coalition",      # NOW TYPE CHECKED - type hints added
   "shap.explainers.other._random",
   "shap.explainers.pytree",
-  "shap.explainers.tf_utils",
+  # "shap.explainers.tf_utils",        # NOW TYPE CHECKED - type hints added
   # "shap.maskers._composite",  # NOW TYPE CHECKED - type hints added
   # "shap.maskers._fixed",      # NOW TYPE CHECKED - type hints added
   "shap.maskers._image",
--- a/shap/explainers/tf_utils.py
+++ b/shap/explainers/tf_utils.py
@@ -1,12 +1,24 @@
 from __future__ import annotations
 
 import warnings
-from typing import TYPE_CHECKING, Any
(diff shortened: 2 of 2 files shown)


CloseChoice (changes requested):
Thanks for the PR but I don't see a lot of value in here. Just ignoring all warnings and using `Any` does not really help, it just silences mypy to a similar point as to disable it entirely.

Therefore I would suggest to use `typing.Protocol` to build a protocol for model, e.g. 

```python
class TFModel(typing.Protocol):
    # define all necessary methods in here
```

SarthakB11:
Pushed a follow-up: defined a `TFModel` `Protocol` over `inputs`/`outputs`/`layers` and applied it to `_get_model_inputs` and `_get_model_output`.

Left `_get_session` and `_get_graph` as `Any` since both treat their args as opaque pass-throughs (no attribute access). Happy to extend if you'd prefer.

SarthakB11:
Rebased onto latest main. Ready for another look when you have a chance.

SarthakB11:
@CloseChoice pushed the TFModel Protocol you asked for and rebased onto latest main. CI green, ready for another look when you have a chance.

PR #4452, opened 2026-04-10
Establish scikit-buiild-core as sole backend

daidahao:
## Overview

Stage 1 of #4327: **Establish scikit-build-core as sole backend**

This PR uses scikit-build-core as the sole build backend, and drops setuptools as backend. 

**Major changes include:**
- Build `shap` for Python 3.12+ stable ABI, which would reduce the number of binary wheels for distribution.
- Compile nanobind extension `cutils` and numpy extension `cext` jointly in `CMakeLists.txt`.
- Mark `shap.cutils` as free-threaded, but it has not been tested yet!
- Cast `PyArrayObject*` to `PyObject*` before calling `Py_XDECREF` (release reference to the array). This has to be done to support Python 3.12+ ABI.
- Drop `_*` prefix from `shap.cutils` (nanobind) functions, e.g., `_compute_grey_code_row_values` becomes `compute_grey_code_row_values`.
- Target macOS build for 10.13+, which is required by nanobind as it uses C++ 17.

**Results:**
1. Running `uv build` locally on macOS 26, Python 3.13 produce two files:
```
Successfully built dist/shap-0.51.1.dev44.tar.gz
Successfully built dist/shap-0.51.1.dev44-cp312-abi3-macosx_26_0_arm64.whl
```
2. Running `uv run pytest tests/explainers/test_kernel.py` and `uv run pytest tests/explainers/test_exact.py` without any fails locally. More unit tests need to be run on CIs.

**What's still missing:**
- `build_wheels` CI has not been updated.
- GPU Tree logic is not being compiled in `CMakeLists.txt`. To enable CUDA build, we would need to transfer the logic in `setup.py` to `CMakeLists.txt`.
- Other CIs like readthedocs have not been updated:

**TODO list:**
- [x] Update readthedocs CI.
- [ ] Trigger CIs on the current branch "move-to-nanobind".
- [ ] Update `run_tests` CI.
- [ ] Update `run_notebooks` CI.
- [ ] ...

**Discussion points:**
- **uv build frontend:** `cibuildwheel` uses pip as default build frontend. Should we switch to uv for faster setup and build?
- **`shap.cutils` visibility:** I suggest renaming the extension to `shap._cutils`.
- **C++ source files:** It is customary to not bundle C++ source files in binary wheels. Should we move `cutils` out of `shap` folder and move into the root folder? We can still distribute the source files in sdist. In the long term, I would also suggest moving `cext` out of `shap` folder.
- **Stable ABI:** nanobind supports building for stable ABI from Python 3.12+. I've enabled it to reduce the number of binary wheels for distribution. Do you agree with the decision here?
- **Free threading:** I have marked the `cutils` extension as `FREE_THREADED`. But this might need to be validated more carefully! For context, [numba v0.65.0](https://numba.readthedocs.io/en/stable/release/0.65.0-notes.html) now supports free threading, so building a fully free-threaded shap is possible. Should we disable the free-threading support for now?
- **Universal macOS build:** macOS can bundle `x86_64` and `arm64` binaries into one `universal2` format, but at the cost of increased size (`shap` binary is less than 1MB actually). Should we enable it?
- **Windows ARM64 support:** Windows on ARM64 is becoming more popular. Should we enable the build support too?

A little bit more context on the Stable ABI and `universal2`, if we opt into Stable ABI, we can reduce 30+ binaries to as few as just six:
- Linux: (x86_64, aarch64) x (manylinux_2_28, musllinux_1_2)
- Windows: AMD64 (+ARM64?)
- macOS: universal2

@CloseChoice Let me know what you think on those points. Thanks.

## Checklist

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -18,9 +18,10 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: "3.12"
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
       - name: Install dependencies for dataset download
         run: |
-          python -m pip install --upgrade uv
           uv pip install --system '.[download_datasets]'
       - name: Pre-download datasets
         run: python tests/datasets_to_cache.py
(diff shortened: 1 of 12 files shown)


daidahao:
readthedocs CI has been updated.

@CloseChoice Could you start the test CIs here?

daidahao:
Just updated the test (`run_tests`) and build (`build_wheels`) CIs to 1) set up and use uv where possible, 2) reduce the number of jobs needed for building wheels.

@CloseChoice

CloseChoice:
thanks @daidahao this looks great. Will merge and try to finish up the remaining things. Great that you checked out the Stable ABI/Limited ABI, that is really helpful for us.

PR #4458, opened 2026-04-10
Resolve: FutureWarning: The NumPy global RNG was seeded by calling `np.random.seed`. In a future version this function will no longer use the global RNG. Pass `rng` explicitly to opt-in to the new behaviour and silence this warning.

shaivimalik:
## Overview

Closes (part of) #3066  

[`summary_legacy`](https://github.com/shap/shap/blob/71d04ab53d3e74b6b47e7709aabfc1ef479adee3/shap/plots/_beeswarm.py#L521) raises this warning when global NumPy random state is set and `rng` is not passed. This warning can be resolved by passing `rng` to the `summary_plot`/`summary_legacy` function calls.

`rng` is also used to generate random numbers wherever needed in `test_summary.py` in place of legacy numpy methods.

`summary_legacy` supports other datatypes such `int` being passed as `rng` as evident from the code and [docstring](https://github.com/shap/shap/blob/d8c78502bdd0369f30617fe7751c5c77b1b7076a/shap/plots/_beeswarm.py#L579). However, the type hint is a bit inaccurate, so I have modified it to support all input types accepted by `np.random.default_rng` in numpy>=2.0

## Checklist


shaivimalik:
```
FAILED tests/plots/test_summary.py::test_summary_violin_with_data - Failed: Error: Image files did not match.
FAILED tests/plots/test_summary.py::test_summary_layered_violin_with_data - Failed: Error: Image files did not match.
```
The warnings have been resolved, but for some reason these two tests are failing.

CloseChoice (changes requested):
Thanks for the PR, this looks good overall. Have you recreated all the pngs or just the ones that fail in the test? I can't review all of them here on github, the ones I viewed looked fine, but could you keep this minimal pls?

shaivimalik:
> Thanks for the PR, this looks good overall. Have you recreated all the pngs or just the ones that fail in the test? I can't review all of them here on github, the ones I viewed looked fine, but could you keep this minimal pls?

Sure, let me restore the pngs that don't fail tests. I did recreate all the pngs earlier.

tests/plots/baseline/test_summary_compact_dot_with_data.png
CloseChoice, line 1: any idea why the images are ordered differently?

tests/plots/baseline/test_summary_dot_with_data.png
CloseChoice, line 1: same here

tests/plots/baseline/test_summary_multi_class.png
CloseChoice, line 1: was there a bug or did you introduce one? Tbh this looks more like you introduced one.

tests/plots/baseline/test_summary_multi_class_legend.png
CloseChoice, line 1: can you check why the switch in features, I assume all of this has the same root cause.

CloseChoice (changes requested):
I understand that we change the tests due to new rng, but some images look suspicious. Can you take another look?

Would it be possible to find the seed that was used before so that we don't have to change images at all?

tests/plots/baseline/test_summary_dot_with_data.png
shaivimalik: Most probably, this line is just shuffling them differently than before: [`rng.shuffle(inds)`](https://github.com/shap/shap/blob/1cd3883962f10d05bd1f1a5ab41a40134a3ce4e4/shap/plots/_beeswarm.py#L795)

tests/plots/baseline/test_summary_multi_class.png
shaivimalik: Might not be a bug, just intended behaviour? Could be result of [this](https://github.com/shap/shap/blob/1cd3883962f10d05bd1f1a5ab41a40134a3ce4e4/shap/plots/_beeswarm.py#L767) and [this](https://github.com/shap/shap/blob/1cd3883962f10d05bd1f1a5ab41a40134a3ce4e4/shap/plots/_beeswarm.py#L1070)

shaivimalik:
> I understand that we change the tests due to new rng, but some images look suspicious. Can you take another look?
> 
> Would it be possible to find the seed that was used before so that we don't have to change images at all?

I don't think we can use the old seed and get the same data. Tha images are changed due to new random data and feature shuffling/ordering logic defined inside of `_beeswarm.py`.

PR #4464, opened 2026-04-10
FIX: Replace bare assert with ValueError in violin()

### LINKED ISSUE #4463: BUG: Replace bare assert with ValueError in shap/plots/_violin.py
### Issue Description

`shap/plots/_violin.py:129` contains a bare `assert` to guard against a 1-D SHAP values array:

```python
assert len(shap_values.shape) != 1, "Violin summary plots need a matrix of shap_values, not a vector."
```

`assert` statements are stripped when Python runs with `-O` (optimized mode) and are considered bad practice for input validation in library code. Every other guard in the same file uses `raise ValueError(emsg)` explicitly — this line is inconsistent.

It should raise `ValueError` unconditionally, matching the pattern already used at lines 121 and 127 of the same file.

```python
import numpy as np
import shap

# This silently passes with python -O, which is wrong
shap.plots.violin(np.random.randn(5), show=False)
```

### Traceback

```
AssertionError: Violin summary plots need a matrix of shap_values, not a vector.
```

### Expected Behavior

```
ValueError: Violin summary plots need a matrix of shap_values, not a vector.
```

### Bug report checklist

### Installed Versions

Latest master.

alikhere:
## Overview

Closes #4463 

Replaced the bare `assert` on line 129 of `shap/plots/_violin.py` with a`raise ValueError` to match the pattern already used at lines 121 and 127 of the same file.

## Checklist

### DIFF
--- a/shap/plots/_violin.py
+++ b/shap/plots/_violin.py
@@ -126,7 +126,9 @@ def violin(
         emsg = f"plot_type: Expected one of ('violin','layered_violin'), received {plot_type} instead."
         raise ValueError(emsg)
 
-    assert len(shap_values.shape) != 1, "Violin summary plots need a matrix of shap_values, not a vector."
+    if len(shap_values.shape) == 1:
+        emsg = "Violin summary plots need a matrix of shap_values, not a vector."
+        raise ValueError(emsg)
 
     # default color:
     if color is None:
--- a/shap/plots/_violin.py
+++ b/shap/plots/_violin.py
@@ -9,6 +9,7 @@ import pandas as pd
 from packaging import version
 from scipy.stats import gaussian_kde
 
+from .. import Explanation
 from ..utils._exceptions import DimensionError
 from . import colors
 from ._labels import labels
@@ -106,7 +107,7 @@ def violin(
     if title is not None:
         warnings.warn("The `title` argument is unused and will be removed in a future release.", DeprecationWarning)
     # support passing an explanation object
-    if str(type(shap_values)).endswith("Explanation'>"):
+    if isinstance(shap_values, Explanation):
         shap_exp = shap_values
         shap_values = shap_exp.values
         if features is None:
@@ -126,7 +127,7 @@ def violin(
         emsg = f"plot_type: Expected one of ('violin','layered_violin'), received {plot_type} instead."
         raise ValueError(emsg)
 
-    if len(shap_values.shape) == 1:
+    if shap_values.ndim < 2:
         emsg = "Violin summary plots need a matrix of shap_values, not a vector."
         raise ValueError(emsg)
 
--- a/tests/plots/test_violin.py
+++ b/tests/plots/test_violin.py
@@ -11,6 +11,12 @@ def test_violin_with_invalid_plot_type():
         shap.plots.violin(np.random.randn(20, 5), plot_type="nonsense")
 
 
+def test_violin_1d_shap_values_raises():
+    """Check that a 1-D shap_values array raises ValueError, not AssertionError."""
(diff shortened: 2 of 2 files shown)


tests/plots/test_violin.py
```diff
@@ -11,6 +11,12 @@ def test_violin_with_invalid_plot_type():
         shap.plots.violin(np.random.randn(20, 5), plot_type="nonsense")
 
 
+def test_violin_1d_shap_values_raises():
+    """Check that a 1-D shap_values array raises ValueError, not AssertionError."""
+    with pytest.raises(ValueError, match="matrix of shap_values"):
+        shap.plots.violin(np.random.randn(5), show=False)
```
tylerjereddy, line 17: I suppose it may not matter much in this particular case, but generally one should pin the random seed of values used for testing.

Consider, for example, the scenario where `violin` does some other kind of input validation on the properties of the value in the array, which randomly leads to a different kind of error here, causing the testsuite to be flaky.

That's perhaps rather picky in this case though.

shap/plots/_violin.py
```diff
@@ -126,7 +126,9 @@ def violin(
         emsg = f"plot_type: Expected one of ('violin','layered_violin'), received {plot_type} instead."
         raise ValueError(emsg)
 
-    assert len(shap_values.shape) != 1, "Violin summary plots need a matrix of shap_values, not a vector."
+    if len(shap_values.shape) == 1:
```
tylerjereddy, line 129: I suppose this is consistent with the original logic. I might intuitively go with `shap_values.ndim < 2`.

shap/plots/_violin.py
```diff
@@ -126,7 +126,9 @@ def violin(
...
+        emsg = "Violin summary plots need a matrix of shap_values, not a vector."
```
tylerjereddy, line 130: I guess this may have been the error message for a long time. This could cause some confusion with the now-discouraged `np.matrix` vs. multi-dimensional array requirement.

tylerjereddy (commented):
Looks sensible I think--I added some nitpicky comments/suggestions.

shap/plots/_violin.py
CloseChoice: Users might hand over either (`explanation_obj` or `shap_values`):

```python
explainer = shap.Explainer(some_model)
explanation_obj = explainer(some_data)
shap_values = explainer.shap_values(some_data)
```
The explanation class exposes the shap values as `explanation_obj.values`. 

@alikhere could you pls check at the very beginning if `isinstance(shap_values, shap.Explanation)` and convert to a numpy array if that's the case? Then we can be sure what we are dealing with here and this would also make @tylerjereddy 's suggestion safe.

CloseChoice (changes requested):

tests/plots/test_violin.py
alikhere: Good catch, switched to np.random.default_rng(0) to pin the seed.

shap/plots/_violin.py
alikhere: Agreed, shap_values.ndim < 2 is cleaner.  Fixed in the [f3091be](https://github.com/shap/shap/pull/4464/commits/f3091be18b8d850262038decda00468d517ba270)

shap/plots/_violin.py
alikhere: Left the message unchanged for now to keep the PR minimal

alikhere:
@CloseChoice Done! added isinstance(shap_values, Explanation) check at the top and imported Explanation from .. matching the pattern in _beeswarm.py and  _bar.py. Feature extraction is preserved so passing an Explanation object  still works end-to-end.

PR #4466, opened 2026-04-11
FIX: Replace bare assert statements with DimensionError in shap/plots/_scatter.py

### LINKED ISSUE #4465: BUG: Replace bare assert statements with proper exceptions in shap/plots/_scatter.py
### Issue Description

`shap/plots/_scatter.py` contains four bare `assert` statements (lines 248, 251, 683, 686) that validate shape compatibility between `shap_values` and `features`. Like the violin issue, `assert` is stripped in optimized mode and produces `AssertionError` rather than an informative exception type.

The pattern already used elsewhere in the plots package is to raise `DimensionError` (from `shap.utils._exceptions`) for shape mismatches, which is what `_violin.py` and `_beeswarm.py` already do.

```python
# current (line 248-252)
assert shap_values_arr.shape[0] == features.shape[0], (
    "'shap_values_arr' and 'features' values must have the same number of rows!"
)
```

Should become:

```python
if shap_values_arr.shape[0] != features.shape[0]:
    raise DimensionError("'shap_values_arr' and 'features' values must have the same number of rows!")
```

```python
import numpy as np
import shap

sv = shap.Explanation(values=np.random.randn(10, 3), data=np.random.randn(5, 3))
shap.plots.scatter(sv[:, 0], show=False)
```

### Expected Behavior

`DimensionError` (or `ValueError`) with a clear message, not `AssertionError`.

### Bug report checklist

alikhere:
## Overview

Closes #4465 

`shap/plots/_scatter.py` had four bare `assert` statements in `scatter()` and `dependence_legacy()` validating shape compatibility between shap values and features. `assert` is stripped in optimized mode (`python -O`) and produces `AssertionError` instead of an informative exception.

Replaced all four with `if / raise DimensionError(...)`, matching the pattern already used in `_violin.py` and `_beeswarm.py`. `DimensionError` was already imported — no new imports needed. Error messages are unchanged.

Added three tests covering row and column mismatch paths in both functions.

## Checklist

### DIFF
--- a/shap/plots/_scatter.py
+++ b/shap/plots/_scatter.py
@@ -245,12 +245,14 @@ def scatter(
         figsize = (7.5, 5) if interaction_index != ind and interaction_index is not None else (6, 5)
         _, ax = plt.subplots(figsize=figsize)
 
-    assert shap_values_arr.shape[0] == features.shape[0], (
-        "'shap_values_arr' and 'features' values must have the same number of rows!"
-    )
-    assert shap_values_arr.shape[1] == features.shape[1], (
-        "'shap_values_arr' must have the same number of columns as 'features'!"
-    )
+    if shap_values_arr.shape[0] != features.shape[0]:
+        raise DimensionError(
+            "'shap_values_arr' and 'features' values must have the same number of rows!"
+        )
+    if shap_values_arr.shape[1] != features.shape[1]:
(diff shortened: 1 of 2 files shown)


alikhere:
Hi @CloseChoice, PTAL

PR #4474, opened 2026-04-11
Fix TreeExplainer crash when input is provided as Python list

### LINKED ISSUE #4473: BUG: TreeExplainer crashes when input is provided as Python list instead of array-like
### Issue Description

When passing input as a plain python list to TreeExplainer.shap_values , the function crashes with an AttributeError instead of handling the input or raising a clear error.
This seems to happen because _validate_inputs assumes the input has a .shape attribute, which is not true for lists.

interestingly, numpy arrays and pandas inputs seem to work fine, so this leads to inconsistent handling of common “array-like” inputs.

```python
import shap
import numpy as np
from sklearn.tree import DecisionTreeRegressor
X = np.random.rand(100, 4)
y = np.random.rand(100)
model = DecisionTreeRegressor().fit(X, y)
sample = list(X[0])
shap_values = explainer.shap_values(sample)
```

### Traceback

```shell
AttributeError: 'list' object has no attribute 'shape'
```

### Expected Behavior

The explainer should either:

- handle list inputs by converting them internally to numpy arrays, or
- raise a clear and user-friendly error message indicating expected input type
instead of failing with an AttributeError

### Bug report checklist

### Installed Versions

- 0.51.0
- python: 3.14

Aryan-Pillai7:
## Overview

Closes #4473

Fixes a crash in TreeExplainer when input is provided as a Python list instead of an array-like object.
 
## Changes
Currently, passing a list (e.g. list(X[0])) to TreeExplainer.shap_values() results in:
`AttributeError: 'list' object has no attribute 'shape'`
this happens because _validate_inputs assumes the input has a .shape attribute.

This PR adds a simple input validation step to convert array-like inputs (such as Python lists) into numpy arrays before further processing by
```
if not hasattr(X, "shape"):	
	X = np.array(X)
```

## Checklist

### DIFF
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -449,6 +449,7 @@ class TreeExplainer(Explainer):
         check_additivity: bool,
     ) -> tuple[npt.NDArray[Any], npt.NDArray[Any] | pd.Series | None, npt.NDArray[np.bool_], bool, int, bool]:
         import numpy as np
+
         if not hasattr(X, "shape"):
             X = np.array(X)
         # see if we have a default tree_limit in place.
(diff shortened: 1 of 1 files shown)


CloseChoice:
Thanks for the PR, be there are type hints in place so if users hand over unsupported data types, I don't think we should try to infer what they actually wanted to do. Am closing this therefore.

PR #4476, opened 2026-04-11
MAINT: Refactor plot tests and internal functions to use local RandomState (Fix #4453)

### LINKED ISSUE #4453: [MAINT]: Refactor internal tests to modern NumPy Generator API (avoiding global RNG seed)
Description
While running the test suite locally for ESoC 2026, I observed several `FutureWarning` logs in `tests/plots/test_summary.py` regarding the use of `np.random.seed`. 
Example warning:
`FutureWarning: The NumPy global RNG was seeded by calling np.random.seed. In a future version this function will no longer use the global RNG. Pass rng explicitly to opt-in to the new behaviour.`

Proposed Change
I noticed that PR #4442 is currently addressing 3rd-party deprecations via suppression. However, for internal SHAP tests (specifically in the plotting test suite), I would like to properly refactor the code to use the modern `np.random.default_rng()` API instead of global seeding. 

I believe refactoring these instances to use a local Generator is a more "future-proof" approach than suppression for internal SHAP code. I have identified the relevant lines in `test_summary.py` and would like to contribute a PR to implement these changes if the maintainers agree.

Pakeeza1508:
### Overview
Closes #4453

This PR fixes the FutureWarning regarding global NumPy seeding in the plotting tests.
I have updated the tests in test_summary.py, test_beeswarm.py, test_force.py, and test_group_difference.py to use local random generators. I specifically used np.random.RandomState to ensure that the plots stay identical to the original versions, so all image comparison tests still pass.
I also updated a few internal calls in _beeswarm.py and _group_difference.py to handle the local generator correctly.

## Checklist

- [x] Unit tests pass locally (Refactored existing plotting tests).


CloseChoice (changes requested):
usage of randomstate is legacy in itself, so we can't use that: https://numpy.org/doc/stable/reference/random/legacy.html

Pakeeza1508:
Hi @CloseChoice, thanks for the feedback. I’ve updated the PR to fully migrate from the legacy RandomState to the modern default_rng / Generator API across the plotting suite as you suggested.
Since the random sequences changed, I've regenerated all the baseline PNGs so the image comparison tests stay green. I also merged the latest master branch and fixed a couple of linting errors in the stub files that popped up during the merge.
I've verified everything locally with a full test run (508 passed). Ready for another look whenever you have time!

PR #4477, opened 2026-04-11
Fix KernelExplainer crash on empty input

### LINKED ISSUE #4475: BUG: KernelExplainer crashes on empty input while TreeExplainer handles it gracefully
### Issue Description

Upon passing an empty input (e.g. array with 0 rows) to KernelExplainer.shap_values, it crashes with an IndexError.
From what I can tell, this happens because no explanations get generated internally, but the code still tries to access the first element of that list (code later tries to access explanations[0] without checking)

What’s a bit odd is that TreeExplainer doesn’t have this problem — it just returns an empty list for the same input. so behavior isn’t entirely consistent across explainers right now.

```python
import shap
import numpy as np
from sklearn.linear_model import LinearRegression

X = np.random.rand(50, 3)
y = np.random.rand(50)
model = LinearRegression().fit(X, y)
empty = X[:0]
explainer.shap_values(empty)
```

### Traceback

```shell
IndexError: list index out of range
```

### Expected Behavior

Shouldn’t crash on empty input. Either 

- return an empty result (like TreeExplainer does)
- raise a clearer error.

### Bug report checklist

### Installed Versions

shap: v0.51.0
python: 3.14

Aryan-Pillai7:
## Overview

Closes #4475

This PR fixes a crash in KernelExplainer.shap_values when the input contains zero samples (empty dataset).
Currently, passing an empty array (e.g. shape (0, n_features)) results in:
`IndexError: list index out of range`
This occurs because the internal explanations list remains empty, but the code attempts to access explanations[0] without validating its length.
In contrast, TreeExplainer handles empty inputs gracefully and returns an empty result. This PR aligns KernelExplainer behavior with that expectation.

## Changes

- Added a guard condition in shap/explainers/_kernel.py to check if explanations is empty before accessing it
- If no explanations are generated, the function now returns an empty list instead of raising an exception
- Ensures consistent behavior across explainers when handling empty inputs

## Behavious After Fix

- Behavior matches TreeExplainer for similar inputs
- no crash occurs

## Checklist


CloseChoice (changes requested):
I'd like to follow fail fast here. Pls rather raise an error if the input matrix is empty, and pls do that for the all explainers and in the generic `shap.Explainer` interface. We'd also need tests for all these cases then.

Aryan-Pillai7:
Thanks for the clarification. My initial approach was aimed at matching the existing `TreeExplainer` behavior for empty inputs and also avoiding hard failure, but I do agree that explicit fail-fast validation was a lot more consistent and standard here.

I’ve updated the PR accordingly and added tests for the cases.

PR #4479, opened 2026-04-11
Replace assert with TypeError for invalid input in KernelExplainer

### LINKED ISSUE #4478: BUG: KernelExplainer uses assert for input validation leading to improper error handling
### Issue Description

While testing and working with KernelExplainer.shap_values, I noticed that invalid input types (for Ex: strings, dicts, lists, None) trigger an AssertionError instead of a more appropriate exception.
This comes from the following line in `shap/explainers/_kernel.py` :
`assert x_type.endswith(arr_type) or scipy.sparse.isspmatrix_lil(X), "Unknown instance type: " + x_type`

using `assert` as has been used frequently here is problematic as:

- assertions can be disabled when Python is run with optimization flags (-O)
- AssertionError is not intended for user-facing input validation
- it leads to inconsistent error handling across the library

```python
import shap
import numpy as np
from sklearn.linear_model import LinearRegression
X = np.random.rand(10, 3)
y = np.random.rand(10)
model = LinearRegression().fit(X, y)
explainer.shap_values("invalid")
```

### Traceback

```shell
AssertionError: Unknown instance type: <class 'str'>
```

### Expected Behavior

Invalid input types should raise a better exception such as `TypeError` which will provide us wit:

- more consistent behavior with standard Python practices
- reliability even when assertions are disabled
- correct semantics for invalid input

### Bug report checklist

### Installed Versions

shap: v0.51.0

Aryan-Pillai7:
## Overview

Closes #4478

This PR improves error handling in` KernelExplainer.shap_values `by replacing a bare assert statement with an explicit exception.
Currently, invalid input types (e.g. str, dict, list, None) trigger an AssertionError due to the use of:

`assert condition, "Unknown instance type"` style format

This leads to multiple issues:
- AssertionError is not intended for user input validation
- assertions can be disabled with Python optimization flags (-O)
- inconsistent error handling compared to other parts of the library

also, error paths like these seem to be encountered in KernelExplainer quite frequently, maybe due to higher user-input values, increasing the likelihood of invalid input types being passed

## Changes

- Replaced the following assertion in shap/explainers/_kernel.py:

` assert x_type.endswith(arr_type) or scipy.sparse.isspmatrix_lil(X), "Unknown instance type: " + x_type`

- with explicit validation by:

```
 if not (x_type.endswith(arr_type) or scipy.sparse.isspmatrix_lil(X)):
    raise TypeError(f"Unknown instance type: {x_type}")
```

## Behaviour after fix

- Invalid input types now raise a TypeError instead of AssertionError
- Behavior is consistent with standard Python practices
- Validation remains active even when assertions are disabled
- Improves reliability for common KernelExplainer usage scenarios

## Checklist


CloseChoice (changes requested):
if we touch this code, can we pls clean this up properly? I am in general not a fan of this `str(type(x)).endswith(...)` construct but would favor a clean `isinstance(x, ...)` of the right base type.

Pls also add a test for this case.

Aryan-Pillai7:
thanks for the suggestion , I followed up by updating the validation logic to remove the string-based type checking and replaced it with explicit `isinstance(...)` checks while preserving sparse matrix support. 

I also did add a regression test covering invalid input types passed to `KernelExplainer.shap_values`, following convention.

PR #4488, opened 2026-04-11
FIX: test for multi row input logic for shap/plots/_text.py

Spandan-Mishra:
## Overview

Progress towards #3690 

- This PR adds a test for shap/plots/_text.py and checks for a condition when multi row inputs are provided
- Increases coverage of above file from 54% -> 59%

## Checklist

### DIFF
--- a/tests/plots/test_text.py
+++ b/tests/plots/test_text.py
@@ -29,3 +29,19 @@ def test_single_text_to_text():
         hierarchical_values=[test_hierarchical_values],
     )
     shap.plots.text(shap_values_test)
+
+
+def test_multi_row_values():
+    """For multi row text plot logic"""
+    test_values = np.random.randn(2, 3)
+
+    test_base_values = np.array([0.1, 0.3])
+
+    test_data = np.array([["Hello", "world", " "], ["Testing", "text", "plot"]])
+
+    test_output_names = "Positive Sentiment"
+
+    shap_values_test = shap.Explanation(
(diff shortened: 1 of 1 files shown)


tests/plots/test_text.py
```diff
@@ -29,3 +29,19 @@ def test_single_text_to_text():
         hierarchical_values=[test_hierarchical_values],
     )
     shap.plots.text(shap_values_test)
+
+
+def test_multi_row_values():
+    """For multi row text plot logic"""
+    test_values = np.random.randn(2, 3)
```
CloseChoice: we need to set the seed here.

CloseChoice (changes requested):

tests/plots/test_text.py
```diff
@@ -29,3 +29,19 @@ def test_single_text_to_text():
...
+    test_base_values = np.array([0.1, 0.3])
+
+    test_data = np.array([["Hello", "world", " "], ["Testing", "text", "plot"]])
+
+    test_output_names = "Positive Sentiment"
+
+    shap_values_test = shap.Explanation(
+        values=test_values, base_values=test_base_values, data=test_data, output_names=test_output_names
+    )
+    shap.plots.text(shap_values_test, display=False)
```
CloseChoice: we also need to assert against concrete values here

CloseChoice (changes requested):

PR #4497, opened 2026-04-12
TEST: improve coverage for `GradientExplainer`

naxatra2:
## Overview

Related to https://github.com/shap/shap/issues/3690

This PR improves test coverage for `shap/explainers/_gradient.py` by adding targeted unit tests for previously untested behaviors in `GradientExplainer`.

## Checklist

- [x ] Unit tests added (if fixing a bug or adding a new feature)

### DIFF
--- a/tests/explainers/test_gradient.py
+++ b/tests/explainers/test_gradient.py
@@ -236,6 +236,11 @@ def test_tf_multi_inputs_multi_outputs():
     predicted = model.predict([input1_data, input2_data])
     explainer = shap.GradientExplainer(model, [input1_data, input2_data])
     shap_values = explainer.shap_values([input1_data, input2_data])
+
+    assert len(shap_values) == 2
+    assert shap_values[0].shape == (32, 3, 3)
+    assert shap_values[1].shape == (32, 4, 3)
+
     np.testing.assert_allclose(shap_values[0].sum(1) + shap_values[1].sum(1) + predicted.mean(0), predicted, atol=1e-1)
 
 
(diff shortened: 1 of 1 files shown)


tests/explainers/test_gradient.py
```diff
@@ -483,3 +483,100 @@ def test_tf_input(random_seed, input_type):
...
+    X = np.random.randn(20, 4)
+    explainer = shap.GradientExplainer(model, X[:10])
+
+    shap_vals, ranks = explainer.shap_values(X[:1], ranked_outputs=2)
+
+    assert ranks.shape[1] == 2
+    assert shap_vals.shape[0] == 1
+
+
+def test_local_smoothing_changes_output():
```
CloseChoice: I don't really see a value in this. There isn't an explicit bug report mentioning problems with this is it? I would prefer to test things like this explicitly by asserting against pre-calculated shap values for different calls of `local_smoothing`. At best the results can be derived analytically before hand.

tests/explainers/test_gradient.py
```diff
@@ -483,3 +483,100 @@ def test_tf_input(random_seed, input_type):
...
+    e2 = shap.GradientExplainer(model, X[:10], local_smoothing=0.1)
+
+    vals1 = e1.shap_values(X[:1])
+    vals2 = e2.shap_values(X[:1])
+
+    assert not np.allclose(vals1, vals2)
+
+
+def test_output_shape_consistency():
+    import numpy as np
```
CloseChoice: Wasn't this branch covered elsewhere? Wouldn't it make sense to simply add the asserts in existing tests?

tests/explainers/test_gradient.py
```diff
@@ -483,3 +483,100 @@ def test_tf_input(random_seed, input_type):
...
+    explainer = shap.GradientExplainer(model, background)
+
+    vals1 = explainer.shap_values(X[:1], rseed=42)
+    vals2 = explainer.shap_values(X[:1], rseed=42)
+
+    np.testing.assert_allclose(vals1, vals2, atol=1e-6)
+
+
+def test_invalid_input_shape_raises():
+    import numpy as np
```
CloseChoice: if this assertion wasn't covered then let's keep this test

tests/explainers/test_gradient.py
```diff
@@ -483,3 +483,100 @@ def test_tf_input(random_seed, input_type):
...
+
+    X = np.random.randn(20, 3)
+    background = X[:10]
+
+    explainer = shap.GradientExplainer(model, background)
+
+    vals1 = explainer.shap_values(X[:1], rseed=42)
+    vals2 = explainer.shap_values(X[:1], rseed=42)
+
+    np.testing.assert_allclose(vals1, vals2, atol=1e-6)
```
CloseChoice, line 494: we don't have anything like this already? I haven't checked but since I see the import statments in each tests I assume this is LLM generated and am therefore critical of test creep.

CloseChoice (changes requested):
Pls keep your changes minimal! Also remove all redundant tests.

tests/explainers/test_gradient.py
naxatra2: Yes, I used an LLM to help draft the test structure. I've now cleaned up the repetitive local imports.
Sorry, I didn't cross-check the import part earlier.

tests/explainers/test_gradient.py
naxatra2: I didn't know this was covered but now I've removed this standalone test and instead added these shape assertions to the end of `test_tf_multi_inputs_multi_outputs`

tests/explainers/test_gradient.py
naxatra2: I've kept this one as it covers a specific error-handling branch that wasn't explicitly tested.

tests/explainers/test_gradient.py
naxatra2: okay, I've just removed the test to keep the PR focused on the core validation.

Issue #652, opened 2019-06-19
Explainer.expected_value changes after calling shap_values() method


BasilBeirouti:
I am using a GBC from sklearn:
`from sklearn.ensemble import GradientBoostingClassifier as GBC`

In both python2 and python3 environments, the `explainer.expected_value` changes from a `array([-0.07683372], dtype=float32)` to a `numpy.float64` after calling `explainer.shap_values(X)`.

In python2 environment, it's actual value also changes:
```
explainer = shap.TreeExplainer(model)
explainer.expected_value
>> array([-0.07683372], dtype=float32)
shap_values = explainer.shap_values(X)
explainer.expected_value
>> -0.32417948156786575
```
In summary: In both python 2 and python 3 environments, the type of `explainer.expected_value` changes. In python 2, I also see that its value changes. This of course results in the `force_plot` output changing as well. 

Python 2 environment was installed with conda, I have attached the relevant 
environment info here (relevant output of conda list): [py2env_shap.txt](https://github.com/slundberg/shap/files/3307497/py2env_shap.txt)

python 3 environment was installed with pip. This is the full output of pip list:
[py3_requirements_shap.txt](https://github.com/slundberg/shap/files/3307508/py3_requirements_shap.txt)

Please let me know if you need more info. I cannot provide the full python 2 environment but I can answer specific questions about it.

slundberg:
Hmmm...the type changing is something I can fix easily. The py2 issue is stranger to me. shap does not support py2 anymore, but I don't want older version to have an issue there either. Do you know which of the before or after is right in the py2 example you saw? I am confused how this would happen since we have a unit test for it (unless you are using a much older version of shap for py2).

BasilBeirouti:
I forgot to include the SHAP version for py2!  Here is the corresponding line:
`shap                      0.28.5           py27h637b7d7_0    conda-forge`

Thank you for the quick response! Could you tell me how to figure out what is the correct value? 
shap is explaining the log(odds-ratio) in this case I believe. Here is the code I think I should use to do so (note I am using weights, not sure if that matters here):

```
params = {'criterion': 'friedman_mse',
 'init': None,
 'learning_rate': 0.1,
 'loss': 'deviance',
 'max_depth': 4,
 'max_features': None,
 'max_leaf_nodes': None,
 'min_impurity_decrease': 0.0,
 'min_impurity_split': None,
 'min_samples_leaf': 1,
 'min_samples_split': 2,
[...]

PR #4498, opened 2026-04-12
Fix/tree explainer expected value initialization

### LINKED ISSUE #4495: BUG: TreeExplainer.shap_values() mutates expected_value for XGBoost models
### Issue Description

`TreeExplainer.expected_value` is computed incorrectly at initialization for certain XGBoost model types, and is silently corrected as a side effect of calling `shap_values()`. This means:
1. `expected_value` before calling `shap_values()` is wrong
2. `expected_value` after calling `shap_values()` is correct
3. Users who read `expected_value` before calling `shap_values()` get an incorrect value with no warning
4. The shape also changes from (1,) to () which is a secondary type inconsistency issue

```python
import shap
import numpy as np
import xgboost as xgb

X, y = make_classification(n_samples=200, n_features=10, random_state=42)
X_train, X_test = X[:150], X[150:]
y_train = y[:150]

model = xgb.XGBClassifier(n_estimators=50, random_state=42)
model.fit(X_train, y_train)

print(f"expected_value before shap_values() : {explainer.expected_value}")
print(f"shape before                         : {np.array(explainer.expected_value).shape}")

shap_values = explainer.shap_values(X_test)

print(f"expected_value after shap_values()   : {explainer.expected_value}")
print(f"shape after                          : {np.array(explainer.expected_value).shape}")

# confirm the post-call value is the correct one
dmatrix = xgb.DMatrix(X_test)
raw_margin = model.get_booster().predict(dmatrix, output_margin=True)
print(f"\nMean raw margin (correct baseline)   : {raw_margin.mean():.8f}")
print(f"expected_value before call           : {np.array(explainer.expected_value - shap_values.mean(0).sum()):.8f}")
```

### Traceback

```shell
No exception or warning is raised. The bug manifests silently as a value and shape change:

    expected_value before shap_values() : [-0.02666822]   shape=(1,)
    expected_value after  shap_values() : -0.01702357     shape=()

Users who read expected_value before calling shap_values() receive an incorrect value with no indication anything is wrong. Users who read it after receive the correct value. The shape inconsistency (1,) vs () is also present regardless of which value is correct.

Additionally confirmed in a real-world dataset by @khteh in #4414:

    Before shap_values(): [0.16934096]   shape=(1,)
    After  shap_values(): 0.18604861     shape=()

```

### Expected Behavior

`expected_value` should be correct immediately after `shap.TreeExplainer(model)` initialization, without requiring a `shap_values()` call first. Its value and shape should remain consistent regardless of whether `shap_values()` has been called.

This issue was first reported in #652 (June 2019) by @BasilBeirouti, who confirmed that for certain XGBoost models the initialization value is incorrect and only becomes correct after calling `shap_values()`. @slundberg tagged it as `todo` at the time:

> "This is something that needs investigated/potentially fixed, 
> so I am adding a todo tag for now."

It was observed again by @mueslo in October 2025 and independently rediscovered through investigation of #4414 in April 2026.

**Important note for anyone attempting a fix:** A naive guard of `if self.expected_value is None` at the mutation sites in 
`shap_values()` breaks additivity for multiple XGBoost model types (XGBRFRegressor, XGBClassifier, XGBRFClassifier, XGBRanker, tweedie regression, dart regression). This was confirmed by running the existing test suite after applying the guard, 10 XGBoost tests failed with additivity errors. The correct fix is to repair the initialization path, so it computes the correct value upfront, rather than preventing the correction during `shap_values()`.

The secondary shape inconsistency - `(1,)` before vs `()` after - should also be normalized as part of any fix, regardless of which value is correct.

### Bug report checklist

- [ ] I have checked that this issue has not already been reported.

### Installed Versions

`0.51.0`

BaibhavKundu2005:
## What does this PR do?
Fixes incorrect `expected_value` initialization for XGBoost models in `TreeExplainer` and normalizes the shape to be consistent before and after calling `shap_values()`.

Closes #4495

When `data=None` and the model is XGBoost with `model_output="raw"`, `__init__` falls into the `node_sample_weight` branch and computes `expected_value` manually from parsed tree weights. This calculation is inaccurate - the correct baseline is embedded in the last column of XGBoost's own `pred_contribs` output from its C++ engine, which 
is the same source used inside `shap_values()`.

This caused two problems:
- `expected_value` value was wrong before calling `shap_values()`
- `expected_value` shape changed from `(1,)` to `()` after calling 
  `shap_values()`, causing broken assertions and confusing behavior

## Fix
At initialization, for XGBoost raw-output models with no background data, we now query XGBoost's C++ engine directly using `original_model.num_features()` for the correct feature count and a single dummy row to obtain the baseline via `pred_contribs`. This is identical to what `shap_values()` always did internally.

Key improvements over naive approaches:
- Uses `self.model.original_model.num_features()` for correct feature 
  count — `self.model.values.shape[1]` returns node count (e.g. 21), 
  not feature count (e.g. 10)
- Wraps scalar in `np.array([...])` to maintain consistent `(1,)` 
  shape before and after `shap_values()`
- Forwards `_xgb_dmatrix_props` to the dummy DMatrix, matching what 
  `shap_values()` does internally
- Explicit `UserWarning` on fallback instead of silently swallowing 
  exceptions

## Evidence
Before fix:
    expected_value at init        : [-0.02666822]  shape=(1,)
    expected_value after call     : -0.01702357    shape=()  ← wrong value AND shape

After fix:
    expected_value at init        : [-0.01702357]  shape=(1,)  ← correct
    expected_value after call     : [-0.01702357]  shape=(1,)  ← unchanged
    Additivity holds              : True
    Max additivity error          : 0.00000143

## History
First reported in #652 (June 2019) and tagged as `todo` by 
@slundberg. Rediscovered by @mueslo in October 2025. Independently 
reproduced and root-caused through investigation of #4414 in April 
2026, tracked in #4495.

## Checklist
- [x] I have read CONTRIBUTING.md
- [x] Pre-commit hooks pass
- [x] Unit test added
- [x] 104 tests pass, 11 skipped

AI Usage: Claude was used to track down the root cause of the issue. The fixes were reviewed, tested and understood by me.


DashratRajpurohit:
Hi @BaibhavKundu2005 — I've opened PR #4499 which takes a different approach 
to the same issue. Rather than normalizing the shape after the mutation, my 
PR tries to compute the correct value at `__init__` time using XGBoost's own 
`pred_contribs` C++ output, which is the same source `shap_values()` already 
uses.

I noticed you found that a naive `if self.expected_value is None` guard broke 
10 XGBoost tests. My fix is different — it doesn't guard the mutation, it 
corrects the source. I haven't been able to run the full test suite locally 
yet. If my approach also breaks tests, I'll close #4499 and your fix would be 
the right one.

Happy to collaborate — let me know what you think once you've had a chance to 
look at it!

BaibhavKundu2005:
@CloseChoice Hello, a ping for whenever you have some time. All the CI checks have passed. I'm an ESoC 2026 batch 2 applicant, contributing to SHAP as part of my evaluation. I'm willing to make changes, if required. Thanks!

BaibhavKundu2005:
@CloseChoice @DashratRajpurohit I'm glad my catch regarding the `original_model.num_features()` bug helped get #4499 across the finish line! Since #4499 has now incorporated that fix and successfully expanded the scope to include LightGBM and CatBoost as well, it definitely makes sense to merge that one.

I will go ahead and close this PR in favor of #4499 to keep the commit history clean. Happy to help to solve this long-standing bug.

PR #4499, opened 2026-04-12
fix: correct expected_value initialization for XGBoost models in TreeExplainer

DashratRajpurohit:
## What does this PR do?

Fixes a 7-year-old silent mutation bug (first reported in #652) where
`TreeExplainer.expected_value` was computed incorrectly at initialization
for XGBoost models when no background dataset is provided, and was then
silently overwritten with the correct value on the first `shap_values()` call.

## Root Cause

When `data=None` and the model is XGBoost, `__init__` falls into the
`node_sample_weight` branch and computes `expected_value` manually from
the parsed tree weights. This path is inaccurate for XGBoost — the correct
baseline is embedded in the **last column of XGBoost's own `pred_contribs`
output** from its C++ engine.

Because `shap_values()` always reads the baseline from that C++ output and
writes it back to `self.expected_value`, the wrong value was silently
corrected on every first call.  Any user who reads `expected_value` before
calling `shap_values()` receives a wrong value (wrong number AND wrong shape
`(1,)` vs `()`), with no warning.

## Fix

At initialization, for XGBoost raw-output models with no background data,
we now do exactly what `shap_values()` always did: ask XGBoost's C++
engine for the baseline via `pred_contribs` on a single dummy row.  A safe
fallback to the old manual calculation is kept in case of any unusual model
configuration.

## Confirmed by

- First silently confirmed in #652 (June 2019, tagged `todo` by @slundberg)
- Rediscovered in October 2025 by @mueslo
- Independently reproduced and root-caused through investigation of #4414 in April 2026, tracked in #4495

## Checklist
- [x] I have read the CONTRIBUTING.md
- [x] Pre-commit hooks pass
- [x] Documentation builds cleanly
- [x] This is a bug fix — existing XGBoost tests should confirm correctness


DashratRajpurohit:
Note: PR #4498 by @BaibhavKundu2005 addresses the same root issue (#4495) 
but with a different approach — normalizing only the shape inconsistency. 
This PR instead tries to fix the incorrect value itself at initialization, 
by querying XGBoost's C++ engine (the same source used inside shap_values()).

Still investigating whether this causes additivity failures in the existing 
test suite. Will update once confirmed.

BaibhavKundu2005:
Great work tracking down the root cause and fixing the initialization path.

I noticed one issue in the implementation: `n_features = self.model.values.shape[1]` returns the number of nodes per tree (21 in my test with 10 features), not the number of input features. The correct source is `self.model.original_model.num_features()`.

I've updated my PR #4496 with a similar initialization fix that uses this correction, also normalizes the shape to `(1,)` consistently, and adds an explicit `UserWarning` on fallback instead of a bare `except Exception`. Happy to coordinate with you and the maintainers on which approach to move forward with.

DashratRajpurohit:
Hi @CloseChoice, I've rebased this PR and resolved the conflicts. Note that this fix is more comprehensive than #4498 as it addresses the expected_value initialization bug across XGBoost, LightGBM, and CatBoost simultaneously, and adds documentation on the silent mutation issue. Ready for final review!

CloseChoice (changes requested):
Thanks for the PR, this looks pretty good and I am quite sure that I ran into this a while back myself. So thanks again. 

Could you pls add a test so that I can debug this properly?

CloseChoice (commented):
I assume you can simply use an example from the linked issue to write the test. Once you done that, could you pls run on master and paste the output in the discussion?

DashratRajpurohit:
Hi @CloseChoice,

I've pushed an update to address the feedback regarding the `expected_value` initialization and the fallback path transparency.

**Summary of updates:**
1. **Regression Test**: I've added a dedicated test case (`test_tree_explainer_expected_value_xgboost_4495`) that captures the mutation bug reported in GH #4495. It verifies that `expected_value` remains stable after multiple calls and maintains mathematical additivity.
2. **UserWarning Implementation**: I've integrated a `UserWarning` that triggers when the explainer is forced into the fallback initialization path for XGBoost models.
3. **Mathematical Consistency**: I've ensured that the fallback path uses the same normalization logic as the standard path to maintain consistency in results.

**The Test Case:**
I'm including the test code here for easier review. This test fails on the current `master` branch but passes with these PR changes:

```python
def test_tree_explainer_expected_value_xgboost_4495():
    """
    Regression test for GH #4495.
    Ensures that TreeExplainer.expected_value is correctly initialized for XGBoost models
    and does not mutate after calling shap_values().
    """
    import numpy as np
    import pytest
    import shap
    xgb = pytest.importorskip("xgboost")

    X, y = make_classification(n_samples=200, n_features=10, random_state=42)
    X_train, X_test = X[:150], X[150:]
    y_train = y[:150]

    model = xgb.XGBClassifier(n_estimators=50, random_state=42)
    model.fit(X_train, y_train)

    expected_value_before = explainer.expected_value

    # Trigger shap_values which previously mutated expected_value
    shap_values = explainer.shap_values(X_test)
    expected_value_after = explainer.expected_value

    # 1. Verify consistency (no mutation)
    assert np.isclose(expected_value_before, expected_value_after), \
        f"expected_value mutated from {expected_value_before} to {expected_value_after}"

    # 2. Verify correctness (matches raw margin mean)
    dmatrix = xgb.DMatrix(X_test)
    raw_margin = model.get_booster().predict(dmatrix, output_margin=True)

    # Check additivity for a few samples
    for i in range(5):
        assert np.isclose(explainer.expected_value + shap_values[i].sum(), raw_margin[i], atol=1e-5)
```

I've also cleaned up the accidental file bloating in the test suite to keep the PR focused and easy to review. Looking forward to your thoughts!

DashratRajpurohit:
Hi @CloseChoice, just a friendly ping when you have bandwidth. All feedback has been addressed and CI is green. Looking forward to your thoughts!

DashratRajpurohit:
Hi @CloseChoice, just a friendly ping when you have bandwidth. All feedback has been addressed and CI is green. Looking forward to your thoughts!

PR #4501, opened 2026-04-12
FIX: Correct file path handling in make_dir (image.py)

### LINKED ISSUE #4500: BUG: Bug in make_dir: incorrect path concatenation causes FileNotFoundError
### Issue Description

The function `make_dir` in `shap/utils/image.py` removes files using string concatenation:

    os.remove(path + file)

This produces incorrect paths (e.g., ".../newfile.txt" instead of ".../new/file.txt"),
leading to FileNotFoundError when clearing a non-empty directory.

Image

```python
import os
from shap.utils.image import make_dir

# create test directory
os.makedirs("test_dir", exist_ok=True)

# create a file inside it
with open("test_dir/file.txt", "w") as f:
    f.write("data")

# call function
make_dir("test_dir")
```

### Traceback

```shell
FileNotFoundError: [Errno 2] No such file or directory: 'test_dirfile.txt'
```

### Expected Behavior

- Files inside the directory should be removed correctly
- The directory should be emptied without errors

### Bug report checklist

### Installed Versions

- shap: latest (master branch)
- python: 3.12
- os: Windows

Abhishek-Kumar-Rai5:
## Overview
Closes #4500 
Fixes incorrect path construction in `make_dir` within `shap/utils/image.py`.

The function was removing files using string concatenation:

```python
os.remove(path + file)
```

## Fix :-  
Replaced string concatenation with proper path joining:
```python
os.remove(os.path.join(path, file))
```
## Validations done :-   
- Reproduced the bug using a directory containing files
- Verified that files are correctly removed after the fix
- Confirmed no errors are raised during execution
## Checklist

Tests for this fix will be added in the follow up PR. I encountered this issue while creating tests for image.py so flagged this first and now will make the follow up PR for it.

## AI usage disclosure 
I used no AI for the fixation. I ran the tests in PR #4503 and encountered this error so I fixed it then and there and raised this PR.

### DIFF
--- a/shap/utils/image.py
+++ b/shap/utils/image.py
@@ -40,7 +40,7 @@ def make_dir(path):
         if os.listdir(path):
             # if exists, empty directory
             for file in os.listdir(path):
-                os.remove(path + file)
+                os.remove(os.path.join(path, file))
 
 
 def add_sample_images(path):


CloseChoice (approved):
Thanks, looks good to me.

PR #4502, opened 2026-04-12
tests(plots): add test coverage for shap.plots._benchmark and fix two source bugs

crocmons:
## Overview

Closes #3690 

This PR introduces a comprehensive test suite for `shap/plots/_benchmark.py`, which previously had **4.80% coverage on Codecov**. Writing the tests also surfaced two latent bugs in the source module, which are fixed here.

### Changes proposed

**New file: `tests/plots/test_benchmark.py`**
- Covers all three plotting branches in `benchmark()`:
  - Single `BenchmarkResult` (non-iterable, line-plot path)
  - Iterable with a single shared metric **with** curves (line-plot path)
  - Iterable with a single shared metric **without** curves (`barh` path)
  - Iterable with multiple metrics (parallel-coordinates path)
- Verifies `BenchmarkResult` properties: `full_name`, `value_sign` defaults for all 8 metric types, custom `value_sign` override, and AUC auto-computation when `value=None`
- Validates the `xlabel_names` dictionary for completeness
- Covers edge cases: empty list input, tuple input, generator input, single-element list, and many-method color assignment
- Uses `matplotlib.use("Agg")` for headless CI compatibility
- Uses an `autouse` fixture (`plt.close("all")`) to prevent figure state leaking between tests
- Patches at the correct import site (`shap.plots._benchmark.colors`) to ensure mocks are effective

**Bug fixes in `shap/plots/_benchmark.py`**
1. **`UnboundLocalError` on empty list input** — an empty iterable skips all plotting sub-branches, leaving `ax` unassigned before it is referenced. Fixed with an early `return` guard.
2. **`ZeroDivisionError` in multi-metric normalization** — when all methods score identically on a metric, `max - min == 0` causes a divide-by-zero. Fixed by short-circuiting to `0.5` (mid-rank) when the range is zero, which is semantically correct since the methods are indistinguishable on that axis.

## Checklist
- [x] Unit tests added (fixing two bugs and adding missing coverage for `shap.plots._benchmark`)


tests/plots/test_benchmark.py
```diff
@@ -0,0 +1,425 @@
...
+
+
+@pytest.fixture(autouse=True)
+def close_figures():
+    """Close all matplotlib figures after every test to avoid state leakage."""
+    yield
+    plt.close("all")
+
+
+class TestBenchmarkResult:
```
CloseChoice: pls remove the class, we prefer function based tests.

tests/plots/test_benchmark.py
```diff
@@ -0,0 +1,425 @@
...
+        r = BenchmarkResult(metric="remove absolute", method="m", value=None, curve_x=x, curve_y=y, curve_y_std=std)
+        assert r.value is not None
+        assert isinstance(r.value, float)
+
+    def test_value_preserved_when_provided(self):
+        r = _result("remove absolute", "m", 0.42)
+        assert r.value == pytest.approx(0.42)
+
+
+class TestXlabelNames:
```
CloseChoice: same here.

CloseChoice (changes requested):

tests/plots/test_benchmark.py
crocmons: ok sure I'll do that

tests/plots/test_benchmark.py
crocmons: ok

PR #4510, opened 2026-04-12
fix: reindex DataFrame columns to match background data in KernelExpl…

### LINKED ISSUE #3958: BUG: Column labels not taken into account
### Issue Description

Perhaps I'm overlooking a crucial piece of information, but I wonder why `shap` does not take into account the column labels of the given data, even though it has access to them and the explainers even store them internally. In essence, this is a condensed version of the issue reported in #3601 with a minimal necessary context, i.e. without `xgboost` or similar.

The danger I see is that there is no mechanism that prevents people from passing unordered dataframes and getting completely wrong explanations for their data. 

Some feedback / guidance would be highly appreciated.

```python
import numpy as np
import pandas as pd
from shap import KernelExplainer
from shap.plots import bar

background_data = pd.DataFrame({"A": [1, 2, 3], "B": [4, 5, 6]})

def model(x: np.ndarray) -> np.ndarray:
    return x[:, 0]


explainer = KernelExplainer(model, data=background_data)
print(explainer.data_feature_names)  # <-- this gives ["A", "B"] as expected
x = explainer(background_data[["B", "A"]])  # <-- columns are permuted
bar(x)  # "incorrectly" identifies "B" as the determining feature
```

### Traceback

_No response_

### Expected Behavior

I would have expected that `shap` internally aligns the given dataset with the background data before computing the SHAP values and then reverts the permutation on the resulting object. In fact, for my own application, I've done so using a custom workaround, which requires some rather nasty lines of code. That is, while indexing seems generally supported a la `explanations[:, idx]`, it does not work for all types of explainers and data since the inner data types differ.

### Bug report checklist

### Installed Versions

shap==0.46.0

neha222222:
…ainer

Closes #3958

## Overview

Closes #XXXX  

## Checklist

### DIFF
--- a/shap/explainers/_kernel.py
+++ b/shap/explainers/_kernel.py
@@ -204,6 +204,10 @@ class KernelExplainer(Explainer):
         start_time = time.time()
 
         if isinstance(X, pd.DataFrame):
+            # reindex columns to match background data order if available
+            if hasattr(self, "data_feature_names") and self.data_feature_names is not None:
+                if list(X.columns) != self.data_feature_names:
+                    X = X.reindex(columns=self.data_feature_names)
             feature_names = list(X.columns)
         else:
             feature_names = getattr(self, "data_feature_names", None)  # type: ignore[assignment]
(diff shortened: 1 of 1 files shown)


CloseChoice (changes requested):
Pls add a test.

PR #4514, opened 2026-04-12
TEST: Improve test coverage for Action base class

Jiya873:
Related to #3690 

The previous test file lacked coverage for the base class's built-in properties, dunder methods, and expected exceptions. So, I have added testcases for these and renamed the `_action.py` filename in the tests directory as these files are often ignored by test discovery tools in CI environments, and renaming it ensures the tests are automatically detected and executed.


CloseChoice (approved):
LGTM, thanks for catching this

PR #4519, opened 2026-04-13
TST: Increase test coverage for shap/maskers/_image.py

bedupako12mas:
Closes #3690 (partial — `shap/maskers/_image.py` only)

## Summary

Adds tests for previously uncovered branches in ```shap/maskers/_image.py```, increasing coverage from 62% to 97% of coverable lines.

The @njit-decorated function _jit_build_partition_tree (lines 180-240) is excluded via `#pragma: no cover` — coverage.py cannot trace into numba JIT-compiled functions (numba/numba#4268).

Lines 15-16 (cv2 import error handler) are not tested — the test module skips entirely when cv2 is absent (via `pytestmark`), making these lines structurally unreachable within the test suite by design.

Tests added (WIP):

- [x] `test_init_string_mask_without_shape`
- [x] `test_init_ndarray_mask_without_shape`
- [x] `test_init_scalar_mask_with_shape`
- [x] `test_call_with_torch_tensor`
- [x] `test_call_image_masker_shape_mismatch`
- ~~`test_call_image_as_single_element_list`~~ — line 90 found unreachable (see Observations)
- [x] `test_call_image_masker_with_no_mask`
- [x] `test_inpaint_fully_masked_image`

## Observations

During test coverage work, an undocumented `mask_value` type was discovered. The `__init__` docstring only lists `np.array` and string options, but the `else` branch (line 62-63) also handles scalar integers/floats by expanding them into a flat array of that value. Passing an incompatible type (e.g. a list) raises a `ValueError` from NumPy with no descriptive error message.

The docstring on line 27 could be updated to:

    mask_value : np.array, int, float, "blur(kernel_xsize, kernel_xsize)", "inpaint_telea", or "inpaint_ns"

No fix is included in this PR — flagging for reviewer discretion.

---

During test coverage work, line 89-90 in `__call__` was found to be unreachable:

    if isinstance(x, list) and len(x) == 1:
        x = x[0]

This list unwrap runs after the shape check on line 79 (`np.prod(x.shape)`), which raises an `AttributeError` if `x` is a list since lists have no `.shape` attribute. Investigation of the codebase shows that multi-input unwrapping is handled at the explainer level (deep_pytorch, deep_tf, gradient), so a list is never passed directly to `Image.__call__`. This code appears to be either dead code or a misplaced check (should come before line 79).

Minimal reproducible example:
```
masker = shap.maskers.Image(np.zeros((5, 5, 3)))
masker(np.ones(48, dtype=bool), [np.zeros((5, 5, 3))])
# AttributeError: 'list' object has no attribute 'shape'
```

No fix is included in this PR — flagging for reviewer discretion.

## AI Disclosure

- **Advised**: Identified which lines were uncovered and suggested test strategies for each branch
- **Assisted**: Provided test templates as guidance; all tests were typed independently by the contributor (not copy-pasted), as reflected by manual typos caught during development
- **Autonomous**: None — all code was written and verified by the contributor


CloseChoice (approved):
Looks good to me. Let's get this in once the pipelines are green.

CloseChoice:
thanks for the effort @bedupako12mas

Issue #3171, opened 2023-08-01
ENH: Documentation : add links between Examples and API Reference


znacer:
### Problem Description

When using the documentation, it would be easier for the users to have a direct link between corresponding pages. Most likely in the top of the page.
This would make it a lot easier than exploring the sidebar tree. Especially knowing that when you try click on a category (API Reference/Example), it will change the page you are on.

### Feature request checklist

- [X] I have checked the issue tracker for duplicate issues.
- [ ] I'd be interested in making a PR to implement this feature

jtimko16:
Hello,

Could you please provide an example, of how it could look like? Perhaps a documentation of another Python package that follows the structure you have in your mind and could serve as an inspiration? Thanks

jtimko16:
@znacer Tagging you, in case you haven't seen the message above :)

znacer:
I think a good example would be the scikit-learn documentation.
When you go in the API documentation of an object or function, there is a section "Examples using {}".
e.g.: [RandomForestClassifier](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html#sklearn.ensemble.RandomForestClassifier)

jtimko16:
Thanks, I will look at it, test some ideas, and give some update in the next few days.

jtimko16:
Hi again @znacer,

Thanks for your patience. I am still trying to get to the bottom of it.

Here we have examples of Tree-based models with links to examples both:

1) In the sidebar 
2) below the section name:

<img width="725" alt="image" src="https://github.com/shap/shap/assets/61096773/4ab0a456-703d-4d44-9fc6-514f64bc4eca">


If we compare to the example of Scikit-learn, there are:

1) Subsection on the current page in the sidebar
2) Image links leading to other pages

<img width="964" alt="image" src="https://github.com/shap/shap/assets/61096773/18e5d723-0021-4519-ada7-78ab237ab8b1">




Therefore are we trying to implement something as the image links from the second screenshot?

Thanks for clarifying.

znacer:
Hi,
The documentation contains some API Examples. The examples are useful but do not covers all the API reference. And there is no direct link between them. 
It would be easier to have a direct link between those 2 page for example :
- [Exact example](https://shap.readthedocs.io/en/latest/example_notebooks/api_examples/explainers/Exact.html)
- [Exact reference](https://shap.readthedocs.io/en/latest/generated/shap.ExactExplainer.html)

[...]

PR #4521, opened 2026-04-13
DOCS: Add example link to SamplingExplainer docstring

Likhith2612:
This PR adds an example link to the SamplingExplainer docstring to improve navigation between the API reference and example notebooks.

This is a small step toward improving the connection between documentation sections, as discussed in #3171.

I’ll continue exploring the codebase and documentation to identify other areas where similar improvements can be made.

Happy to make any changes based on feedback!


shap/explainers/_sampling.py
```diff
@@ -40,6 +40,9 @@ class SamplingExplainer(KernelExplainer):
         KernelExplainer, this data can be the whole training set, even if that is a large set. This
         is because SamplingExplainer only samples from this background dataset.
 
+    Examples
+    --------
+    See `SHAP example notebooks <https://shap.readthedocs.io/en/latest/api_examples/>`_
```
CloseChoice, line 45: Hmm, I don't like this, since this can go out of sync and there is now reliable way to test this in the CI. Can you simply add an example in here?

CloseChoice (changes requested):

PR #4522, opened 2026-04-13
docs: clarify node_sample_weight should use data count, not hessian sum

### LINKED ISSUE #4166: Which method is better for TreeSHAP expected value: data count or hessian sum?
I noticed that LightGBM and XGBoost use different ways to calculate the expected value for TreeSHAP.

- LightGBM uses the [data count](https://github.com/microsoft/LightGBM/blob/336a77dfdecd68a528bc9b13a47e6ff93cca0ff6/src/io/tree.cpp#L953).
- XGBoost uses the [hessian sum](https://github.com/dmlc/xgboost/blob/2417e5bece34435e2b0ea8f774e403ae0ad133f5/src/predictor/treeshap.cc#L176).

The two methods will be different for classification task.

Which method is more accurate for calculating the expected value in SHAP?

When we pass a [custom tree ensemble model](https://shap.readthedocs.io/en/latest/example_notebooks/tabular_examples/tree_based_models/Example%20of%20loading%20a%20custom%20tree%20model%20into%20SHAP.html) into SHAP, we need to provide `node_sample_weights`. In this case, are we suggested to provide **data count** or **hessian sum**?

deveshbervar:
Closes #4166

## Summary

This PR answers the open question in #4166 by clarifying the correct usage of `node_sample_weight` in TreeSHAP and adding regression tests.

## Answer to the question

**Data count is correct** for `node_sample_weight`.

- For **regression** tasks, data count and hessian sum are equivalent.
- For **classification** tasks, they differ significantly (empirically ~45× on the breast cancer dataset).

Using hessian sum instead of data count can violate the fundamental SHAP local accuracy property:

```python
sum(shap_values) + expected_value == model_output

### DIFF
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -1775,7 +1775,20 @@ class SingleTree:
 
     node_sample_weight : numpy.array
         A 1d numpy array of length #nodes. The index ``i`` contains the number of
-        records (usually from the training data) that falls into node ``i``.
+        training samples (data count) that fall into node ``i``.
+
+        .. note::
+            This should be the **raw sample count**, not the hessian sum.
+            For regression tasks these are equivalent, but for classification
+            tasks they differ. Using the hessian sum (as stored internally by
+            XGBoost and some LightGBM objectives) instead of the data count
+            can violate the SHAP local accuracy property:
+            ``sum(shap_values) + expected_value == model_output``.
+
+            When constructing a custom tree dict to pass to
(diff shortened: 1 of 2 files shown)


deveshbervar:
Hi maintainers,
I’ve submitted this PR to address issue #4166 by clarifying the usage of `node_sample_weight` and adding tests.
I would really appreciate any feedback. Happy to make any changes needed.

Thank you!

PR #4524, opened 2026-04-13
TEST:Test coefficient coverage

ariba-sayed:
Added unit tests for the Coefficient explainer to improve test coverage.

Changes

1)Added tests for valid model initialization
2)Added tests for invalid model 
3) Verified correctness of attribution values
4)Verified output shape consistency
5) Covered edge case with empty input

The Coefficient explainer did not have dedicated tests.

All code and test logic have been reviewed and understood before submission.

### DIFF
--- a/tests/explainers/other/test_coefficient.py
+++ b/tests/explainers/other/test_coefficient.py
@@ -1,0 +1,71 @@
+import numpy as np
+
+from shap.explainers.other._coefficient import Coefficient
+
+
+# Checks that the class accepts a valid model
+class DummyModel:
+    coef_ = np.array([1, 2, 3])
+
+
+def test_init_valid_model():
+    model = DummyModel()
+    explainer = Coefficient(model)
+    assert explainer.model is model
+
+
+class BadModel:
+    pass
+
+
+# Checks that invalid input is rejected
+def test_init_invalid_model():
+    model = BadModel()
+    try:
+        Coefficient(model)
+        assert False, "Expected AssertionError"
+    except AssertionError:
+        assert True
+
+
+# Checks core functionality of attributions method with basic input
+def test_attributions_basic():
+    class DummyModel:
+        coef_ = np.array([1, 2, 3])
+
+    model = DummyModel()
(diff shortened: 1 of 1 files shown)


tests/explainers/other/test_coefficient.py
```diff
@@ -0,0 +1,71 @@
...
+
+    result = explainer.attributions(X)
+
+    expected = np.array([[1, 2, 3], [1, 2, 3]])
+
+    assert np.array_equal(result, expected)
+
+
+# Ensures output shape is correct
+def test_attributions_shape():
```
CloseChoice, line 49: pls get rid of this test, it does not provide any value above `test_attributions_basic`

tests/explainers/other/test_coefficient.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class BadModel:
+    pass
+
+
+# Checks that invalid input is rejected
+def test_init_invalid_model():
+    model = BadModel()
+    try:
```
CloseChoice, line 24: can you pls do this with `pytest.raises`?

tests/explainers/other/test_coefficient.py
```diff
@@ -0,0 +1,71 @@
...
+
+# Checks that the class accepts a valid model
+class DummyModel:
+    coef_ = np.array([1, 2, 3])
+
+
+def test_init_valid_model():
+    model = DummyModel()
+    explainer = Coefficient(model)
+    assert explainer.model is model
```
CloseChoice, line 14: can you pls check this in one of the tests below and get rid of this test?

CloseChoice (changes requested):

tests/explainers/other/test_coefficient.py
24f3004086: yes i will do it

tests/explainers/other/test_coefficient.py
24f3004086: yes

PR #4531, opened 2026-04-13
FIX: raise ValueError for unsupported method in shap.plots.embedding

### LINKED ISSUE #4394: BUG: UnboundLocalError in embedding () when passing unsupported method

officialasishkumar:
## Overview

Closes #4394

When an unsupported `method` string (e.g. `"tsne"`) was passed to
`shap.plots.embedding`, the `else` branch only printed a message and
let execution continue.  Because `embedding_values` was never assigned
in that branch, the subsequent `plt.scatter` call raised an
`UnboundLocalError: local variable 'embedding_values' referenced before
assignment` — a confusing crash that hid the real problem.

**Fix:** replace the `print()` call with a descriptive `ValueError` so
callers receive a clear, actionable error instead.

Also adds `tests/plots/test_embedding.py` with three tests:
- `pca` path works without error
- custom numpy-array embedding works without error
- unsupported method raises `ValueError` (regression for this issue)

## Checklist


CloseChoice (approved):
Thanks for the fix and the great tests. LGTM.

Pls ping on green pipelines

PR #4537, opened 2026-04-13
FIX: Prevent memory over-allocation in PermutationExplainer for constant features

### LINKED ISSUE #4527: BUG: Incorrect memory allocation and error calculation when there are no varying inputs in PermutationExplainer
### Issue Description

In `shap/shap/explainers/_permutation.py`, on line 160, `inds = fm.varying_inputs()`, if there are no varying inputs, then `len(inds) = 0`. The default value of `max_evals` is 500 (line 80). 

`npermutations` is calculated as `npermutations = max_evals // (2 * len(inds) + 1)`.
Therefore, `npermutations` = 500 // (2 * 0 + 1) = 500.

In the `else` branch (line 229) for the case `len(inds) = 0` and `error_bounds = True`:
`row_values_history = (2 * npermutations, len(fm))` -> (1000, num_features)

Also, in the `else` branch, the model is only evaluated once => there are no loops. This means that `rows_values_history` remains all zeros and never gets updated.

The error, `error_std` = `row_values_history.std(0)` = 0.0
Memory allocation = 1000 permutations
Memory used = 2 passes (1 model eval)
Efficiency = 0.2%

```python
import numpy as np
import shap

# Create a simple model
def model(X):
    """Simple model that always returns the same value"""
    return np.sum(X, axis=1, keepdims=True)

# Create background data (all same values)
background = np.array([[1.0, 2.0, 3.0]] * 5)

# Create test data (all same values as background - NO VARYING INPUTS)
X = np.array([[1.0, 2.0, 3.0]] * 2)

print("=" * 80)
print("MRE: Incorrect Memory Allocation with No Varying Inputs + error_bounds=True")
print("=" * 80)

# Create explainer with error_bounds=True
explainer = shap.explainers.PermutationExplainer(model, background)

print("\n### Input Data ###")
print(f"Background shape: {background.shape}")
print(f"Test data shape: {X.shape}")
print(f"All features same between X and background: {np.allclose(X[0], background[0])}")

# Call with error_bounds=True and track what happens
explanation = explainer(X, max_evals=500, error_bounds=True, silent=True)

print("\n### Result ###")
print(f"SHAP values shape: {explanation.values.shape}")

if hasattr(explanation, 'error_std'):
    print(f"Error std shape: {explanation.error_std.shape}")
    print(f"Error std values: {explanation.error_std}")

print("\n### Memory Inefficiency ###")
actual_allocated = 1000  # 2 * 500 
actual_used = 2  # 2 forward/backward passes in the else branch loop
print(f"Memory allocated for history: {actual_allocated} permutations")
print(f"Memory actually used: {actual_used} passes (only 1 model eval)")
print(f"Efficiency: {(actual_used / actual_allocated) * 100:.1f}%")
print(f"Wasted allocation: {actual_allocated - actual_used} permutation slots!")
```

### Traceback

```shell

```

### Expected Behavior

When there's only 1 evaluation `(len(inds) == 0),` either `error_std` should not be computed, or `npermutations` should be set to 1 thus making the shape of `rows_values_history`, `(2, num_features)`.

### Bug report checklist

### Installed Versions

`0.51.0`

AI Usage: The possibility of len(inds) being 0 was observed by me. The repercussions were explained by Claude and the fix was simply not computing error_std in this particular case since it has been handled in code.

BaibhavKundu2005:
## What does this PR do?
Prevents incorrect memory allocation and misleading error output in `PermutationExplainer.explain_row()` when there are no varying input features (i.e., when all input features are identical to the background data).

Closes #4527 

When `len(inds) == 0` (no features vary from background) and 
`error_bounds = True`:

- `npermutations` was calculated as `max_evals // (2 * 0 + 1)` = 500
- `row_values_history` was allocated as `(2 * 500, num_features)` = `(1000, num_features)`
- The code only evaluates the model once when there are no varying inputs (the history loop never executes)
- The history array remained all zeros and was never populated
- `error_std` was computed from this uninitialized array, producing misleading zeros. A user cannot distinguish between genuine zero variance and uncomputed variance
- 99.8% of allocated memory was wasted

## Fix
Set `row_values_history = None` directly in the `else` branch instead of allocating an unused array. The return statement already handles `None` correctly:

    "error_std": None if row_values_history is None else row_values_history.std(0)

This produces `error_std=None` instead of a misleading array of zeros, which is semantically correct, error bounds cannot be computed when there are no varying inputs.

## Evidence
Before fix:
- `row_values_history.shape`: (1000, 3) — allocated but never written
- `error_std`: array of zeros — misleading

After fix:
- `row_values_history`: None — no allocation
- `error_std`: None — correctly signals uncomputed

## Checklist
- [x] I have read CONTRIBUTING.md
- [x] Pre-commit hooks pass
- [x] test_permutation.py passes

### DIFF
--- a/shap/explainers/_permutation.py
+++ b/shap/explainers/_permutation.py
@@ -236,13 +236,7 @@ class PermutationExplainer(Explainer):
             expected_value = outputs[0]
             row_values = np.zeros((len(fm),) + outputs.shape[1:])
             if error_bounds:
-                row_values_history = np.zeros(
-                    (
-                        2 * npermutations,
-                        len(fm),
-                    )
-                    + outputs.shape[1:]
-                )
+                row_values_history = None  # No varying inputs, no history to track
 
         return {
             "values": row_values / (2 * npermutations),  # type: ignore[operator]
--- a/shap/explainers/_permutation.py
+++ b/shap/explainers/_permutation.py
@@ -163,10 +163,6 @@ class PermutationExplainer(Explainer):
         masks = np.zeros(2 * len(inds) + 1, dtype=int)
         masks[0] = MaskedModel.delta_mask_noop_value
         npermutations = max_evals // (2 * len(inds) + 1)
-        # When there are no varying inputs, there's only one evaluation, so set npermutations to 1
-        # to avoid massive memory over-allocation of row_values_history
-        if len(inds) == 0:
-            npermutations = 1
         row_values = None
         row_values_history = None
         history_pos = 0
--- a/shap/explainers/_permutation.py
+++ b/shap/explainers/_permutation.py
@@ -163,6 +163,10 @@ class PermutationExplainer(Explainer):
         masks = np.zeros(2 * len(inds) + 1, dtype=int)
         masks[0] = MaskedModel.delta_mask_noop_value
         npermutations = max_evals // (2 * len(inds) + 1)
+        # When there are no varying inputs, there's only one evaluation, so set npermutations to 1
+        # to avoid massive memory over-allocation of row_values_history
+        if len(inds) == 0:
+            npermutations = 1
         row_values = None
         row_values_history = None
         history_pos = 0


BaibhavKundu2005:
Hello @CloseChoice. A ping for whenever you can spare some time to review this and provide your valuable feedback.

PR #4366, opened 2026-04-05
Move to nanobind

CloseChoice:
## Overview

This is just the first step and should serve as a PoC to implement numba functions in C++.

supports #4327

PR #4539, opened 2026-04-13
Add nanobind stub

daidahao:
## Overview

Closes #XXXX  

Add automatic nanobind stub generation to fix CI fails.

@CloseChoice 

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,6 @@ node_modules
 mpl-results
 # intermediate AI skill files
 .claude/disclosures/
+
+# stub files
+*.pyi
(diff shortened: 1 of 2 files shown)


CloseChoice (approved):
LGTM, very nice thanks. Wasn't aware that this works automatically.

CloseChoice:
Interestingly mypy does not accept the auto generated stubs for the overloaded function in grey_code_utils.h. I mean that makes sense, the overload simply happens on the dimension of an input, so the auto generated annotations are exactly the same.

daidahao:
> Interestingly mypy does not accept the auto generated stubs for the overloaded function in grey_code_utils.h. I mean that makes sense, the overload simply happens on the dimension of an input, so the auto generated annotations are exactly the same.

Yep. See my comments here: https://github.com/shap/shap/pull/4366#issuecomment-4239699709. I think there is no need to run mypy over auto-generated pyi file, so we could just exclude it.

PR #4570, opened 2026-04-14
ENH: Migrate _reverse_window to C++ via nanobind

bedupako12mas:
Part of #4327 (Stage 2) — targets `move-to-nanobind` branch.

## Summary

Migrates `_reverse_window` from `shap/utils/_clustering.py` (numba `@njit`) to C++ via nanobind.

- Added `shap/cutils/clustering_utils.h` with C++ implementation
- Registered `reverse_window` in `shap/cutils/cutils.cpp`
- Replaced `@njit _reverse_window` in `_clustering.py` with call to `shap._cutils.reverse_window`

**Tests:** Added 5 unit tests in `tests/utils/test_clustering.py`

- [x] `test_reverse_window_basic` — reverses a mid-array window
- [x] `test_reverse_window_full_array` — reverses the entire array
- [x] `test_reverse_window_single_element` — length=1, array unchanged
- [x] `test_reverse_window_empty` — length=0, array unchanged
- [x] `test_reverse_window_at_end` — window at the tail of the array

## AI Disclosure

- **Advised**: Explained nanobind type mapping and codebase conventions
- **Assisted**: Reviewed implementation for correctness and style consistency
- **Autonomous**: None — all code was written and verified by the contributor


CloseChoice (changes requested):
This looks pretty good. Could you pls remove the corresponding numba njitted function and replace it with your c++ implementation?

bedupako12mas:
Done — replaced @njit _reverse_window in _clustering.py with the C++ implementation. All existing test_clustering.py tests pass.

bedupako12mas:
Removed `@njit` from `delta_minimization_order` — numba's nopython mode cannot type nanobind C++ functions, so any function that calls a nanobind function must be a regular Python function. The remaining `@njit` functions (`_reverse_window_score_gain`, `_mask_delta_score`) don't call nanobind functions and can stay jitted for now — they are candidates for follow-up C++ ports as part of issue #4327.

CloseChoice (changes requested):
Thanks, looks good so far. I am not a fan of testing internal functions like you do here, but I see the point for cutils. Therefore I would suggest moving this to a new test_cutils file.

tests/utils/test_clustering.py
```diff
@@ -40,3 +41,36 @@ def test_hclust_errors_on_unknown_linkages():
...
+def test_reverse_window_basic():
+    order = np.array([0, 1, 2, 3, 4], dtype=np.int64)
+    reverse_window(order, 1, 3)
+    assert list(order) == [0, 3, 2, 1, 4]
+
+
+def test_reverse_window_full_array():
+    order = np.array([0, 1, 2, 3, 4], dtype=np.int64)
+    reverse_window(order, 0, 5)
+    assert list(order) == [4, 3, 2, 1, 0]
```
CloseChoice: could you pls use `np.testing.assert_array_equal` for this?

daidahao:
> Removed `@njit` from `delta_minimization_order` — numba's nopython mode cannot type nanobind C++ functions, so any function that calls a nanobind function must be a regular Python function. The remaining `@njit` functions (`_reverse_window_score_gain`, `_mask_delta_score`) don't call nanobind functions and can stay jitted for now — they are candidates for follow-up C++ ports as part of issue #4327.

That could lead to performance issues if you remove it. My advice is to migrate all numba functions in scope here.

Also, we should set up performance monitoring asv (Stage 2) before we merge any new nanobind functions. Otherwise, we would not know the performance impact. 

For the migrated functions here, you should provide at least a profiling script so we can ensure that the performance is not degraded.

cc: @CloseChoice

tests/utils/test_clustering.py
bedupako12mas: Done. Tests now present in tests/test_cutils.py

bedupako12mas:
> > Removed `@njit` from `delta_minimization_order` — numba's nopython mode cannot type nanobind C++ functions, so any function that calls a nanobind function must be a regular Python function. The remaining `@njit` functions (`_reverse_window_score_gain`, `_mask_delta_score`) don't call nanobind functions and can stay jitted for now — they are candidates for follow-up C++ ports as part of issue #4327.
> 
> That could lead to performance issues if you remove it. My advice is to migrate all numba functions in scope here.
> 
> Also, we should set up performance monitoring asv (Stage 2) before we merge any new nanobind functions. Otherwise, we would not know the performance impact.
> 
> For the migrated functions here, you should provide at least a profiling script so we can ensure that the performance is not degraded.
> 
> cc: @CloseChoice

Thanks for flagging this. I agree that leaving `delta_minimization_order` without `@njit` while `_reverse_window_score_gain` and `_mask_delta_score` remain jitted is an incomplete migration. I'd like to port all three remaining functions (`_mask_delta_score`, `_reverse_window_score_gain`, and cleaning up `delta_minimization_order`) in this PR to complete the clustering stack migration and address the performance concern. I'll wait for @CloseChoice to confirm before I proceed further.

CloseChoice:
> > Removed `@njit` from `delta_minimization_order` — numba's nopython mode cannot type nanobind C++ functions, so any function that calls a nanobind function must be a regular Python function. The remaining `@njit` functions (`_reverse_window_score_gain`, `_mask_delta_score`) don't call nanobind functions and can stay jitted for now — they are candidates for follow-up C++ ports as part of issue #4327.
> 
> That could lead to performance issues if you remove it. My advice is to migrate all numba functions in scope here.
> 
> Also, we should set up performance monitoring asv (Stage 2) before we merge any new nanobind functions. Otherwise, we would not know the performance impact.
> 
> For the migrated functions here, you should provide at least a profiling script so we can ensure that the performance is not degraded.
> 
> cc: @CloseChoice

totally agree, I somehow missed this link between the functions. Thanks for flagging. Totally agree implementing asv as well (and potentially a workflow that check that no serious performance decreases happen)

bedupako12mas:
Thanks for confirming — I'll go ahead and port `_mask_delta_score` and `_reverse_window_score_gain` to C++ in this PR.

On the asv side — the repo doesn't have a benchmarks directory or any asv config yet, so this would need to be set up from scratch. Should this be a blocker for merging, or can it be tracked separately in issue #4327? I ask because the asv setup and any CI workflow around it would affect the entire `move-to-nanobind` project, not just this PR.

On performance — C++ should be competitive with numba here since the arrays are small (no JIT warmup cost either), but I agree benchmarking is important

CloseChoice:
> Thanks for confirming — I'll go ahead and port `_mask_delta_score` and `_reverse_window_score_gain` to C++ in this PR.
> 
> On the asv side — the repo doesn't have a benchmarks directory or any asv config yet, so this would need to be set up from scratch. Should this be a blocker for merging, or can it be tracked separately in issue #4327? I ask because the asv setup and any CI workflow around it would affect the entire `move-to-nanobind` project, not just this PR.
> 
> On performance — C++ should be competitive with numba here since the arrays are small (no JIT warmup cost either), but I agree benchmarking is important

I see asv after we merged the `move-to-nanobind` branch. So implementing a profiling suite should be the highest priority once phase 1 is complete.
I let claude run benchmarks using nanobind and it was even faster than numba, we simply safe the AOT-compilation overhead. But this needs to be confirmed and continously monitored. So I consider a full profiling suite a blocker for this PR.

PR #4577, opened 2026-04-14
Remove optional dependencies

daidahao:
Co-authored-by: Zhihao Dai <zhihao.dai@eng.ox.ac.uk>

## Overview

Closes #XXXX  

Supports #4327.

Add-on to #4366.

This PR moves all optional dependency groups into `dependency-groups` so that they are not exposed to users via `pip`.

They can still be installed via `uv sync --group GROUP_NAME` or `pip install --group GROUP_NAME`. But they are no longer available via `pip install shap[GROUP_NAME]`.

## Checklist


CloseChoice:
We can directly target master here. There is no need for this to be first merged into `move-to-nanobind` IMO. Objections?

.github/workflows/run_notebooks.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade uv
-          uv pip install --system '.[test,plots,test_notebooks,nbtest]'
+          uv pip install --system --group test --group plots --group nbtest
```
CloseChoice: is a dot missing here?

.github/workflows/run_tests.yml
```diff
@@ -122,7 +122,7 @@ jobs:
       - name: Install dependencies
         # Do regular install NOT editable install: see GH #3020
         run: |
-          uv pip install --system '.[${{ matrix.extras }},plots]'
+          uv pip install --system --group ${{matrix.extras}} --group plots
```
CloseChoice: also a dot missing?

.github/workflows/run_tests.yml
```diff
@@ -190,7 +190,7 @@ jobs:
         uses: astral-sh/setup-uv@v7
       - name: Install dependencies
         run: |
-          uv pip install --system numpy==2.0 '.[test-core,plots]'
+          uv pip install --system numpy==2.0 --group test-core --group plots
```
CloseChoice: same

.github/workflows/run_tests.yml
```diff
@@ -208,7 +208,7 @@ jobs:
         uses: astral-sh/setup-uv@v7
       - name: Install dependencies
         run: |
-          uv pip install --system '.[test-core]'
+          uv pip install --system --group test-core
```
CloseChoice: same

CloseChoice (changes requested):
Thanks for the PR. 

I think some dots slipped throught, I could see a way how this could even work without installing shap explicitly if the folder structure exactly mirrors how shap imports work. Would suggest using the dots though, since I am already thinking of moving the source code into a `src` directory!

daidahao:
> Thanks for the PR.
> 
> I think some dots slipped throught, I could see a way how this could even work without installing shap explicitly if the folder structure exactly mirrors how shap imports work. Would suggest using the dots though, since I am already thinking of moving the source code into a `src` directory!

Thank you for that! Yes, some dots slipped through. I've applied fixes.

I would not recommend moving the source code into `src` because we would lose almost all of file tracking. A lot of other projects do not use `src` layout and I don't see how it affects the build & distribution here.

daidahao:
> We can directly target master here. There is no need for this to be first merged into `move-to-nanobind` IMO. Objections?

Yes, we can do that. I need to investigate this more locally.

Two questions:
- Do we still need `requirements.txt`? I cannot find any reference to the file in code.
- What is the origin of `requirements.ci.txt`? It is being used for mypy tests, but I could not find how versions are decided.

CloseChoice:
> > We can directly target master here. There is no need for this to be first merged into `move-to-nanobind` IMO. Objections?
> 
> Yes, we can do that. I need to investigate this more locally.
> 
> Two questions:
> 
>     * Do we still need `requirements.txt`? I cannot find any reference to the file in code.
> 
>     * What is the origin of `requirements.ci.txt`? It is being used for mypy tests, but I could not find how versions are decided.

we can get rid of both files. These are legacy files used to install a subset of dependencies. I introduced `requirements.ci.txt` a while ago to make our CI reliable but we changed it back sometime later, so it's obsolete.

I am thinking about restructuring to `src` since having a folder with the same name as the package can shadow some import paths. Will need to investigate more thoroughly though what the implications are. Happy for any links if you have any readily available.

daidahao:
Closing this issue to merge into main branch directly instead #4589. @CloseChoice

PR #4589, opened 2026-04-14
ENH: Move optional-dep into dep-groups

daidahao:
Co-authored-by: Zhihao Dai <zhihao.dai@eng.ox.ac.uk>

## Overview

Replaces #4577 and merges into main branch directly.

This PR moves all optional dependency groups into `dependency-groups` so that they are not exposed to users via `pip`.

- They can still be installed via `uv sync --group GROUP_NAME` or `pip install --group GROUP_NAME`. But they are no longer available via `pip install shap[GROUP_NAME]`.
- Also update `CONTRIBUTING.md` to recommend using `uv` for development. For installing from source, however, `pip` is still the preferred way for most users.
- Remove `requirements.txt` no longer needed.
- <s>Fix a few mypy errors previously shadowed by relative paths.</s>
- Fix a build bug in `CMakeLists.txt` where CUDA extension could still be built with `SHAP_ENABLE_CUDA=0`.

## Checklist


daidahao:
The errors in https://github.com/shap/shap/pull/4589/commits/9badd4646f920914547e9540fa875c460be5874b  appear to be caused by `DeepExplainer` using background data as masker. However, both `DeepExplainer`s do not appear to use `self.masker` anywhere. So I've set it to `None` instead.

Do you think the fix is correct here? @CloseChoice

daidahao:
Any thoughts on this? @CloseChoice

.github/workflows/run_notebooks.yml
```diff
@@ -34,8 +34,7 @@ jobs:
         uses: astral-sh/setup-uv@v7
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade uv
-          uv pip install --system '.[test,plots,test_notebooks,nbtest]'
+          uv pip install --system . --group test --group plots --group nbtest
```
CloseChoice, line 37:
```suggestion
          uv pip install --system . --group test --group plots --group nbtest --group test_notebooks
```

shap/explainers/_deep/__init__.py
```diff
@@ -83,8 +83,7 @@ def __init__(self, model, data, session=None, learning_phase_flags=None):
             except Exception:
                 framework = "tensorflow"
 
-        masker = data
-        super().__init__(model, masker)
+        super().__init__(model, masker=None)
```
CloseChoice: looks like something is missing here?

shouldn't this be
```python
        super().__init__(model, masker=data)
```
?

CloseChoice (changes requested):
Two main things here:
1. You are mixing style changes with the change to dependency groups. I would suggest splitting this PR into two: one just making sure to have the dependency groups available, the other implementing the style changes.

daidahao:
> Two main things here:
> 
> 1. You are mixing style changes with the change to dependency groups. I would suggest splitting this PR into two: one just making sure to have the dependency groups available, the other implementing the style changes.

@CloseChoice 
Apologies for mixing the two in one PR. This PR was from a while ago and IIRC there were some style changes which had to be applied at the same time as dependency groups.

Nevertheless, I will remove the style changes from this PR so it is dependency group-focused and start a new PR for style changes only.

.github/workflows/run_notebooks.yml
daidahao: I've merged `test_notebooks` into `nbtest` group because the two only appeared together.

daidahao:
@CloseChoice I've removed any style changes to py files. However, mypy should fail because of existing typing errors. We can address them in a separate PR.

CONTRIBUTING.md
```diff
@@ -118,56 +117,51 @@ changes you make to the source code are immediately reflected in your
 environment.
 
 ```bash
-pip install --editable '.[test,plots,docs]'
+uv sync --group test-core --group plots
```
CloseChoice, line 122:
```suggestion
pip install -e . --group test-core --group plots
# or using uv
uv sync --group test-core --group plots
```

CONTRIBUTING.md
```diff
@@ -118,56 +117,51 @@ changes you make to the source code are immediately reflected in your
...
-Note: When installing from source, shap will attempt to build the C extension
-and the CUDA extension. If CUDA is not available, shap will retry the build
-without CUDA support.
+To use the CUDA extension for ``GPUTreeExplainer``, set the ``SHAP_ENABLE_CUDA`` environment variable to `1` when installing:
 
-Consequently, is is quite normal to see warnings such as `WARNING: Could not
-compile cuda extensions` when building from source if you do not have CUDA
-available.
+```bash
+SHAP_ENABLE_CUDA=1 uv sync --group test-core --group plots
```
CloseChoice, line 138: I would either remove `uv sync` here and replace it with a bare pip command or simply show both commands.

CONTRIBUTING.md
```diff
@@ -118,56 +117,51 @@ changes you make to the source code are immediately reflected in your
...
 
 ### Code checks with precommit
 
 We use [pre-commit hooks](https://pre-commit.com/#install) to run code checks.
 Enable `pre-commit` in your local environment with:
 
 ```bash
-pip install pre-commit
-pre-commit install
+uv run pre-commit install
```
CloseChoice, line 149: this is equilvalent two the old command if we assume that the project venv is activated and up to date, therefore just

```suggestion
pip install pre-commit
# or using uv
uv run pre-commit install
```

CONTRIBUTING.md
```diff
@@ -118,56 +117,51 @@ changes you make to the source code are immediately reflected in your
...
 ```
 
 To run the checks on all files, use:
 
 ```bash
-pre-commit install
-pre-commit run --all-files
+uv run pre-commit run --all-files
```
CloseChoice, line 157: same here

```suggestion
pre-commit run --all-files
# or using uv
uv run pre-commit run --all-files
```

CONTRIBUTING.md
```diff
@@ -118,56 +117,51 @@ changes you make to the source code are immediately reflected in your
...
 ```
 
 [Ruff](https://beta.ruff.rs/docs/) is used as a linter, and it is enabled as a
 pre-commit hook. You can also run `ruff` locally with:
 
 ```bash
-pip install ruff
-ruff check .
+uv run ruff check .
```
CloseChoice, line 166: same here, no difference if local venv is activated and up to date, therefore being tool specific is not making life easier IMO. Would either just keep as is or showing uv as an alternative

CloseChoice (changes requested):
Thanks for the PR, some minor doc updates, then we're good to merge.

CONTRIBUTING.md
daidahao: I think the main benefit of using `uv` for development is dependencies are "locked into" `pyproject.toml`, rather than being incremental as in a vanilla venv, i.e., redundant dependencies are not automatically removed from venv.

Personally, I would further recommend tracking `uv.lock` so all developers are locked into the same dependencies. But I would leave this for your discretion and a possible future PR.

daidahao:
Tests are failing on newer XGBoost...

CloseChoice (approved):

PR #4596, opened 2026-04-15
ENH: Removed the show argument in shap.plots

### LINKED ISSUE #3819: ENH: Remove the `show` argument in shap.plots

Anushreebasics:
## Overview

Closes #3819

- Added a shared deprecation/behavior helper: [_show.py]

- Updated exported matplotlib-based shap.plots functions so: default behavior is no implicit show call , passing show still works for now, but emits DeprecationWarning

## Checklist

### DIFF
--- a/shap/plots/_bar.py
+++ b/shap/plots/_bar.py
@@ -10,6 +10,7 @@ from .. import Cohorts, Explanation
 from ..utils import format_value, ordinal_str
 from ..utils._exceptions import DimensionError
 from ._labels import labels
+from ._show import resolve_show
 from ._style import get_style
 from ._utils import convert_ordering, dendrogram_coords, get_sort_order, merge_nodes, sort_inds
 
@@ -27,7 +28,7 @@ def bar(
     clustering_cutoff=0.5,
     show_data="auto",
     ax=None,
-    show=True,
+    show=None,
 ):
     """Create a bar plot of a set of SHAP values.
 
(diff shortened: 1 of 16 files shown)


Anushreebasics:
@connortann @CloseChoice please review

Anushreebasics:
@CloseChoice please review

Anushreebasics:
@CloseChoice @amishhaa please review

Anushreebasics:
@CloseChoice please review

PR #4604, opened 2026-04-15
fix: prevent IndexError in waterfall plots

### LINKED ISSUE #3553: BUG: Waterfall feature names IndexError
### Issue Description

When a shapley explainer is given an `n`x`n` feature matrix, the waterfall plot fails to be created because the feature names are not a list of the `n` feature names, but only the single feature at the sample position given.

The `Explanation` object has this code in `__init__`:
```py
        if len(_compute_shape(feature_names)) == 1: # TODOsomeday: should always be an alias once slicer supports per-row aliases
            if len(values_shape) >= 1 and len(feature_names) == values_shape[0]:
                feature_names = Alias(list(feature_names), 0)
            elif len(values_shape) >= 2 and len(feature_names) == values_shape[1]:
                feature_names = Alias(list(feature_names), 1)
```
The `Alias` `dim` of the `feature_names` is set to 0 when the features and samples match which results in just the single feature "Feature_1" instead of the list of features "['Feature_1', 'Feature_2', 'Feature_x']". The `Explanation` object then saves the `Slicer` object in `self._s` which when the waterfall plot is created with just a single sample `shap_explainer[0]`, `__get_item__` is called and then the slice uses the `dim` from above to either select the index given (in this case `0`) when `dim=0`, or `slice(None, None, None)` for `dim=1`.

```python
import pandas as pd
import numpy as np
import shap
import xgboost

model = xgboost.XGBClassifier()
n_samples, n_features = 400, 30
df = pd.DataFrame(np.random.random((n_samples, n_features)), columns=[f"Feature_{i}" for i in range(n_features)])

model.fit(df, np.random.choice([0, 1], n_samples))

shap_explainer = explainer(df.sample(n_features))

shap.plots.waterfall(shap_explainer[0])
```

### Traceback

```shell
  File "<string>", line 1, in <module>
  File "/home/luke/oxcan-fs/venv/lib/python3.10/site-packages/shap/plots/_waterfall.py", line 139, in waterfall
    yticklabels[rng[i]] = format_value(float(features[order[i]]), "%0.03f") + " = " + feature_names[order[i]]
IndexError: string index out of range
```

### Expected Behavior

The slice should always get the full set of features when creating a waterfall plot.

### Bug report checklist

### Installed Versions

0.42.1

DashratRajpurohit:
Fixes #3553. Added a safety check to pad feature_names when they are shorter than the number of SHAP values, preventing a common crash on malformed input. Verified with local reproduction


DashratRajpurohit:
"I’ve extended the safety padding from `waterfal`l to both `summary_legacy` and` bar_legacy `for full consistancy across the library. 

By standardizing this padding, we ensure that all legacy plotting functions are now robust agains`t IndexError` crashes when provided wit insufficient` feature_names. `This solidifies the fix for #3553 and improves the overall reliability of the SHAP plotting suite."

DashratRajpurohit:
Updated with the latest master. This fix addresses the IndexError in waterfall plots mentioned in #3553. Ready for CI approval and review.

DashratRajpurohit:
I've added a regression test in `tests/plots/test_issue_3553.py` to verify this fix. It confirms that `waterfall_legacy` now correctly handles cases where `feature_names` is shorter than shap_values by padding with default labels instead of raising an `IndexError`.

DashratRajpurohit:
Hi @CloseChoice, rebased and resolved docstring conflicts. Ready for review whenever you have a moment.

DashratRajpurohit:
Hi @CloseChoice, just a friendly ping when you have bandwidth. All feedback has been addressed and CI is green. Looking forward to your thoughts!

PR #4620, opened 2026-04-16
FIX: pair even middle-size subsets in KernelExplainer random sampling

### LINKED ISSUE #4117: Inconsistent Paired Sampling for Even-Feature Datasets in `KernelExplainer.explain()`
In [_kernel.py](https://github.com/shap/shap/blob/master/shap/explainers/_kernel.py), within the `KernelExplainer.explain()` method, there appears to be an inconsistency in the paired sampling logic when the dataset has an even number of features.

Specifically, when `num_full_subsets != num_subset_sizes`, the sampling procedure does **not** perform paired sampling for coalitions of size `num_subset_sizes`. This is due to the conditional check on line 460, which skips pairing when `num_subset_sizes` exceeds `num_paired_subset_sizes`.

This issue arises only for datasets with an even number of features, where a coalition and its complement can be of equal size. In such cases, `num_subset_sizes` is one larger than `num_paired_subset_sizes`, causing the final subset size to be excluded from paired sampling.

This behavior leads to reduced precision in the approximated Shapley values for datasets with even number of features, as demonstrated in https://[arxiv.org/pdf/2410.04883](https://arxiv.org/pdf/2410.04883).

Suggested Fix:
Update`remaining_weight_vector` and the condition on line 460 to ensure that coalitions of size `num_subset_sizes` are also included in the paired sampling process when the number of features is even.

amkr6207:
## Overview

Closes #4117.

Fixes even-feature paired sampling in `KernelExplainer.explain()` random sampling.

## Changes

- Include the even-`M` middle subset size in random paired sampling.
- Update random-stage weight handling accordingly.
- Replace implicit complement `+1` indexing with explicit complement-mask lookup.
- Add regression test: `test_kernel_even_feature_middle_subset_is_paired`.

## Validation

- `python -m pytest -q tests/explainers/test_kernel.py`

## Checklist
- [x] Unit test added (bug fix)

### DIFF
--- a/tests/explainers/test_kernel.py
+++ b/tests/explainers/test_kernel.py
@@ -391,31 +391,22 @@ def test_explainer_non_number_dtype(dt):
     np.testing.assert_allclose(shap_values.values.max(), 0.26548, rtol=1e-2)
 
 
-def test_kernel_even_feature_middle_subset_is_paired(monkeypatch):
+def test_kernel_even_feature_middle_subset_is_paired():
     """For even feature counts, random sampling should pair middle-size subsets with complements."""
     X_background = np.zeros((1, 6))
     X_eval = np.ones((1, 6))
 
-    # Force random sampling to choose subset_size=3 repeatedly (the even middle size for M=6).
-    def _fixed_choice(a, size, p):
-        return np.full(size, 2, dtype=int)
-
-    # Always select the first 3 feature indices.
-    def _fixed_permutation(n):
-        return np.arange(n)
-
-    monkeypatch.setattr(np.random, "choice", _fixed_choice)
-    monkeypatch.setattr(np.random, "permutation", _fixed_permutation)
+    # Use a fixed seed that naturally samples the even middle subset size (M/2) in random sampling.
+    np.random.seed(4)
 
     explainer = shap.KernelExplainer(lambda data: data.sum(axis=1), X_background)
-    explainer.shap_values(X_eval, nsamples=2)
+    explainer.shap_values(X_eval, nsamples=2, silent=True)
 
     masks = explainer.maskMatrix[: explainer.nsamplesAdded]
-    middle_subset_mask = np.array([1, 1, 1, 0, 0, 0], dtype=float)
-    complement_mask = 1.0 - middle_subset_mask
+    middle_masks = masks[np.sum(masks, axis=1) == 3]
 
-    assert np.any(np.all(masks == middle_subset_mask, axis=1))
-    assert np.any(np.all(masks == complement_mask, axis=1))
+    np.testing.assert_allclose(middle_masks.shape[0], 2)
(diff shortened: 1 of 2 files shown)


amkr6207:
@CloseChoice, PTAL. 

Thanks!

tests/explainers/test_kernel.py
```diff
@@ -391,6 +391,33 @@ def test_explainer_non_number_dtype(dt):
...
+
+    explainer = shap.KernelExplainer(lambda data: data.sum(axis=1), X_background)
+    explainer.shap_values(X_eval, nsamples=2)
+
+    masks = explainer.maskMatrix[: explainer.nsamplesAdded]
+    middle_subset_mask = np.array([1, 1, 1, 0, 0, 0], dtype=float)
+    complement_mask = 1.0 - middle_subset_mask
+
+    assert np.any(np.all(masks == middle_subset_mask, axis=1))
+    assert np.any(np.all(masks == complement_mask, axis=1))
```
tylerjereddy: Could we not use `assert_allclose` here? The feedback in case of failure is much better if so.

tests/explainers/test_kernel.py
```diff
@@ -391,6 +391,33 @@ def test_explainer_non_number_dtype(dt):
...
+    # Force random sampling to choose subset_size=3 repeatedly (the even middle size for M=6).
+    def _fixed_choice(a, size, p):
+        return np.full(size, 2, dtype=int)
+
+    # Always select the first 3 feature indices.
+    def _fixed_permutation(n):
+        return np.arange(n)
+
+    monkeypatch.setattr(np.random, "choice", _fixed_choice)
+    monkeypatch.setattr(np.random, "permutation", _fixed_permutation)
```
tylerjereddy: This is pretty hacky--the matching issue seems to claim that the problem can arise naturally so drafting a more "real world" regression test might be nice.

tylerjereddy (changes requested):

tests/explainers/test_kernel.py
amkr6207: Thanks, you’re right. Updated to np.testing.assert_allclose for clearer failure feedback.

tests/explainers/test_kernel.py
amkr6207: Agreed. I removed the monkeypatching and switched to a deterministic seeded test that hits this path through normal KernelExplainer sampling. 

Thanks!

amkr6207:
@tylerjereddy, PTAL. Thanks!

amkr6207:
@CloseChoice, @tylerjereddy, PTAL. Thanks!

amkr6207:
@CloseChoice, @tylerjereddy, could you please review this PR? Thanks!

amkr6207:
@CloseChoice, @tylerjereddy, could you please review this PR? Thanks!

PR #4641, opened 2026-04-17
  fix(Explanation): feature_names corrupted when slicing by output name string

### LINKED ISSUE #4633: BUG: feature_names assigned wrong value in Explanation.__getitem__ when slicing by output name string
### Issue Description

When slicing a multi-output Explanation object by an output name string (e.g. shap_values["positive"]), the resulting Explanation has its feature_names silently replaced with the raw sample data array instead of the original feature names.                                       
  
### Root Cause                                                                                                                                  
                                                            
  In shap/_explanation.py at line 370, the Explanation constructor is called with:

  ```python
feature_names=np.array(new_data),
```  
  new_data is the list of per-sample input data values, not feature names. The correct value is self.feature_names.                           
                                                                                                                                                                                                                                                                                       
  
 #### Impact                                                                                                                                      
                                                            
  - All plots (beeswarm, bar, waterfall, force) produce wrong or garbled axis labels for multi-output models (e.g. multi-class classifiers,   
  text-generation token-level SHAP).
  - Silently corrupts results — no exception is raised, making it hard to detect. 

```python
import numpy as np                                                                                                                          
  import shap                                               

  rs = np.random.RandomState(0)
  feature_names = ["f0", "f1", "f2", "f3"]
  output_names = [["pos", "neg"]] * 3
                                                                                                                                              
  exp = shap.Explanation(
      values=[rs.rand(4, 2) for _ in range(3)],                                                                                               
      base_values=[rs.rand(2) for _ in range(3)],           
      data=[rs.rand(4) for _ in range(3)],                                                                                                    
      feature_names=feature_names,
      output_names=output_names,                                                                                                              
  )                                                         

  sliced = exp["pos"]
  print(sliced.feature_names)  # prints raw data arrays, not ["f0","f1","f2","f3"]
```

### Expected Behavior
  sliced.feature_names should equal ["f0", "f1", "f2", "f3"].                                                                                 
  
### Actual Behavior  

 sliced.feature_names contains raw numerical data arrays (the contents of new_data), making every downstream plot label incorrect or crash.  

### Bug report checklist

### Installed Versions

 0.1.dev2960

ritankarsaha:
## Overview

Closes #4633 

- **Bug:** `Explanation.__getitem__`  assigned `feature_names=np.array(new_data)` — the raw sample data — instead of
    `feature_names=self.feature_names` when slicing a multi-output `Explanation` by an output name string.
  - **Impact:** Every downstream plot (`beeswarm`, `bar`, `waterfall`, etc.) for multi-output or text-generation models received corrupted, meaningless feature names, silently producing wrong axis labels.
  - **Fix:** One-line change: replace `np.array(new_data)` with `self.feature_names`.
  - **Test:** Added `test_feature_names_preserved_when_slicing_by_output_name` in `tests/test_explanation.py` to prevent regression.

## Checklist


ritankarsaha:
Hey @CloseChoice made fixes for this issue properly, can you take a look and let me know if anything is needed to be modified or not. 
Thanks !!

DashratRajpurohit:
Great to see a fix for the feature_names assignment!

While testing this locally with a few more edge cases, I noticed two related issues in this same block that might be worth addressing here to make the fix more robust:

1. **NoneType Check**: If `base_values` is None (which happens with some model types), line 362 will still raise a TypeError because it tries to index into a None object. Adding a if `self.base_values is not None`: check would prevent this.

2. **Logic Flow**: The method currently continues down to line 422, where it calls `__getitem__` again on the already-sliced new_self. This "double-slicing" is what causes the IndexError mentioned in the related issues, since the dimensions have already been reduced by the manual workaround.

Adding a return new_self at the end of these workaround blocks (around line 380 and 406) would solve the redundant slicing issue. What do you think?

PR #4653, opened 2026-04-18
Enhance backward compatibility in render method,  Fix React root handling and improve render safety for SHAP visualizers

Aman-00007:
- Added safety check to handle null or undefined container in render method
- Implemented root caching using a Map to prevent multiple React roots per container
- Improved stability and performance by ensuring a single root instance is reused
- Maintained backward compatibility with legacy ReactDom.render API
- Ensured compatibility with React 18 createRoot API

These changes make the rendering logic more robust and prevent potential crashes or unexpected behavior during repeated renders.

### DIFF
--- a/javascript/index.jsx
+++ b/javascript/index.jsx
@@ -6,17 +6,33 @@ import {
   AdditiveForceArrayVisualizer
 } from "./visualizers";
 
-// Save some globals for the inline scripts to access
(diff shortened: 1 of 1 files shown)


Aman-00007:
Fix root reuse, add null check, and improve React 18 render handling

PR #4654, opened 2026-04-18
ENH: Migrate clustering functions to C++ via nanobind

bedupako12mas:
Part of #4327 (Stage 2) — continuation of #4570, retargeted to master after Stage 1 merged.

## Summary

Migrates the full clustering stack from `shap/utils/_clustering.py` (numba `@njit`) to C++ via nanobind. Removes the numba dependency from `_clustering.py` entirely.

Functions ported:
- `_reverse_window` → `reverse_window`
- `_mask_delta_score` → `mask_delta_score`
- `_reverse_window_score_gain` → `reverse_window_score_gain`
- `delta_minimization_order` → `delta_minimization_order`
- `_pt_shuffle_rec` → `pt_shuffle_rec`

## Changes

- `shap/cutils/clustering_utils.h` — C++ implementations
- `shap/cutils/cutils.cpp` — nanobind registrations for all 5 functions
- `shap/utils/_clustering.py` — removed all `@njit` functions and `from numba import njit`; `partition_tree_shuffle` retained as a thin Python wrapper calling `pt_shuffle_rec`
- `shap/utils/__init__.py` — updated `delta_minimization_order` re-export to pull from `_cutils` instead of `_clustering`
- `shap/_cutils.pyi` — type stubs for all 5 new functions
- `tests/test_cutils.py` — 20 unit tests across all 5 functions

## Design decisions

**`detail::` namespace for internal helpers**
`detail::row_delta` and `detail::pt_shuffle_rec` are not bound to Python — they live in an internal namespace. `row_delta` exists because nanobind does not support row slicing (`masks[i]` on a 2D array), so per-row Hamming distance is computed via a `.view()` accessor loop. `pt_shuffle_rec` is separated so the recursive worker takes `std::mt19937&` by reference without exposing the RNG to Python.

**`.view()` for 2D array access**
All 2D array access uses `arr.view()(i, j)` rather than raw pointers with `nb::c_contig`. This is stride-safe and consistent with the existing `grey_code_utils.h` convention in the codebase.

**`bool` dtype for mask arrays**
Mask parameters are typed as `nb::ndarray<bool, ...>` matching the actual dtype used throughout the codebase (`npt.NDArray[np.bool_]`). Using `int64_t` for bool masks would be incorrect.

**`nb::capsule` for owned return arrays**
`delta_minimization_order` allocates a heap array and returns it as a numpy array. Ownership is transferred via `nb::capsule` with a custom deleter — the standard nanobind pattern for returning C++-allocated data without copying.

**Raw pointer in `delta_minimization_order` inner loop**
The inner loop of `delta_minimization_order` accesses `data[i]` directly on the owned `int64_t*` pointer rather than going through a nanobind accessor. This is safe because we own the allocation, and avoids accessor overhead in the hot path.

**`thread_local std::mt19937` for RNG**
One RNG instance per thread, seeded from `std::random_device`. Safe for multi-threaded use without a mutex. Matches numba's behaviour — numba's internal `@njit` RNG is also independent of `np.random.seed()`.

**`std::bernoulli_distribution(0.5)` for coin flip**
The Python implementation uses `np.random.randn() < 0` — a normal draw checked for sign — which is a 50/50 coin flip. `std::bernoulli_distribution(0.5)` is semantically identical but makes the intent explicit.

## Tests

20 unit tests in `tests/test_cutils.py` using `np.testing.assert_array_equal` as requested in #4570 review. Tests cover correctness, edge cases, and structural invariants. For `pt_shuffle_rec` specifically, exact output cannot be asserted (non-deterministic RNG) so tests verify set correctness, tree-contiguity, and contiguity stability across 20 runs.

Note: `_build_tree` in the test file calls `np.random.seed(0)` to produce a deterministic partition tree. This resets numpy's global random state but is acceptable here since the C++ RNG is independent of numpy state.

## AI Disclosure

Advised: Explained nanobind type mapping, codebase conventions, and scipy linkage matrix format  
Assisted: Reviewed implementation for correctness and style consistency  
Autonomous: None — all code was written and verified by the contributor  

### DIFF
--- a/shap/cutils/clustering_utils.h
+++ b/shap/cutils/clustering_utils.h
@@ -8,6 +8,24 @@
 namespace nb = nanobind;
 using namespace nb::literals;
 
+namespace detail {
+
+inline int row_delta(
+    const nb::ndarray<bool, nb::shape<-1, -1>, nb::device::cpu>& masks,
+    int64_t row_a,
+    int64_t row_b
+) {
+    auto m = masks.view();
+    int score = 0;
+    for (size_t k = 0; k < masks.shape(1); k++) {
+        score += m(row_a, k) ^ m(row_b, k);
+    }
+    return score;
+}
+
+} // namespace detail
+
+
 void reverse_window(
     nb::ndarray<int64_t, nb::shape<-1>, nb::device::cpu>& order,
     const int start,
@@ -33,6 +51,21 @@ int mask_delta_score(
         score += m1(i) ^ m2(i);
     }
     return score;
+}
+
+int reverse_window_score_gain(
+    const nb::ndarray<bool, nb::shape<-1, -1>, nb::device::cpu>& masks,
+    const nb::ndarray<int64_t, nb::shape<-1>, nb::device::cpu>& order,
+    const int start,
+    const int length
+) {
+    assert(start >= 1);
+    assert(start + length < (int)order.shape(0));
+    int forward_score = detail::row_delta(masks, order(start - 1), order(start))
+                      + detail::row_delta(masks, order(start + length - 1), order(start + length));
+    int reverse_score = detail::row_delta(masks, order(start - 1), order(start + length - 1))
+                      + detail::row_delta(masks, order(start), order(start + length));
+    return forward_score - reverse_score;
 }
 
 #endif // CLUSTERING_UTILS_H
--- a/shap/cutils/clustering_utils.h
+++ b/shap/cutils/clustering_utils.h
@@ -1,0 +1,38 @@
+#ifndef CLUSTERING_UTILS_H
+#define CLUSTERING_UTILS_H
+
+#include <cassert>
+#include <nanobind/nanobind.h>
+#include <nanobind/ndarray.h>
+
+namespace nb = nanobind;
+using namespace nb::literals;
+
+void reverse_window(
+    nb::ndarray<int64_t, nb::shape<-1>, nb::device::cpu>& order,
+    const int start,
+    const int length
+) {
+    assert(length >= 0);
+    assert(start >= 0);
+    assert((size_t)start + (size_t)length <= order.shape(0));
+    for (size_t i = 0; i < (size_t)length / 2; i++) {
+        int64_t tmp = order(start + i);
+        order(start + i) = order(start + length - i - 1);
+        order(start + length - i - 1) = tmp;
+    }
+}
+
+int mask_delta_score(
+    const nb::ndarray<bool, nb::shape<-1>, nb::device::cpu>& m1,
+    const nb::ndarray<bool, nb::shape<-1>, nb::device::cpu>& m2
+) {
+    assert(m1.shape(0) == m2.shape(0));
+    int score = 0;
+    for (size_t i = 0; i < m1.shape(0); i++) {
+        score += m1(i) ^ m2(i);
+    }
+    return score;
+}
+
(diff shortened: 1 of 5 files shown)


shauryam2807:
Yes My friend @bedupako12mas but I see somthing in your code
Took a look through the C++ implementation — nice work on the overall structure! A few things I noticed during review:

1. **`assert()` in `reverse_window`/`reverse_window_score_gain`**: These are compiled away in release builds (`NDEBUG`), so production users would get silent UB on invalid inputs instead of a helpful error. Consider using `if (...) throw std::out_of_range(...)` or nanobind's argument validation instead.

2. **`detail::pt_shuffle_rec` passes `nb::ndarray` references through recursion**: For a tree of depth `O(M)`, each recursive frame carries the nanobind reference-counted wrappers. Unpacking to raw pointers at the wrapper level (like the existing `grey_code_utils.h` pattern) and recursing with raw `const double*` / `int64_t*` would avoid the per-call overhead.

3. **`row_delta` / `mask_delta_score` return `int` instead of `int64_t`**: Minor, but the original numba functions return 64-bit integers. `int` is 32-bit on most platforms.

4. **Missing `#pragma once`** in `clustering_utils.h`.

Happy to help address any of these if useful!

bedupako12mas:
Thanks for the review @shauryam2807! Going through each point:

**1. `assert()` in `reverse_window` / `reverse_window_score_gain`**

`grey_code_utils.h` and `kernel_explainer_utils.h` both use `assert()` and include `<cassert>` — it's the established pattern across all three headers in this codebase. Switching to `throw std::out_of_range` in `clustering_utils.h` alone would be inconsistent with the existing headers written by the maintainers. If the project wants to move to exception-based validation uniformly, that's a codebase-level discussion — happy to follow whatever direction the maintainers set there.

**2. `detail::pt_shuffle_rec` passes `nb::ndarray` references through recursion**

Fair observation. Unpacking to raw pointers at the public wrapper boundary and recursing with plain `const bool*` / `const double*` / `int64_t*` would avoid carrying the reference-counted wrappers through every recursive frame (currently lines 28–54 in `clustering_utils.h`). I can make that change — would the maintainers like to see it before merge, or is it acceptable as a follow-up?

**3. `row_delta` / `mask_delta_score` return `int` instead of `int64_t`**

`int` matches the established convention in this codebase. `kernel_explainer_utils.h` line 13 returns `int` from `compute_exp_val`, and `grey_code_utils.h` uses `int` for local variables and parameters throughout (e.g. `int M` line 25, `int delta_ind` line 27, `int noop_code` parameter). The return values here are Hamming distances bounded by `n_features` (4–13 in practice), so overflow is not a concern. Changing to `int64_t` would be inconsistent with the existing headers.

**4. Missing `#pragma once`**

Both existing headers use `#ifndef`/`#define` guards — `grey_code_utils.h` lines 1–2 and `kernel_explainer_utils.h` lines 1–2. `clustering_utils.h` follows the same convention intentionally (lines 1–2). Both approaches serve the same purpose; `#pragma once` would be inconsistent with the rest of the codebase.

shauryam2807:
Fair points on matching existing `grey_code_utils.h` conventions — consistency with the codebase is the right call. Glad the recursion wrapper observation was useful though!

I'll get started on porting `_masked_model.py` functions next.

CloseChoice:
Thanks for the PR.

Superseded by #5088

Issue #3796, opened 2024-08-12
Plot overhaul: embedding


connortann:
This issue is part of the [Visualisation overhaul](https://github.com/orgs/shap/projects/1) project.

The things we wish to overhaul are detailed in #3411, and the current status of these TODO items for this plot is tracked in the project fields (see sidebar).

techaadii:
Hi, I'm contributing to #3411 — I have open PRs for scatter, waterfall (#4401) and group_difference. After going through the code base I felt I'd like to pick up embedding (#3796). The file has no ax parameter, uses implicit plt.xxx() throughout, and has no tests at all. My plan: add ax parameter with plt.gca() default, convert all implicit calls, add type hints, complete the docstring, and create a new test file. I'll disclose AI usage per policy. cc @connortann

PR #4660, opened 2026-04-18
ENH: Refactor embedding plot to use explicit matplotlib interface

### LINKED ISSUE #4659: ENH: Plot overhaul for shap/plots/_embedding.py — explicit matplotlib interface
## Overview

Part of the ongoing plotting API consistency overhaul (see #3411, specifically sub-task #3796).

`shap/plots/_embedding.py` currently uses the **implicit pyplot interface** throughout and does not accept or return an `ax` parameter, making it impossible to compose with other matplotlib code.

## Current problems

```python
# Current implementation uses implicit interface
plt.scatter(...)
plt.axis("off")
cb = plt.colorbar()
plt.gcf().set_size_inches(...)
```

- No `ax` parameter: callers cannot place the plot on a specific axes
- Returns nothing (implicitly `None` even when `show=False`)
- Uses `plt.gcf()` / `plt.scatter()` / `plt.colorbar()` throughout
- No type hints
- Docstring has typos and is missing parameters

## Proposed changes

Following the pattern established in #3523 (bar plot overhaul):

1. **Add `ax` parameter** defaulting to `plt.gca()` so callers can pass an existing axes
2. **Return `ax`** when `show=False`, consistent with the rest of the refactored plots
3. **Replace implicit interface** with explicit `ax.scatter()`, `ax.axis()`, `fig.colorbar()`, `fig.set_size_inches()`
4. **Add type hints** for all parameters and the return type
5. **Update docstring** — fix typos, add `ax` and return-value documentation

## Example of proposed API

```python
fig, ax = plt.subplots()
shap.plots.embedding(0, shap_values, ax=ax, show=False)
ax.set_title("My custom title")
```

## References

- Meta-issue: #3411 (Improve plotting API for consistency)
- Sub-task: #3796 (Plot overhaul: embedding)
- Reference implementation: #3523 (bar plot overhaul)

magic-peach:
## Overview

Closes #4659
Supports #3411
Supports #3796

This PR refactors `shap/plots/_embedding.py` to follow the explicit matplotlib interface established for the SHAP plotting API overhaul (see #3523 as the reference implementation).

### `shap/plots/_embedding.py`

- **Added `ax` parameter** (type `matplotlib.axes.Axes | None`, default `None`). When `None`, `plt.gca()` is used, which preserves the previous behaviour of writing to the current figure.
- **Returns `ax`** when `show=False`, consistent with the other refactored plots. Returns `None` when `show=True` (avoids cluttering notebook output with the axes repr).
- **Replaced implicit interface** with explicit calls:
  - `plt.scatter(...)` → `ax.scatter(...)`
  - `plt.axis("off")` → `ax.axis("off")`
  - `plt.colorbar()` → `ax.get_figure().colorbar(sc, ax=ax)` (attaches to the correct axes)
  - `plt.gcf().set_size_inches(...)` → `ax.get_figure().set_size_inches(...)`
- **Added type hints** for all parameters and the return value.
- **Updated docstring**: fixed typos (`it's` → `its`, `is should` → `should`), added documentation for `ax` and the return value, added rst cross-reference for `show`.
- Moved `import matplotlib.pyplot as plt` inside the function body (pattern used by other refactored plots to keep module-level imports lean).

### `tests/plots/test_embedding.py` (new file)

Added 18 unit tests covering:

| Test | What it checks |
|------|---------------|
| `test_embedding_pca_returns_ax` | Returns `Axes` when `show=False` |
| `test_embedding_pca_no_return_when_show_true` | Returns `None` when `show=True` |
| `test_embedding_custom_ax` | Caller-supplied `ax` is used and returned |
| `test_embedding_draws_on_given_ax` | Scatter drawn on correct axes |
| `test_embedding_ind_by_integer` | `ind` as integer |
| `test_embedding_ind_by_feature_name` | `ind` as feature name string |
| `test_embedding_ind_sum` | `ind="sum()"` colors by sum |
| `test_embedding_ind_rank` | `ind="rank(0)"` selects by rank |
| `test_embedding_custom_embedding_array` | Custom `(n x 2)` array as `method` |
| `test_embedding_unsupported_method_prints_and_returns` | Bad `method` prints message, returns `ax` |
| `test_embedding_auto_feature_names` | Auto-generated feature labels |
| `test_embedding_explicit_feature_names` | Explicit `feature_names` list |
| `test_embedding_alpha` | Custom `alpha` accepted |
| `test_embedding_axis_off` | `ax.axison` is `False` after call |
| `test_embedding_colorbar_label_uses_feature_name` | Colorbar axes present |
| `test_embedding_colorbar_label_sum` | Colorbar present for `sum()` |

## Rationale

I studied the existing code, the reference PR #3523, and the discussion in #3411 before making changes. The key insight from @connortann's note in #3411 is that we should use `plt.gca()` rather than `plt.figure()` as the default, to preserve the pattern where callers can do:

```python
f = plt.figure()
shap.plots.embedding(0, shap_values, show=False)
return f
```

The `plt.gca()` default ensures `ax.get_figure()` always resolves to the caller's figure.

The unsupported-method branch previously referenced `embedding_values` before it was defined (a latent `UnboundLocalError`). I fixed this by returning early from that branch.

## Checklist

- [x] Uses explicit object-oriented matplotlib interface
- [x] Accepts an optional `ax` parameter and returns `ax`
- [x] Unit tests added
- [x] Docstring updated
- [x] Type hints added

### DIFF
--- a/shap/plots/_embedding.py
+++ b/shap/plots/_embedding.py
@@ -1,39 +1,74 @@
-import matplotlib.pyplot as plt
+from __future__ import annotations
+
+import numpy as np
 import sklearn
+from matplotlib.axes import Axes
 
 from ..utils import convert_name
 from . import colors
 from ._labels import labels
 
 
-def embedding(ind, shap_values, feature_names=None, method="pca", alpha=1.0, show=True):
+def embedding(
+    ind: int | str,
+    shap_values: np.ndarray,
+    feature_names: list[str] | None = None,
+    method: str | np.ndarray = "pca",
+    alpha: float = 1.0,
+    show: bool = True,
+    ax: Axes | None = None,
+) -> Axes | None:
     """Use the SHAP values as an embedding which we project to 2D for visualization.
 
     Parameters
     ----------
-    ind : int or string
+    ind : int or str
         If this is an int it is the index of the feature to use to color the embedding.
         If this is a string it is either the name of the feature, or it can have the
-        form "rank(int)" to specify the feature with that rank (ordered by mean absolute
-        SHAP value over all the samples), or "sum()" to mean the sum of all the SHAP values,
-        which is the model's output (minus it's expected value).
+        form ``"rank(int)"`` to specify the feature with that rank (ordered by mean
+        absolute SHAP value over all the samples), or ``"sum()"`` to mean the sum of
+        all the SHAP values, which is the model's output (minus its expected value).
 
-    shap_values : numpy.array
+    shap_values : numpy.ndarray
         Matrix of SHAP values (# samples x # features).
 
-    feature_names : None or list
+    feature_names : list of str, optional
         The names of the features in the shap_values array.
 
-    method : "pca" or numpy.array
-        How to reduce the dimensions of the shap_values to 2D. If "pca" then the 2D
-        PCA projection of shap_values is used. If a numpy array then is should be
-        (# samples x 2) and represent the embedding of that values.
+    method : str or numpy.ndarray, optional
+        How to reduce the dimensions of the shap_values to 2D. If ``"pca"`` then the
+        2D PCA projection of shap_values is used. If a numpy array then it should be
+        ``(# samples x 2)`` and represent the embedding of those values.
 
-    alpha : float
-        The transparency of the data points (between 0 and 1). This can be useful to the
+    alpha : float, optional
+        The transparency of the data points (between 0 and 1). This can be useful to
         show density of the data points when using a large dataset.
 
+    show : bool, optional
+        Whether :external+mpl:func:`matplotlib.pyplot.show()` is called before
+        returning. Setting this to ``False`` allows the plot to be customised further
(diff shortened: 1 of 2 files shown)


magic-peach:
@slundberg @connortann could you please share more details about ESOC'26

DashratRajpurohit:
It looks like the` pre-commit.ci `checks are failing because numpy and Axes are only used for type hinting in `shap/plots/_embedding.py.`

Since you have `from __future__ import annotations `at the top, you can move those third-party imports into a TYPE_CHECKING block to satisfy the` TC002 `linter rule. This should fix the CI:

```
from typing import TYPE_CHECKING 
 #... other runtime imports
if TYPE_CHECKING:
    import numpy as np
    from matplotlib.axes import Axes
```
Note that` import matplotlib.pyplot as plt` should stay outside the block because it's used at runtime, but numpy and Axes can definitely be moved inside.

PR #4646, opened 2026-04-17
POC: migrate _clustering.py numba functions to C++ via nanobind

shauryam2807:
## Overview

Proof-of-concept migration of 5 performance-critical `@njit` functions from 
`shap/utils/_clustering.py` to C++ using `nanobind`, as proposed in the 
ESoC 2026 Batch 2 project ("Move performance-critical functions from Numba to C++").

## Migrated Functions

| Function | Description |
|----------|-------------|
| `_mask_delta_score` | XOR-based binary mask delta scoring |
| `_reverse_window` | In-place sub-array reversal |
| `_reverse_window_score_gain` | Score gain computation for window reversal |
| `delta_minimization_order` | Greedy ordering optimization |
| `_pt_shuffle_rec` | Recursive partition tree shuffling |

## Results

- **Correctness:** All 8 test cases pass — C++ output matches Numba exactly.
- **Performance:** **12.9x speedup** over Numba on `delta_minimization_order` benchmark.

## Technical Details

- Built with `scikit-build-core` + `nanobind` (aligns with Stage 1 build migration)
- Uses `nb::ndarray` for zero-copy numpy interop
- Thread-safe RNG via `std::mt19937` (replaces `np.random.randn`)
- Inlined hot loops in `delta_minimization_order` for additional performance

## Related

- Test coverage PR: #4634
- ESoC project: [Move performance-critical functions from Numba to C++](https://github.com/shap/shap/wiki/ESoC-2026-Project-Ideas)

> **Note:** This is a Draft PR to demonstrate the migration approach. 
> Not intended for merge until the `scikit-build-core` infrastructure (Stage 1) is finalized.

PR #4661, opened 2026-04-18
ENH: add ASV benchmarking infrastructure for nanobind migration

bedupako12mas:
Part of #4327 (Stage 2) — performance monitoring infrastructure for the nanobind migration.

## What this PR does

Sets up [ASV (Airspeed Velocity)](https://asv.readthedocs.io/en/stable/) benchmarking infrastructure and adds the first benchmark file covering the 5 clustering functions being ported in #4654.

Following the discussion in #4570, both @CloseChoice and @daidahao identified ASV as a blocker before merging further nanobind migrations — see [this comment](https://github.com/shap/shap/pull/4570#issuecomment-4254720468) and the Stage 2 priority note in #4327. This PR delivers that infrastructure.

The benchmark structure mirrors the full migration scope from #4327, with one benchmark file per source file. This PR establishes the pattern; subsequent porting PRs add their corresponding benchmark class.

| Benchmark file | Source `.py` file | C++ header | Functions |
|---|---|---|---|
| `benchmarks_clustering.py` ✅ | `utils/_clustering.py` | `clustering_utils.h` ✅ | 5 — this PR |
| `benchmarks_exact.py` | `explainers/_exact.py` | `grey_code_utils.h` ✅ (partial) | 2 |
| `benchmarks_partition.py` | `explainers/_partition.py` | `partition_utils.h` (new) | 1 |
| `benchmarks_tabular.py` | `maskers/_tabular.py` | `tabular_utils.h` (new) | 2 |
| `benchmarks_masked_model.py` | `utils/_masked_model.py` | `masked_model_utils.h` (new) | 4 |
| `benchmarks_image.py` | `maskers/_image.py` | `image_utils.h` (new) | 1 |

**Files added/changed:**
- `asv.conf.json` — ASV project config (`--no-build-isolation` required for scikit-build-core)
- `benchmarks/__init__.py` — marks directory as benchmark package
- `benchmarks/benchmarks_clustering.py` — benchmarks for all 5 clustering functions
- `pyproject.toml` — adds `benchmark = ["asv"]` to `[dependency-groups]`
- `.gitignore` — excludes `.asv/` build artifacts

### One class per function

The alternative — one class per file with all functions benchmarked together — was rejected because the functions in `_clustering.py` have structurally different inputs that cannot share a `setup()`:

- `mask_delta_score` takes two 1D bool vectors of length `n_features`
- `reverse_window` takes a 1D int64 array of length `n_masks = 2**n_features`
- `reverse_window_score_gain` takes a 2D bool matrix `(n_masks, n_features)` plus an index array
- `delta_minimization_order` takes the same 2D matrix but runs a full 2-opt pass over it
- `pt_shuffle_rec` takes a partition tree built from scipy linkage

A shared `setup()` would have to allocate all of these simultaneously and pass the right subset to each method — fragile, hard to read, and wrong for functions like `delta_minimization_order` whose input size grows as `2**n_features` not `n_features`. One class per function keeps each `setup()` minimal and correct.

```
benchmarks/
└── benchmarks_clustering.py
    ├── BenchmarkMaskDeltaScore
    ├── BenchmarkReverseWindow
    ├── BenchmarkReverseWindowScoreGain
    ├── BenchmarkDeltaMinimizationOrder
    └── BenchmarkPtShuffleRec
```

As more `@njit` functions are ported from other source files (`_masked_model.py`, `_tabular.py`, etc.), each source file gets its own benchmark file following the same pattern.

### Input sizes anchored to real SHAP datasets

- `n_features=4` — iris
- `n_features=8` — california housing
- `n_features=13` — adult

`n_masks = 2**n_features` to reflect the exact explainer mask matrix size.

### Numba reference implementations copied inline

The `@njit` functions are copied verbatim from `shap/utils/_clustering.py` as `time_numba` baselines. This serves two purposes:

1. Like-for-like steady-state comparison excluding JIT warmup overhead (numba's warmup costs 200–500ms per function and would dominate any cold-start measurement).
2. When the source file has its `@njit` functions removed as part of the migration, the benchmark retains a numba baseline to compare against.

These copies will be deleted once numba is fully eliminated from the codebase.

### JIT warmup in `setup()`

ASV calls `setup()` once before timing begins, then calls the benchmark method repeatedly. We exploit this by calling each `@njit` function once inside `setup()` — this triggers numba's JIT compilation before the timer starts. Without this, the first timed call would include 200–500ms of compilation overhead, making numba look catastrophically slow against C++ and invalidating all steady-state comparisons.

```python
def setup(self, n_features):
    rng = np.random.default_rng(0)
    self.masks = rng.integers(0, 2, (2**n_features, n_features)).astype(bool)
    _delta_minimization_order(self.masks)  # force JIT compilation

def time_numba(self, n_features):
    _delta_minimization_order(self.masks)  # already compiled — measures steady-state only
```

### Degradation via `skip_benchmark_if`

Both `time_numba` and `time_cpp` are guarded independently using [`@skip_benchmark_if`](https://asv.readthedocs.io/en/stable/benchmarks.html#skipping-benchmarks) from `asv_runner.benchmarks.mark`. This is the correct skip pattern for ASV 0.6.x — the older approach of raising `NotImplementedError` inside the benchmark body still runs `setup()` and wastes time; `@skip_benchmark_if` is evaluated at collection time and skips the method entirely.

numba and each C++ function are imported under separate `_HAS_*` flags so a missing dependency skips only the affected method without touching any other benchmark.

```python
try:
    from numba import njit
    _HAS_NUMBA = True
except ImportError:
    _HAS_NUMBA = False
    def njit(fn):
        return fn

try:
    from shap._cutils import mask_delta_score as _cpp_mask_delta_score
    _HAS_MASK_DELTA_SCORE = True
except ImportError:
    _HAS_MASK_DELTA_SCORE = False

class BenchmarkMaskDeltaScore:
    @skip_benchmark_if(not _HAS_NUMBA)
    def time_numba(self, n_features): ...

    @skip_benchmark_if(not _HAS_MASK_DELTA_SCORE)
    def time_cpp(self, n_features): ...
```

**Note:** on `master`, all `time_cpp` methods currently show as SKIPPED because the C++ implementations live in unmerged #4654. This is intentional — the infrastructure is safe to merge independently.

## How to run

```bash
# All benchmarks in a file
asv run --python=same -b benchmarks_clustering

# All methods in one class
asv run --python=same -b BenchmarkDeltaMinimizationOrder

# Specific method
asv run --python=same -b BenchmarkDeltaMinimizationOrder.time_cpp
```

## Results (measured against #4654)

| Function | n=4 | n=8 | n=13 | Verdict |
|---|---|---|---|---|
| `mask_delta_score` | C++ 472ns / numba 438ns | C++ 486ns / numba 436ns | C++ 503ns / numba 444ns | numba ~7% — ns-scale, negligible |
| `reverse_window` | C++ 622ns / numba 581ns | C++ 786ns / numba 706ns | C++ 4.87μs / numba 3.32μs | numba faster; gap grows with array size |
| `reverse_window_score_gain` | C++ 536ns / numba 590ns | C++ 555ns / numba 563ns | C++ 559ns / numba 568ns | C++ ~9% at n=4, tied at n≥8 |
| `delta_minimization_order` | C++ 4.02μs / numba 43.6μs | C++ 989μs / numba 3.89ms | C++ 56.3ms / numba 173ms | **C++ ~11x / 3.9x / 3.1x faster** |
| `pt_shuffle_rec` | C++ 1.06μs / numba 1.04μs | C++ 1.16μs / numba 1.24μs | C++ 1.27μs / numba 1.52μs | C++ wins n≥8, tied at n=4 |

`delta_minimization_order` is the hot path — called once per `Explainer` invocation and dominates clustering time. C++ wins by 3–11x depending on feature count. `reverse_window` is the one function where numba consistently wins; LLVM auto-vectorises the tight in-place loop over large arrays in a way our plain C++ for-loop does not. This is expected and acceptable — `reverse_window` is called inside `delta_minimization_order`, and the overall function is still 3–11x faster in C++.

## AI Disclosure

- **Advised:** Explained ASV configuration and benchmarking methodology
- **Assisted:** Reviewed implementation for correctness and style consistency
- **Autonomous:** None — all code was written and verified by the contributor

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@ node_modules
 mpl-results
 # intermediate AI skill files
 .claude/disclosures/
+.asv/
--- a/asv.conf.json
+++ b/asv.conf.json
@@ -8,8 +8,5 @@
     "install_command": ["python -mpip install --no-build-isolation {build_dir}"],
     "show_commit_url": "https://github.com/shap/shap/commit/",
     "pythons": ["3.12"],
-    "benchmark_dir": "benchmarks",
-    "results_dir": ".asv/results",
-    "html_dir": ".asv/html",
-    "build_cache_size": 2
+    "benchmark_dir": "benchmarks"
 }
--- a/asv.conf.json
+++ b/asv.conf.json
@@ -1,0 +1,15 @@
+{
+    "version": 1,
+    "project": "shap",
+    "project_url": "https://github.com/shap/shap",
+    "repo": ".",
+    "branches": ["master"],
+    "environment_type": "virtualenv",
+    "install_command": ["python -mpip install --no-build-isolation {build_dir}"],
+    "show_commit_url": "https://github.com/shap/shap/commit/",
+    "pythons": ["3.12"],
+    "benchmark_dir": "benchmarks",
+    "results_dir": ".asv/results",
+    "html_dir": ".asv/html",
+    "build_cache_size": 2
+}
--- a/benchmarks/benchmarks_clustering.py
+++ b/benchmarks/benchmarks_clustering.py
@@ -43,6 +43,13 @@ try:
 except ImportError:
     _HAS_DELTA_MINIMIZATION_ORDER = False
 
+try:
+    from shap._cutils import pt_shuffle_rec as _cpp_pt_shuffle_rec
+
+    _HAS_PT_SHUFFLE_REC = True
+except ImportError:
+    _HAS_PT_SHUFFLE_REC = False
+
 
 # --- Numba reference implementations ---
 
@@ -69,6 +76,25 @@ def _delta_minimization_order(all_masks, max_swap_size=100, num_passes=2):
                 if _reverse_window_score_gain(all_masks, order, i, length) > 0:
                     _reverse_window(order, i, length)
     return order
+
+
+@njit
+def _pt_shuffle_rec(i, indexes, index_mask, partition_tree, M, pos):
+    if i < 0:
+        if index_mask[i + M]:
+            indexes[pos] = i + M
+            return pos + 1
+        else:
+            return pos
+    left = int(partition_tree[i, 0] - M)
+    right = int(partition_tree[i, 1] - M)
+    if np.random.randn() < 0:
+        pos = _pt_shuffle_rec(left, indexes, index_mask, partition_tree, M, pos)
+        pos = _pt_shuffle_rec(right, indexes, index_mask, partition_tree, M, pos)
+    else:
+        pos = _pt_shuffle_rec(right, indexes, index_mask, partition_tree, M, pos)
+        pos = _pt_shuffle_rec(left, indexes, index_mask, partition_tree, M, pos)
+    return pos
 
 
 @njit
@@ -158,3 +184,49 @@ class BenchmarkDeltaMinimizationOrder:
     @skip_benchmark_if(not _HAS_DELTA_MINIMIZATION_ORDER)
     def time_cpp(self, n_features):
         _cpp_delta_minimization_order(self.masks)
+
+
+class BenchmarkPtShuffleRec:
+    params = [4, 8, 13]
+    param_names = ["n_features"]
+
+    def setup(self, n_features):
+        from scipy.cluster.hierarchy import ward
+        from scipy.spatial.distance import pdist
+
+        rng = np.random.default_rng(0)
+        data = rng.random((n_features, n_features))
+        self.partition_tree = ward(pdist(data)).astype(np.float64)
+        self.M = n_features
+        self.index_mask = np.ones(n_features, dtype=bool)
+        self.indexes = np.zeros(n_features, dtype=np.int64)
+        # force JIT compilation
+        _pt_shuffle_rec(
+            self.partition_tree.shape[0] - 1,
+            self.indexes.copy(),
+            self.index_mask,
+            self.partition_tree,
+            self.M,
+            0,
+        )
+
+    def time_numba(self, n_features):
+        _pt_shuffle_rec(
+            self.partition_tree.shape[0] - 1,
+            self.indexes.copy(),
+            self.index_mask,
+            self.partition_tree,
+            self.M,
+            0,
+        )
+
+    @skip_benchmark_if(not _HAS_PT_SHUFFLE_REC)
+    def time_cpp(self, n_features):
+        _cpp_pt_shuffle_rec(
+            self.partition_tree.shape[0] - 1,
+            self.indexes.copy(),
+            self.index_mask,
+            self.partition_tree,
+            self.M,
+            0,
+        )
--- a/benchmarks/benchmarks_clustering.py
+++ b/benchmarks/benchmarks_clustering.py
@@ -213,8 +213,7 @@ class BenchmarkPtShuffleRec:
         self.M = n_features
         self.index_mask = np.ones(n_features, dtype=bool)
         self.indexes = np.zeros(n_features, dtype=np.int64)
-        # force JIT compilation
-        _pt_shuffle_rec(
+        _pt_shuffle_rec(  # force JIT compilation
             self.partition_tree.shape[0] - 1,
             self.indexes.copy(),
             self.index_mask,
--- a/benchmarks/benchmarks_clustering.py
+++ b/benchmarks/benchmarks_clustering.py
@@ -1,0 +1,52 @@
+"""Benchmarks for clustering utility functions.
+
+Numba reference implementations are copied from shap/utils/_clustering.py
+for migration validation. They will be removed once numba is fully eliminated.
+
+Input sizes are anchored to real SHAP datasets:
+  n_features=4  — iris (4 features)
+  n_features=8  — california housing (8 features)
+  n_features=13 — adult (13 features)
+
+n_masks is derived as 2**n_features to reflect the exact explainer mask matrix size.
+"""
+
+import numpy as np
+from asv_runner.benchmarks.mark import skip_benchmark_if
+from numba import njit
+
+try:
+    from shap._cutils import mask_delta_score as _cpp_mask_delta_score
+
+    _HAS_MASK_DELTA_SCORE = True
+except ImportError:
+    _HAS_MASK_DELTA_SCORE = False
+
+
+# --- Numba reference implementations ---
+
+
+@njit
+def _mask_delta_score(m1, m2):
+    return (m1 ^ m2).sum()
+
+
+# --- Benchmarks ---
+
+
+class BenchmarkMaskDeltaScore:
+    params = [4, 8, 13]
+    param_names = ["n_features"]
+
+    def setup(self, n_features):
+        rng = np.random.default_rng(0)
+        self.m1 = rng.integers(0, 2, n_features).astype(bool)
+        self.m2 = rng.integers(0, 2, n_features).astype(bool)
+        _mask_delta_score(self.m1, self.m2)  # force JIT compilation
+
+    def time_numba(self, n_features):
+        _mask_delta_score(self.m1, self.m2)
+
+    @skip_benchmark_if(not _HAS_MASK_DELTA_SCORE)
+    def time_cpp(self, n_features):
+        _cpp_mask_delta_score(self.m1, self.m2)
--- a/benchmarks/benchmarks_clustering.py
+++ b/benchmarks/benchmarks_clustering.py
@@ -13,7 +13,19 @@ n_masks is derived as 2**n_features to reflect the exact explainer mask matrix s
 
 import numpy as np
 from asv_runner.benchmarks.mark import skip_benchmark_if
-from numba import njit
+from scipy.cluster.hierarchy import ward
+from scipy.spatial.distance import pdist
+
+try:
+    from numba import njit
+
+    _HAS_NUMBA = True
+except ImportError:
+    _HAS_NUMBA = False
+
+    def njit(fn):
+        return fn
+
 
 try:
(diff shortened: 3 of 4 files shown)


shauryam2807:
Nice work on the production-ready port and ASV setup! My POC in #4646 validated the all-5-together approach — glad to see it taken to completion here. I'll focus on porting the next set of numba functions from `_masked_model.py` to keep the migration moving.

bedupako12mas:
Thanks @shauryam2807! Happy to coordinate — _masked_model.py is all yours. I'll take _compute_grey_code_row_values_st (extends the existing grey_code_utils.h) and the two tabular  
  masker functions (_single_delta_mask, _delta_masking) as my next PRs once #4654 merges and the maintainers are happy with the pattern.

shauryam2807:
Hy my friend @bedupako12mas I see somthing in your code
Nice initiative setting up the ASV infrastructure! Reviewed the benchmarks and noticed a few issues:

1. **Dtype mismatch in numba baselines**: The inline `@njit` copies use `bool` arrays (line 128: `.astype(bool)`), but in production the masks passed to these functions are `np.bool_` arrays created via `np.zeros(..., dtype=bool)` in `_exact.py:334` (`all_masks = np.array(all_masks)` where each mask is already `dtype=bool`). So the C++ side is correct with `bool`, but the numba copies here are also being fed `bool` — which means the original `int64` XOR path that numba JIT-compiles in production (`(m1 ^ m2).sum()` on int64) is **not** being benchmarked. The numba baseline is benchmarking a `bool` XOR path that may have different performance characteristics from what users actually experience.

2. **`BenchmarkReverseWindow` benchmarks `.copy()` overhead, not just `reverse_window`**: Both `time_numba` and `time_cpp` (lines 154, 158) call `self.order.copy()` inside the timed method. ASV measures wall time of the entire method body, so every iteration includes `numpy.copy()` overhead. This makes the benchmark partially a `numpy.copy` benchmark rather than a pure `reverse_window` benchmark. Consider copying in `setup()` and operating in-place, or using ASV's `setup_cache` pattern.

3. **`asv.conf.json` hardcodes `"pythons": ["3.12"]`**: This prevents contributors on 3.11 or 3.13 from running benchmarks without editing the config. Consider omitting the field entirely (ASV defaults to system Python) or using `["3.11", "3.12", "3.13"]`.

4. **`from asv_runner.benchmarks.mark import skip_benchmark_if` at module level (line 19)**: This is a hard import — if `asv_runner` is not installed, the entire benchmark file fails to import and ASV reports a collection error rather than gracefully skipping. A `try/except` fallback (like you did for numba) would be more robust for contributors running benchmarks outside a full ASV environment.

Happy to help with any of these!

daidahao:
@bedupako12mas Thank you very much for the contribution here. Please allow me up to a week time to review and evaluate the monitoring stack here, as this is an important addition to shap.

@shauryam2807 Hi, I would recommend using the review function on GitHub for logging specific comments and suggestions, and leaving only general comments here.

bedupako12mas:
Thanks for the review @shauryam2807! Going through each point:

**1. Dtype mismatch in numba baselines**

The dtype is consistent throughout. The actual production call site for `delta_minimization_order` is `_exact.py` — line 306 creates masks with `np.zeros(M, dtype=bool)`, line 313 converts the list with `np.array(all_masks)` preserving that dtype, and line 320 passes the result directly to `delta_minimization_order`. The benchmarks match this with `.astype(bool)` (lines 132–133 in `benchmarks_clustering.py`). There is no "int64 XOR path" in the production call sites; both the numba and C++ baselines exercise the same bool path that production uses.

**2. `.copy()` overhead in `BenchmarkReverseWindow`**

This is the correct pattern for benchmarking an in-place function in ASV. ASV calls `setup()` once before all iterations, not before each individual call — there is no per-iteration reset hook. Moving the copy to `setup()` would mean the second iteration operates on an already-reversed array (lines 158 and 162 in `benchmarks_clustering.py`), corrupting the measurement.

NumPy uses the same approach in their own benchmarks for exactly this reason — `numpy/benchmarks/benchmarks/bench_function_base.py` lines 247–251 explicitly uses `np.sort()` (out-of-place) instead of `arr.sort()` (in-place), with the comment: *"This is important because the data is prepared once per benchmark, but used across multiple runs."*

**3. `"pythons": ["3.12"]` in `asv.conf.json`**

This is intentional. `pyproject.toml` line 21 declares `requires-python = ">=3.12"`, so 3.11 is not a supported Python for this project. Pinning to 3.12 keeps benchmarks consistent with the project's minimum supported version.

**4. `asv_runner` hard import**

`asv_runner` is installed as a direct dependency of `asv` itself — it is not an optional package. Any environment that can run `asv run` already has `asv_runner` present (line 15 in `benchmarks_clustering.py`). The per-function `_HAS_*` flags already handle the actual optional dependencies (numba, the C++ extension).

shauryam2807:
Hello
Thanks for the detailed clarifications, @bedupako12mas — you're right about the NumPy ASV convention for in-place functions and the dtype tracking in `_exact.py`.

@daidahao Understood, apologies for the noise — I'll use the Files Changed review tab for code-specific comments going forward.

daidahao:
Hi @bedupako12mas and all,

Thanks again for this PR. After reviewing the code and "benchmarking" (I prefer calling it performance monitoring to distinguish from current `shap.benchmark`), I have two main concerns regarding this current approach (none of which is a critique of your code because I think it is a fine demo of what asv could deliver!):

1. How do we ensure hardware consistency when monitoring?

All asv code run locally and we cannot guarantee every dev has the same hardware as others. Monitoring relies on self-reporting by the developers, and there is no "standard" hardware system to ensure consistency.

With that in mind, [codespeed](https://codspeed.io) seems more applicable. It integrates with `pytest` which we already use, runs everything on CIs, and uses GitHub Actions runners to ensure same hardware. However, I would caution that system/hardware configuration on GitHub Actions change over time. Also, CI runs can be very flaky sometimes due to networking issues.

2. Do we monitor individual functions or end-to-end SHAP pipeline?

The current PR measures runtime of individual numba functions like `delta_minimization_order`. While the monitoring set-up is simple, it does not really tell us directly how code changes affect explainer runtime (`ExactExplainer` in this case). Does the new nanobind function really speed up the end-to-end explainer runtime? Or does it only add marginal speedup? Should we prefer monitoring end-to-end explainer runtime?

I think we all benefit from a robust and realistic performance monitoring infrastructure that can produce consistent metrics and serve us in the long term.

@CloseChoice @bedupako12mas @shauryam2807 

Looking forward to hearing what you think best solutions might be, before proceeding further with this PR.

shauryam2807:
Hello
Thanks for raising these points, @daidahao — both are important for getting this right long-term.

On hardware consistency: I agree that local ASV runs make cross-developer comparisons unreliable. CodSpeed looks like a strong fit — CI-based execution on GitHub Actions runners gives us reproducible hardware, and the pytest integration means we don't need to maintain a separate benchmarking framework alongside our existing test suite. One thing worth noting is that GitHub Actions runner specs can shift over time, so absolute numbers may drift — but relative comparisons within the same CI run (e.g., numba vs nanobind on the same commit) should stay consistent, which is what we actually care about for the migration.

On function-level vs end-to-end: I think we need both. Function-level benchmarks are useful during development to catch regressions in individual ports and pinpoint where speedups/slowdowns come from. But end-to-end explainer benchmarks (e.g., ExactExplainer.shap_values() on a standard dataset like adult/iris) are what actually matter to users and answer the question "did this migration make SHAP faster?" I'd suggest we start with a small set of end-to-end benchmarks for ExactExplainer and PartitionExplainer alongside the function-level ones, so we can report both granular and user-facing metrics.

Happy to help set up the CodSpeed integration or write the end-to-end benchmark cases if that direction sounds good!

bedupako12mas:
Thanks for this @daidahao — both concerns are valid and I hadn't fully thought through either of them.

**1. Hardware consistency**

You're right that local ASV runs make cross-developer comparisons unreliable — there's no way to know whether a timing difference reflects the implementation or the machine. I was thinking about the nanobind migration workflow (catching regressions per-port) and extensibility across the full migration scope (#4327) — the structure maps one benchmark file per source file as the porting progresses — rather than a project-wide monitoring standard. Looking at CodSpeed, I can see why it fits better here — it runs on fixed GitHub Actions runners so the numbers are reproducible across contributors and commits, and the pytest integration means we don't need a separate benchmarking framework alongside the existing test suite. I'd support moving in that direction.

**2. Function-level vs end-to-end**

Also fair. I benchmarked `delta_minimization_order` in isolation and assumed the 3–11x speedup would translate to `ExactExplainer` runtime — but that assumption has two problems. First, I don't actually know what fraction of total explainer runtime that function accounts for. Second, even if the fraction is significant, function-level benchmarks don't capture Python overhead, the cost of passing data between Python and C++ at each call boundary, or cache effects from the surrounding explainer computation. A user running `ExactExplainer.shap_values()` cares whether their script runs faster, not whether `delta_minimization_order` is 11x faster in isolation. End-to-end benchmarks on standard datasets (adult, iris) would answer the question that actually matters to users.

Happy to adapt this PR, close it in favour of a CodSpeed-based approach, or contribute wherever is most useful. What direction do you and @CloseChoice want to take this?

daidahao:
@shauryam2807 @bedupako12mas 

Hi both, thank you for your suggestions.

Let us wait and see what @CloseChoice says before proceeding further.

shauryam2807:
Hello @daidahao @CloseChoice @bedupako12mas 

I want to know is there any update regarding this?

CloseChoice:
thanks for the PR. We merged https://github.com/shap/shap/pull/5071 though, therefore I close this one.

PR #4666, opened 2026-04-18
Restored League of Legends notebook with mock data fallback

archittmittal:
This PR restores the League of Legends Win Prediction notebook by adding a mock data generation fallback when the external Kaggle dataset is missing. This allows the notebook to pass in CI environments without requiring large external downloads. The notebook has also been removed from the CI exclusion list.


archittmittal:
Hi @CloseChoice and @dsgibbons, I've restored the League of Legends Win Prediction notebook by adding a mock data fallback for CI environments and enabled it in the notebook test suite. Could you please take a look when you have a moment? Thanks!

PR #4680, opened 2026-04-19
fix(TreeExplainer): warn on ignored link and linearize_link params

### LINKED ISSUE #3513: BUG: TreeExplainer ignores "link" argument passed by the Explainer API
The current implementation of TreeExplainer entirely ignores the `link` and `linearize_link` arguments which are passed from `Explainer` here:

https://github.com/shap/shap/blob/b3573bd333cb24090deab373319ec0a662dc5805/shap/explainers/_explainer.py#L188

They are currently caught by the "deprecated options" catchall, and subsequently ignored:

https://github.com/shap/shap/blob/b3573bd333cb24090deab373319ec0a662dc5805/shap/explainers/_tree.py#L63-L72

This seems rather concerning. It's not obvious to me where precisely the link function should be applied in TreeExplainer, if at all. As far as I can make out, the TreeExplainer uses the "model_output" inferred from the explained model rather than the user-defined "link" variable.

ritankarsaha:
## Overview

Closes #3513 

  - `link=` and `linearize_link=` were silently ignored by `TreeExplainer`, producing wrong SHAP attributions with no diagnostic message 
  - Now raises `FutureWarning` when either is passed as non-`None`, pointing users to `model_output='probability'` as the correct alternative.
  - Adds 3 regression tests covering both warning cases and the no-warning baseline.

## Checklist

### DIFF
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -142,7 +142,6 @@ class TreeExplainer(Explainer):
         feature_perturbation: Literal["auto", "interventional", "tree_path_dependent"] = "auto",
         feature_names: list[str] | None = None,
         approximate: Any = DEPRECATED_APPROX,
-        # FIXME: The `link` and `linearize_link` arguments are ignored. GH #3513
         link: Any = None,
         linearize_link: Any = None,
     ) -> None:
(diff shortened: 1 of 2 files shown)


ritankarsaha:
@CloseChoice can you please take a look into this PR, thanks !

PR #4682, opened 2026-04-19
ENH: Migrate MaskedModel backend to nanobind (removes Numba dependency & adds custom link support)

shauryam2807:
## Description
This PR completes the next stage of the C++ extension migration by porting the `MaskedModel` logic from Numba to high-performance C++ using `nanobind`. 

Previously, `shap.utils._masked_model` relied heavily on Numba's `@njit` decorators. While performant, this created a strict dependency on Numba and introduced a significant architectural limitation: Numba could not compile arbitrary Python functions, preventing the use of custom user-defined link functions in `MaskedModel`.

By migrating the core computational bottleneck (`_build_fixed_output`) to native C++, we have entirely eliminated the Numba dependency for this module while simultaneously unlocking support for dynamic link functions via `nanobind`'s Python interoperability capabilities.

### Key Changes
* **C++ Extension (`masked_model_utils.h`)**: Implemented high-performance C++ variants of `init_masks`, `rec_fill_masks`, `build_fixed_single_output`, and `build_fixed_multi_output`. 
* **Custom Link Function Support**: Passed arbitrary Python callables into the C++ hot-path using `nb::callable`. This resolves long-standing issues where custom link functions (or unsupported NumPy functions) would cause Numba JIT compilation crashes.
* **Universal Dtype Support & Casting**: Handled both `float32` and `float64` execution paths dynamically in C++. Implemented an explicit Python-side `_upcast_array` utility to safely promote `float16` inputs to `float32` before they cross the C++ boundary, avoiding underlying array cast failures.
* **Robust Dispatching**: Implemented a variadic `nb::args` fallback in `cutils.cpp` to ensure robust overload resolution across different OS platforms when dealing with complex 8-argument Python-to-C++ boundaries containing mixed `ndarray` and `None` types.
* **Removed Numba**: Dropped the Numba dependency and all `@njit` decorators from `_masked_model.py`.

## Checklist
- [x] All existing tests in `tests/utils/test_masked_model.py` pass successfully.
- [x] Verified correctness for single-output arrays, multi-output arrays, and `float16` types.
- [x] Verified that custom link functions execute safely within the C++ loop.
- [x] Code style matches the surrounding `cutils` architecture.

## Related Issues
- Progress towards removing the Numba dependency completely from the SHAP ecosystem (GSoC 2026).

### DIFF
--- a/shap/_cutils.pyi
+++ b/shap/_cutils.pyi
@@ -13,7 +13,6 @@ def compute_grey_code_row_values(
     extended_delta_indexes: Annotated[NDArray[numpy.uint64], dict(shape=(None,), device="cpu")],
     noop_code: int,
 ) -> None: ...
-
 @overload
 def compute_grey_code_row_values(
     row_values: Annotated[NDArray[numpy.float64], dict(shape=(None, None), device="cpu")],
@@ -24,7 +23,6 @@ def compute_grey_code_row_values(
     extended_delta_indexes: Annotated[NDArray[numpy.uint64], dict(shape=(None,), device="cpu")],
     noop_code: int,
 ) -> None: ...
-
 def compute_exp_val(
     nsamples_run: int,
     nsamples_added: int,
--- a/shap/_cutils.pyi
+++ b/shap/_cutils.pyi
@@ -12,8 +12,7 @@ def compute_grey_code_row_values(
     shapley_coeff: Annotated[NDArray[numpy.float64], dict(shape=(None,), device="cpu")],
     extended_delta_indexes: Annotated[NDArray[numpy.uint64], dict(shape=(None,), device="cpu")],
     noop_code: int,
-) -> None:
-    """Compute the row values for the grey code algorithm in 1D"""
+) -> None: ...
 
 @overload
 def compute_grey_code_row_values(
@@ -24,8 +23,7 @@ def compute_grey_code_row_values(
     shapley_coeff: Annotated[NDArray[numpy.float64], dict(shape=(None,), device="cpu")],
     extended_delta_indexes: Annotated[NDArray[numpy.uint64], dict(shape=(None,), device="cpu")],
     noop_code: int,
-) -> None:
-    """Compute the row values for the grey code algorithm in 2D"""
+) -> None: ...
 
 def compute_exp_val(
     nsamples_run: int,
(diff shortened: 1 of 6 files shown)


daidahao:
Hi @shauryam2807, thanks for the PR here. Sorry I would not be able to provide full review until performance monitoring like the one proposed in #4661 is fully set up.

From a quick glance, your PR seems to have included #4654 by @bedupako12mas. Please do not include others' PRs in your PR, as this would make both review and merge impossible.

shauryam2807:
Hello
Hi @daidahao, thanks for the feedback! Apologies for the delayed response.

You're absolutely right — I accidentally included commits from #4654 in my branch history. I'll rebase against upstream master and drop those commits so this PR only contains the MaskedModel nanobind migration work. I'll force-push the cleaned branch shortly.

Regarding performance monitoring, that makes total sense. Happy to wait for #4661 to be fully set up before a full review. In the meantime, I'll also work on fixing the remaining CI/Mypy failures so the PR is in a clean state when you're ready to review.

Thanks again for your guidance!

shauryam2807:
Hey @daidahao,

Just a quick update following up on my last comment: the git history has been cleaned up, and **all CI checks are now fully passing!** ✅

I spent a lot of time doing a deep-dive to track down the root cause of the previous test regressions in the tabular explainers. It was a challenging debugging process, but I learned a ton about how Python and C++ interact under the hood! It turned out to be a combination of two low-level issues in the new C++ backend:

1. **Dtype Mismatches (`std::bad_cast`):** Models like XGBoost often return `float32` outputs. The Python buffer allocation defaulted to `float64`, causing `nanobind` to throw a catastrophic cast exception when bridging the `float32` outputs to the C++ `double` signature.
2. **Out-Of-Bounds Memory Reads (Segfault):** During partial mask updates (where some features are identical to the background), the C++ initialization loop was attempting to read the full `sample_count` from the `outputs` buffer, reading past the allocated memory and causing a segfault.

Both of these have been resolved by strictly enforcing dtype alignment in Python before bridging, and relying on explicit zero-fills for safe buffer initialization in C++.

The codebase is clean and stable now. I'll stand by for PR #4661 to merge so we can get the performance monitoring set up for a final review!

CloseChoice:
Thanks for the PR, this was superseded by #5092

PR #4698, opened 2026-04-20
 feat(plots): add ax parameter to shap.plots.waterfall

### LINKED ISSUE #4697: ENH: shap.plots.waterfall does not accept an ax parameter, making subplot grids impossible
### Problem Description

Every modern SHAP plot (beeswarm, bar, scatter, heatmap, violin) accepts an ax parameter so it can be drawn into a caller-supplied matplotlib.axes.Axes object. shap.plots.waterfall does not — it hard-codes plt.gcf() / plt.gca() throughout, which means users cannot embed waterfall plots in plt.subplots() grids. A common use case (comparing two predictions side-by-side with fig, (ax1, ax2) = plt.subplots(1, 2)) is currently impossible without ugly monkey-patching.

### Alternative Solutions

_No response_

### Additional Context

_No response_

### Feature request checklist

ritankarsaha:
## Overview

Closes #4697 

Adds ax=None to shap.plots.waterfall, matching the interface already offered by bar, beeswarm, scatter, heatmap, and violin.
Changes:
  - shap/plots/_waterfall.py: Added ax=None parameter. When ax is None (default), behaviour is identical to before — plt.gca() is used and the figure is auto-sized. When an Axes is passed, the plot renders into it and the figure is not resized. All internal plt.xxx() calls replaced with ax.xxx() / fig.xxx() equivalents. Return value changed from plt.gca() to ax (still the same object in the default case, now guaranteed to be the passed-in axes).
  - tests/plots/test_waterfall.py: Added test_waterfall_ax_parameter which creates a 1×2 subplot grid, draws two waterfall plots into each cell, asserts the returned axes is the one passed in, and asserts figure size was not altered.
- Backward compatibility: Fully backward-compatible — ax=None default preserves all existing behaviour.

## Checklist

### DIFF
--- a/shap/plots/_waterfall.py
+++ b/shap/plots/_waterfall.py
@@ -13,7 +13,7 @@ from ._style import get_style
 # plot that is associated with that feature get overlaid on the plot...it would quickly allow users to answer
 # why a feature is pushing down or up. Perhaps the best way to do this would be with an ICE plot hanging off
 # of the bar...
-def waterfall(shap_values, max_display=10, show=True):
+def waterfall(shap_values, max_display=10, show=True, ax=None):
     """Plots an explanation of a single prediction as a waterfall plot.
 
     The SHAP value of a feature represents the impact of the evidence provided by that feature on the model's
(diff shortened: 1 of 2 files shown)


ritankarsaha:
@CloseChoice can you please take a look into this PR ?

PR #4719, opened 2026-04-21
Add type stubs for C extension utilities

Avichal296:
Add type stubs for C extension utilities module


daidahao:
@Avichal296 Thanks for the PR. But I would advise NOT tracking cutils stub because it is auto generated from nanobind. Also, the PR contains unrelated code.

Avichal296:
@daidahao Thanks for the feedback! I've updated the PR to include only relevant test cases for shap.utils.image, improved assertions, and added edge case coverage.

PR #4726, opened 2026-04-21
FIX: avoid mutable default argument in SequentialPerturbation

### LINKED ISSUE #4723: Improve code quality in SequentialPerturbation (mutable default argument and minor clarity fixes)
Hi! While going through the benchmarking code, I noticed a small improvement in the SequentialPerturbation implementation.

In the __call__ method, the parameter `indices=[]` uses a mutable default argument. In Python, mutable defaults can lead to unexpected behavior across multiple function calls since the same object is reused.

It might be safer to replace it with:
    indices=None
and initialize it inside the function:
    if indices is None:
        indices = []

Additionally, I noticed a minor improvement opportunity in one of the error messages:
"DataFrame arguments dont iterate correctly"

→ could be updated for clarity and grammar.

These are small changes but can improve code safety and readability.

I'd be happy to work on this and submit a PR if this sounds good.

ashwinkothavade:
## Summary

Closes #4723.

Two small code-quality fixes in `shap/benchmark/_sequential.py`:

1. **Mutable default argument** — `SequentialPerturbation.__call__` had `indices=[]`. This is the classic Python footgun: the default list is shared across all calls. It is currently latent because the parameter is not consumed in the function body, but leaving it as `[]` is a trap for the next person who adds code that mutates it. Changed to `indices=None`, matching the recommendation in the issue. No caller is affected.
2. **Error message grammar** — `"DataFrame arguments dont iterate correctly, pass numpy arrays instead!"` → `"DataFrame arguments don't iterate correctly; please pass numpy arrays instead."`

I noticed `indices` (and `y`) appear to be unused within `__call__`. I left them in place to keep this PR minimal; happy to open a follow-up to remove the dead parameters if maintainers agree.

## Tests

Added `tests/benchmark/test_sequential.py` with two regression tests:
- `test_sequential_perturbation_indices_default_is_not_mutable` — pins the default via `inspect.signature`.
- `test_sequential_masker_rejects_dataframe_with_clear_message` — exercises the `TypeError` path and matches the new message.

## Local verification

- `ruff check` and `ruff format --check` both pass on the modified files.
- Full pytest was not run locally (the C extension requires CMake which I don't have set up on this machine). Relying on CI for the test run.

### DIFF
--- a/shap/benchmark/_sequential.py
+++ b/shap/benchmark/_sequential.py
@@ -17,7 +17,7 @@ class SequentialMasker:
     def __init__(self, mask_type, sort_order, masker, model, *model_args, batch_size=500):
         for arg in model_args:
             if isinstance(arg, pd.DataFrame):
-                raise TypeError("DataFrame arguments dont iterate correctly, pass numpy arrays instead!")
+                raise TypeError("DataFrame arguments don't iterate correctly; please pass numpy arrays instead.")
 
         # convert any DataFrames to numpy arrays
         # self.model_arg_cols = []
@@ -91,7 +91,7 @@ class SequentialPerturbation:
         explanation,
         *model_args,
         percent=0.01,
-        indices=[],
+        indices=None,
         y=None,
         label=None,
         silent=False,
--- a/tests/benchmark/test_sequential.py
+++ b/tests/benchmark/test_sequential.py
@@ -1,0 +1,32 @@
+"""Tests for shap.benchmark._sequential."""
+
+import inspect
+
+import numpy as np
+import pandas as pd
+import pytest
+
+from shap.benchmark._sequential import SequentialMasker, SequentialPerturbation
+from shap.maskers import Independent
+
+
+def _model(x):
+    return x.sum(axis=-1)
+
+
+def test_sequential_perturbation_indices_default_is_not_mutable():
+    # Regression test for https://github.com/shap/shap/issues/4723.
+    # A mutable default argument would be shared across calls and could cause
+    # cross-call state bleed if the function body is ever changed to mutate it.
+    default = inspect.signature(SequentialPerturbation.__call__).parameters["indices"].default
+    assert default is None
+
+
+def test_sequential_masker_rejects_dataframe_with_clear_message():
+    rs = np.random.RandomState(0)
+    X = rs.random((4, 3))
+    df = pd.DataFrame(X, columns=["a", "b", "c"])
+    masker = Independent(X)
+
+    with pytest.raises(TypeError, match=r"don't iterate correctly"):
+        SequentialMasker("keep", "positive", masker, _model, df)
--- a/tests/benchmark/test_sequential.py
+++ b/tests/benchmark/test_sequential.py
@@ -1,25 +1,15 @@
 """Tests for shap.benchmark._sequential."""
 
-import inspect
-
 import numpy as np
 import pandas as pd
 import pytest
 
-from shap.benchmark._sequential import SequentialMasker, SequentialPerturbation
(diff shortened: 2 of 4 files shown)


tests/benchmark/test_sequential.py
```diff
@@ -0,0 +1,32 @@
...
+
+def _model(x):
+    return x.sum(axis=-1)
+
+
+def test_sequential_perturbation_indices_default_is_not_mutable():
+    # Regression test for https://github.com/shap/shap/issues/4723.
+    # A mutable default argument would be shared across calls and could cause
+    # cross-call state bleed if the function body is ever changed to mutate it.
+    default = inspect.signature(SequentialPerturbation.__call__).parameters["indices"].default
```
tylerjereddy: I'd recommend just using the `ruff` rule that prevents this instead of rolling your own detection for it--see: https://docs.astral.sh/ruff/rules/mutable-argument-default/

That rule is used by many Python OSS projects.

shap/benchmark/_sequential.py
```diff
@@ -91,7 +91,7 @@ def __call__(
         explanation,
         *model_args,
         percent=0.01,
-        indices=[],
+        indices=None,
```
tylerjereddy, line 94: Just a drive by review, but usually when making this change, one needs to set the "old default" in the source below when `None` is present.

tylerjereddy (changes requested):

tests/benchmark/test_sequential.py
ashwinkothavade: @tylerjereddy 
Good call — `B` isn't currently in `[tool.ruff.lint.select]` in `pyproject.toml`, so I'll enable `B006` via `extend-select` and drop my custom `inspect.signature` test (the lint rule makes it redundant).

Running `ruff check --select B006 .` on current master flags 4 other sites:

- `shap/benchmark/_explanation_error.py:92` — same pattern, `indices=[]` in `__call__`, and `indices` is likewise unused in the body
- 2 cells in `notebooks/overviews/Be careful when interpreting predictive models...ipynb`
- 1 cell in `notebooks/tabular_examples/model_agnostic/Simple California Demo.ipynb`

Plan for this PR:

- fix `_explanation_error.py` the same way as `_sequential.py`
- add a per-file-ignore for the notebook cells (they're reader-facing example code; I'd rather not quietly change behaviour there as part of a lint-adoption PR — happy to open a separate notebook PR if useful)
- remove my `test_sequential_perturbation_indices_default_is_not_mutable` — superseded by the lint rule
- keep `test_sequential_masker_rejects_dataframe_with_clear_message` (covers the grammar fix)

If you'd rather I enable the full `B` ruleset instead of just `B006`, say the word — I just didn't want to bundle surprises into this PR.

shap/benchmark/_sequential.py
ashwinkothavade: @tylerjereddy 
Agreed on the general pattern. In this specific spot `indices` isn't referenced anywhere in the body of `__call__` — same story at `shap/benchmark/_explanation_error.py:92` (also flagged by B006). So `if indices is None: indices = []` here would be dead code shadowing a dead parameter.

Two options:

a) Leave as `indices=None` in both sites and flag the unused parameter in the PR description as a candidate for a follow-up cleanup (minor breaking change if anyone is passing it as a kwarg, though since it's a no-op the behavioural impact is zero).

b) Remove `indices` (and `y` in `_sequential.__call__`, which is also unused) in this PR.

Leaning toward (a) to keep the PR surgical and let maintainers weigh in on the removal separately, but happy to do (b) here if you'd prefer. What's your call?

Issue #3662, opened 2024-05-16
Demangle pytorch and tensorflow dependencies


CloseChoice:
### Issue Description

Currently our dependencies for the deep explainer are mangled together somewhere deep down. This best visible when running the notebook `notebooks/image_examples/image_classification/PyTorch Deep Explainer MNIST example.ipynb` without `tf_keras` installed, then one gets an error. This is completely counter intuitive since this notebook just needs pytorch related dependencies. We should definitely get this into a state where pytorch and tensorflow code can coexist without the other framework being installed.

### Minimal Reproducible Example

```python
pip uninstall tf_keras

then run `notebooks/image_examples/image_classification/PyTorch Deep Explainer MNIST example.ipynb`

then this throws an error in the cell



# since shuffle=True, this is a random sample of test data
batch = next(iter(test_loader))
images, _ = batch

background = images[:100]
test_images = images[100:103]

e = shap.DeepExplainer(model, background)
shap_values = e.shap_values(test_images)
``
```

### Traceback

```shell
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
File ~/programming/github/shap_master/.venv/lib/python3.10/site-packages/transformers/activations_tf.py:22
     21 try:
---> 22     import tf_keras as keras
     23 except (ModuleNotFoundError, ImportError):

ModuleNotFoundError: No module named 'tf_keras'

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
File ~/programming/github/shap_master/.venv/lib/python3.10/site-packages/transformers/utils/import_utils.py:1472, in _LazyModule._get_module(self, module_name)
   1471 try:
-> 1472     return importlib.import_module("." + module_name, self.__name__)
   1473 except Exception as e:

File /usr/lib/python3.10/importlib/__init__.py:126, in import_module(name, package)
    125         level += 1
--> 126 return _bootstrap._gcd_import(name[level:], package, level)

File <frozen importlib._bootstrap>:1050, in _gcd_import(name, package, level)

File <frozen importlib._bootstrap>:1027, in _find_and_load(name, import_)
...
   1476         f" traceback):\n{e}"
   1477     ) from e

RuntimeError: Failed to import transformers.modeling_tf_utils because of the following error (look up to see its traceback):
Your currently installed version of Keras is Keras 3, but this is not yet supported in Transformers. Please install the backwards-compatible tf-keras package with `pip install tf-keras`.
```

### Expected Behavior

[...]

PR #4728, opened 2026-04-21
TST: guard against TF imports on the PyTorch DeepExplainer path

ashwinkothavade:
## Summary

Refs #3662.

I tried to reproduce the original bug in a clean env (`torch` installed, no `tensorflow`/`tf_keras`/`keras`) on current master and couldn't — `shap.DeepExplainer` runs through a PyTorch model without pulling any TF module into `sys.modules`. Greps confirm that every `import tensorflow` in `shap/explainers/**` and `shap/models/**` is already function-local, and `shap/explainers/_deep/__init__.py` lazy-dispatches to `deep_tf` vs `deep_pytorch`. So the demangling CloseChoice asked for already happened at some point between May 2024 and today.

Instead of a no-op fix, this PR adds a regression test so a future change can't silently re-entangle the deps:

- New test `tests/explainers/test_deep_torch_only.py` spawns a subprocess, sets `sys.modules["tensorflow"] = None` (same for `tf_keras`, `keras`) — which makes those imports raise `ImportError` even if the packages are actually installed in the test environment.
- In that subprocess it `import shap`, builds a tiny `nn.Linear` model, constructs `shap.DeepExplainer`, and calls `shap_values(X, check_additivity=False)`.
- After each step it asserts none of the blocked names leaked into `sys.modules`.

`check_additivity=False` is used because there's an unrelated additivity-tolerance assertion on simple torch models; that's not what this test is guarding against.

If maintainers prefer closing #3662 outright I'm happy with that too — I'd just like to leave this test in place either way so the property is codified.

## Local verification

- `ruff check` and `ruff format --check` clean.
- Confirmed the subprocess body passes against PyPI `shap==0.51.0` with only `torch` installed.
- Full `pytest` wasn't run locally (C extension requires CMake, which I don't have set up).

### DIFF
--- a/tests/explainers/test_deep_torch_only.py
+++ b/tests/explainers/test_deep_torch_only.py
@@ -1,0 +1,82 @@
+"""Regression guardrail for https://github.com/shap/shap/issues/3662.
+
+The PyTorch code path of ``shap.DeepExplainer`` must not pull in tensorflow /
+tf_keras / keras. This test runs the full PyTorch flow in a subprocess where
+those modules are blocked in ``sys.modules`` so that a later regression
+reintroducing an eager TF import would surface here.
+"""
+
+from __future__ import annotations
+
+import subprocess
+import sys
+import textwrap
+
+import pytest
+
+pytest.importorskip("torch")
+
+
+BLOCKED = ("tensorflow", "tf_keras", "keras")
+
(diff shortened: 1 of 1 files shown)


ashwinkothavade:
@CloseChoice 
Can you please review the PR

PR #4629, opened 2026-04-16
ENH: Remove unused typing-extensions dependency

samudraneel05:
## Overview

Fixes #4628
As discussed in #4623, `typing-extensions` appears to be an unused dependency. 

Description of the changes proposed in this pull request:

Went ahead and removed typing-extensions from requirement files wherever present, including `pyproject.toml`, `requirements.ci.txt` and removed the direct requirement trace in `docs/requirements-docs.txt`.

## Checklist

- [x] All [pre-commit checks](https://pre-commit.com/#install) pass.
- [x] Unit tests added (if fixing a bug or adding a new feature)

PR #3812, opened 2024-08-13
MAINT: Enable mypy on further modules

connortann:
## Overview

Further progress towards #3730
Closes #3793

Runs mypy on a few more modules. Adds a few type hints and `type: ignore` comments on false positives.

PR #4735, opened 2026-04-22
ENH: ship a PEP 561 py.typed marker

### LINKED ISSUE #3860: ENH: Declare that package supports typing with a py.typed marker
## Overview

We could add a py.typed marker to the library, so that downstream users can start to make use of type hints.

## Discussion

Reference: https://peps.python.org/pep-0561/

> Package maintainers who wish to support type checking of their code MUST add a marker file named py.typed to their package supporting typing.

It seems like a helpful idea; however, I note that many other major libraries do _not_ include this marker, which makes me a little hesitant. So, as a precaution I think we should try to find out from other maintainers if there are any downsides or reasons to avoid.

ashwinkothavade:
Closes #3860 (pending maintainer greenlight — see discussion on issue).

## Context

shap has a mature internal typing posture:

- `[tool.mypy]` configured in `pyproject.toml`
- type hints landed across `shap.models`, `shap.plots.scatter`, `shap.datasets` and the wider production surface (#3753, #3811, #3812, #3815, #4217)
- `typing-extensions` was cleaned up when confirmed unused (#4629)

Without a PEP 561 `py.typed` marker, downstream users running mypy get none of that — every `shap.*` symbol is treated as `Any`.

## Changes

1. Added empty `shap/py.typed` marker
2. Added `"shap/py.typed"` to `sdist.include` in `pyproject.toml` so the marker ships in both wheels and sdists under scikit-build-core's packaging rules

No code changes, no runtime behaviour changes. Strictly additive for downstream type-checker users; invisible to everyone else.

## Verification

Build artifacts need verifying by a maintainer (I don't have CMake set up locally to run `python -m build`). Expected: `unzip -l dist/shap-*.whl | grep py.typed` should show the file.

## Re maintainer hesitation in the original issue

The library-adoption landscape has shifted since Sept 2024 — numpy and xgboost ship it, many others (matplotlib, pandas) now do too. shap's typing work is already exercised via mypy in CI, so exposing it is lower-risk than shipping experimental stubs like the scipy-stubs case.

Happy to hold this open pending any maintainer consultation you'd like to run per @CloseChoice's earlier question in the issue thread.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -257,7 +257,7 @@ wheel.packages = [
 ]
 
 metadata.version.provider = "scikit_build_core.metadata.setuptools_scm"
-sdist.include = ["shap/_version.py"]
+sdist.include = ["shap/_version.py", "shap/py.typed"]
 sdist.exclude = ["docs/", "notebooks/"]
 
 # Protect the configuration against future changes in scikit-build-core


daidahao:
@ashwinkothavade 
Please stop spamming with AI agents. Focus on one or two PRs at a time and learn how shap works.

PR #4756, opened 2026-04-22
DOCS: Update Explaining_the_Loss_of_a_Model notebook to use current S…

Favourj-bit:
## Overview

Fixes the `Explaining_the_Loss_of_a_Model.ipynb` notebook as part of #3036.

### Bug fixes
- Fixed `feature_perturbation="independent"` which is no longer a valid option in the current SHAP version. Valid options are now `"auto"`, `"interventional"`, or `"tree_path_dependent"`. Updated to `"interventional"` which is consistent with the notebook's use of background data

### API updates
- Replaced `explainer.shap_values()` with `explainer()` to return an `Explanation` object
- Updated SHAP value summation to use `shap_values.values.sum(1)` to reflect the new `Explanation` object structure

### Documentation improvements
- Removed the placeholder line "This notebook will be fleshed out once we post a full write-up of this method"
- Updated the intro markdown to reflect that `feature_perturbation="interventional"` is the correct option (was incorrectly documented as `"independent"`)
- Added a markdown section and comment explaining the log-loss computation in Cell 3
- Added inline comment clarifying that the summation cell verifies SHAP values sum to the model loss

### New examples
- Added a beeswarm plot to visualize which features drive the model's log-loss across all explained samples

## Checklist


Favourj-bit:
A note for reviewers: 

`shap.plots.waterfall` and `shap.plots.bar` are currently incompatible 
with `model_output="log_loss"` as `base_values` is returned as a method rather than a scalar 
when using the new Explanation-based API. This may be worth addressing upstream.

Favourj-bit:
Please, I also have two questions:

1. The original notebook used feature_perturbation="independent" which is no longer a valid
option in the current SHAP version. I updated it to "interventional" as it seemed the closest
match, but wanted to confirm this is the right choice for the log-loss use case.

2. The original notebook had a placeholder: "This notebook will be fleshed out once we post a
full write-up of this method." Is there a write-up available now that we can link to or expand
from? 

I would be happy to flesh out the notebook further if there is more material to work with.

PR #4761, opened 2026-04-22
TYP: enable mypy check_untyped_defs for shap/maskers/_image

SarthakB11:
## Summary

- Adds type annotations to all functions and classes in `shap/maskers/_image.py`
- Removes `shap.maskers._image` from the `check_untyped_defs = false` override in `pyproject.toml`
- `mypy shap/maskers/_image.py` passes with zero new errors (only the pre-existing `numba.njit` attr-defined error shared with `_clustering.py` and `_masked_model.py`, which disappears in CI where numba is not installed)

## Changes

**`shap/maskers/_image.py`**
- `from __future__ import annotations` + `numpy.typing as npt` added
- Class-level variable declarations added so `_blur_value_cache` is typed as `npt.NDArray[Any] | None` rather than inferred as `None`
- `__init__` annotated; `int(np.prod(...))` wrap added (returns `np.intp`, base class expects `int`); targeted `# type: ignore` for `issubclass`-based guard (mypy cannot narrow on `issubclass`, changing to `isinstance` would be a behavior change) and for `tuple(map(int, ...))` returning variable-length tuple
- `assert self._blur_value_cache is not None` added before index access — logically safe (the `if self.last_xid != id(x)` block above always assigns it), narrows the type for mypy
- `__call__` annotated with `x: Any` (accepts `torch.Tensor` + ndarray), `mask: npt.NDArray[Any] | bool | None`, `# type: ignore[override]` since base class does not allow `None` mask
- `inpaint`, `build_partition_tree`, `save`, `load` annotated
- `_jit_build_partition_tree` annotated with `# type: ignore[misc]` on `@njit` (project convention, same as `_clustering.py`)

**`pyproject.toml`**
- `"shap.maskers._image"` commented out following the existing `# NOW TYPE CHECKED` convention

## Test

```bash
.venv/bin/mypy shap/maskers/_image.py --no-error-summary 2>&1 | grep "_image.py"
# shap/maskers/_image.py:10: error: Module "numba" has no attribute "njit"  [attr-defined]
# (pre-existing, shared with _clustering.py and _masked_model.py, gone in CI)
```

## AI Disclosure

Type annotations were drafted with AI assistance (Claude) and verified manually against the source code and mypy output.

Part of #3730.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -196,7 +196,7 @@ module = [
   "shap.explainers.tf_utils",
   # "shap.maskers._composite",  # NOW TYPE CHECKED - type hints added
   # "shap.maskers._fixed",      # NOW TYPE CHECKED - type hints added
-  "shap.maskers._image",
+  # "shap.maskers._image",  # NOW TYPE CHECKED - type hints added
   # "shap.maskers._masker",     # NOW TYPE CHECKED - type hints added
   "shap.maskers._tabular",
   "shap.maskers._text",
--- a/shap/maskers/_image.py
+++ b/shap/maskers/_image.py
@@ -1,8 +1,12 @@
 # TODO: heapq in numba does not yet support Typed Lists so we can move to them yet...
+from __future__ import annotations
+
 import heapq
+from typing import Any, BinaryIO
 
 import numba.typed
 import numpy as np
+import numpy.typing as npt
 from numba import njit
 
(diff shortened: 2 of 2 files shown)


SarthakB11:
@CloseChoice bumping this too, same #3730 series as #4445/#4760/#4772. CI green, no review yet.

PR #4766, opened 2026-04-23
FIX: assign return value of np.flipud() in AdditiveForceArrayVisualizer

### LINKED ISSUE #4765: Bug: `np.flipud()` result not assigned in stacked force plot ordering
## Summary

When rendering a stacked force plot via `shap.plots.force()` for multiple observations, the code intends to ensure that samples with higher predictions appear first (top of the plot). However, this logic is ineffective because the result of `np.flipud()` is not assigned back, leaving `clustOrder` unchanged.

---

## Steps to Reproduce
```python
import numpy as np
import shap

np.random.seed(0)
X_low  = np.random.randn(10, 5) - 2
X_high = np.random.randn(10, 5) + 2
X = np.vstack([X_low, X_high])

model = lambda x: x.sum(axis=1)
shap_values = explainer(X)

shap.plots.force(shap_values)
```

---

## Expected Behavior
Samples with higher predictions should consistently appear at the top of the stacked force plot.

---

## Actual Behavior
The ordering is not adjusted, and higher-prediction samples are not guaranteed to appear first.

---

## Root Cause
In `shap/plots/_force.py`:

```python
    np.flipud(clustOrder)  # result is not assigned
```

`np.flipud()` returns a new array and does not modify in-place, so `clustOrder` remains unchanged.

---

## Proposed Fix
```diff
- np.flipud(clustOrder)
+ clustOrder = np.flipud(clustOrder)
```

---

## Impact
- Affects only multi-sample (stacked) force plots  
- Does not raise errors, but leads to inconsistent visual ordering  
- Single-sample plots are unaffected  

---

## Additional Notes
A simple regression test verifying ordering consistency would help prevent this issue in the future.

maniktyagi04:
Closes #4765

## Summary
Fixes an issue in stacked force plots where sample ordering was not correctly applied.

---

## Problem
In `shap/plots/_force.py`, the code attempts to reverse the cluster order so that higher predictions appear first:

```python
    np.flipud(clustOrder)
```

However, `np.flipud()` returns a new array and does not modify in-place. Since the result was not assigned back, the ordering remained unchanged.

---

## Fix
```diff
- np.flipud(clustOrder)
+ clustOrder = np.flipud(clustOrder)
```

---

## Impact
- Affects only multi-sample stacked force plots
- Ensures higher prediction samples appear first as intended
- No impact on single-sample plots or other visualizations

### DIFF
--- a/shap/plots/_beeswarm.py
+++ b/shap/plots/_beeswarm.py
@@ -513,7 +513,7 @@ def shorten_text(text, length_limit):
 
 
 def is_color_map(color):
-    safe_isinstance(color, "matplotlib.colors.Colormap")
+    return safe_isinstance(color, "matplotlib.colors.Colormap")
 
 
 # TODO: remove unused title argument / use title argument
--- a/shap/plots/_force.py
+++ b/shap/plots/_force.py
@@ -541,7 +541,7 @@ class AdditiveForceArrayVisualizer(BaseVisualizer):
 
         # make sure that we put the higher predictions first...just for consistency
         if sum(arr[clustOrder[0]].effects) < sum(arr[clustOrder[-1]].effects):
-            np.flipud(clustOrder)  # reverse
+            clustOrder = np.flipud(clustOrder)  # reverse
 
         # build the json data
         clustOrder = np.argsort(clustOrder)  # inverse permutation


maniktyagi04:
@CloseChoice 
Just following up on PR #4765  that’s been open for a while.

Would appreciate a review whenever possible, or feedback if something needs improvement.

PR #4798, opened 2026-04-23
FIX: small fix to remove unreachable return statement.

amishhhaaaa:
## Overview

Small fix to remove unreachable return statement.

## Checklist


amishhaa:
cc: @CloseChoice

PR #4799, opened 2026-04-23
_tree: reduce time complexity for queue pop operation from O(n) to O(1).

amishhhaaaa:
## Overview

Fixes TODO.
From various benchmarks and articles we know that the theoretically time complexity
of queue.popleft() - O(1) is less that queue.pop(0) - O(n), we can effectively make an
improvement and reduce the time complexity from O(N^2) to O(N) for this BFS traversal.
references: https://wiki.python.org/moin/TimeComplexity

## Checklist


amishhaa:
cc: @CloseChoice

Issue #4358, opened 2026-04-04
Discussions on testcase quality vs coverage expectations.


naxatra2:
I’ve noticed some recent PRs that add a large number of small tests to increase coverage. While that improves the coverage metrics, I am concerned that it may adversely affect the code quality. 

For example: https://github.com/shap/shap/pull/4349 https://github.com/shap/shap/pull/4348 https://github.com/shap/shap/pull/4347

These 3 PRs single handedly add like 45+ testcases. I think we should some more emphasis on realistic test cases instead of just coverage %

tylerjereddy:
Seems fair to at least expect folks to respond to requests for revisions if tests are duplicated or of low quality (I saw some of your reviews). Some folks tend to prefer not to respond to AI-driven PRs at all, but I think Tobias is broadly in favor of helpful AI contributions.

If the test cases are feeling like "fuzzing," or property-based testing that isn't based on real-world scenarios (i.e., `hypothesis` type tests), I know a lot of projects also consider reviewing such PRs quite low priority.

tylerjereddy:
On other hand, for some tricky `shap` bugs I think I've had trouble reproducing them without using either a huge dataset or something tricky produced by `hypothesis`, so may need case-by-case consideration, as usual..

naxatra2:
I completely get your point.
However, if we allow a single PR to address dozens of small issues at once, especially if they are of lower quality. it creates a high barrier for other new contributors. It can be discouraging for the community when one person single handedly clears the issue tracker with many low effort tests. These test cases may not even get pushed to the repo, yet no other new contributor will try to take them up because there is already a PR addressing that issue.

Ultimately, it could be detrimental to new contributors and the overall maintainability of the project.

tylerjereddy:
That's true!

CloseChoice:
I am very much in favor of going through the `shap.<specific>Explainer` when building test cases, just adding tests for some more or less internal functions (even though they are not prefixed with `_`) does not really help IMO. But ad  @tylerjereddy there are some hard edge cases that are simply impractical to catch through this interface. In those cases I think using the shortcuts through specific functions is totally legit. 

But nailing which case applies is hard and actually quite a burden on maintainers especially with the recent influx of AI driven PRs.

tarun-227:
Thanks for the feedback — these are valid points. I've gone through #4347 and reworked the clustering tests based on the discussion here and @CloseChoice's guidance about preferring the `shap.<Explainer>` interface.

[...]

PR #4802, opened 2026-04-24
TEST: add public-API behavior tests in shap/explainers/_additive.py

SuMayaBee:
## Overview

Addresses #3690

While going through the coverage report for the warm-up task, I noticed `shap/explainers/_additive.py` was sitting at 37% but had no dedicated test file at all. The 37% was almost entirely imports, the class signature, and docstrings, the actual bodies were never exercised. So this PR adds the first dedicated test file for `shap.AdditiveExplainer`.

I read through the discussion in #4358 before writing anything and tried to keep the tests aligned with what was asked there. Everything goes through the public `shap.AdditiveExplainer` API and asserts on the returned `Explanation` object, none of the tests poke at internal helpers.

The file `test_additive.py` adds six tests and **Coverage goes from 37% to 84%**. 

**Screenshot**:
---
image

---
The remaining 7 lines (50-62, 116-118) are the EBM-specific path that raises `NotImplementedError` and needs the optional `interpret` library to reach. Happy to add tests for those if reviewers want them.

## AI Usage Disclosure

As per the SHAP AI Usage Policy:

- **What the AI did**: Helped me explore the file faster and generated the boilerplate for the test functions.
- **What I did**: Picked the target file myself after reviewing the coverage report and the #4358 discussion, decided which behaviors were worth pinning down, and ran the full test suite locally to confirm nothing regressed. I take full responsibility for every line in this PR.

## Checklist

- [x] All `pre-commit` checks pass.
- [x] Unit tests added (six new tests, no regressions in the full suite).

### DIFF
--- a/tests/explainers/test_additive.py
+++ b/tests/explainers/test_additive.py
@@ -1,11 +1,4 @@
-"""Behavior tests for ``shap.AdditiveExplainer`` going through the public API.
-
-These tests cover the explainer contract for purely additive models, the
-masker integration, the shape consistency required by downstream plot
-functions, and the explicit error path for unsupported masker types. They
-follow the guidance in #4358 to test through ``shap.<SpecificExplainer>``
-rather than internal helpers.
-"""
+"""Tests for shap.AdditiveExplainer."""
 
 import numpy as np
 import pytest
--- a/tests/explainers/test_additive.py
+++ b/tests/explainers/test_additive.py
@@ -1,0 +1,141 @@
+"""Behavior tests for ``shap.AdditiveExplainer`` going through the public API.
+
+These tests cover the explainer contract for purely additive models, the
+masker integration, the shape consistency required by downstream plot
+functions, and the explicit error path for unsupported masker types. They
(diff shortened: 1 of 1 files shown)


tests/explainers/test_additive.py
```diff
@@ -0,0 +1,141 @@
+"""Behavior tests for ``shap.AdditiveExplainer`` going through the public API.
```
amisha-chhajed: I think we do not have docstrings in any of the test files above import statements, IMHO we can include this information in commit messages and pr descriptions because when anyone extends this test file further to write tests for _additive apart from "Behavior tests for ``shap.AdditiveExplainer`` going through the public API." they would have to update this doc-string, what do you think?

SuMayaBee: Actually most files in tests/explainers/ do have module docstrings such as test_exact.py, test_linear.py, test_tree.py all have them, so I was following the existing pattern. But agreed the wording was too specific, simplified it.

PR #4806, opened 2026-04-24
FIX: scatter multi-feature show=False now returns Figure

### LINKED ISSUE #4805: BUG: scatter plot with multiple features returns None when show=False
## Issue Description

`shap.plots.scatter` is documented to return a `matplotlib.axes.Axes` object when `show=False`, allowing callers to further customize the plot. However, when multiple features are passed (causing subplots to be created), the function always returns `None` regardless of the `show` parameter.

This is inconsistent with the single-feature path, which correctly returns `ax` when `show=False`.

## Minimal Reproducible Example

```python
import numpy as np
import shap

# Create a small synthetic dataset with two features
X = np.random.randn(50, 2)
model_output = X[:, 0] + 0.5 * X[:, 1]
explainer = shap.Explainer(lambda x: x[:, 0] + 0.5 * x[:, 1], X)
sv = explainer(X)

# Single feature – works correctly, returns an Axes object
ax = shap.plots.scatter(sv[:, 0], show=False)
print(type(ax))   # <class 'matplotlib.axes._axes.Axes'>

# Multiple features – broken, returns None
result = shap.plots.scatter(sv, show=False)
print(result)     # None  ← should be a Figure object
```

## Traceback

No exception is raised; the return value is silently `None`, making the plot impossible to retrieve or further customise.

## Expected Behavior

When `show=False`, `shap.plots.scatter` should return the matplotlib `Figure` object for the multi-feature subplot case (analogous to returning `ax` in the single-feature case). This would allow callers to, for example, save the figure:

```python
fig = shap.plots.scatter(sv, show=False)
fig.savefig("dependence_plots.png", dpi=150)
```

## Bug Report Checklist

- [x] I have confirmed this bug exists on the master branch of shap.

## Installed Versions

shap 0.47.2

magic-peach:
## Overview

Closes #4805

`shap.plots.scatter` documents that it returns a `matplotlib.axes.Axes` when `show=False`, enabling callers to further customise or save the figure. The single-feature code path correctly returns `ax`, but the multi-feature subplot path always returned `None` due to a bare `return` statement after the `if show: plt.show()` block.

**Root cause** (`shap/plots/_scatter.py`, multi-feature branch):
```python
# before
_ = plt.subplots(...)   # figure reference discarded
...
if show:
    plt.show()
return              # always None
```

**Fix**: capture the figure from `plt.subplots` and return it when `show=False`:
```python
# after
fig, _ = plt.subplots(...)
...
if show:
    plt.show()
else:
    return fig      # callers can now save / further modify the figure
```

This makes the multi-feature path consistent with the single-feature path and with the documented return contract.

## Checklist

- [x] Unit tests added (regression test `test_scatter_multifeature_show_false_returns_figure` in `tests/plots/test_scatter.py`)

### DIFF
--- a/shap/plots/_scatter.py
+++ b/shap/plots/_scatter.py
@@ -140,7 +140,7 @@ def scatter(
         ymin = parse_axis_limit(ymin, shap_values.values, is_shap_axis=True)
         ymax = parse_axis_limit(ymax, shap_values.values, is_shap_axis=True)
         ymin, ymax = _suggest_buffered_limits(ymin, ymax, shap_values.values)
-        _ = plt.subplots(1, len(inds), figsize=(min(6 * len(inds), 15), 5))
+        fig, _ = plt.subplots(1, len(inds), figsize=(min(6 * len(inds), 15), 5))
         for i in inds:
             ax = plt.subplot(1, len(inds), i + 1)
             scatter(shap_values[:, i], color=color, show=False, ax=ax, ymin=ymin, ymax=ymax)
@@ -160,7 +160,8 @@ def scatter(
             plt.legend()
         if show:
             plt.show()
-        return
+        else:
+            return fig
 
     if len(shap_values.shape) != 1:
         raise DimensionError(
--- a/shap/plots/_scatter.py
+++ b/shap/plots/_scatter.py
@@ -120,8 +120,13 @@ def scatter(
     Returns
     -------
     ax : matplotlib Axes object
-        Returns the :external+mpl:class:`~matplotlib.axes.Axes` object with the plot drawn onto it. Only
-        returned if ``show=False``.
+        When a **single feature** is plotted and ``show=False``, returns the
+        :external+mpl:class:`~matplotlib.axes.Axes` object so the plot can be
+        further customised.
+    fig : matplotlib Figure object
+        When **multiple features** are plotted and ``show=False``, returns the
+        :external+mpl:class:`~matplotlib.figure.Figure` object containing all
+        subplots. Use ``fig.axes`` to access the individual Axes.
 
     Examples
     --------
--- a/tests/plots/test_scatter.py
+++ b/tests/plots/test_scatter.py
@@ -73,7 +73,6 @@ def test_scatter_multifeature_show_false_returns_figure(explainer):
     result = shap.plots.scatter(explanation[:, ["Age", "Workclass"]], show=False)
     assert result is not None, "scatter() returned None with show=False; expected a matplotlib Figure"
     assert hasattr(result, "savefig"), f"Expected a Figure object, got {type(result)}"
-    plt.close("all")
 
 
 @pytest.fixture()
--- a/tests/plots/test_scatter.py
+++ b/tests/plots/test_scatter.py
@@ -64,6 +64,18 @@ def test_scatter_custom(explainer):
     return plt.gcf()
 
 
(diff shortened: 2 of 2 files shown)


shap/plots/_scatter.py
```diff
@@ -140,7 +140,7 @@ def scatter(
         ymin = parse_axis_limit(ymin, shap_values.values, is_shap_axis=True)
         ymax = parse_axis_limit(ymax, shap_values.values, is_shap_axis=True)
         ymin, ymax = _suggest_buffered_limits(ymin, ymax, shap_values.values)
-        _ = plt.subplots(1, len(inds), figsize=(min(6 * len(inds), 15), 5))
+        fig, _ = plt.subplots(1, len(inds), figsize=(min(6 * len(inds), 15), 5))
```
amisha-chhajed, line 148: fig is a matplotlib.figure.Figure object and this method documents returning a matplotlib Axes object refer: https://github.com/shap/shap/blob/master/shap/plots/_scatter.py#L122
I think _, ax would give you that object which you can return, see, https://github.com/shap/shap/blob/master/shap/plots/_scatter.py#L246
However if returning fig is necessary in that case the documentation line needs to be updated.

tests/plots/test_scatter.py
```diff
@@ -64,6 +64,18 @@ def test_scatter_custom(explainer):
...
+def test_scatter_multifeature_show_false_returns_figure(explainer):
+    """scatter() with multiple features and show=False must return a Figure, not None.
+
+    Regression test for GH #4805.
+    """
+    explanation = explainer(explainer.data)
+    result = shap.plots.scatter(explanation[:, ["Age", "Workclass"]], show=False)
+    assert result is not None, "scatter() returned None with show=False; expected a matplotlib Figure"
+    assert hasattr(result, "savefig"), f"Expected a Figure object, got {type(result)}"
+    plt.close("all")
```
amisha-chhajed: you dont need to close, we handle cleanup in https://github.com/shap/shap/blob/master/tests/conftest.py#L79 for all test files.

amisha-chhajed (changes requested):

shap/plots/_scatter.py
magic-peach: Good catch. Since the multi-feature path creates several subplots, returning a single `Axes` isn't possible — the natural container is the `Figure`. I've updated the docstring to document both return paths explicitly: single-feature returns the `Axes`, multi-feature returns the `Figure` (with a note that `fig.axes` gives access to the individual `Axes` objects). Pushed in the latest commit.

tests/plots/test_scatter.py
magic-peach: Removed — thanks for the pointer to conftest.py. The latest commit drops the `plt.close('all')` call.

magic-peach:
Thanks for the review, @amishhaa!

**Return type (`shap/plots/_scatter.py`)** — Since the multi-feature path creates N subplots, returning a single `Axes` object isn't meaningful. The `Figure` is the natural container for the whole plot. I've updated the docstring to document both paths explicitly: single-feature returns `Axes` (unchanged), multi-feature returns `Figure` and notes that `fig.axes` gives access to the individual `Axes` objects.

**`plt.close("all")` in test** — Removed; didn't realise `conftest.py` already handles this for all plot tests.

Both changes are in the latest commit. Let me know if anything else needs adjusting!

magic-peach:
@amishhaa could you please review my pr

PR #4808, opened 2026-04-24
TST: increase coverage for shap/plots/_group_difference.py

keerthana-ar:
Towards #3690

Adds unit tests for shap/plots/_group_difference.py, increasing coverage from 80% to 100%. The file had an existing image comparison test covering 80% of lines; the new tests cover the remaining 20% to reach 100%.

New tests cover the previously missing lines:
- 1D input with no feature_names (lines 48-50)
- Auto-generated feature names (line 54)
- sort=False path (line 61)
- max_display limiting bars (line 64)
- Auto figure creation when no ax provided (lines 70-71)
- plt.show() called when show=True (line 92)

Note: AI tools were used to assist in writing this PR, as per the AI usage policy.

### DIFF
--- a/tests/plots/test_group_difference.py
+++ b/tests/plots/test_group_difference.py
@@ -62,9 +62,7 @@ def test_sort_false(shap_values_2d, group_mask):
     """Line 61: sort=False path."""
     feature_names = ["a", "b", "c", "d", "e"]
     ax = plt.subplots()[1]
-    shap.plots.group_difference(
-        shap_values_2d, group_mask, feature_names=feature_names, sort=False, show=False, ax=ax
-    )
+    shap.plots.group_difference(shap_values_2d, group_mask, feature_names=feature_names, sort=False, show=False, ax=ax)
     tick_labels = [t.get_text() for t in ax.get_yticklabels()]
     assert set(tick_labels) == set(feature_names)
 
--- a/tests/plots/test_group_difference.py
+++ b/tests/plots/test_group_difference.py
@@ -1,6 +1,9 @@
+import matplotlib
 import matplotlib.pyplot as plt
 import numpy as np
 import pytest
+
+matplotlib.use("Agg")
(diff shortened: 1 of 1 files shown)


tests/plots/test_group_difference.py
```diff
@@ -1,19 +1,88 @@
...
     group_mask = np.random.randint(2, size=shap_values.shape[0])
     feature_names = explainer.data_feature_names
     fig, ax = plt.subplots()
     shap.plots.group_difference(shap_values, group_mask, feature_names, show=False, ax=ax)
     plt.tight_layout()
     return fig
+
+
+@pytest.fixture(autouse=True)
+def close_plots():
```
amisha-chhajed: We dont need this, we handle the cleanup for all test files in https://github.com/shap/shap/blob/master/tests/conftest.py#L79.

tests/plots/test_group_difference.py
```diff
@@ -1,19 +1,88 @@
...
+def test_auto_feature_names(shap_values_2d, group_mask):
+    """Line 54: auto-generate Feature i names."""
+    ax = plt.subplots()[1]
+    shap.plots.group_difference(shap_values_2d, group_mask, feature_names=None, show=False, ax=ax)
+    tick_labels = [t.get_text() for t in ax.get_yticklabels()]
+    assert any(label.startswith("Feature") for label in tick_labels)
+
+
+def test_sort_false(shap_values_2d, group_mask):
+    """Line 61: sort=False path."""
```
amisha-chhajed: Line {number} is IMHO a bit brittle because when new code is merged into _group_difference the line numbers would change for all the code whose lines we are referencing like this.

keerthana-ar: Thanks for the feedback! I've removed the close_plots fixture and updated the docstrings to remove line number references.

PR #4823, opened 2026-04-25
fix: replace assert with raise ValueError/TypeError in _tree.py

### LINKED ISSUE #4820: BUG: assert used for input validation in _tree.py raises AssertionError instead of ValueError/TypeError
### Issue Description

`_tree.py` uses `assert` for user-facing input validation in multiple places. This causes two problems:

1. Users get `AssertionError` instead of `ValueError`/`TypeError`, which is unexpected and hard to catch
2. Running with `python -O` silently strips the checks, letting invalid input through

I verified the second point: passing `approximate=True` to CatBoost raises `AssertionError` normally, but with `python -O` it silently runs and returns results (see MRE below).

Here are the affected asserts I found:

**Unfitted model checks** (L1035, L1075, L1119): `assert hasattr(model, "estimators_")` on sklearn models. Reachable when calling `TreeEnsemble(model)` directly.

**NGBoost checks** (L1472, L1481): `assert model.base_models` and `assert safe_isinstance(...)` for unfitted/wrong learner type.

**Unsupported parameter combos** (L611, L633, L634, L781, L817): `approximate=True` with LightGBM/CatBoost, `tree_limit` with CatBoost, wrong `model_output` with interaction values.

**Internal consistency** (L1519, L1568, L1600): output dimension mismatch across trees, base_offset length, xgboost n_outputs.

**Input type/shape in _validate_inputs and predict** (L466, L467, L1687, L1688): these are mostly unreachable in normal flow because earlier code handles the conversion, but they'd be wrong exception type if hit.

I'm leaving the 5 asserts inside `XGBTreeModelLoader` (L2250, L2316, L2365, L2403, L2409) alone since those check internal parsed data, not user input.

```python
import numpy as np
from catboost import CatBoostRegressor
import shap

model = CatBoostRegressor(iterations=10, verbose=0, random_seed=0)
X = np.random.randn(100, 4)
model.fit(X, np.random.randn(100))

# Normal python: raises AssertionError (should be ValueError)
try:
    explainer.shap_values(X[:5], approximate=True)
except ValueError:
    print("caught ValueError")  # never runs
except AssertionError as e:
    print(f"got AssertionError: {e}")
# Output: got AssertionError: approximate=True is not supported for CatBoost models!

# python -O: assert is stripped, invalid input silently accepted
# Run: python -O this_script.py
# Output: no error, returns results silently


Second example hitting the unfitted model assert:


import numpy as np
from sklearn.ensemble import ExtraTreesRegressor
from shap.explainers._tree import TreeEnsemble

model = ExtraTreesRegressor()  # not fitted
try:
    te = TreeEnsemble(model)
except ValueError:
    print("caught ValueError")  # never runs
except AssertionError as e:
    print(f"got AssertionError: {e}")
# Output: got AssertionError: Model has no `estimators_`! Have you called `model.fit`?
```

### Traceback

```shell
# CatBoost approximate=True:
  File "test.py", line 9, in <module>
    explainer.shap_values(X[:5], approximate=True)
  File "/Users/ojchug/pc/ESoC/shap/shap/explainers/_tree.py", line 633, in shap_values
    assert not approximate, "approximate=True is not supported for CatBoost models!"
AssertionError: approximate=True is not supported for CatBoost models!

# With python -O: no error at all, assert is stripped
```

### Expected Behavior

These checks should use `raise ValueError` / `raise TypeError` so they:
- Can be caught with `except ValueError`
- Are not stripped by `python -O`

```python
# Instead of:
assert not approximate, "approximate=True is not supported for CatBoost models!"

# Should be:
if approximate:
    raise ValueError("approximate=True is not supported for CatBoost models!")
```

I'll open a PR for this fix.

### Bug report checklist

### Installed Versions

```
shap: 0.51.1.dev49 (master branch)
python: 3.12.12
numpy: 2.3.5
catboost: (installed, used for MRE)
```

ojuschugh1:
## Overview

Closes #4820

`_tree.py` uses `assert` for input validation in 17 places. This is a problem because:

1. Users get `AssertionError` instead of `ValueError`/`TypeError`, which they can't catch with `except ValueError`
2. `python -O` strips assert statements entirely, so the checks silently disappear

For example, passing `approximate=True` to a CatBoost model raises `AssertionError` normally but silently runs and returns results under `python -O`:

```python
# normal python: AssertionError
# python -O: no error, returns results
explainer.shap_values(X, approximate=True)
```

This PR replaces the 17 user-facing asserts with `raise ValueError` / `raise TypeError` / `raise AttributeError` as appropriate. Error messages are kept the same (two got slightly more detail added).

The 5 asserts inside `XGBTreeModelLoader` are left alone since those check internal parsed data structures, not user input.

**What changed:**

- Unfitted model checks : `assert hasattr(model, "estimators_")` -> `raise ValueError`
- Input validation in `_validate_inputs` and `predict` : `assert isinstance` -> `raise TypeError`, `assert len(X.shape)` -> `raise ValueError`
- Unsupported param combos  `assert not approximate` -> `raise ValueError`
- Internal consistency : output dimension mismatch, base_offset length, xgboost n_outputs

No behavior change in normal mode. Same error messages, just the right exception types.

## Checklist

### DIFF
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -463,8 +463,10 @@ class TreeExplainer(Explainer):
         if X.dtype != self.model.input_dtype:
             X = X.astype(self.model.input_dtype)
         X_missing = np.isnan(X, dtype=bool)
-        assert isinstance(X, np.ndarray), "Unknown instance type: " + str(type(X))
-        assert len(X.shape) == 2, "Passed input data matrix X must have 1 or 2 dimensions!"
+        if not isinstance(X, np.ndarray):
+            raise TypeError("Unknown instance type: " + str(type(X)))
+        if len(X.shape) != 2:
+            raise ValueError("Passed input data matrix X must have 1 or 2 dimensions!")
 
         if self.model.model_output == "log_loss":
             if y is None:
@@ -608,7 +610,8 @@ class TreeExplainer(Explainer):
                     )
 
             elif self.model.model_type == "lightgbm":
-                assert not approximate, "approximate=True is not supported for LightGBM models!"
+                if approximate:
+                    raise ValueError("approximate=True is not supported for LightGBM models!")
                 phi = self.model.original_model.predict(X, num_iteration=tree_limit, pred_contrib=True)
                 # Note: the data must be joined on the last axis
                 if (
@@ -630,8 +633,10 @@ class TreeExplainer(Explainer):
                         raise ValueError(emsg) from e
 
             elif self.model.model_type == "catboost":  # thanks to the CatBoost team for implementing this...
-                assert not approximate, "approximate=True is not supported for CatBoost models!"
-                assert tree_limit == -1, "tree_limit is not yet supported for CatBoost models!"
+                if approximate:
+                    raise ValueError("approximate=True is not supported for CatBoost models!")
+                if tree_limit != -1:
+                    raise ValueError("tree_limit is not yet supported for CatBoost models!")
                 import catboost
 
                 if not isinstance(X, catboost.Pool):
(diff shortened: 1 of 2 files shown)


ojuschugh1:
Hi @CloseChoice  could you kindly review it? Thanks!

Issue #4709, opened 2026-04-20
BUG: Explanation.__rsub__ computes self - other instead of other - self


SuMayaBee:
While computing residuals to measure how far each feature's SHAP contribution deviates from a reference value, I noticed that `value - explanation` returns the same result as `explanation - value`, which is wrong.

I was doing something like:

```python
baseline = 0.5
residuals = baseline - shap_values  # how far each SHAP value is from 0.5
```

But the result was identical to `shap_values - baseline`, which made no sense. 

### Minimal Reproducible Example

```python
!pip install shap -q

import numpy as np
import shap

exp = shap.Explanation(
    values=np.array([1.0, 2.0, 3.0]),
    base_values=0.0,
    data=np.array([0.1, 0.2, 0.3]),
    feature_names=["f0", "f1", "f2"]
)

result_sub  = (exp - 1.0).values   # exp.__sub__(1.0)  -> should be [0, 1, 2]
result_rsub = (1.0 - exp).values   # exp.__rsub__(1.0) -> should be [0, -1, -2]

print("exp - 1.0 :", result_sub)
print("1.0 - exp :", result_rsub)
print("Are they equal?", np.allclose(result_sub, result_rsub))  # True <- bug

expected = 1.0 - np.array([1.0, 2.0, 3.0])
print("Expected:", expected)    # [ 0. -1. -2.]
print("Got:     ", result_rsub) # [ 0.  1.  2.] <- wrong
```

## Screenshot (bug confirmed):

<img width="704" height="388" alt="Image" src="https://github.com/user-attachments/assets/ecdbec1a-f7a1-4e40-ac77-7c21d8a619b1" />

### Bug report checklist

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the latest release of shap.
- [x] I have confirmed this bug exists on the master branch of shap.
- [x] I'd be interested in making a PR to fix this bug

PR #4833, opened 2026-04-26
FIX: Implement missing __rtruediv__ operator for Explanation objects

### LINKED ISSUE #4785: BUG: Missing Explanation.__rtruediv__ causes TypeError for scalar / explanation operations
### Issue Description

The `shap.Explanation` class implements a full suite of arithmetic magic methods to support algebraic manipulation of SHAP values. It correctly implements right-hand operators for addition and subtraction (`__radd__`, `__rsub__`), but is missing `__rtruediv__`.

This means that while `shap_values / 2.0` works, the mathematically equivalent inverse operation `1.0 / shap_values` raises a `TypeError`, making certain normalization patterns impossible without workarounds and breaking algebraic symmetry with the existing right-hand operator support.

This is the same class of missing operator as identified in #4709 (`__rsub__`), where a right-hand operator was either absent or incorrectly implemented on the `Explanation` class.

```python
import numpy as np
import shap

exp = shap.Explanation(
    values=np.array([1.0, 2.0, 4.0]),
    base_values=0.0,
    data=np.array([0.1, 0.2, 0.3]),
    feature_names=["f0", "f1", "f2"]
)

# These all work correctly:
print(exp / 2.0)   # __truediv__  ✓
print(1.0 * exp)   # __rmul__     ✓
print(1.0 + exp)   # __radd__     ✓

# This crashes:
print(1.0 / exp)   # __rtruediv__ — missing
```

### Traceback

```shell
TypeError: unsupported operand type(s) for /: 'float' and 'Explanation'
```

### Expected Behavior

`1.0 / exp` should return a new `Explanation` object where each value in `exp.values` is used as the denominator:

```python
values: [1/1.0, 1/2.0, 1/4.0] = [1.0, 0.5, 0.25]
```

This is mathematically consistent with how `__rmul__` and `__radd__` already behave on the same class.

**Proposed fix** — add to `shap/_explanation.py`:

```python
def __rtruediv__(self, other: Explanation | npt.NDArray[Any] | float | int) -> Explanation:
    return self._apply_binary_operator(other, lambda a, b: b / a, "__rtruediv__")
```

Note: passing `operator.truediv` directly would repeat the same operand-order bug present in `__rsub__` before #4709, computing `self / other` instead of the correct `other / self`. The operands must be explicitly reversed.

### Bug report checklist

### Installed Versions

shap: current master
python: 3.11
numpy: 2.x

GobinathMaruthalingam:
Fixes #4785.

### Description
The `Explanation` class implements `__truediv__` but is missing `__rtruediv__`. This causes `1.0 / shap_values` to crash with a `TypeError`, breaking algebraic symmetry with the existing right-hand operator support (`__radd__`, `__rmul__`).

This PR adds `__rtruediv__` with explicitly reversed operands (`lambda a, b: b / a`) to prevent repeating the same operand-order bug seen in `__rsub__` before #4709.

### Tests added
- `float / explanation` — core behaviour
- `int / explanation` — non-float scalar edge case
- Operand-order regression guard confirming `other / self` is computed, not `self / other`

### Note on NumPy array behaviour
A test for `numpy_array / explanation` was considered but omitted. When the left operand is an ndarray, NumPy's own `__truediv__` takes precedence and `Explanation.__rtruediv__` is never invoked. This matches the existing behaviour across all other right-hand operators in the class (`__radd__`, `__rmul__`) and is not specific to this PR.

### Checklist
- [x] Unit tests added covering float and int divisors
- [x] Operand-order regression guard included
- [x] All existing tests pass locally

### DIFF
--- a/tests/test_explanation.py
+++ b/tests/test_explanation.py
@@ -245,3 +245,28 @@ def test_cohorts_generation_with_one_feature():
     cohorts = exp.cohorts(3)
     assert isinstance(cohorts, shap.Cohorts)
     assert len(cohorts.cohorts) == 3
+
+
+def test_explanation_rtruediv():
+    """Checks that __rtruediv__ correctly computes other / self, not self / other."""
+    import numpy as np
+    import shap
+
+    exp = shap.Explanation(values=np.array([1.0, 2.0, 4.0]))
+
+    # Test 1: float / explanation gives correct inverted values
+    result = 4.0 / exp
+    assert isinstance(result, shap.Explanation)
+    assert np.allclose(result.values, [4.0, 2.0, 1.0])
+
+    # Test 2: int / explanation (covers non-float scalars)
+    result_int = 8 / exp
+    assert isinstance(result_int, shap.Explanation)
+    assert np.allclose(result_int.values, [8.0, 4.0, 2.0])
+
+    # Test 3: operand-order regression guard — must NOT equal explanation / scalar
+    forward = (exp / 4.0).values
+    assert not np.allclose(result.values, forward), (
+        "__rtruediv__ returned same result as __truediv__ - operand order is wrong"
+    )
+
--- a/tests/test_explanation.py
+++ b/tests/test_explanation.py
@@ -249,10 +249,6 @@ def test_cohorts_generation_with_one_feature():
 
 def test_explanation_rtruediv():
     """Checks that __rtruediv__ correctly computes other / self, not self / other."""
-    import numpy as np
-
-    import shap
-
     exp = shap.Explanation(values=np.array([1.0, 2.0, 4.0]))
 
     # Test 1: float / explanation gives correct inverted values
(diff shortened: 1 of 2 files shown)


tests/test_explanation.py
```diff
@@ -245,3 +245,28 @@ def test_cohorts_generation_with_one_feature():
     cohorts = exp.cohorts(3)
     assert isinstance(cohorts, shap.Cohorts)
     assert len(cohorts.cohorts) == 3
+
+
+def test_explanation_rtruediv():
+    """Checks that __rtruediv__ correctly computes other / self, not self / other."""
+    import numpy as np
+
+    import shap
```
SuMayaBee: We don't need those imports, numpy and shap are already imported at the top of the file, these imports inside the function body are redundant.

GobinathMaruthalingam: Oops, didn't see that
Removed in **bb2cef96**, dropped the redundant imports

PR #4835, opened 2026-04-26
ENH: Implement support specific python SPEC 1

### LINKED ISSUE #4076: ENH: Support scientific python SPEC 1
### Problem Description

Since we had issues with import times previously and the scientific python packages move into this direction adoption of [SPEC 1](https://scientific-python.org/specs/spec-0001/) would be great. 

Note that for our tests we should use the EAGER_IMPORT mode to find import errors. Also note that we need to take care of mypy and code completion as outlined in https://scientific-python.org/specs/spec-0001/#type-checkers

### Alternative Solutions

_No response_

### Additional Context

_No response_

### Feature request checklist

justin212407:
## Overview

Closes #4076   

## Description
This PR implements lazy loading for the SHAP package following the [Scientific Python SPEC-0001](https://scientific-python.org/specs/spec-0001/) specification. This optimization reduces import time and memory footprint by deferring the import of expensive submodules (e.g., `shap.explainers._tree`) until they are actually needed.

### 1. Core Import Infrastructure
- Migrated package imports to use lazy_loader for dynamic module loading.
- Introduced stub files as the single source of truth for the public API.
- Consolidated all public exports into stub definitions, which are parsed at runtime by the lazy loading mechanism.

### 2. Lazy Loading Throughout Package
- **Explainers**: Introduced lazy loading for explainer modules and added backwards-compatible short-form aliases (e.g. Kernel, Tree).
- **Plots**: Introduced lazy loading for plotting modules and added legacy aliases for existing plot imports (e.g. bar_plot, decision_plot).
- **Utils**: Added kmeans export to public API 
- **Actions**: Added ActionOptimizer export to public API 

### 3. Deferred dependency loading in core logic
- Deferred heavy dependencies (pandas, scipy, sklearn) until runtime when actually needed.
- Moved heavyweight initialization paths to lazy initialization.
- Shifted imports to point-of-use within methods.

## Testing
Added comprehensive test suite `test_lazy_imports.py`  with:
- Validation that heavy modules are not eagerly imported on import shap
- Verification that `EAGER_IMPORT=1` env var forces immediate resolution (for CI early error detection)
- Assertion that all public API names remain accessible

Run lazy import tests with:
```bash
pytest tests/test_lazy_imports.py -v
```

Or force eager import in CI:

```bash
EAGER_IMPORT=1 pytest tests/
```

## Screenshots:
Screenshot From 2026-04-25 15-45-02

Screenshot From 2026-04-25 15-45-43

Screenshot From 2026-04-25 15-46-04

Screenshot From 2026-04-25 15-47-23

## Checklist


justin212407:
@CloseChoice Do you think the approach for the implementation is what was desired from the relevant issue?

shap/_explanation.py
Karan-Palan, line 1: `Cohorts` and `Explanation` are still being loaded eagerly at the package boundary. That means a simple `import shap` still pulls in the entire explanation stack.

It’d be better to make these imports lazy as well, and keep the scope focused

shap/plots/__init__.pyi
```diff
@@ -0,0 +1,66 @@
+# ruff: noqa: F401
```
Karan-Palan, line 1: Might not be required here

PR #4853, opened 2026-04-27
fix: handle tuple outputs in DeepExplainer for models with recurrent layers

### LINKED ISSUE #3773: How can explainer analyze just several specific layers or sub-models we chooosed?
For example, I have a model doing inference like:

    def forward(
            self,
            s,
            select_mask,
            state = None,
        ):
            s,targets=self.selector(s,select_mask)  #len(s.shape)
            logits, h = self.preprocess(s, state)
            logits = self.last(logits)
            if self.softmax_output:
                logits = F.softmax(logits, dim=-1)
            return logits, h, targets, s

So my model has 3 sub-models (selector, preprocess, last), where selector and last is MLP and preprocess is LSTM. Now I want to analyze just preprocess and last, so I call:

     shap.DeepExplainer(torch.nn.Sequential(model.preprocess,model.last), s)

However, the preprocess has two outputs and only logits will be input to last, so it got an error here.

How can I make it work? Or can I customize anohter function to replace forward when I call explainer?

Mahaveerjain-18:
## Summary

Fixes #3773

`DeepExplainer` crashes when models return **tuple outputs** (e.g., LSTM returning `(output, hidden_state)`). This PR adds proper handling for such models.

## Problem

When a user tries to explain sub-models composed via `nn.Sequential`, and an intermediate sub-module returns a tuple (common with LSTM/GRU/RNN layers), `DeepExplainer` crashes in multiple places:

1. **`__init__`** — `outputs.device` / `outputs.shape[1]` fails on tuples (`AttributeError`)
2. **`gradient()`** — `outputs[:, idx]` fails on tuples
3. **`nn.Sequential`** — passes the full tuple between sub-modules, causing `TypeError` in the next layer

## Changes

| Change | Description |
|--------|-------------|
| `_extract_tensor_output()` | Helper to safely unwrap tuple outputs by selecting the first tensor |
| `_model_contains_recurrent_layers()` | Detects LSTM/GRU/RNN in model tree to warn users and adjust behavior |
| `_wrap_sequential_with_tuple_handling()` | Auto-wraps `nn.Sequential` models to handle tuple outputs between sub-modules |
| `__init__` / `gradient()` / `shap_values()` | Use `_extract_tensor_output` at all model output access points |
| `op_handler` | Added LSTM, GRU, RNN as linear operations |
| Additivity check | Skipped for models with recurrent layers (fused kernels cannot be fully decomposed by DeepLIFT) |

## Test Results

| Scenario | Before | After |
|----------|--------|-------|
| `nn.Sequential(preprocess, last)` with LSTM | ❌ `TypeError` | ✅ Pass |
| Model returning tuple directly | ❌ `AttributeError` | ✅ Pass |
| Wrapper model workaround | ❌ Additivity assertion | ✅ Pass |
| Normal single-tensor model (regression) | ✅ | ✅ Pass |

## Reproduction

```python
import torch
import torch.nn as nn
import shap

class Preprocess(nn.Module):
    def __init__(self):
        super().__init__()
        self.lstm = nn.LSTM(10, 16, batch_first=True)
    def forward(self, s):
        s = s.unsqueeze(1)
        logits, h = self.lstm(s)
        return logits.squeeze(1), h  # tuple output

class Last(nn.Module):
    def __init__(self):
        super().__init__()
        self.fc = nn.Linear(16, 3)
    def forward(self, x):
        return self.fc(x)

# This used to crash, now works:
model = nn.Sequential(Preprocess(), Last())
background = torch.randn(5, 10)
explainer = shap.DeepExplainer(model, background)
shap_values = explainer.shap_values(torch.randn(2, 10))
```

### DIFF
--- a/shap/explainers/_deep/deep_pytorch.py
+++ b/shap/explainers/_deep/deep_pytorch.py
@@ -7,6 +7,103 @@ from .._explainer import Explainer
 from .deep_utils import _check_additivity
 
 
+def _extract_tensor_output(outputs):
+    """Extract a single tensor from model outputs.
+
+    Some models (e.g., those containing LSTM or GRU layers) return tuples
+    of tensors instead of a single tensor. This helper extracts the primary
+    output tensor (the first element) from such tuple outputs.
+
+    Parameters
+    ----------
+    outputs : torch.Tensor or tuple
+        The raw output from a model's forward pass.
+
+    Returns
+    -------
+    torch.Tensor
+        The primary output tensor.
+    """
+    if isinstance(outputs, tuple):
+        return outputs[0]
+    return outputs
+
+
+_RECURRENT_MODULE_TYPES = {"LSTM", "GRU", "RNN"}
+
+
+def _model_contains_recurrent_layers(model):
+    """Check if a model contains any recurrent layers (LSTM, GRU, RNN).
+
+    These layers have internal gate operations (sigmoid, tanh, element-wise
+    multiplications) that are fused into optimized kernels and cannot be
+    decomposed by DeepLIFT's module-level hooks.
+
+    Parameters
+    ----------
+    model : torch.nn.Module
+        The model to check.
+
+    Returns
+    -------
+    bool
+        True if the model contains any recurrent layers.
+    """
+    for module in model.modules():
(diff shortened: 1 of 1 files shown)


Mahaveerjain-18:
@CloseChoice 
Will you please review it

PR #4867, opened 2026-04-27
Add edge-case and behavioral tests for waterfall plot

Sonika-19:
## Overview

Closes #3690 (partial)

This PR adds edge-case and behavioral tests for the waterfall plot to improve test coverage and robustness across different input scenarios.

### Key additions:

- Verified that `show=False` correctly returns the current matplotlib axis
- Added test for unsupported empty SHAP values
- Tested fallback behavior when `feature_names=None`
- Added support tests for pandas Series inputs
- Verified behavior for all-positive and all-negative SHAP values
- Added test for `max_display` parameter behavior
- Added validation for invalid (non-Explanation) inputs

These tests focus on functional behavior and avoid brittle assertions tied to matplotlib internals.  
They are implemented in a separate test module to complement existing mpl-based visual tests without modifying them.

---

## Checklist

### DIFF
--- a/tests/plots/test_waterfall_edge_cases.py
+++ b/tests/plots/test_waterfall_edge_cases.py
@@ -1,21 +1,11 @@
 """Edge-case and behavioral tests for the waterfall plot."""
 
-import matplotlib
-matplotlib.use("Agg")
-
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
 import pytest
 
 import shap
-
-
-@pytest.fixture(autouse=True)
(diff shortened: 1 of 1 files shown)


.gitignore
SuMayaBee, line 1: The .gitignore shows up as a modified binary file in the diff, clearly unintentional, should not be in this PR.

Sonika-19: Thanks for pointing that out! I've reverted the unintended .gitignore change so the PR now only includes the test updates.

Issue #4420, opened 2026-04-09
bug: is_color_map() always returns None


maniktyagi04:
## Description

The function `is_color_map(color)` in `shap/plots/_beeswarm.py` calls `safe_isinstance(...)` but does not return its result. As a result, it always returns `None` instead of a boolean.

## Location

`shap/plots/_beeswarm.py` — `is_color_map` (around line 515)

## Affected Code

    def is_color_map(color):
        safe_isinstance(color, "matplotlib.colors.Colormap")  # return value discarded

## Expected Behavior

Returns `True` or `False` depending on whether `color` is a `matplotlib.colors.Colormap`.

## Actual Behavior

Always returns `None`.

PR #4893, opened 2026-04-28
FIX: add missing return statement to is_color_map in beeswarm

### LINKED ISSUE #4892: [BUG] is_color_map in beeswarm always returns None due to missing return statement
## Problem
`is_color_map` in `shap/plots/_beeswarm.py` calls `safe_isinstance` 
but discards the result — it always returns `None` regardless of input, 
making it completely broken for all callers.

## Reproduction
```python
from shap.plots._beeswarm import is_color_map
cmap = plt.get_cmap("viridis")
print(is_color_map(cmap))  # None — should be True
print(is_color_map("blue"))  # None — should be False
```

## Expected behavior
Returns `True` for colormaps, `False` otherwise.

## Fix
I have a fix ready and will open a PR immediately.

Sandrakimiring:
#### Reference Issues/PRs
Fixes #4892

#### What does this implement/fix? Explain your changes.
`is_color_map` in `shap/plots/_beeswarm.py` called `safe_isinstance` 
but discarded the return value, so the function always returned `None` 
regardless of input — making it completely broken for all callers.

Fixed by adding `return` before the `safe_isinstance` call.

#### Does your contribution introduce a new dependency? If yes, which one?
No new dependencies.

#### What should a reviewer concentrate their feedback on?
The one-line fix in `shap/plots/_beeswarm.py` and the 6 new tests 
in `tests/plots/test_beeswarm.py` — especially 
`test_is_color_map_return_type_is_bool` which would have failed 
against the old code since `None` is not a bool.

#### PR checklist
- [x] I've added unit tests and made sure they pass locally.

### DIFF
--- a/shap/plots/_beeswarm.py
+++ b/shap/plots/_beeswarm.py
@@ -513,7 +513,7 @@ def shorten_text(text, length_limit):
 
 
 def is_color_map(color):
-    safe_isinstance(color, "matplotlib.colors.Colormap")
+    return safe_isinstance(color, "matplotlib.colors.Colormap")
 
 
 # TODO: remove unused title argument / use title argument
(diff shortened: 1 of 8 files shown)


Sandrakimiring:
Closing as duplicate of #4420 which has already been resolved.

PR #4909, opened 2026-04-29
TEST: add mismatched feature_names handling test

### LINKED ISSUE #4826: TEST: Add test for mismatched feature_names handling in Explanation
### Problem Description

While working with the Explanation class, I noticed that it allows flexible usage of feature_names, even when the length does not match the number of features.

Instead of enforcing strict validation (which may break existing behavior), this issue proposes adding a test to ensure that mismatched feature_names inputs do not cause crashes or unexpected failures.

## Proposed change
- Add a test case where feature_names length does not match values shape
- Ensure the Explanation object is still created successfully

## Why this is useful
- Improves test coverage
- Ensures robustness for edge cases
- Aligns with the current flexible design of Explanation

Happy to work on a PR for this 🙂

### Alternative Solutions

_No response_

### Additional Context

_No response_

### Feature request checklist

onkar717:
## Summary

Add a regression test for mismatched `feature_names` handling in `Explanation`.

This PR adds test coverage for the current flexible behavior where `Explanation` accepts `feature_names` whose length does not match the shape of `values`, and still creates the object successfully.

Fixes #4826

## Changes Made

* Added `test_mismatched_feature_names_does_not_crash()` to `tests/test_explanation.py`
* Verifies that `Explanation(...)` initializes successfully when:

  * `values.shape == (10, 4)`
  * `feature_names` contains 3 entries
* Confirms `feature_names` is preserved as provided

## Why This Matters

The current implementation allows non-matching `feature_names` lengths instead of enforcing strict validation. This test documents and protects that behavior.

Benefits:

* Improves test coverage
* Prevents accidental regressions
* Ensures edge-case robustness
* Aligns tests with existing API behavior

## Testing

Verified that the following works without errors:

```python id="a1m9xq"
values = np.ones((10, 4))
names = ["a", "b", "c"]

exp = shap.Explanation(values=values, feature_names=names)
```

Assertions:

* `isinstance(exp, shap.Explanation)`
* `exp.feature_names == names`

## Notes

This PR adds tests only. No production code behavior was changed.

### DIFF
--- a/tests/test_explanation.py
+++ b/tests/test_explanation.py
@@ -245,3 +245,18 @@ def test_cohorts_generation_with_one_feature():
     cohorts = exp.cohorts(3)
     assert isinstance(cohorts, shap.Cohorts)
     assert len(cohorts.cohorts) == 3
+
+
+def test_mismatched_feature_names_does_not_crash():
+    """GH #4826: Explanation allows mismatched feature_names without crashing.
+
+    When feature_names length does not match any axis of values, the object
+    should still be created successfully with feature_names stored as provided.
+    """
+    values = np.ones((10, 4))
+    mismatched_names = ["a", "b", "c"]  # 3 names for 4 features
+
+    exp = shap.Explanation(values=values, feature_names=mismatched_names)
+
+    assert isinstance(exp, shap.Explanation)
+    assert exp.feature_names == mismatched_names


onkar717:
cc : @Sowmyadevalla2005

tests/test_explanation.py
```diff
@@ -245,3 +245,18 @@ def test_cohorts_generation_with_one_feature():
...
+
+
+def test_mismatched_feature_names_does_not_crash():
+    """GH #4826: Explanation allows mismatched feature_names without crashing.
+
+    When feature_names length does not match any axis of values, the object
+    should still be created successfully with feature_names stored as provided.
+    """
+    values = np.ones((10, 4))
+    mismatched_names = ["a", "b", "c"]  # 3 names for 4 features
```
tylerjereddy, line 257: Some context may be helpful--for example, do things also "work" if there are more names than features? May want to parametrize both of those cases if the team decides this is desirable to support.

Sowmyadevalla2005 (commented):
Thanks for the suggestion.
That makes sense, testing the case where feature_names has more entries than features would be useful.
Parametrizing both cases (fewer and more names) would cover both mismatch scenarios.

Happy to help extend this if needed.

Issue #3793, opened 2024-08-12
Plot overhaul: beeswarm


connortann:
This issue is part of the [Visualisation overhaul](https://github.com/orgs/shap/projects/1) project.

The things we wish to overhaul are detailed in #3411, and the current status is tracked in the project fields (see sidebar).

### Related

Handing fig and ax:
- #3561
- #3573

Multiclass:
- #3756

Regarding "sum of other values":
- #2225
- #1984

PR #4937, opened 2026-04-30
Standardize matplotlib Axes handling across beeswarm, scatter, and bar plots

### LINKED ISSUE #3411: [Meta-issue] Improve plotting API for consistency

gouravgangwardev:
## Overview

Closes #3411 

### Summary
This PR adds support for the `ax` parameter to `beeswarm`, `scatter`, and `bar` plots, aligning them with modern matplotlib usage patterns and ensuring consistent behavior across SHAP plotting APIs.

These changes remove implicit reliance on global pyplot state and enable proper integration with subplot-based workflows.

### Key Changes
- Added `ax` parameter to:
  - `shap.plots.beeswarm`
  - `shap.plots.scatter`
  - `shap.plots.bar`
- Updated internal matplotlib rendering to respect externally provided axes
- Standardized return behavior across all three plots:
  - `show=False` → returns `matplotlib.axes.Axes`
  - `show=True` → returns `None`
- Ensured figure resizing only occurs when `ax` is not provided
- Preserved backward compatibility for existing usage patterns

### Behavioral Guarantees
Across all three plots:

- Axis identity is preserved (`result is ax` when `ax` is provided)
- No unintended modification of sibling subplots
- Legends and colorbars attach only to the target axes
- No unexpected axes are created (except valid inset axes such as colorbars)
- Pyplot global state is not improperly altered

### Tests Added / Updated
For each plot (`beeswarm`, `scatter`, `bar`):

- Return contract validation:
  - `show=False` returns Axes
  - `show=True` returns None
- Axis identity preservation
- Subplot isolation (no leakage to sibling axes)
- Figure resizing behavior:
  - Resizing occurs only when `ax` is not provided
- Multiple calls on the same axes remain stable
- Legend/colorbar placement correctness (where applicable)
- Default behavior when `ax=None` uses `plt.gca()`
- Removal of ineffective assertions and strengthening of test conditions

Visual regression tests were retained and validated where applicable.

### Why This Matters
These changes enable:

- Seamless integration with matplotlib subplot workflows
- Improved composability for multi-plot dashboards and reports
- Elimination of implicit global state reliance
- Consistent developer experience across SHAP plotting functions

### Notes for Reviewers
- Follows the established `ax` handling pattern used in other SHAP plots
- Emphasizes strict subplot isolation and predictable matplotlib behavior
- Backward compatibility verified through existing and new tests
- Focused on minimal API surface change while improving correctness and flexibility

## Checklist

### DIFF
--- a/shap/plots/_bar.py
+++ b/shap/plots/_bar.py
@@ -240,11 +240,11 @@ def bar(
     if num_features < len(values[0]):
         yticklabels[-1] = f"Sum of {num_cut} other features"
 
-    if ax is None:
-        ax = plt.gca()
-        # Only modify the figure size if ax was not passed in
+    
+    _ax_provided = ax is not None
+    if not _ax_provided:
+        fig, ax = plt.subplots()
         # compute our figure size based on how many features we are showing
-        fig = plt.gcf()
         row_height = 0.5
         fig.set_size_inches(8, num_features * row_height * np.sqrt(len(values)) + 1.5)
 
@@ -339,9 +339,6 @@ def bar(
     else:
         ax.set_xlim(xmin, xmax + x_buffer)
 
-    # if features is None:
-    #     plt.xlabel(labels["GLOBAL_VALUE"], fontsize=13)
-    # else:
     ax.set_xlabel(xlabel, fontsize=13)
 
     if len(values) > 1:
@@ -390,6 +387,7 @@ def bar(
 
     if show:
         plt.show()
+        return None
     else:
         return ax
 
--- a/shap/plots/_bar.py
+++ b/shap/plots/_bar.py
@@ -240,7 +240,6 @@ def bar(
     if num_features < len(values[0]):
         yticklabels[-1] = f"Sum of {num_cut} other features"
 
-    
     _ax_provided = ax is not None
     if not _ax_provided:
         fig, ax = plt.subplots()
--- a/shap/plots/_bar.py
+++ b/shap/plots/_bar.py
@@ -239,13 +239,13 @@ def bar(
             yticklabels.append(feature_names[i])
     if num_features < len(values[0]):
         yticklabels[-1] = f"Sum of {num_cut} other features"
+    
+    if ax is None:
+    fig, ax = plt.subplots()
 
-    _ax_provided = ax is not None
-    if not _ax_provided:
-        fig, ax = plt.subplots()
-        # compute our figure size based on how many features we are showing
-        row_height = 0.5
-        fig.set_size_inches(8, num_features * row_height * np.sqrt(len(values)) + 1.5)
+    # Only modify the figure size if we created the figure.
+    row_height = 0.5
+    fig.set_size_inches(8, num_features * row_height * np.sqrt(len(values)) + 1.5)
 
     # if negative values are present then we draw a vertical line to mark 0, otherwise the axis does this for us...
     negative_values_present = np.sum(values[:, feature_order[:num_features]] < 0) > 0
--- a/shap/plots/_bar.py
+++ b/shap/plots/_bar.py
@@ -239,7 +239,7 @@ def bar(
             yticklabels.append(feature_names[i])
     if num_features < len(values[0]):
         yticklabels[-1] = f"Sum of {num_cut} other features"
-    
+
     if ax is None:
     fig, ax = plt.subplots()
 
--- a/shap/plots/_bar.py
+++ b/shap/plots/_bar.py
@@ -241,11 +241,9 @@ def bar(
         yticklabels[-1] = f"Sum of {num_cut} other features"
 
     if ax is None:
-    fig, ax = plt.subplots()
-
-    # Only modify the figure size if we created the figure.
-    row_height = 0.5
-    fig.set_size_inches(8, num_features * row_height * np.sqrt(len(values)) + 1.5)
+        fig, ax = plt.subplots()
+        row_height = 0.5
+        fig.set_size_inches(8, num_features * row_height * np.sqrt(len(values)) + 1.5)
 
     # if negative values are present then we draw a vertical line to mark 0, otherwise the axis does this for us...
     negative_values_present = np.sum(values[:, feature_order[:num_features]] < 0) > 0
--- a/shap/plots/_beeswarm.py
+++ b/shap/plots/_beeswarm.py
@@ -123,7 +123,7 @@ def beeswarm(
         )
         raise ValueError(emsg)
 
-    if ax and plot_size:
+    if ax is not None and plot_size is not None and plot_size != "auto":
         emsg = (
             "The beeswarm plot does not support passing an axis and adjusting the plot size. "
             "To adjust the size of the plot, set plot_size to None and adjust the size on the original figure the axes was part of"
@@ -193,8 +193,10 @@ def beeswarm(
     if feature_names is None:
         feature_names = np.array([labels["FEATURE"] % str(i) for i in range(num_features)])
 
-    if ax is None:
+    _ax_provided = ax is not None
+    if not _ax_provided:
         ax = plt.gca()
+    plt.sca(ax)
     fig = ax.get_figure()
     assert isinstance(fig, Figure)  # type narrowing for mypy
 
@@ -366,12 +368,13 @@ def beeswarm(
         yticklabels[-1] = f"Sum of {num_cut} other features"
 
     row_height = 0.4
-    if plot_size == "auto":
-        fig.set_size_inches(8, min(len(feature_order), max_display) * row_height + 1.5)
-    elif isinstance(plot_size, (list, tuple)):
-        fig.set_size_inches(plot_size[0], plot_size[1])
-    elif plot_size is not None:
-        fig.set_size_inches(8, min(len(feature_order), max_display) * plot_size + 1.5)
+    if not _ax_provided:
+        if plot_size == "auto":
+            fig.set_size_inches(8, min(len(feature_order), max_display) * row_height + 1.5)
+        elif isinstance(plot_size, (list, tuple)):
+            fig.set_size_inches(plot_size[0], plot_size[1])
+        elif plot_size is not None:
+            fig.set_size_inches(8, min(len(feature_order), max_display) * plot_size + 1.5)
     ax.axvline(x=0, color="#999999", zorder=-1)
 
     # make the beeswarm dots
@@ -513,7 +516,7 @@ def shorten_text(text, length_limit):
 
 
 def is_color_map(color):
-    return safe_isinstance(color, "matplotlib.colors.Colormap")
+    safe_isinstance(color, "matplotlib.colors.Colormap")
 
 
 # TODO: remove unused title argument / use title argument
@@ -781,6 +784,7 @@ def summary_legacy(
         plt.gcf().set_size_inches(plot_size[0], plot_size[1])
     elif plot_size is not None:
         plt.gcf().set_size_inches(8, len(feature_order) * plot_size + 1.5)
+    _legacy_ax = plt.gca()
     plt.axvline(x=0, color="#999999", zorder=-1)
 
     if plot_type == "dot":
@@ -1122,7 +1126,7 @@ def summary_legacy(
 
         m = cm.ScalarMappable(cmap=cmap if plot_type != "layered_violin" else plt.get_cmap(color))
         m.set_array([0, 1])
-        cb = plt.colorbar(m, ax=plt.gca(), ticks=[0, 1], aspect=80)
+        cb = plt.colorbar(m, ax=_legacy_ax, ticks=[0, 1], aspect=80)
         cb.set_ticklabels([labels["FEATURE_VALUE_LOW"], labels["FEATURE_VALUE_HIGH"]])
         cb.set_label(color_bar_label, size=12, labelpad=0)
         cb.ax.tick_params(labelsize=11, length=0)
--- a/shap/plots/_scatter.py
+++ b/shap/plots/_scatter.py
@@ -140,27 +140,32 @@ def scatter(
         ymin = parse_axis_limit(ymin, shap_values.values, is_shap_axis=True)
         ymax = parse_axis_limit(ymax, shap_values.values, is_shap_axis=True)
         ymin, ymax = _suggest_buffered_limits(ymin, ymax, shap_values.values)
-        _ = plt.subplots(1, len(inds), figsize=(min(6 * len(inds), 15), 5))
-        for i in inds:
-            ax = plt.subplot(1, len(inds), i + 1)
-            scatter(shap_values[:, i], color=color, show=False, ax=ax, ymin=ymin, ymax=ymax)
+        _, axes = plt.subplots(1, len(inds), figsize=(min(6 * len(inds), 15), 5))
+        if len(inds) == 1:
+            axes = [axes]
+        last_ax = None
+        for idx, i in enumerate(inds):
+            cur_ax = axes[idx]
+            scatter(shap_values[:, i], color=color, show=False, ax=cur_ax, ymin=ymin, ymax=ymax)
             if overlay is not None:
                 line_styles = ["solid", "dotted", "dashed"]
                 for j, name in enumerate(overlay):
                     vals = overlay[name]
                     if isinstance(vals[i][0][0], (float, int)):
-                        plt.plot(vals[i][0], vals[i][1], color="#000000", linestyle=line_styles[j], label=name)
-            if i == 0:
-                ax.set_ylabel(ylabel)
+                        cur_ax.plot(vals[i][0], vals[i][1], color="#000000", linestyle=line_styles[j], label=name)
+            if idx == 0:
+                cur_ax.set_ylabel(ylabel)
             else:
-                ax.set_ylabel("")
-                ax.set_yticks([])
-                ax.spines["left"].set_visible(False)
+                cur_ax.set_ylabel("")
+                cur_ax.set_yticks([])
+                cur_ax.spines["left"].set_visible(False)
+            last_ax = cur_ax
         if overlay is not None:
-            plt.legend()
+            last_ax.legend()
         if show:
             plt.show()
-        return
+            return None
+        return last_ax
 
     if len(shap_values.shape) != 1:
         raise DimensionError(
@@ -241,7 +246,9 @@ def scatter(
     categorical_interaction = False
 
     # create a matplotlib figure, if `ax` hasn't been specified.
-    if ax is None:
+    # Never call plt.sca() — it mutates global pyplot state and breaks subplot isolation.
+    _ax_provided = ax is not None
+    if not _ax_provided:
         figsize = (7.5, 5) if interaction_index != ind and interaction_index is not None else (6, 5)
         _, ax = plt.subplots(figsize=figsize)
 
@@ -292,7 +299,7 @@ def scatter(
         elif clow % 1 == 0 and chigh % 1 == 0 and chigh - clow < 10:
             categorical_interaction = True
 
-        # discritize colors for categorical features
+        # discretize colors for categorical features
         if categorical_interaction and clow != chigh:
             clow = np.nanmin(cv.astype(float))
             chigh = np.nanmax(cv.astype(float))
@@ -348,15 +355,15 @@ def scatter(
         p = ax.scatter(xv, s, s=dot_size, linewidth=0, color=color, alpha=alpha, rasterized=len(xv) > 500)
 
     if interaction_index != ind and interaction_index is not None:
-        # draw the color bar
+        # draw the color bar — use ax= so matplotlib steals from the correct axes only
         if isinstance(cd[0], str):
             tick_positions = np.array([cname_map[n] for n in cnames])
             tick_positions *= 1 - 1 / len(cnames)
             tick_positions += 0.5 * (chigh - clow) / (chigh - clow + 1)
-            cb = plt.colorbar(p, ticks=tick_positions, ax=ax, aspect=80)
+            cb = ax.figure.colorbar(p, ticks=tick_positions, ax=ax, aspect=80)
             cb.set_ticklabels(cnames)
         else:
-            cb = plt.colorbar(p, ax=ax, aspect=80)
+            cb = ax.figure.colorbar(p, ax=ax, aspect=80)
 
         # Type narrowing for mypy
         assert isinstance(interaction_index, (int, np.integer)), f"Unexpected {type(interaction_index)=}"
@@ -366,8 +373,6 @@ def scatter(
             cb.ax.tick_params(length=0)
         cb.set_alpha(1)
         cb.outline.set_visible(False)  # type: ignore
-    #         bbox = cb.ax.get_window_extent().transformed(fig.dpi_scale_trans.inverted())
-    #         cb.ax.set_aspect((bbox.height - 0.7) * 20)
 
     xmin = parse_axis_limit(xmin, xv, is_shap_axis=False)
     xmax = parse_axis_limit(xmax, xv, is_shap_axis=False)
@@ -403,8 +408,6 @@ def scatter(
     if hist:
         _plot_histogram(ax, xv, xv_no_jitter)
 
-    plt.sca(ax)
-
     # make the plot more readable
     ax.set_xlabel(name, color=axis_color, fontsize=13)
     ax.set_ylabel(labels["VALUE_FOR"] % name, color=axis_color, fontsize=13)
@@ -424,6 +427,7 @@ def scatter(
         with warnings.catch_warnings():  # ignore expected matplotlib warnings
             warnings.simplefilter("ignore", RuntimeWarning)
             plt.show()
+        return None
     else:
         return ax
 
@@ -451,18 +455,15 @@ def _suggest_x_jitter(values: np.ndarray) -> float:
         diffs = np.diff(unique_vals)
         min_dist = np.min(diffs[diffs > 1e-8])
     except (TypeError, ValueError):
-        # If unique_vals contains non-numeric values or all differences are to small, set arbitrarily at 1
+        # If unique_vals contains non-numeric values or all differences are too small, set arbitrarily at 1
         min_dist = 1
 
     num_points_per_value = len(values) / len(unique_vals)
     if num_points_per_value < 10:
-        # categorical = False
         x_jitter = 0
     elif num_points_per_value < 100:
-        # categorical = True
         x_jitter = min_dist * 0.1
     else:
-        # categorical = True
         x_jitter = min_dist * 0.2
     return x_jitter
 
@@ -636,7 +637,8 @@ def dependence_legacy(
     categorical_interaction = False
 
     # create a matplotlib figure, if `ax` hasn't been specified.
-    if not ax:
+    _ax_provided = ax is not None
+    if not _ax_provided:
         figsize = (7.5, 5) if interaction_index != ind and interaction_index is not None else (6, 5)
         fig = plt.figure(figsize=figsize)
         ax = fig.gca()
@@ -654,7 +656,8 @@ def dependence_legacy(
 
         # there is no interaction coloring for the main effect
         if ind1 == ind2:
-            fig.set_size_inches(6, 5, forward=True)
+            if not _ax_provided:
+                fig.set_size_inches(6, 5, forward=True)
 
         # TODO: remove recursion; generally the functions should be shorter for more maintainable code
         dependence_legacy(
@@ -678,7 +681,8 @@ def dependence_legacy(
 
         if show:
             plt.show()
-        return
+            return None
+        return ax
 
     assert shap_values.shape[0] == features.shape[0], (
         "'shap_values' and 'features' values must have the same number of rows!"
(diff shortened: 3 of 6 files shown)


gouravgangwardev:
This PR completes the final remaining work for meta issue #3411.

All plotting functions now consistently support the `ax` parameter and follow standardized matplotlib behavior:
- Correct axis handling with no global state leakage
- Consistent return contract (`show=False`, `show=True`, `return_objects`)
- Subplot-safe rendering
- Preserved backward compatibility

With this, the plotting API is now fully aligned and #3411 can be considered resolved.

shap/plots/_bar.py
```diff
@@ -240,11 +240,10 @@ def bar(
     if num_features < len(values[0]):
         yticklabels[-1] = f"Sum of {num_cut} other features"
 
-    if ax is None:
-        ax = plt.gca()
-        # Only modify the figure size if ax was not passed in
+    _ax_provided = ax is not None
+    if not _ax_provided:
```
adamwitmer: The logic here seems overcomplicated. Is there a reason for using a double negative here instead of just keeping the original logic for gating the `fig, ax` initialization?

shap/plots/_bar.py
```diff
@@ -339,9 +338,6 @@ def bar(
     else:
         ax.set_xlim(xmin, xmax + x_buffer)
 
-    # if features is None:
-    #     plt.xlabel(labels["GLOBAL_VALUE"], fontsize=13)
-    # else:
```
adamwitmer, line 344: Its probably not necessary to make unnecessary changes (however small) that are unrelated to the scope of the PR (https://github.com/shap/shap/blob/master/CONTRIBUTING.md#etiquette-for-creating-prs).

gouravgangwardev: Makes sense. I removed those commented lines as a small cleanup, but they're unrelated to the purpose of this PR.

shap/plots/_bar.py
gouravgangwardev: Good point. There wasn't a strong reason for introducing _ax_provided—I was aiming to make the "axes provided by the caller" state more explicit, but since it's only used once it ends up adding unnecessary indirection. I'll switch it back to if ax is None: while keeping the plt.subplots() change

shap/plots/_beeswarm.py
adamwitmer, line 1: What is the motivation for the changes proposed here given that the associated issue(s),  i.e. #3793 have already been addressed/closed in previous PR's? As mentioned in https://github.com/shap/shap/blob/master/CONTRIBUTING.md#etiquette-for-creating-prs:

> Before starting on a PR, please make a proposal by opening an Issue, checking for any duplicates.

This PR does not reference any open issues besides the ongoing meta-issue #3411, where other PR's related to the `_beeswarm.py` module have aleady been merged and corresponding issues have been closed. If you have an issue open you should reference here, otherwise you should open an issue to provide proper motivation for the need for these changes.

In general, this PR has an issue with scoping (also mentioned explicitly in the `CONTRIBUTING.md` docs), so if the changes are genuinely necessary/well motivated, they should be addressed in a separate, focused PR, rather than being grouped together with changes to other functions/modules.

gouravgangwardev: Thanks for pointing this out. You're right that these changes are no longer directly tied to an open issue. My intention was to address some remaining inconsistencies I noticed while working in this area, but I agree that they should have been better scoped and motivated.

I'll separate the unrelated changes and, if they still make sense independently, open a dedicated issue/PR with a clear rationale rather than bundling them here.

shap/plots/_scatter.py
adamwitmer, line 1: same comment here re: scope of/motivation for the changes in this PR given that the issue related to `_scatter.py` plot overhaul (#3804) has been closed for quite some time, and the only issue referenced here is the plot overhaul meta-issue #3411. If these changes are genuinely needed/well motivated, they should be addressed in a separate PR that describes the reasoning for the changes.

gouravgangwardev: You're right—I should have scoped this more carefully. I'll split the unrelated changes into focused PRs and open issues where needed so each change has a clear motivation. Thanks for pointing this out.

shap/plots/_scatter.py
```diff
@@ -241,7 +246,9 @@ def scatter(
     categorical_interaction = False
 
     # create a matplotlib figure, if `ax` hasn't been specified.
-    if ax is None:
+    # Never call plt.sca() — it mutates global pyplot state and breaks subplot isolation.
```
adamwitmer, line 249: You explicitly added a call to `plt.sca()` in the `_beeswarm.py` module, but here is says to never call it. Is there a difference in its usage there that allows it to be called without the breaking changes you mention here?

shap/plots/_scatter.py
```diff
@@ -348,15 +355,15 @@ def scatter(
         p = ax.scatter(xv, s, s=dot_size, linewidth=0, color=color, alpha=alpha, rasterized=len(xv) > 500)
 
     if interaction_index != ind and interaction_index is not None:
-        # draw the color bar
+        # draw the color bar — use ax= so matplotlib steals from the correct axes only
```
adamwitmer, line 358: inline comment change here seems unnecessary given the context provided by calling `ax.figure.colorbar`

shap/plots/_scatter.py
gouravgangwardev: That comment was left over from an earlier iteration while I was experimenting with the implementation, and I forgot to update it after the changes. I'll revise it so that it accurately reflects the final implementation.

shap/plots/_scatter.py
gouravgangwardev: Good catch. I added this comment while figuring out the changes and comparing different approaches, but after the final implementation it no longer adds any useful context. I forgot to remove it—I'll clean it up.

shap/plots/_scatter.py
```diff
@@ -292,7 +299,7 @@ def scatter(
         elif clow % 1 == 0 and chigh % 1 == 0 and chigh - clow < 10:
             categorical_interaction = True
 
-        # discritize colors for categorical features
+        # discretize colors for categorical features
```
adamwitmer, line 302: here and below, improperly scoped (albeit valid) changes unrelated to the plot overhaul changes performed in this PR. Perhaps adding a relevant `numpy` style prefix to the PR title (as is requested in in the `CONTRIBUTING.md` documentation: https://github.com/shap/shap/blob/master/CONTRIBUTING.md#etiquette-for-creating-prs) and commit messages would be appropriate given the scoping issues within this PR. Example prefixes are shown here for context: https://numpy.org/devdocs/dev/development_workflow.html#writing-the-commit-message.

shap/plots/_scatter.py
```diff
@@ -366,8 +373,6 @@ def scatter(
             cb.ax.tick_params(length=0)
         cb.set_alpha(1)
         cb.outline.set_visible(False)  # type: ignore
-    #         bbox = cb.ax.get_window_extent().transformed(fig.dpi_scale_trans.inverted())
-    #         cb.ax.set_aspect((bbox.height - 0.7) * 20)
```
adamwitmer, line 370: Unrelated changes to the scope of this PR.

shap/plots/_scatter.py
```diff
@@ -451,18 +455,15 @@ def _suggest_x_jitter(values: np.ndarray) -> float:
         diffs = np.diff(unique_vals)
         min_dist = np.min(diffs[diffs > 1e-8])
     except (TypeError, ValueError):
-        # If unique_vals contains non-numeric values or all differences are to small, set arbitrarily at 1
+        # If unique_vals contains non-numeric values or all differences are too small, set arbitrarily at 1
```
adamwitmer, line 458: unrelated (albeit valid) change to inline comment.

shap/plots/_scatter.py
```diff
@@ -451,18 +455,15 @@ def _suggest_x_jitter(values: np.ndarray) -> float:
...
         min_dist = 1
 
     num_points_per_value = len(values) / len(unique_vals)
     if num_points_per_value < 10:
-        # categorical = False
```
adamwitmer, line 459: unrelated changes to the scope of this PR here and below re: `# categorical = True`

shap/plots/_scatter.py
```diff
@@ -241,7 +246,9 @@ def scatter(
     categorical_interaction = False
 
     # create a matplotlib figure, if `ax` hasn't been specified.
-    if ax is None:
+    # Never call plt.sca() — it mutates global pyplot state and breaks subplot isolation.
+    _ax_provided = ax is not None
+    if not _ax_provided:
```
adamwitmer, line 251: Similar comment as before re: use of double negative here. Although `_ax_provided` is used in more than one place in this module, the logic is a bit confusing in relation to the original logic of simply checking for the existence of `ax`.

shap/plots/_scatter.py
```diff
@@ -636,7 +637,8 @@ def dependence_legacy(
     categorical_interaction = False
 
     # create a matplotlib figure, if `ax` hasn't been specified.
-    if not ax:
+    _ax_provided = ax is not None
+    if not _ax_provided:
```
adamwitmer, line 641: same comment as above re: double negative here.

shap/plots/_scatter.py
```diff
@@ -728,7 +732,7 @@ def dependence_legacy(
         elif clow % 1 == 0 and chigh % 1 == 0 and chigh - clow < 10:
             categorical_interaction = True
 
-        # discritize colors for categorical features
+        # discretize colors for categorical features
```
adamwitmer, line 735: unrelated (albeit valid) change to inline comment.

shap/plots/_scatter.py
```diff
@@ -781,19 +785,17 @@ def dependence_legacy(
             if len(tick_positions) == 2:
                 tick_positions[0] -= 0.25
                 tick_positions[1] += 0.25
-            cb = plt.colorbar(p, ticks=tick_positions, ax=ax, aspect=80)
+            cb = fig.colorbar(p, ticks=tick_positions, ax=ax, aspect=80)
```
adamwitmer, line 788: is there a reason why you use `fig.colorbar` here and below as opposed to using `ax.figure.colorbar` as you do above?

shap/plots/_scatter.py
```diff
@@ -781,19 +785,17 @@ def dependence_legacy(
...
+            cb = fig.colorbar(p, ax=ax, aspect=80)
 
         cb.set_label(feature_names[interaction_index], size=13)
         cb.ax.tick_params(labelsize=11)
         if categorical_interaction:
             cb.ax.tick_params(length=0)
         cb.set_alpha(1)
         cb.outline.set_visible(False)  # type: ignore
-    #         bbox = cb.ax.get_window_extent().transformed(fig.dpi_scale_trans.inverted())
-    #         cb.ax.set_aspect((bbox.height - 0.7) * 20)
```
adamwitmer, line 796: unrelated changes to remove commented code here re: scoping.

tests/plots/test_bar.py
```diff
@@ -110,3 +110,121 @@ def test_bar_raises_error_for_empty_explanation(explainer):
     shap_values = explainer(explainer.data)
     with pytest.raises(ValueError, match="The passed Explanation is empty"):
         shap.plots.bar(shap_values[0:0], show=False)
+
+
+def _make_explanation(n=50, n_features=5, seed=0):
```
adamwitmer, line 115: repeated use of this function in multiple tests (across different testing modules) suggests that it should be used as a `pytest` fixture and possibly added to the appropriate `conftest.py` module with the appropriate `scope` parameter, i.e. https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session.

tests/plots/test_bar.py
```diff
@@ -110,3 +110,121 @@ def test_bar_raises_error_for_empty_explanation(explainer):
...
+        values=values,
+        data=data,
+        feature_names=feature_names,
+        base_values=np.zeros(n),
+    )
+
+
+def test_show_false_returns_axes():
+    """show=False must return a matplotlib Axes object (not None, not some other type)."""
+    plt.close("all")
```
adamwitmer, line 131: Is there a good reason to call `plt.close("all")` at the beginning and end of each test? Seems somewhat redundant  given that all the tests call `plt.close("all")` at the end too.

tests/plots/test_bar.py
```diff
@@ -110,3 +110,121 @@ def test_bar_raises_error_for_empty_explanation(explainer):
...
+    expln = _make_explanation()
+    result = shap.plots.bar(expln, show=False)
+    assert isinstance(result, plt.Axes), f"Expected plt.Axes, got {type(result)}"
+    plt.close("all")
+
+
+def test_show_true_returns_none():
```
adamwitmer, line 138: In general, the number of tests and redundancy in logic suggests that there may be opportunity to parametrize some of these tests (where possible/appropriate) to reduce the overall size of the diff here.

tests/plots/test_bar.py
```diff
@@ -110,3 +110,121 @@ def test_bar_raises_error_for_empty_explanation(explainer):
...
+
+def test_figure_size_unchanged_when_ax_provided():
+    """When an ax is provided, bar() must not alter the figure size."""
+    plt.close("all")
+    original_size = (4.0, 3.0)
+    fig, ax = plt.subplots(figsize=original_size)
+    expln = _make_explanation()
+    shap.plots.bar(expln, ax=ax, show=False)
+    actual_size = tuple(fig.get_size_inches())
+    assert actual_size == pytest.approx(original_size, rel=1e-3), (
```
adamwitmer, line 181: What is the motivation for using `pytest.approx` here, if the figure tuple should be the exact same, than you can just make a direct assertion on the expected output.

tests/plots/test_bar.py
```diff
@@ -110,3 +110,121 @@ def test_bar_raises_error_for_empty_explanation(explainer):
...
+def test_figure_size_set_when_no_ax_provided():
+    """When no ax is provided, bar() must resize the figure to fit the features."""
+    plt.close("all")
+    expln = _make_explanation(n_features=5)
+    result = shap.plots.bar(expln, show=False)
+    fig = result.get_figure()
+    w, h = fig.get_size_inches()
+    # Default logic: 8 wide, height depends on num_features — just verify it's not the mpl default (6.4 x 4.8)
+    assert w == pytest.approx(8.0, rel=1e-3), f"Expected width 8.0, got {w}"
+    assert h != pytest.approx(4.8, rel=0.1), "Figure height was not resized from the mpl default"
```
adamwitmer, line 196: The logic here is a bit confusing. We are testing that the height of the image should scale with the number of features but never be close to 4.8 with a relative tolerance of 0.1, however if I change the number of input features to `n_features=7` above, the output of `h` is 5.0 and the test fails with the following traceback:

```
E       AssertionError: Figure height was not resized from the mpl default
E       assert np.float64(5.0) != 4.8 ± 0.48
E        +  where 4.8 ± 0.48 = <function approx at 0x10144be20>(4.8, rel=0.1)
E        +    where <function approx at 0x10144be20> = pytest.approx
```

So either there is a flaw in your logic or the test is not robust to changes in `n_features` given the relative tolerance value used here.

tests/plots/test_beeswarm.py
```diff
@@ -119,3 +119,181 @@ def test_beeswarm_colors_values_with_data(color):
...
     )
     shap.plots.beeswarm(explanation, show=False, color_bar=True, color=color)
+
+
+# -----------------------------------------------------------------------
+# API consistency tests
+# -----------------------------------------------------------------------
+
+
+def _make_explanation(n=50, n_features=5, seed=0):
```
adamwitmer, line 129: same comment here as above re making a `pytest` fixture for this function.

tests/plots/test_beeswarm.py
```diff
@@ -119,3 +119,181 @@ def test_beeswarm_colors_values_with_data(color):
...
+def test_figure_size_unchanged_when_ax_provided():
+    """When an ax is provided, the figure size must not be altered by beeswarm()."""
+    plt.close("all")
+    original_size = (4.0, 3.0)
+    fig, ax = plt.subplots(figsize=original_size)
+    expln = _make_explanation()
+    # plot_size must be None when ax is provided (otherwise ValueError is raised)
+    shap.plots.beeswarm(expln, ax=ax, plot_size=None, show=False)
+    actual_size = tuple(fig.get_size_inches())
+    assert actual_size == pytest.approx(original_size, rel=1e-3), (
```
adamwitmer, line 186: similar comment here as above re: the expectation that we should be sensitive to small changes in floating point variation of the output size.

tests/plots/test_beeswarm.py
```diff
@@ -119,3 +119,181 @@ def test_beeswarm_colors_values_with_data(color):
...
+    rs = np.random.RandomState(seed)
+    values = rs.randn(n, n_features)
+    data = rs.randn(n, n_features)
+    feature_names = [f"feat{i}" for i in range(n_features)]
+    return shap.Explanation(values=values, data=data, feature_names=feature_names)
+
+
+def test_show_false_returns_axes():
+    """show=False must return a matplotlib Axes object."""
+    plt.close("all")
```
adamwitmer, line 140: same comment here as above re: multiple redundant calls to `plt.close("all")`

tests/plots/test_beeswarm.py
```diff
@@ -119,3 +119,181 @@ def test_beeswarm_colors_values_with_data(color):
...
+
+    shap.plots.beeswarm(expln, ax=ax, show=False)
+    n_collections_after_second = len(ax.collections)
+    assert n_collections_after_second > n_collections_after_first, (
+        "Second beeswarm() call did not add new artists to the axes"
+    )
+    plt.close("all")
+
+
+def test_colorbar_compatibility():
```
adamwitmer, line 211: same general comment here as above re: opportunity to parameterize tests with similar logic/assertions.

tests/plots/test_scatter.py
```diff
@@ -112,3 +112,196 @@ def test_scatter_plot_value_input(input):
...
     plt.tight_layout()
     return plt.gcf()
+
+
+# -----------------------------------------------------------------------
+# API consistency tests
+# -----------------------------------------------------------------------
+
+
+def _make_explanation(n=50, seed=0):
```
adamwitmer, line 122: same comment here as above re: `pytest` fixture.

tests/plots/test_scatter.py
```diff
@@ -112,3 +112,196 @@ def test_scatter_plot_value_input(input):
...
+    """Create a simple single-column Explanation for unit tests."""
+    rs = np.random.RandomState(seed)
+    values = rs.randn(n)
+    data = rs.randn(n)
+    return shap.Explanation(values=values, data=data, feature_names="feature1")
+
+
+def _make_explanation_2col(n=50, seed=0):
```
adamwitmer, line 130: This could be another, separate fixture.

tests/plots/test_scatter.py
```diff
@@ -112,3 +112,196 @@ def test_scatter_plot_value_input(input):
...
+def test_show_false_returns_axes():
+    """show=False must return a matplotlib Axes object."""
+    plt.close("all")
+    expln = _make_explanation()
+    result = shap.plots.scatter(expln, show=False)
+    assert isinstance(result, plt.Axes), f"Expected Axes, got {type(result)}"
+    plt.close("all")
+
+
+def test_show_true_returns_none():
```
adamwitmer, line 147: I am able to delete a significant portion of the logic used in this test without breaking the tests, i.e.:

```diff
diff --git a/tests/plots/test_scatter.py b/tests/plots/test_scatter.py
index 9689e38e..abd8a53d 100644
--- a/tests/plots/test_scatter.py
+++ b/tests/plots/test_scatter.py
@@ -148,13 +148,7 @@ def test_show_true_returns_none():
     """show=True must return None (pyplot handles display)."""
     plt.close("all")
     expln = _make_explanation()
-    # Patch plt.show to be a no-op so the test doesn't open a window
-    original_show = plt.show
-    plt.show = lambda: None
-    try:
-        result = shap.plots.scatter(expln, show=True)
-    finally:
-        plt.show = original_show
+    result = shap.plots.scatter(expln, show=True)
     assert result is None, f"Expected None when show=True, got {type(result)}"
     plt.close("all")
```

Which may warrant a bit more explanation via inline comments for the necessity of this logical complexity.

tests/plots/test_scatter.py
```diff
@@ -112,3 +112,196 @@ def test_scatter_plot_value_input(input):
...
+    plt.close("all")
+
+
+def test_figure_size_set_when_ax_not_provided():
+    """When no ax is provided, scatter() must create a figure with a reasonable default size."""
+    plt.close("all")
+    expln = _make_explanation()
+    result = shap.plots.scatter(expln, show=False)
+    fig_size = result.get_figure().get_size_inches()
+    # Default single-feature size should be approximately (6, 5)
```
adamwitmer, line 211: Is there a need to account for floating point tolerances here? What is the scenario where the figure size will be approximately 6 x 5?

shap/plots/_scatter.py
gouravgangwardev: Good point. I'll remove these unrelated cleanup changes from this PR and keep it focused on the plot behavior changes.

shap/plots/_scatter.py
gouravgangwardev: I introduced _ax_provided to make the intent explicit, but it doesn't improve readability here. I'll switch it back to if ax is None:. like in previous

shap/plots/_scatter.py
gouravgangwardev: No strong reason—I used fig because it was already available in that code path. Using ax.figure.colorbar(...) is more consistent with the rest of the module,

shap/plots/_scatter.py
gouravgangwardev: This comment update is unrelated to the main purpose of the PR, so I'll remove it and keep it for a separate cleanup change.

shap/plots/_scatter.py
gouravgangwardev: This cleanup is outside the scope of the PR, so I'll remove it to keep the changes focused.

shap/plots/_scatter.py
gouravgangwardev: Good point. I introduced `_ax_provided` to make the intent explicit, but it doesn't improve readability here. I'll switch it back to `if ax is None:`.
like previous

shap/plots/_scatter.py
gouravgangwardev: same

shap/plots/_scatter.py
gouravgangwardev: just made for personal clearity and forgot to remove

shap/plots/_scatter.py
gouravgangwardev: for personal understanding only

tests/plots/test_bar.py
gouravgangwardev: Thanks for the suggestion. I hadn't considered moving these helpers into shared fixtures, but I agree it would reduce duplication. I'll look into consolidating them appropriately.

tests/plots/test_bar.py
gouravgangwardev: The extra plt.close("all") isn't necessary here. I'll remove the redundant calls. it was a overlook on my side

tests/plots/test_bar.py
gouravgangwardev: Thanks for the suggestion. I agree there's room to reduce duplication here by parameterizing the tests where it improves readability. I'll look into consolidating the similar cases.

tests/plots/test_scatter.py
gouravgangwardev: Good point. The figure size should be deterministic here, so I'll replace the approximate assertions with direct equality checks.

tests/plots/test_scatter.py
gouravgangwardev: . I added that while working through the test implementation

tests/plots/test_scatter.py
gouravgangwardev: Good suggestion. I'll consider making this a separate fixture to better reflect its distinct purpose.

tests/plots/test_scatter.py
gouravgangwardev: Thanks for the suggestion. I'll look into extracting these helpers into shared fixtures as well.

tests/plots/test_beeswarm.py
gouravgangwardev: ok  I'll review the tests and consolidate similar cases where parameterization improves readability.

tests/plots/test_beeswarm.py
gouravgangwardev: Good catch. I'll remove the redundant `plt.close("all")` calls here as well.

tests/plots/test_beeswarm.py
gouravgangwardev: Good point. Since the expectation is that the figure size remains unchanged, a direct equality assertion is more appropriate here. I'll update it.

tests/plots/test_beeswarm.py
gouravgangwardev: Thanks for the suggestion. I'll look into moving these shared helpers into fixtures to reduce duplication across the test modules.

tests/plots/test_bar.py
gouravgangwardev: Thanks for pointing this out. You're right that this assertion isn't robust to changes in the number of features. I'll revisit the test so it verifies the intended behavior without depending on this tolerance.

tests/plots/test_bar.py
gouravgangwardev: Good point. Since the figure size should remain unchanged in this case, a direct equality check is more appropriate. I'll update the assertion.

PR #4963, opened 2026-05-03
FIX: Apply title argument in legacy summary plot

### LINKED ISSUE #4962: BUG: summary_plot ignores title argument for standard summary plots
### Issue Description

`shap.summary_plot` accepts a `title` argument, but for a standard summary plot the title is not applied to the matplotlib plot.

This makes the argument look supported from the public API, but passing it has no visible effect. I noticed this while checking the legacy summary plot path in `_beeswarm.py`.

Expected behavior: the provided title should be shown on the plot.
Actual behavior: the plot title stays empty.

```python
import numpy as np
import shap

title = "Custom summary title"

shap.summary_plot(
    np.random.randn(20, 5),
    np.random.randn(20, 5),
    title=title,
    show=False,
)

print(plt.gca().get_title())
```

### Traceback

```shell
No traceback is raised.
```

### Expected Behavior

The provided title should be applied to the matplotlib plot.

For the example above, this should print:

Custom summary title

### Bug report checklist

### Installed Versions

Python: 3.12.13 
shap: 0.51.1.dev67
numpy: 2.4.4
matplotlib: 3.10.9

Tharun-10Dragneel:
## Overview

Closes #4962 

- Apply the `title` argument in the legacy `summary_plot` path.
- Add a regression test for `shap.summary_plot(..., title=...)`.

## Checklist

### DIFF
--- a/shap/plots/_beeswarm.py
+++ b/shap/plots/_beeswarm.py
@@ -1147,6 +1147,10 @@ def summary_legacy(
         plt.xlabel(labels["GLOBAL_VALUE"], fontsize=13)
     else:
         plt.xlabel(labels["VALUE"], fontsize=13)
+
+    if title is not None:
+        plt.title(title)
+
     plt.tight_layout()
     if show:
         plt.show()
--- a/tests/plots/test_summary.py
+++ b/tests/plots/test_summary.py
@@ -31,6 +31,19 @@ def test_summary_with_data():
     return fig
 
 
+def test_summary_plot_applies_title():
+    title = "Custom summary title"
+
+    shap.summary_plot(
+        np.random.randn(20, 5),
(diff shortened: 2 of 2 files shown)


mohityadav8 (changes requested):
Verified this fix locally. Reproduced the bug on master first: title comes back as '' via plt.gca().get_title() when passed to summary_plot. Confirmed this branch resolves it: the title is now correctly applied, and the added test_summary_plot_applies_title test passes.

One small cleanup: the stale TODO comment at line 519 ("# TODO: remove unused title argument / use title argument") should probably be removed as part of this fix, since the title argument is now used. Right now it's a bit misleading to leave it in.

Otherwise this looks good to merge.

Issue #4966, opened 2026-05-04
BUG: waterfall_legacy() uses duplicate y-tick positions causing matplotlib 3.3+ to collapse ticks


onkar717:
### Issue Description

`waterfall_legacy()` in `shap/plots/_waterfall.py` (line 657) calls:

```python
plt.yticks(list(range(num_features)) * 2, ...)
```

`list(range(n)) * 2` produces **duplicate** positions — e.g. for 5 features:
`[0, 1, 2, 3, 4, 0, 1, 2, 3, 4]`

Matplotlib 3.3+ detects duplicate tick positions and collapses them, causing the
dual-layer tick rendering (gray value labels + black feature names) to silently fail.
Only one layer of labels is drawn instead of two, making the plot look wrong.

### Minimal Reproducible Example

```python
import shap
import numpy as np
import matplotlib.pyplot as plt

expected_value = 0.5
shap_values = np.array([0.1, -0.2, 0.3, 0.05, -0.15])
features = np.array([1.0, 2.0, 3.0, 4.0, 5.0])
feature_names = ["a", "b", "c", "d", "e"]

shap.plots._waterfall.waterfall_legacy(
    expected_value, shap_values,
    features=features, feature_names=feature_names, show=False
)
plt.show()
# Result: only one set of tick labels rendered (missing gray value labels or black feature names)
```

### Traceback

```shell

```

### Expected Behavior

Both tick layers render correctly gray `value = feature_name` labels on the left,
and black `feature_name`-only labels on the right as in the modern `waterfall()`.

### Bug report checklist

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the [latest release](https://github.com/shap/shap/releases) of shap.
- [x] I have confirmed this bug exists on the [master branch](https://github.com/shap/shap/blob/master/CONTRIBUTING.md#installing-from-the-master-branch) of shap.
- [x] I'd be interested in making a PR to fix this bug

### Installed Versions

- shap version: latest main
- matplotlib >= 3.3

shuklapraveen1:
I’d be interested in working on this. I’ll first compare the current `waterfall()` implementation with `waterfall_legacy()` and try to backport the y-tick handling fix mentioned in #4967.

mohityadav8:
Reproduced on current master (matplotlib >= 3.3, via the exact repro in the issue).

[...]

PR #4967, opened 2026-05-04
FIX: backport duplicate y-tick position fix from waterfall() to waterfall_legacy()

onkar717:
## Summary

- `waterfall_legacy()` used `list(range(num_features)) * 2` for y-tick positions,
  producing duplicates (e.g. `[0,1,2,3,4,0,1,2,3,4]`).
- Matplotlib 3.3+ collapses duplicate tick positions, so the second layer of labels
  (black feature-name-only ticks) was silently dropped from the plot.
- The modern `waterfall()` already has the correct pattern at line 298 with an
  explanatory comment. This PR backports it to `waterfall_legacy()`.
 
 ## Issue

https://github.com/shap/shap/issues/4966

## Change

```diff
-    plt.yticks(
-        list(range(num_features)) * 2,
-        yticklabels[:-1] + [label.split("=")[-1] for label in yticklabels[:-1]],
-        fontsize=13,
-    )
+    # The 1e-8 is so matplotlib 3.3 doesn't try and collapse the ticks
+    ytick_pos = list(range(num_features)) + list(np.arange(num_features) + 1e-8)
+    plt.yticks(ytick_pos, yticklabels[:-1] + [label.split("=")[-1] for label in yticklabels[:-1]], fontsize=13)
```

`shap/plots/_waterfall.py`, line 656–660

## Testing

The existing `test_waterfall_legacy` in `tests/plots/test_waterfall.py` (line 40)
uses `@pytest.mark.mpl_image_compare` and exercises this exact code path.

Run:
```bash
pytest tests/plots/test_waterfall.py::test_waterfall_legacy -v
```

The baseline image needs to be regenerated since the tick positions now correctly
produce two distinct rendered layers:
```bash
pytest tests/plots/test_waterfall.py::test_waterfall_legacy --mpl-generate-path=tests/plots/baseline
```

## Related

- Fixes: [issue link]
- The fix in the modern `waterfall()` was introduced in the same file at line 298
  with the comment `"The 1e-8 is so matplotlib 3.3 doesn't try and collapse the ticks"`
  but was never applied to the legacy function.
```

### DIFF
--- a/shap/plots/_waterfall.py
+++ b/shap/plots/_waterfall.py
@@ -653,11 +653,9 @@ def waterfall_legacy(expected_value, shap_values=None, features=None, feature_na
             )
 
     # draw the y-ticks twice, once in gray and then again with just the feature names in black
-    plt.yticks(
-        list(range(num_features)) * 2,
-        yticklabels[:-1] + [label.split("=")[-1] for label in yticklabels[:-1]],
-        fontsize=13,
-    )
+    # The 1e-8 is so matplotlib 3.3 doesn't try and collapse the ticks
+    ytick_pos = list(range(num_features)) + list(np.arange(num_features) + 1e-8)
+    plt.yticks(ytick_pos, yticklabels[:-1] + [label.split("=")[-1] for label in yticklabels[:-1]], fontsize=13)
 
     # put horizontal lines for each feature row
     for i in range(num_features):


mohityadav8 (commented):
Verified this fix locally. Confirmed the bug on master first (only one label layer renders due to matplotlib collapsing the duplicate tick positions), then confirmed this branch fixes it: ax.get_yticklabels() now correctly returns two distinct layers (e.g. "4 = d" and "d"), and the rendered plot shows both gray value labels and black feature-name labels as expected.

Note: #5021 also fixes this same issue with the same underlying approach (the 1e-8 offset), and additionally includes a regression test. Since both PRs resolve the bug correctly, might be worth deciding which one to merge to avoid duplicate work, possibly this one plus #5021's test.

CloseChoice:
can we get a test for this pls?

PR #4975, opened 2026-05-05
Enable brute-force iteration in LinearExplainer for cases where permutations < n_samples

Jatinbhardwaj-093:
## Overview

Add a TODO in `LinearExplainer._estimate_transforms` by adding brute-force enumeration of all feature permutations when `M! ≤ nsamples`.

Previously, the method always used random sampling (Monte Carlo) to estimate the transform matrices, even when the number of features was small enough to enumerate all permutations cheaply. For example, a model with 3 features has only 3! = 6 permutations, yet the code would randomly sample 1,000 times — wasting computation and introducing unnecessary noise.

## Checklist

## AI Disclosure 
- No coding agents were used; only code completion tools were utilized.
- Tests were generated using Gemini 3.1.


Jatinbhardwaj-093:
@CloseChoice can you please review this? It has been pending for a week.

PR #5021, opened 2026-05-28
FIX: prevent duplicate y-tick collapse in waterfall_legacy

shuklapraveen1:
## Overview

Related to #4966

This updates `waterfall_legacy()` to avoid duplicate y-tick positions when rendering the dual-layer labels.

Previously the implementation used:

```python
list(range(num_features)) * 2
```

which produces duplicate tick positions such as:

```python
[0, 1, 2, 3, 4, 0, 1, 2, 3, 4]
```

Matplotlib 3.3+ collapses duplicate tick positions, causing one layer of labels to disappear.

This patch mirrors the approach already used in the modern `waterfall()` implementation by applying a small offset (`1e-8`) to the second tick layer.

## Changes

* updates `waterfall_legacy()` y-tick handling
* adds a regression test for duplicate y-tick collapse behavior

## Testing

```bash
pytest tests/plots/test_waterfall.py
```

All tests pass successfully.

## Checklist

* [x] All pre-commit checks pass.
* [x] Unit tests added (if fixing a bug or adding a new feature)


mohityadav8 (changes requested):
One gap in the added test: it only checks the tick count (len(tick_labels) == 10), not that the two label layers actually differ. A regression where both halves end up identical (e.g. both showing just the feature name, or both showing "value = name") would still pass this assertion as long as the count stayed at 10.

Suggest strengthening it to check the label text itself, e.g.:

    def test_waterfall_legacy_duplicate_yticks():
        expected_value = 0.5
        shap_values = np.array([0.1, -0.2, 0.3, 0.05, -0.15])
        features = np.array([1.0, 2.0, 3.0, 4.0, 5.0])
        feature_names = ["a", "b", "c", "d", "e"]
        fig = plt.figure()
        shap.plots._waterfall.waterfall_legacy(
            expected_value,
            shap_values,
            features=features,
            feature_names=feature_names,
            show=False,
        )
        ax = plt.gca()
        tick_labels = [t.get_text() for t in ax.get_yticklabels()]

        assert len(tick_labels) == 10

        gray_layer = tick_labels[:5]
        black_layer = tick_labels[5:]

        # the two layers should not be identical, otherwise the dual-layer
        # rendering has silently collapsed back into one
        assert gray_layer != black_layer

        # the gray layer should contain the "=" separator (value = feature name),
        # the black layer should just be feature names
        assert all("=" in label for label in gray_layer)
        assert all("=" not in label for label in black_layer)

        plt.close(fig)

This would have actually caught the original bug (duplicated tick positions collapsing both layers into the same text), not just a positional count.

mohityadav8 (commented):
Verified this fix locally as well. Same result as #4967: ax.get_yticklabels() returns 10 distinct positions with two correct label layers, and the added test_waterfall_legacy_duplicate_yticks test passes.

Both this PR and #4967 apply the same fix (offsetting the second tick layer by 1e-8). This one has the advantage of including a regression test, which #4967 doesn't have. Might be worth maintainers picking one PR to merge, or porting this test over to #4967, to avoid merging two near-identical fixes.

PR #5024, opened 2026-05-29
FIX: raise ValueError for invalid MAPLE fe_type

shuklapraveen1:
> [!IMPORTANT]
>   
# Overview

This PR fixes an issue in `MAPLE` where passing an invalid `fe_type` causes the library to print an error message and terminate the Python interpreter using `sys.exit(0)`.

Currently:

```python
MAPLE(..., fe_type="invalid_type")
```

prints:

```text
Unknown FE type None
```

and exits the process.

Instead, invalid input should raise a Python exception that users can catch and handle.

## Changes

* Replaced `sys.exit(0)` with a `ValueError`
* Added a clear error message indicating the supported values (`"rf"` and `"gbrt"`)
* Added a regression test to verify invalid `fe_type` values raise `ValueError`

## Tests

Added a regression test that verifies invalid `fe_type` values raise `ValueError` instead of terminating the interpreter.

```python
def test_maple_invalid_fe_type():
    with pytest.raises(ValueError, match="Unknown FE type"):
        MAPLE(..., fe_type="invalid_type")
```

## Checklist

* [x] All pre-commit checks pass
* [x] Unit test added
* [x] Existing MAPLE tests pass


shuklapraveen1:
@CloseChoice please look into the fix

PR #5026, opened 2026-05-29
Use TreeEnsemble directly in TreeExplainer and add cpu gpu equivalence tests for categories

CloseChoice:
## Overview

This PR does two things:
1. enables the TreeExplainer to use TreeEnsembles directly
2. tests cpu/gpu equivalence tests for categories which uses 1.

The direct handling of the TreeEnsembles in the TreeExplainer is just used for testing here but could also be improved on later.

## Checklist

- [x] Unit tests added ~(if fixing a bug or adding a new feature)~


RAMitchell (approved):
Very nice

PR #5030, opened 2026-06-02
DOCS: replace deprecated K.get_session() in VGG16 example

shuklapraveen1:
Related to #3036.

## Summary

This PR updates the VGG16 GradientExplainer example to use TensorFlow 2.x compatible APIs and removes the dependency on the deprecated `keras.backend.get_session()` workflow.

## Background

The current example uses:

```python
K.get_session().run(...)
```

which is no longer available in modern TensorFlow/Keras environments and causes the example to fail when executed with recent TensorFlow versions.

## Changes

* Added `import tensorflow as tf`.
* Replaced the session-based implementation of `map2layer()` with a TensorFlow 2.x compatible approach using `tf.keras.Model`.
* Removed reliance on TensorFlow 1.x graph/session execution.

### Before

```python
def map2layer(x, layer):
    feed_dict = dict(zip([model.layers[0].input], [preprocess_input(x.copy())]))
    return K.get_session().run(model.layers[layer].input, feed_dict)
```

### After

```python
def map2layer(x, layer):
    intermediate_model = tf.keras.Model(
        inputs=model.input,
        outputs=model.layers[layer].input,
    )

    return intermediate_model(
        preprocess_input(x.copy()),
        training=False,
    ).numpy()
```

## Validation

* Verified that the updated implementation returns the expected intermediate-layer activations.
* Confirmed that the output shape matches the original behavior.
* Tested successfully with a modern TensorFlow 2.x environment.

## Impact

This change improves compatibility with current TensorFlow/Keras releases while preserving the original functionality of the example.


shuklapraveen1:
Contributes to #3036.

This comment the VGG16 GradientExplainer example in the README to use TensorFlow 2.x APIs instead of the deprecated `keras.backend.get_session()` workflow.

## Changes

* Replace `import keras.backend as K` with `import tensorflow as tf`.
* Replace `K.get_session().run(...)` with a `tf.keras.Model` based implementation.
* Remove reliance on TensorFlow 1.x session execution.

shuklapraveen1:
Hi! Just checking in on this PR. The compatibility update is complete, all checks are passing, and I've verified it works with TensorFlow 2.x. I'd appreciate a review whenever someone has time. Thanks!

PR #5044, opened 2026-06-15
Fix/gradient explainer tf get session removed

Ebube22:
## Overview

TensorFlow removed `tf.compat.v1.get_session()` and the global Keras session helpers in recent versions, which caused `_get_session` in `shap/explainers/tf_utils.py` to raise an `AttributeError` when no session was explicitly passed in.

This PR updates `_get_session` to first check whether a default TensorFlow session already exists before attempting to create a new one, falling back gracefully on newer TF versions where the old session APIs are no longer available.

## Checklist

### DIFF
--- a/shap/explainers/tf_utils.py
+++ b/shap/explainers/tf_utils.py
@@ -25,17 +25,12 @@ def _get_session(session):
         try:
             session = tf.compat.v1.keras.backend.get_session()
         except AttributeError:
-            # get_session was removed from Keras in 2023. Fall back to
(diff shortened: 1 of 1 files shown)


Ebube22:
Note: the failing checks appear to be pre-existing CI issues unrelated to this change (sklearn `DOUBLE` import error in test_tree.py and matplotlib image comparison mismatches in the plots tests) — this PR only modifies `_get_session` in `tf_utils.py`.

PR #5078, opened 2026-07-11
FIX: shap.Explainer hard-fails on InterpretML EBMs with interaction terms

### LINKED ISSUE #4942: BUG: shap.Explainer hard-fails on default InterpretML EBMs with interaction terms
### Issue Description

`shap.Explainer` hard-fails on default/current InterpretML Explainable Boosting Machines because `AdditiveExplainer` raises
  `NotImplementedError` whenever `model.interactions != 0`.

  SHAP has a specialized `AdditiveExplainer` path for `interpret.glassbox.ExplainableBoostingClassifier`, but in
  `shap/explainers/_additive.py`, `supports_model_with_masker()` raises:

  ```python
  NotImplementedError("Need to add support for interaction effects!")
```

  whenever model.interactions != 0. Because this is raised from supports_model_with_masker() during auto-selection, it crashes the unified    
  shap.Explainer(...) API rather than falling back to a model-agnostic explainer.

  This is significant in practice because current InterpretML EBMs enable interactions by default:

  - ExplainableBoostingClassifier(..., interactions="3x", ...)
  - ExplainableBoostingRegressor(..., interactions="5x", ...)

  So a default-constructed EBM cannot be explained via shap.Explainer.

  Additionally, ExplainableBoostingRegressor is not handled in the current active code path at all — only ExplainableBoostingClassifier is    
  checked.

```python
import shap
  from interpret.glassbox import ExplainableBoostingClassifier

  X, y = make_classification(n_samples=300, n_features=6, random_state=0)

  model = ExplainableBoostingClassifier(random_state=0)
  model.fit(X, y)

```

### Traceback

```shell
    File "repro.py", line 29, in <module>
    File ".../shap/explainers/_explainer.py", line 193, in __init__
      elif explainers.AdditiveExplainer.supports_model_with_masker(model, self.masker):
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
    File ".../shap/explainers/_additive.py", line 117, in supports_model_with_masker
      raise NotImplementedError("Need to add support for interaction effects!")
  NotImplementedError: Need to add support for interaction effects!
```

### Expected Behavior

SHAP should either:

  1. Support InterpretML EBMs with interaction terms in `AdditiveExplainer`, producing SHAP values that satisfy additivity against the model's
   raw additive score; or
  2. Gracefully fall back to a model-agnostic explainer when interactions are present, instead of raising from `supports_model_with_masker()` 
  during auto-selection.

### Bug report checklist

### Installed Versions

- shap: 0.51.0 (latest PyPI release) — also reproduced on `master` @ 0.51.1.dev67
- interpret: 0.7.8
- scikit-learn: 1.8.0
- Python: 3.14.4
- OS: Windows 11

04pallav:
Fixes #4942

## Problem

`shap.Explainer` hard-fails on default InterpretML Explainable Boosting Machines because `AdditiveExplainer.supports_model_with_masker()` raises `NotImplementedError` when `model.interactions != 0`. This crashes the unified `shap.Explainer(...)` API during auto-selection rather than falling back to a model-agnostic explainer.

Additionally, `ExplainableBoostingRegressor` was completely unhandled in `__init__`.

### `shap/explainers/_additive.py`
- `supports_model_with_masker`: return `False` instead of `raise NotImplementedError` when interactions are present — lets auto-selector skip AdditiveExplainer gracefully
- `supports_model_with_masker`: add support for `ExplainableBoostingRegressor`
- `__init__`: add `elif` for `ExplainableBoostingRegressor` binding `self.model = model.predict`

### `shap/explainers/_explainer.py`
- Auto-selector fallback (`else` block): detect EBM models and wrap them into callables (`decision_function`/`predict`), then recursively re-invoke `__init__`. This follows the same pattern used for transformers pipeline wrapping (lines 136-158).

## Why both changes are needed

Returning `False` from `supports_model_with_masker` alone is insufficient — EBM objects are not callable (`callable(ebm)` returns `False`), so the auto-selector still raises `TypeError` in the fallback path. The `_explainer.py` change ensures EBM models become callable before the fallback check.

## Testing

- EBC with interactions (default) → model-agnostic explainer ✓
- EBR with interactions (default) → model-agnostic explainer ✓
- EBC without interactions (`interactions=0`) → AdditiveExplainer ✓
- EBR without interactions (`interactions=0`) → AdditiveExplainer ✓
- Existing additive tests: 3/3 passed (1 pre-existing failure, `test_additive_explainer_with_ebm_regressor` uses deprecated `shap_values()` API)
- Explainer tests: 1 passed, 4 skipped

### DIFF
--- a/shap/explainers/_additive.py
+++ b/shap/explainers/_additive.py
@@ -64,6 +64,9 @@ class AdditiveExplainer(Explainer):
                 )
                 return
 
+        elif safe_isinstance(model, "interpret.glassbox.ExplainableBoostingRegressor"):
+            self.model = model.predict
+
         # here we need to compute the offsets ourselves because we can't pull them directly from a model we know about
         assert safe_isinstance(self.masker, "shap.maskers.Independent"), (
             "The Additive explainer only supports the Tabular masker at the moment!"
@@ -114,7 +117,12 @@ class AdditiveExplainer(Explainer):
         """
         if safe_isinstance(model, "interpret.glassbox.ExplainableBoostingClassifier"):
             if model.interactions != 0:
-                raise NotImplementedError("Need to add support for interaction effects!")
+                return False
+            return True
+
+        if safe_isinstance(model, "interpret.glassbox.ExplainableBoostingRegressor"):
+            if model.interactions != 0:
+                return False
             return True
 
         return False
--- a/shap/explainers/_explainer.py
+++ b/shap/explainers/_explainer.py
@@ -214,10 +214,36 @@ class Explainer(Serializable):
 
                 # if we get here then we don't know how to handle what was given to us
                 else:
-                    raise TypeError(
-                        "The passed model is not callable and cannot be analyzed directly with the given masker! Model: "
-                        + str(model)
-                    )
+                    # Check if model is an interpret EBM — wrap into a callable and retry
+                    if safe_isinstance(self.model, "interpret.glassbox.ExplainableBoostingClassifier"):
+                        return self.__init__(
+                            self.model.decision_function,
+                            self.masker,
+                            link=link,
+                            algorithm=algorithm,
+                            output_names=output_names,
+                            feature_names=feature_names,
+                            linearize_link=linearize_link,
+                            seed=seed,
+                            **kwargs,
+                        )
+                    elif safe_isinstance(self.model, "interpret.glassbox.ExplainableBoostingRegressor"):
+                        return self.__init__(
+                            self.model.predict,
+                            self.masker,
+                            link=link,
+                            algorithm=algorithm,
+                            output_names=output_names,
+                            feature_names=feature_names,
+                            linearize_link=linearize_link,
+                            seed=seed,
+                            **kwargs,
+                        )
+                    else:
+                        raise TypeError(
+                            "The passed model is not callable and cannot be analyzed directly with the given masker! Model: "
+                            + str(model)
+                        )
 
             # build the right subclass
             if algorithm == "exact":
--- a/tests/explainers/test_additive.py
+++ b/tests/explainers/test_additive.py
@@ -6,7 +6,7 @@ import shap
 
 @pytest.fixture
 def ebm_classifier_no_interaction():
-    interpret = pytest.importorskip("interpret")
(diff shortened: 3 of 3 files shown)


mohityadav8 (changes requested):
Verified the core fix locally.

- shap.Explainer(model, X) on a default EBM classifier (interactions="3x") no longer raises NotImplementedError, correctly falls back to ExactExplainer.
- EBM classifier with interactions=0 still correctly resolves to AdditiveExplainer (fast path preserved).
- 3 of 4 new tests pass.

One test fails: test_additive_explainer_with_ebm_regressor, with AttributeError: 'AdditiveExplainer' object has no attribute 'shap_values'. Confirmed this is pre-existing and unrelated to this PR — on master, shap_values() is commented out on AdditiveExplainer (shap/explainers/_additive.py line 196) and only __call__ remains as the current API. Since this test currently fails on CI, either:
1. Update the test to use the current API (e.g. explainer(X) instead of explainer.shap_values(X)), or
2. Mark it xfail with a note that it's pre-existing and unrelated to this fix, so CI doesn't show a false red X against this PR.

Also flagging, unrelated to correctness: in _explainer.py, the fallback branch uses `return self.__init__(...)` recursively when wrapping EBM models into callables. This works, but recursing into __init__ and returning its value is unconventional (__init__ should return None). Might be cleaner to resolve the model into a callable in a small helper before the single __init__ call.

Lastly: none of the 4 new tests exercise the actual reported bug end-to-end (shap.Explainer(model, X) not raising) — they only test supports_model_with_masker() and one direct AdditiveExplainer construction. Worth adding something like:

    def test_ebm_with_interactions_falls_back_via_explainer():
        X, y = make_classification(n_samples=150, n_features=6, random_state=0)
        model = ExplainableBoostingClassifier(interactions=2, random_state=0)
        model.fit(X, y)
        explainer = shap.Explainer(model, X)  # should not raise
        sv = explainer(X[:5])
        assert sv.values.shape == (5, 6)

to lock in the actual fix from the issue, not just the supporting logic.

PR #5081, opened 2026-07-15
PERF: Migrate lower_credit to nanobind

daidahao:
## Overview

@CloseChoice

**IMPORTANT**: I ran unit tests in debug mode and can confirm that test_tabular_single_output ran through the 1D path, while test_tabular_multi_output ran through the 2D path.

Migrates the Partition explainer's recursive `lower_credit` routine from Numba/Python to the nanobind C++ extension.

The change:

- adds dimension-constrained C++ implementations for one- and two-dimensional values;
- exposes them as nanobind overloads selected from the dimensionality of `values`;
- removes the Numba implementation and calls the extension from `PartitionExplainer`;
- preserves the existing in-place credit distribution and validates parent/child cluster sizes.

### Overload dispatch

Both recursive implementations now use the same Python-facing call shape: the current node, feature count, previous node, scaling factor, values, and clustering. Nanobind dispatches to the correct implementation using explicit `nb::ndim<1>` and `nb::ndim<2>` constraints on `values`.

The dimensional constraints are required. Without them, the overload signatures are indistinguishable and a multi-output array can enter the 1D implementation, producing incorrect credit distribution.

## Performance monitoring

Command:

```console
uv run asv continuous upstream/master HEAD --bench partition_explainer
```

ASV reported `BENCHMARKS NOT SIGNIFICANTLY CHANGED`.

| Change | Before [846f28c7] <master> | After [66e64546] <pr/daidahao/5081> | Ratio | Benchmark |
|---|---:|---:|---:|---|
| | 3.70±0.05s | 3.71±0.02s | 1.00 | `partition_explainer.PartitionSuite.time_multi_output` |
| | 3.68±0.01s | 3.67±0.01s | 1.00 | `partition_explainer.PartitionSuite.time_single_output` |

## Validation

- [x] Pre-commit hooks passed for the committed changes.
- [x] `HF_HUB_OFFLINE=1 OMP_NUM_THREADS=1 VECLIB_MAXIMUM_THREADS=1 uv run pytest tests/explainers/test_partition.py -q` — 8 passed.
- [x] `uv run asv continuous upstream/master HEAD --bench partition_explainer` — no significant benchmark changes.

No new unit tests were added because the existing single- and multi-output baseline tests cover the overload dispatch and output equivalence.

## AI usage disclosure

Codex assisted with the migration and PR preparation. It inspected the implementation and repository guidance, diagnosed the overload-dispatch regression, ran tests and benchmarks, pushed the prepared branch, and updated this PR description. The contributor remains responsible for, and has reviewed, the submitted code and description.


shap/cutils/partition_explainer_utils.h
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        const double left_factor = static_cast<double>(lsize) / static_cast<double>(group_size);
+        const double right_factor = static_cast<double>(rsize) / static_cast<double>(group_size);
+
+        lower_credit_2d(li, M, i, left_factor, values, clustering);
+        lower_credit_2d(ri, M, i, right_factor, values, clustering);
+    }
+
+    void lower_credit(
```
CloseChoice: okay, I see. You are doing the dispatch manually here. I wonder if we better follow one consistent approach that we already took in here:
https://github.com/shap/shap/blob/master/shap/cutils/grey_code_utils.h
and let nanobind handle the dispatching.

CloseChoice (changes requested):

shap/cutils/partition_explainer_utils.h
daidahao: Thanks! I have switched to compiler dispatch / overload, so manual dispatch is no longer needed.

daidahao:
@CloseChoice 
I've switched to nanobind overloads as requested. The PR is ready for review again.

shap/explainers/_partition.py
```diff
@@ -264,7 +264,14 @@ def explain_row(
         # drop the interaction terms down onto self.values
         self.values[:] = self.dvalues
 
-        lower_credit(len(self.dvalues) - 1, 0, M, self.values, self._clustering)
+        lower_credit(
+            i=len(self.dvalues) - 1,
+            M=M,
+            prev_i=len(self.dvalues) - 1,
```
CloseChoice, line 270: I don't quite understand the use of `prev_i` if it is exactly the same as `i`.

shap/explainers/_partition.py
```diff
@@ -264,7 +264,14 @@ def explain_row(
...
+            factor=0.0,
```
CloseChoice, line 271: what is factor for? The whole addition does not have an effect with `factor=0.0`, or am I missing something?

CloseChoice (changes requested):

CloseChoice (changes requested):
Claude found a parity break:
check the newly added test for that.

daidahao:
@CloseChoice I've inspected this "parity break" which exposes a floating-point operation-order difference. In theory, we could preserve integer operands (child size and group size) to produce an exact match, but that means using two arguments rather just one double as `factor`. However, the current implementation is already at parity with the original and practical impact is negligible (1.6e-16 for the test).

Instead, I would suggest changing the test assertion to:

```
    np.testing.assert_allclose(values[2], 4.2 * 1 / 3, rtol=0, atol=1e-15)
```

What do you think?

shap/explainers/_partition.py
daidahao: `prev_i` is the parent node if current node `i` and is needed for for recursive calls. The `lower_credit` function recursively visits root node's children and assign value based on the `factor` (size of the node divided by the size of the parent/group). The `prev_i` changes when the function calls itself recursively.

shap/explainers/_partition.py
daidahao: See above comment https://github.com/shap/shap/pull/5081#discussion_r3830451094

CloseChoice:
@daidahao I removed the additional parameters you added, and restructured the output. Also ran ~200 parity tests and they matched after slight refactoring, before we had some differences at 1e-10 scales.

PR #5089, opened 2026-07-17
PERF: Remove Numba from image masker

daidahao:
## Overview

Removes the remaining Numba usage from the image masker's partition-tree construction while preserving its behavior and steady-state performance.

The change:

- replaces `numba.typed.List` with the built-in Python list for the heap;
- removes the `@njit` decorator and obsolete Numba imports from the image masker;
- adds image masker initialization and image `PartitionExplainer` performance monitoring;
- adds OpenCV to the ASV benchmark environment; and
- preprocesses benchmark images using the return value expected by ResNet50.

This simplifies image masker initialization without a measurable performance change in the new focused benchmark.

## Performance monitoring

Comparison command:

```console
uv run asv continuous upstream/master HEAD
```

ASV ran the full suite of 12 benchmarks across both revisions and reported `BENCHMARKS NOT SIGNIFICANTLY CHANGED.`

The image masker initialization benchmark remained stable:

| Change | Before [846f28c7] <master> | After [2c2f23db] <feature/nanobind-image> | Ratio | Benchmark |
|---|---:|---:|---:|---|
| | 215+/-3ms | 210+/-2ms | 0.98 | partition_explainer.ImagePartitionSuite.time_masker_init |

## Validation

- [x] `uv run pytest tests/maskers/test_image.py -q` - 11 passed
- [x] `uv run pytest tests/explainers/test_partition.py::test_tabular_single_output tests/explainers/test_partition.py::test_tabular_multi_output -q` - 2 passed
- [x] `uv run asv check` - no problems found
- [x] `uv run asv continuous upstream/master HEAD` - no significant changes across 12 benchmarks
- [x] repository pre-commit hooks on all changed files

## AI usage disclosure

Codex assisted with branch review, benchmark validation, commit preparation, and PR drafting. It identified and corrected discarded ResNet50 preprocessing output, ran the focused tests and ASV comparison, pushed the prepared branch, and drafted/opened this PR. The contributor remains responsible for, and has reviewed, the submitted code and description.


shap/maskers/_image.py
```diff
@@ -171,7 +168,7 @@ def load(cls, in_file, instantiate=True):
         return kwargs
 
 
-@njit  # pragma: no cover
+# @njit  # pragma: no cover
```
CloseChoice: any reason why you didn't remove this?

CloseChoice (changes requested):
Removing the numba jit compilation does not decrease the performance? Did I get that right? I haven't looked deeply into what `q` is here but I am a bit puzzled.

shap/maskers/_image.py
daidahao: Apologies, but it should not affect the results here because njit is commented out. I've removed the line now.

daidahao:
@CloseChoice That is correct.

According to the ASV monitoring, removing JIT from `_jit_build_partition_tree` does not affect its speed and the function is only called once at `Image` masker initialisation. So I decided to keep it in Python.

`q` is a priority queue for building the partition tree. Each entry on the queue is an image region that needs to be partitioned.

CloseChoice (approved):
LGTM, thanks.

PR #5142, opened 2026-09-05
ENH: Add Pyodide/JupyterLite support via PyEmscripten wheel (PEP 783)

### LINKED ISSUE #5124: ENH: Pyodide build to add support for JupyterLite
### Problem Description

Hi, we tried to use SHAP in JupyterLite, and it didn't work. It seems SHAP has no Pyodide build, nor is pure-python package, and so Pyodide bails with this error:
```
ValueError: Can't find a pure Python 3 wheel for 'shap'.
See: https://pyodide.org/en/stable/usage/faq.html#why-can-t-micropip-find-a-pure-python-wheel-for-a-package
You can use `await micropip.install(..., keep_going=True)` to get a list of all packages with missing wheels.
```

Are you planning adding support for Pyodide? Quick search suggests that this might be a blocker - https://github.com/shap/shap/issues/4327.

Below is analysis from ChatGPT 5.6 Sol xhigh re how this could be done:

----

Short answer: Pyodide support looks feasible, but no usable SHAP wheel exists yet. A direct cross-build proved that both CPU extensions can compile and link as WebAssembly after a diagnostic workaround. Installation and runtime behavior remain unproven.

## How SHAP builds wheels

At SHAP commit 79f287fa (https://github.com/shap/shap/commit/79f287fa8939e21a610026368ff0af214f62b38d), then master:

- pyproject.toml (https://github.com/shap/shap/blob/79f287fa8939e21a610026368ff0af214f62b38d/pyproject.toml) uses the scikit-build-core PEP 517 backend. It runs CMake, while
  setuptools-scm generates the version.

- CMakeLists.txt (https://github.com/shap/shap/blob/79f287fa8939e21a610026368ff0af214f62b38d/CMakeLists.txt) always builds two CPU extensions:
    - _cutils: a nanobind C++ module used by the Exact and Kernel explainers, maskers, and clustering utilities.
    - _cext: the Tree SHAP extension built against the CPython and NumPy C APIs.

- _cext_gpu is optional and built only when SHAP_ENABLE_CUDA is set.
- Regular wheels request CPython’s 3.12 stable ABI. Free-threaded builds use their interpreter-specific ABI.
- The release workflow (https://github.com/shap/shap/blob/79f287fa8939e21a610026368ff0af214f62b38d/.github/workflows/build_wheels.yml) uses cibuildwheel 3.2.1 for Linux, macOS,
  and Windows. There is currently no WebAssembly job.

- The plotting JavaScript bundle is prebuilt; wheel creation does not invoke npm or webpack.

The current system was introduced in SHAP 0.52.0 (https://github.com/shap/shap/releases/tag/v0.52.0), replacing the older setup.py and Cython build.

## What currently blocks Pyodide

1. Numba and llvmlite

   SHAP declares both as unconditional runtime dependencies, but Pyodide 314.0.6 includes neither. SHAP still has three direct numba.njit import sites. The ongoing Numba-to-C++
   migration (https://github.com/shap/shap/issues/4327) has not finished.

   Pyodide does include NumPy, SciPy, pandas, scikit-learn, cloudpickle, packaging, and tqdm. slicer==0.0.8 has a pure-Python wheel, but a complete dependency-resolving
   installation has not been tested. Current Pyodide package list (https://pyodide.org/en/stable/usage/packages-in-pyodide.html)

2. Nanobind stub generation

   nanobind_add_stub() generates _cutils.pyi by importing the compiled _cutils extension. During cross-compilation, host macOS Python cannot import a WebAssembly .so. This
   behavior follows nanobind’s documented stub-generation mechanism (https://nanobind.readthedocs.io/en/latest/typing.html).

   SHAP also generates py.typed, but currently copies only _cutils.pyi into the source package and explicitly installs neither typing file.

3. A nanobind/Emscripten compiler issue

   In the tested toolchain, nanobind 3.0.0 failed because stderr was undeclared. Adding CXXFLAGS="-include cstdio" bypassed this error, but that is only a diagnostic workaround.

## Direct build result

I tested:

- pyodide-build 0.39.0
- Pyodide 314.0.6
- Emscripten 5.0.3
- Target CPython 3.14.2
- nanobind 3.0.0

After the diagnostic cstdio workaround, both targets compiled and linked:

_cext.abi3.so:   WebAssembly binary module
_cutils.abi3.so: WebAssembly binary module

The build then stopped when host Python tried to import _cutils.abi3.so for stub generation.

So the experiment confirms native compilation and linking. It does not confirm wheel completion, dependency installation, import shap, or execution of an explainer inside
Pyodide.

## Changes needed for a Pyodide wheel

1. Finish removing Numba, or add a tested no-JIT decorator fallback and mark numba and llvmlite with sys_platform != "emscripten".
2. Check in _cutils.pyi and py.typed, explicitly install them, and skip target-module stub generation during Emscripten builds.
3. Fix the missing standard-I/O declaration upstream, or add a narrow Emscripten-only workaround with a regression test.
4. Add a dedicated cibuildwheel 4.x Pyodide job. Pin compatible cibuildwheel, pyodide-build, and Pyodide versions rather than relying on defaults. PEP 783 permits publishing
   standard pyemscripten wheels directly to PyPI, which is now preferable to maintaining a separate recipe. PEP 783 (https://peps.python.org/pep-0783/), cibuildwheel Pyodide
   support (https://cibuildwheel.pypa.io/en/stable/platforms/#pyodidewebassembly)

5. Test normal installation and import, wheel contents, and small Exact, Kernel, Partition, and scikit-learn TreeExplainer cases in Node-backed Pyodide, followed by a browser
   smoke test.

The sensible initial scope is those CPU explainers. CUDA-based GPUTreeExplainer cannot target Pyodide, while DeepExplainer and GradientExplainer remain outside the initial scope
because current Pyodide includes neither TensorFlow nor PyTorch. Nontrivial explanations should run in a Web Worker so they do not block the UI.

Even though SHAP uses abi3, each PyEmscripten platform generation still needs its own tagged wheel because Emscripten does not promise ABI compatibility across generations. A
Pyodide package request for Numba and SHAP (https://github.com/pyodide/pyodide-recipes/issues/273) has been open since 2023.

Verdict: promising but unproven. The next milestone is a patched wheel that installs with normal dependency resolution and runs at least one explainer under Pyodide.

### Alternative Solutions

_No response_

### Additional Context

_No response_

### Feature request checklist

mohityadav8:
## Summary

This PR adds Pyodide/JupyterLite support to SHAP, allowing installation
via micropip in browser-based Python environments.

## Changes

**pyproject.toml**
- Remove numba and llvmlite as hard dependencies (the primary blocker
  for micropip install in Pyodide)
- Add WebAssembly OS classifier per PEP 783
- Add [tool.cibuildwheel.pyodide] section with CXXFLAGS=-include cstdio
  and SHAP_SKIP_STUB_GEN=1 environment variables

**CMakeLists.txt**
- Skip nanobind stub-gen during Emscripten cross-builds
- Read NUMPY_INCLUDE_DIR from environment instead of querying host Python
- Add -include cstdio compiler flag for Emscripten builds to fix missing
  stderr declaration in nanobind <= 3.0.0
- Skip CUDA/GPU extension when building for Pyodide

**New files**
- shap/_cutils.pyi: Pre-committed type stub for Pyodide wheels
- shap/py.typed: PEP 561 marker file
- .github/workflows/build_pyodide_wheel.yml: CI workflow for nightly
  and release PyEmscripten wheel builds
- tests/test_pyodide.py: Smoke tests skipped on non-Emscripten platforms

## Testing

Smoke tests cover ExactExplainer, KernelExplainer, PartitionExplainer,
and TreeExplainer with scikit-learn models.

Closes #5124

### DIFF
--- a/.github/workflows/build_pyodide_wheel.yml
+++ b/.github/workflows/build_pyodide_wheel.yml
@@ -1,0 +1,158 @@
+name: Build Pyodide wheel
+
+# Builds a PyEmscripten wheel (per PEP 783) that can be installed into
+# Pyodide / JupyterLite via micropip.
+#
+# The wheel is published to PyPI alongside the regular wheels when a release
+# tag is pushed.  It is also uploaded as a workflow artifact on every run so
+# it can be smoke-tested in a browser or Node-backed Pyodide environment.
+
+on:
+  release:
+    types: [published]
+
+  # Nightly, aligned with the regular build_wheels schedule
+  schedule:
+    - cron: "30 4 * * *"
+
+  # Enable manual triggering for development / debugging
+  workflow_dispatch:
+
+jobs:
+  # ---------------------------------------------------------------------------
+  # Skip nightly builds when there are no new commits (mirrors build_wheels.yml)
+  # ---------------------------------------------------------------------------
+  check_changes:
+    name: Check for new commits
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    outputs:
+      should_build: ${{ steps.check.outputs.should_build }}
+    steps:
+      - name: Compare HEAD with last successful nightly run
+        id: check
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          if [ "${{ github.event_name }}" != "schedule" ]; then
+            echo "Not a scheduled run, building."
+            echo "should_build=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          last_sha=$(gh run list \
+            --repo "${{ github.repository }}" \
+            --workflow "${{ github.workflow }}" \
+            --event schedule \
+            --status success \
+            --branch "${{ github.ref_name }}" \
+            --limit 1 \
+            --json headSha \
+            --jq '.[0].headSha // ""')
+          echo "Current HEAD:          ${{ github.sha }}"
+          echo "Last successful build: ${last_sha:-<none>}"
+          if [ "$last_sha" = "${{ github.sha }}" ]; then
+            echo "No new commits since last nightly, skipping."
+            echo "should_build=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "should_build=true" >> "$GITHUB_OUTPUT"
+          fi
+
+  # ---------------------------------------------------------------------------
+  # Build the Pyodide / PyEmscripten wheel
+  # ---------------------------------------------------------------------------
+  build_pyodide_wheel:
+    name: Build Pyodide wheel
+    needs: [check_changes]
+    if: needs.check_changes.outputs.should_build == 'true'
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+
+    steps:
+      # Fetch full history so setuptools-scm can derive the version from tags.
+      - name: Check out the repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Fetch all tags
+        run: git fetch --tags
+
+      - name: Set up Python 3.12
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+
+      # cibuildwheel 4.x ships Pyodide support out of the box.
+      # Pin versions for reproducibility; bump intentionally.
+      - name: Build Pyodide wheel via cibuildwheel
+        uses: pypa/cibuildwheel@v4.2.0
+        env:
+          # ---- target platform ----
+          CIBW_PLATFORM: pyodide
+
+          # ---- verbosity ----
+          CIBW_BUILD_VERBOSITY: 2
+
+          # ---- Emscripten workaround: nanobind ≤ 3.0.0 uses bare `stderr`
+          # without including <cstdio>; inject the header for the cross build.
+          # This flag is only passed to the C++ compiler (clang/em++) via
+          # scikit-build-core's CMAKE_CXX_FLAGS forwarding.
+          CIBW_ENVIRONMENT: >-
+            CXXFLAGS="-include cstdio"
+
+          # ---- skip stub generation that needs host-Python import of WASM .so
+          # The pre-committed _cutils.pyi is used instead (see CMakeLists.txt).
+          CIBW_ENVIRONMENT_PASS_LINUX: SHAP_SKIP_STUB_GEN
+
+          # ---- Pyodide smoke-test (runs inside Node-backed Pyodide)
+          # Only test the CPU explainers that Pyodide can support.
+          CIBW_TEST_COMMAND: >-
+            python -c "
+            import pyodide_js, asyncio, micropip, pathlib, sys
+            # micropip is the micro-installer used inside Pyodide
+            wheel = str(pathlib.Path('{wheel}').resolve())
+            async def _run():
+                await micropip.install('file://' + wheel)
+                import shap
+                import numpy as np
+                from sklearn.ensemble import RandomForestClassifier
+                X = np.random.randn(20, 4)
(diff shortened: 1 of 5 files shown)


mohityadav8:
cc @CloseChoice done

.github/workflows/build_pyodide_wheel.yml
```diff
@@ -0,0 +1,158 @@
...
+    - cron: "30 4 * * *"
+
+  # Enable manual triggering for development / debugging
+  workflow_dispatch:
+
+jobs:
+  # ---------------------------------------------------------------------------
+  # Skip nightly builds when there are no new commits (mirrors build_wheels.yml)
+  # ---------------------------------------------------------------------------
+  check_changes:
```
CloseChoice, line 25: This duplicates code. Can you move the core logic to build_wheels.yml?

shap/_cutils.pyi
CloseChoice, line 1: Why are there whitespace changes here?

tests/explainers/test_exact.py
```diff
@@ -4,9 +4,9 @@
 
 import numpy as np
 from conftest import compare_numpy_outputs_against_baseline
-from shap._cutils import compute_grey_code_row_values_st
 
 import shap
+from shap._cutils import compute_grey_code_row_values_st
```
CloseChoice, line 9: pls don't reformat anything that's not strictly needed in a PR that is dedicated to adding new functionality, we want to keep the changes as minimal as possible.

tests/utils/test_masked_model.py
```diff
@@ -1,7 +1,7 @@
 import numpy as np
 import pytest
-from shap._cutils import _init_masks  # type: ignore[attr-defined]
 
+from shap._cutils import _init_masks  # type: ignore[attr-defined]
```
CloseChoice, line 4: see comment above

tests/test_pyodide.py
```diff
@@ -0,0 +1,171 @@
+r"""Smoke tests for the Pyodide/JupyterLite build.
```
CloseChoice, line 1: Why do we have this? Shouldn't we be able to run the normal test suite using pyodide?

pyproject.toml
```diff
@@ -241,7 +243,9 @@ minimum-version = "build-system.requires"
 # Setuptools-style build caching in a local directory
 build-dir = "build/{wheel_tag}"
 
-# Exclude C/C++/CUDA source files from the wheel (compiled .so is sufficient)
+# Exclude C/C++/CUDA source files from the wheel (compiled .so is sufficient).
```
CloseChoice, line 246: I don't see any need to change this comment.

pyproject.toml
```diff
@@ -361,3 +365,27 @@ archs = ["x86_64", "arm64"]
 [tool.cibuildwheel.macos.environment]
 # Needed for C++17 support
 MACOSX_DEPLOYMENT_TARGET = "10.13"
+
+# ==============================================================================
```
CloseChoice, line 369: pls remove this comment. If there is a link explaining that just refer there.

CloseChoice (changes requested):
Pls see my comments.

CloseChoice:
I would like to integrate pyodide in the existing workflows:
- build_wheels.yml
- run_tests.yml
Before officially supporting pyodide we need to ensure that everything works as for CPython.

mohityadav8:
hi @CloseChoice you have maked a new pr #5162 with a cleanup so should i close my pr 🥺sorry for late response i was busy 🥺🥺

CloseChoice:
yes, let's close this. I added your changes and just build on top of that. So thanks a lot for this effort.

mohityadav8:
@CloseChoice i reviewed issue and pr and i reproduce those issssue and they are good should i mention you their WDYT?

PR #5144, opened 2026-09-05
Update actions/checkout to v7.0.1 and pin its commit SHA

CloseChoice:
## Overview

Closes #XXXX  

## Checklist

### DIFF
--- a/.github/workflows/build_gpu.yml
+++ b/.github/workflows/build_gpu.yml
@@ -14,7 +14,7 @@ jobs:
     permissions:
       id-token: write   # for PyPI trusted publishing
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v7
         with:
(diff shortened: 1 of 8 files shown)


.github/workflows/build_gpu.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     permissions:
       id-token: write   # for PyPI trusted publishing
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v7
```
tylerjereddy, line 17: just an unsolicited drive-by suggestion, but in the future you might want to consider using hashes that match the version for supply chain security reasons (hashes are harder to mutate); there's an example of a bot doing it upstream here: https://github.com/numpy/numpy/pull/32139

CloseChoice: Thanks tyler, will adapt the numpy pattern soon.

PR #5152, opened 2026-09-06
DOCS: Document categorical split limitation in TreeExplainer

### LINKED ISSUE #5132: DOCS: TreeExplainer does not document that models with categorical splits are unsupported with background data
## Summary

`TreeExplainer`'s docstring lists LightGBM as supported and documents `data` and `feature_perturbation` in detail, but never mentions that a model trained with **native categorical splits** cannot be used with a background dataset — i.e. not with `feature_perturbation="interventional"`, and therefore not with `model_output="log_loss"`, which requires one.

There are **zero occurrences of "categor"** in the class or `__init__` docstring (76 lines of parameter documentation).

This is not an obscure corner: LightGBM's pandas-`category` support is a headline feature, so "LightGBM is supported" reads as covering it.

## What the docs currently say

- **`model`** — "XGBoost, LightGBM, CatBoost, Pyspark and most tree-based scikit-learn models are supported." No caveat.
- **`feature_perturbation`** — explains interventional vs tree-path-dependent, background-set sizing, and the causal-inference rationale. This is the natural home for the constraint, and it isn't there.

## Reproduction

shap 0.51.0, lightgbm 4.6.0, pandas 2.3.3, numpy 2.4.6, Python 3.11.

```python
import numpy as np, pandas as pd, lightgbm as lgb, shap

rng = np.random.default_rng(0)
n = 2000
X = pd.DataFrame({
    "num": rng.standard_normal(n),
    "cat": pd.Categorical(rng.choice(["a", "b", "c"], n)),
})
y = (X["num"] + (X["cat"] == "a") + rng.standard_normal(n) * 0.5 > 0).astype(int)
model = lgb.LGBMClassifier(n_estimators=20, verbose=-1).fit(X, y)
bg = X.iloc[:100]

# 1. works
shap.TreeExplainer(model).shap_values(X.iloc[:5])

# 2. ExplainerError at construction — clear, names the remedy
shap.TreeExplainer(model, data=bg)

# 3. constructs fine, then AttributeError from shap_values — confusing
shap.TreeExplainer(model, data=bg, model_output="log_loss").shap_values(X.iloc[:5], y.iloc[:5])
```

| Construction | Result |
| --- | --- |
| `TreeExplainer(model)` | works |
| `TreeExplainer(model, data=bg)` | `ExplainerError`, names the remedy |
| `TreeExplainer(model, data=bg, model_output="log_loss")` | constructs OK, then `AttributeError: 'TreeEnsemble' object has no attribute 'values'` |

## Secondary problem: the existing guard is bypassed

Case 2 raises a genuinely helpful error:

```
ExplainerError: Currently TreeExplainer can only handle models with categorical splits when
feature_perturbation="tree_path_dependent" and no background data is passed. Please try again
using shap.TreeExplainer(model, feature_perturbation="tree_path_dependent").
```

Adding `model_output="log_loss"` slips past it. The LightGBM branch of `TreeEnsemble.__init__` swallows the underlying failure:

```python
try:
    self.trees = [SingleTree(e, data=data, data_missing=data_missing) for e in tree_info]
except Exception:
    self.trees = None  # we get here because the cext can't handle categorical splits yet
```

so the ensemble is left half-built — the block assigning `.values`, `.children_left` etc. is guarded by `if self.trees is not None and self.trees:` and never runs.

The clear error in case 2 comes from `self.expected_value = self.model.predict(self.data).mean(0)` in `TreeExplainer.__init__`. The `log_loss` path skips that line (its expected value depends on `y`), so nothing checks, and the failure only surfaces in `_validate_inputs` at `tree_limit > self.model.values.shape[0]` — several frames from the actual cause.

## Suggested change

Add a sentence to `feature_perturbation` (and ideally to `model`):

> Models with native categorical splits (e.g. LightGBM trained on pandas `category` dtypes) are only supported with `feature_perturbation="tree_path_dependent"` and no background data. `model_output="log_loss"` requires background data and is therefore unavailable for such models.

Raising the existing `ExplainerError` on the `log_loss` path as well would turn case 3 into the same clear message.

JuanSeguraA:
## Overview

Closes #5132 

Added documentation regarding the use of models with native categorical splits and how they only work with feature_perturbation="tree_path_dependent" and no background data. Also noted that this caveat extends to model_output="log_loss", which requires feature_perturbation="interventional" and therefore a background dataset, making it incompatible with models that have native categorical splits.

I also added similar documentation in the feature_perturbation section of the docstring, since that's probably where a user debugging this specific issue would look.

Regarding the bug fix, if you used a LightGBM model with categorical splits and passed background data with model_output="log_loss", TreeExplainer would build fine but then crash later with a confusing AttributeError. The same setup without log_loss already gave a clear error, but this path just wasn't checked. Therefore, added the same check to the log_loss path, so it now raises the same clear error right away instead of failing later.

## Checklist

### DIFF
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -354,6 +354,12 @@ class TreeExplainer(Explainer):
 
         # compute the expected value if we have a parsed tree for the cext
         if self.model.model_output == "log_loss":
+            if self.model.trees is None:
+                raise ExplainerError(
+                    "Currently TreeExplainer can only handle models with categorical splits when "
+                    'feature_perturbation="tree_path_dependent" and no background data is passed. Please try again using '
+                    'shap.TreeExplainer(model, feature_perturbation="tree_path_dependent").'
+                )
             self.expected_value = self.__dynamic_expected_value
         elif data is not None:
             try:
--- a/shap/explainers/_tree.py
+++ b/shap/explainers/_tree.py
@@ -179,7 +179,10 @@ class TreeExplainer(Explainer):
         model : model object
             The tree based machine learning model that we want to explain.
             XGBoost, LightGBM, CatBoost, Pyspark and most tree-based
-            scikit-learn models are supported.
+            scikit-learn models are supported. Notably, models with native categorical splits
+            only work with ``feature_perturbation="tree_path_dependent"`` and no background data.
+            Therefore, such models cannot be used with ``model_output="log_loss"``, which requires
+            ``feature_perturbation="interventional"`` and therefore a background dataset.
 
         data : numpy.array or pandas.DataFrame
             The background dataset to use for integrating out features.
@@ -207,6 +210,10 @@ class TreeExplainer(Explainer):
             - if ``"auto"``, the "interventional" approach will be used when a
               background is provided, otherwise the "tree_path_dependent" approach will
               be used.
+
+            Models with native categorical splits (e.g. LightGBM trained on pandas ``category`` dtype columns)
+            are only supported with ``feature_perturbation="tree_path_dependent"``. Passing a background dataset
+            with such a model is not currently supported.
 
             .. versionadded:: 0.47
                The `"auto"` option was added.


mohityadav8 (changes requested):
Thanks for tackling both the docs and the underlying bug here.

The `ExplainerError` added at line 358 reuses the exact message from the `tree_path_dependent`-without-background guard at line 368, including the remedy "Please try again using `shap.TreeExplainer(model, feature_perturbation="tree_path_dependent")`." That remedy doesn't actually work in this branch: this code only runs when `model_output="log_loss"`, and `tree_path_dependent` requires `model_output="raw"` (enforced at line 333-334). So the suggested fix means "give up on log_loss," but the message doesn't say that, it just points the user at a setup that will never give them log_loss output.

Could you write a log_loss-specific message instead, something like:

    raise ExplainerError(
        "TreeExplainer cannot compute log_loss for models with native categorical splits: "
        "log_loss requires background data and feature_perturbation=\"interventional\", "
        "which are unsupported for such models."
    )

Also, no test currently covers this exact case (categorical-splits model with model_output="log_loss" and background data). Since this PR fixes a real bug (case 3 in #5132, an AttributeError that should be a clear error), it would be good to lock that in with something like:

    def test_tree_explainer_categorical_log_loss_raises():
        model = lgb.LGBMClassifier(...).fit(X_with_categorical, y)
        with pytest.raises(ExplainerError, match="log_loss"):
            shap.TreeExplainer(model, data=background, model_output="log_loss")

Everything else looks good, the docstring additions cover what the issue asked for.

JuanSeguraA:
@mohityadav8 good catch, thanks. Updated the message to explain that log_loss isn't supported for models with native categorical splits.

Also added test_lightgbm_categorical_log_loss_raises covering the exact case you described (categorical-splits LightGBM model + model_output="log_loss" + background data --> ExplainerError).

PR #5164, opened 2026-09-17
FIX: match KernelExplainer DataFrame columns to the background by name

### LINKED ISSUE #3958: BUG: Column labels not taken into account

Atishyy27:
## Overview

Closes #3958

`shap_values` read a DataFrame by position, so reordered columns were scored on the wrong features. A DataFrame with the background's column names in another order is now explained in background order and mapped back. Missing background columns raise a `ValueError` instead of an `IndexError`. Series, arrays, extra or renamed columns, duplicate background names and `shap.kmeans` backgrounds stay positional as before; happy to follow up on Series and kmeans if wanted. Follows #4510, which had no test.

Written with Claude (Opus 5).

## Checklist


shap/explainers/_kernel.py
```diff
@@ -226,6 +226,30 @@ def __call__(  # type: ignore[override]
             compute_time=time.time() - start_time,
         )
 
+    def _background_order(self, X: pd.DataFrame) -> list[int] | None:
```
CloseChoice: This looks like the correct idea but I think this is overly complicated. Couldn't we simply use `X.loc[:, background.columns]`?

CloseChoice (changes requested):
Thanks for the analysis and the PR.

CloseChoice:
@Atishyy27 I made some changes to simplify the handling. I hope this is fine. Let me know what you think.

EDIT: my changes made this technically breaking since duplicating columns worked before. Though I think we can certainly break this edge case and require users to set unique column names.

Atishyy27:
Looks good to me, the stricter contract reads better than mine. One case it drops: a numpy background with feature_names passed, where a reordered frame goes back to positional. Happy to leave that out of scope.

